PR #1917, opened 2026-04-13
Daemon transport stability updates - loopback host, serialization, timeouts, error IDs

### LINKED ISSUE #1938: ECONNREFUSED on daemon MCP hop due to IPv4/IPv6 mismatch on Linux CI
Description: On Linux CI runners, the daemon's internal MCP HTTP server binds to localhost, which may resolve to ::1 (IPv6). Meanwhile, fetch() resolves localhost to 127.0.0.1 (IPv4). This mismatch causes ECONNREFUSED on the internal Unix socket → Streamable HTTP MCP hop, making all tool calls fail intermittently depending on DNS resolution order.

Reproduction: Run JUnit tests against the daemon on a Linux host where localhost resolves to ::1 in /etc/hosts. The daemon starts successfully but tool calls fail with ECONNREFUSED.

Expected: The daemon's internal HTTP server and client should agree on the loopback address.

Impact: CI test runs fail intermittently with no clear cause. Retrying the same job may succeed if DNS resolution happens to be consistent.

RyeBread1026:
Closes #1938, #1939, #1942, #1943, #1944

## Summary

The daemon's internal MCP HTTP server and Unix socket forwarding layer had multiple failure modes causing CI test timeouts, `ECONNREFUSED`, `Operation cancelled`, and silently lost responses. This PR fixes five distinct transport-layer issues that together caused frequent, hard-to-diagnose CI failures.

### A. Align loopback host to IPv4 (`127.0.0.1`)

**Problem:** On Linux CI, `localhost` can resolve to `::1` (IPv6) for one component and `127.0.0.1` (IPv4) for another. Bun's `fetch` and Node's `http.Server.listen` disagree, causing `ECONNREFUSED` on the internal Unix-socket → Streamable HTTP MCP hop.

**Fix:** Default `this.host` to `"127.0.0.1"` instead of `"localhost"`. The `--host` CLI flag still overrides (e.g. `--host localhost` or `--host ::1` restores old behavior). Also threads `--host` through `DaemonManager` arg building and parsing so the override propagates to spawned daemon processes.

**Mechanism:** In `index.ts`, `daemonHost` changed from `string` (always `"localhost"`) to `string | undefined` (only set when `--host` is explicitly passed). The Daemon constructor only receives `host` in its options when the user passes `--host`, so the new `"127.0.0.1"` default in the constructor actually takes effect. Previously `index.ts` always passed `host: "localhost"`, which would have overridden any constructor default.

**Risk:** Low. Both values only accept loopback connections. The only breaking scenario would be an IPv6-only loopback host with no IPv4 — effectively impossible for Android emulator environments. macOS resolves `localhost` to `127.0.0.1` anyway so this is a no-op locally.

**Also in this change:** Clears the device-pool initialization timeout after the `Promise.race` completes (via `try/finally`). Previously, if device discovery finished before the timeout, the stale `setTimeout` callback would still fire and log a misleading "timed out" warning. The `clearTimeout` is a no-op if the timeout already fired, so there's no risk to the timeout-wins path.

**Files:** `src/daemon/daemon.ts`, `src/daemon/manager.ts`, `src/index.ts`, `docs/.../ci-integration.md` (expanded with this-repo emulator.wtf setup and consumer-app adaptation guide; no code changes)

---

### B. Serialize MCP socket forwards

**Problem:** Parallel JUnit test threads each have their own Unix socket connection, but the daemon shares a single Streamable HTTP client session to its own MCP HTTP server. Concurrent `callTool` calls on that one session corrupt the HTTP transport, producing `Operation cancelled` on in-flight `executePlan` and random transport errors.

**Fix:** Introduce `runExclusiveMcpForward` - a promise-chain serializer that ensures only one `callTool` is in flight at a time on the shared MCP client. Add `[McpForward]` debug logging (queue wait time, tool name, duration). Tag `executionTracker.cancelSessionExecutions` with `onclose` vs `onerror` reason strings for diagnosability.

**Why `describeUnknownError`:** The `StreamableHTTPServerTransport.onerror` callback receives `unknown`-typed errors that are often empty objects `{}` or non-`Error` values - useless when reading CI daemon logs. `describeUnknownError` handles all the weird cases (empty objects, `.cause` chains, non-Error throws) and produces a single readable log line. It feeds into both the `onerror` log message and the `cancelSessionExecutions` reason string so you can see *why* a session was torn down.

**`executionTracker.ts`:** Added an optional `reason` parameter (default `"unspecified"`) to `cancelSessionExecutions` and the internal `cancelExecutionsForKey`. The reason is included in the log line. Existing callers outside this PR are unaffected since the parameter has a default value.

**Risk:** The serialization queue means parallel JUnit threads' MCP calls are now sequential. If thread A runs a 5-minute `executePlan`, thread B's calls wait in the queue. However, the old concurrent behavior was broken - it corrupted the shared Streamable HTTP session - so this replaces "sometimes corrupts" with "correctly queued." Single-fork configurations (`maxParallelForks = 1`) see no overhead.

**Files:** `src/daemon/socketServer.ts`, `src/daemon/daemon.ts`, `src/server/executionTracker.ts`, `src/utils/describeUnknownError.ts` (new), `test/daemon/socketServerMcpForwardSerialization.test.ts` (new), `test/utils/describeUnknownError.test.ts` (new)

---

### C. Disable Streamable HTTP SSE auto-reconnect

**Problem:** The MCP SDK's Streamable HTTP client can auto-open a second standalone GET (SSE listener) after a disconnect. The server transport allows only one SSE stream per session - a second GET while the first is still mapped returns HTTP 409 and tears down the entire session, cancelling any in-flight tool calls.

**Fix:** Set `maxRetries: 0` on the loopback client's `StreamableHTTPReconnectionOptions`. Stale sessions are recovered via the existing "Session not found" retry in `runExclusiveMcpForward`.

**Risk:** None. The previous auto-reconnect behavior was actively harmful (caused 409 teardowns). Disabling it is strictly better; stale sessions already have a recovery path.

**Files:** `src/daemon/socketServer.ts`

---

### D. Floor `executePlan` MCP timeout to 10 minutes

**Problem:** The JUnit runner could pass a short `timeoutMs` (30s or 180s) which propagated to the inner Streamable HTTP `callTool`. When the timeout fired mid-plan, the SDK aborted the HTTP request, the server transport fired `onclose`, and `executionTracker` cancelled the execution — producing `Operation cancelled` even though the plan was progressing normally.

**Fix:** Enforce a 10-minute floor (`MIN_EXECUTE_PLAN_MCP_TIMEOUT_MS = 600_000`) on both the JUnit side (`effectiveExecutePlanTimeoutMs()`) and the daemon socket server side (`resolveMcpRequestTimeoutMs()`). Extracted timeout logic into `src/daemon/mcpRequestTimeout.ts` with unit tests. `resolveMcpRequestTimeoutMs` also guards against non-finite inputs (`NaN`, `Infinity`, negative, zero) - falling back to the 30s default rather than propagating a broken timeout into `callTool`.

**Scope:** The floor applies **only to `executePlan`** - `resolveMcpRequestTimeoutMs` checks `request.method === "tools/call" && request.params?.name === "executePlan"` and only then applies `Math.max(base, 600_000)`. All other tools (`observe`, `tapOn`, `listTools`, etc.) keep whatever timeout the client sends, defaulting to 30s. The old code applied a flat 30s to everything including `executePlan`, which is what caused premature `Operation cancelled` on plans that were progressing normally.

**Kotlin side:** `AutoMobilePlanExecutionOptions.timeoutMs` default changed from `30_000L` to `600_000L`, and a new `effectiveExecutePlanTimeoutMs()` extension enforces the same floor. This means existing tests using default options will now wait up to 10 minutes instead of 30 seconds for `executePlan`. The 30s default was never viable for plan execution (plans typically run 1-10 minutes), so the old default always led to premature timeouts.

**Cross-language sync:** The `600_000` constant is defined in both TypeScript (`MIN_EXECUTE_PLAN_MCP_TIMEOUT_MS` in `src/daemon/mcpRequestTimeout.ts`) and Kotlin (`MIN_EXECUTE_PLAN_TIMEOUT_MS` in `AutoMobilePlanTypes.kt`). Both files carry `Keep in sync with ...` comments pointing to their counterpart to prevent drift.

**Files:** `src/daemon/socketServer.ts`, `src/daemon/mcpRequestTimeout.ts` (new), `test/daemon/mcpRequestTimeout.test.ts` (new), `android/.../AutoMobilePlanExecutor.kt`, `android/.../AutoMobilePlanTypes.kt`, `android/.../AutoMobileTestDemo.kt` (test), `android/.../SimpleIntegrationTest.kt` (test)

---

### E. Disable HTTP server timeouts + fix `id:unknown` error responses

**Problem 1 — HTTP timeout:** The daemon's loopback HTTP server is created with `node:http.createServer()` with no timeout configuration. Node.js 18+ defaults `requestTimeout` to 300,000ms (5 minutes). When `executePlan` runs for ~5 minutes via the Streamable HTTP transport, the server terminates the HTTP connection. The `StreamableHTTPServerTransport.onclose` fires, but the tool result is lost because the response stream is already torn down. The MCP SDK client never receives the result and waits until its own timeout. The JUnit runner also times out - even though the plan completed successfully on the daemon.

**Fix:** Set `requestTimeout = 0`, `headersTimeout = 0`, and `timeout = 0` on the loopback HTTP server. This is safe because the server only accepts connections from the in-process MCP client, never from external sources.

**Problem 2 - id:unknown:** The socket server's `handleConnection` data handler had a single outer `try/catch` that sent error responses with `id: "unknown"`. The JUnit client keys its `CompletableFuture` map by request ID, so an `id: "unknown"` response never completes any pending future - causing a silent timeout instead of a proper error.

**Fix:** Move error handling inside the per-line loop. Each line gets its own `try/catch` with the actual `request.id` (falling back to `"unknown"` only if JSON parsing itself fails). This ensures the JUnit client receives a matchable error response and fails fast instead of hanging for the full timeout duration.

**Risk:** Strictly better. Errors that previously caused mysterious 10-minute timeouts will now surface immediately as proper error responses. A bad line no longer aborts processing of subsequent lines in the same data event.

**Files:** `src/daemon/daemon.ts`, `src/daemon/socketServer.ts`

---

## Testing

- All existing daemon tests pass (191 tests across 23 files)
- New tests: `socketServerMcpForwardSerialization.test.ts` (serialization + queue timeout), `mcpRequestTimeout.test.ts` (timeout floor logic), `describeUnknownError.test.ts`
- Build passes cleanly

## How these interact

The five fixes address different stages of the same request path:

```
JUnit runner
  → Unix socket (per-thread)
    → UnixSocketServer.handleConnection [E: id:unknown fix]
      → runExclusiveMcpForward [B: serialization, D: timeout floor]
        → StreamableHTTPClientTransport [C: SSE reconnect disabled]
          → HTTP server [A: 127.0.0.1 bind, E: request timeout disabled]
            → StreamableHTTPServerTransport
              → MCP tool handler (executePlan)
```

Without all five, fixing one layer just moves the failure to the next.

### DIFF
--- a/android/junit-runner/src/main/kotlin/dev/jasonpearson/automobile/junit/AutoMobilePlanExecutor.kt
+++ b/android/junit-runner/src/main/kotlin/dev/jasonpearson/automobile/junit/AutoMobilePlanExecutor.kt
@@ -339,7 +339,11 @@ internal object AutoMobilePlanExecutor {
       DaemonHeartbeat.registerSession(sessionUuid)
       response =
           try {
-            DaemonSocketClientManager.callTool("executePlan", JsonObject(args), options.timeoutMs)
+            DaemonSocketClientManager.callTool(
+                "executePlan",
+                JsonObject(args),
+                options.effectiveExecutePlanTimeoutMs(),
+            )
           } finally {
             DaemonHeartbeat.unregisterSession(sessionUuid)
           }
--- a/android/junit-runner/src/main/kotlin/dev/jasonpearson/automobile/junit/AutoMobilePlanTypes.kt
+++ b/android/junit-runner/src/main/kotlin/dev/jasonpearson/automobile/junit/AutoMobilePlanTypes.kt
@@ -6,14 +6,27 @@ import dev.jasonpearson.automobile.validation.ToolResponse
 import dev.jasonpearson.automobile.validation.ToolResult
 import dev.jasonpearson.automobile.validation.ToolResultEntry
 
+/**
+ * Minimum time the JUnit runner waits on the daemon for one `executePlan` (socket read + inner MCP
+ * `callTool`). Lower values caused premature inner HTTP cancellation and `Operation cancelled` on
+ * long UI plans.
+ *
+ * Keep in sync with `MIN_EXECUTE_PLAN_MCP_TIMEOUT_MS` in `src/daemon/mcpRequestTimeout.ts`.
+ */
+const val MIN_EXECUTE_PLAN_TIMEOUT_MS: Long = 600_000L
+
 /** Configuration options for AutoMobile plan execution. */
 data class AutoMobilePlanExecutionOptions(
-    val timeoutMs: Long = 30000L, // 30 second default
+    val timeoutMs: Long = MIN_EXECUTE_PLAN_TIMEOUT_MS,
     val device: String = "auto",
     val aiAssistance: Boolean = true,
     val maxRetries: Int = 0,
     val debugMode: Boolean = System.getProperty("automobile.debug", "false").toBoolean(),
 )
+
+/** Effective wait budget for a single `executePlan` daemon call (never below [MIN_EXECUTE_PLAN_TIMEOUT_MS]). */
+fun AutoMobilePlanExecutionOptions.effectiveExecutePlanTimeoutMs(): Long =
+    maxOf(timeoutMs, MIN_EXECUTE_PLAN_TIMEOUT_MS)
 
 /** Result of AutoMobile plan execution. */
 data class AutoMobilePlanExecutionResult(
--- a/android/junit-runner/src/test/kotlin/dev/jasonpearson/automobile/junit/AutoMobileTestDemo.kt
+++ b/android/junit-runner/src/test/kotlin/dev/jasonpearson/automobile/junit/AutoMobileTestDemo.kt
@@ -23,10 +23,26 @@ class AutoMobileTestDemo {
   @Test
   fun testAutoMobilePlanExecutionOptionsDefaults() {
     val options = AutoMobilePlanExecutionOptions()
-    assertEquals(30000L, options.timeoutMs)
+    assertEquals(MIN_EXECUTE_PLAN_TIMEOUT_MS, options.timeoutMs)
     assertEquals("auto", options.device)
     assertTrue(options.aiAssistance)
     assertEquals(0, options.maxRetries)
+  }
+
+  @Test
+  fun effectiveExecutePlanTimeoutNeverDropsBelowFloor() {
+    assertEquals(
+        MIN_EXECUTE_PLAN_TIMEOUT_MS,
+        AutoMobilePlanExecutionOptions().effectiveExecutePlanTimeoutMs(),
+    )
+    assertEquals(
+        MIN_EXECUTE_PLAN_TIMEOUT_MS,
+        AutoMobilePlanExecutionOptions(timeoutMs = 30_000L).effectiveExecutePlanTimeoutMs(),
+    )
+    assertEquals(
+        900_000L,
+        AutoMobilePlanExecutionOptions(timeoutMs = 900_000L).effectiveExecutePlanTimeoutMs(),
+    )
   }
 
   @Test
--- a/android/junit-runner/src/test/kotlin/dev/jasonpearson/automobile/junit/SimpleIntegrationTest.kt
+++ b/android/junit-runner/src/test/kotlin/dev/jasonpearson/automobile/junit/SimpleIntegrationTest.kt
@@ -15,7 +15,7 @@ class SimpleIntegrationTest {
   @Test
   fun testAutoMobilePlanExecutionOptionsDefaults() {
     val options = AutoMobilePlanExecutionOptions()
-    assertEquals(30000L, options.timeoutMs)
+    assertEquals(MIN_EXECUTE_PLAN_TIMEOUT_MS, options.timeoutMs)
     assertEquals("auto", options.device)
     assertTrue(options.aiAssistance)
     assertEquals(0, options.maxRetries)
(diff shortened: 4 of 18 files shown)


kaeawc (approved):
Verified locally: build, lint, and 3562 tests pass. Exercised observe, tapOn, swipeOn, launchApp, terminateApp, inputText, clearText, and pressButton on Android emulator — all working correctly.

Code review notes (non-blocking):

1. **Serialization queue scope** — `runExclusiveMcpForward` serializes all MCP forwards, so a long `executePlan` blocks subsequent calls. Understood this is intentional and addressed in #1918.

2. **HTTP timeout disabling** — disabling all three timeouts (`requestTimeout`, `headersTimeout`, `timeout`) is safe since this is loopback-only and MCP has its own timeouts via `resolveMcpRequestTimeoutMs`. Consider a brief comment noting that.

3. **Kotlin/TS timeout sync** — `MIN_EXECUTE_PLAN_MCP_TIMEOUT_MS` says "keep in sync with" `AutoMobilePlanTypes.kt` but there's no enforcement. A cross-language test or shared constant would prevent silent drift.

LGTM — solid fixes for real CI pain points, good test coverage, well-documented rationale.

PR #1918, opened 2026-04-13
Fail-Fast on Device Disconnect + Screenshot Pressure

### LINKED ISSUE #1945: Emulator crash leaves active sessions hanging until full MCP timeout
Description: When an Android emulator crashes or goes offline during test execution, the daemon's device disconnect monitor detects the disappearance but only cleans up recordings and miss counters. The session and any active executePlan calls are left intact, waiting for the full MCP request timeout (up to 10 minutes) before failing. This produces long, unexplained CI hangs where the emulator is gone but the test runner is still waiting.

Reproduction: Kill an emulator mid-test (e.g., adb emu kill during an executePlan). The test hangs for the remaining timeout duration instead of failing immediately.

Expected: When a device disconnects, the daemon should cancel all active executions for the affected session and release the session immediately, so the test fails fast with a clear cancellation error.

Impact: Emulator crashes in CI produce 5-10 minute silent hangs before failure, wasting CI time and making it hard to distinguish between slow tests and infrastructure failures.

RyeBread1026:
Closes #1945, #1946, #1947

**Base:** [PR #1917 - Daemon Transport Stability](https://github.com/kaeawc/auto-mobile/pull/1917)

## Summary

When an Android emulator crashes or goes offline, the daemon had no mechanism to detect this and cancel in-flight work. Active `executePlan` calls would hang until the full MCP request timeout (up to 10 minutes), producing long, unexplained CI failures. Additionally, the MCP forward serialization queue (introduced in PR #1917) meant queued requests burned their timeout budget while waiting, then immediately timed out when they got their turn. Finally, navigation screenshots added constant I/O pressure to the emulator, contributing to instability.

This PR addresses all three:
- Detect device disconnects and immediately cancel + release the affected session
- Account for queue wait time in MCP forward timeout budgets
- Add a `navigation-screenshots` feature flag to gate screenshot capture

### A. Device disconnect cancels session

**Problem:** The daemon's device disconnect monitor detected device disappearance but only cleaned up recordings and cleared miss counters. The session and any active executions were left intact, waiting for the full MCP request timeout.

**Fix:** After cleaning up recordings for a disconnected device, do a reverse lookup (`sessionManager.getSessionForDevice(deviceId)`) to find the session assigned to that device. If found, call `cancelAndReleaseSession(sessionId)` which cancels all active `executionTracker` executions and releases the session + device assignment. The test fails immediately with a cancellation error instead of hanging.

**New:** `SessionManager.getSessionForDevice(deviceId)` - a reverse lookup from `deviceSessionMap` (the map already existed; this just exposes the lookup).

**Risk:** Low. `cancelAndReleaseSession` is already used by the heartbeat timeout path. The only new behavior is triggering it from device disconnect. If the device comes back (e.g., transient ADB hiccup), the session will already be released - but the disconnect monitor already uses a miss-counter with threshold to avoid single-miss false positives.

**Test coverage:** `getSessionForDevice` has 3 unit tests (assigned lookup, unknown device, post-release). The full disconnect-monitor → cancel integration requires mocking `MultiPlatformDeviceManager`, video recordings, `executionTracker`, and `devicePool` - deferred to a follow-up.

**Files:** `src/daemon/daemon.ts`, `src/daemon/sessionManager.ts`, `test/daemon/sessionManager.test.ts`

---

### B. MCP forward queue timeout budget accounting

**Problem:** PR #1917 serialized MCP forwards through `runExclusiveMcpForward`. A side effect: if request A holds the lock for 5 minutes, request B's timeout budget ticks down while it's queued. By the time B gets its turn, it may have ≤0ms left and either immediately times out in `callTool` (wasting a round-trip) or starts work it can't finish.

**Fix:** Capture the total timeout budget (`resolveMcpRequestTimeoutMs(request)`) *before* entering the queue. Inside the serialized callback:
1. Calculate `remainingTimeoutMs = totalTimeoutMs - queueWaitMs`
2. If `remainingTimeoutMs <= 0`, throw immediately: `MCP forward timeout: spent Xms waiting in queue (budget was Yms)`
3. Pass `remainingTimeoutMs` directly to `handleIdeRequest` as the `timeoutMs` parameter - no redundant recomputation
4. The "Session not found" retry path also deducts elapsed time and fails fast if no budget remains rather than using an arbitrary grace period

**Why this matters:** Without budget accounting, a queued request with a 30s timeout that waited 29s in the queue would start a `callTool` with a fresh 30s timeout - incorrect. Now it gets 1s, or fails immediately if its budget is spent.

**Test:** New test case `queued request fails fast when queue wait exceeds its timeout` - a blocking request holds the lock, a second request with a 500ms timeout is queued, time advances past the budget, and the second request fails with `"waiting in queue"`.

**Risk:** Low. The fail-fast path is strictly better than the old behavior (waiting for a full timeout that can't succeed). The only observable change is that queued requests may fail *faster* than before - which is the desired outcome.

**Files:** `src/daemon/socketServer.ts`, `test/daemon/socketServerMcpForwardSerialization.test.ts`

---

### C. `navigation-screenshots` feature flag

**Problem:** The `NavigationScreenshotManager` captures a screenshot on every navigation event and hierarchy change. In CI environments with resource-constrained emulators, this adds constant I/O pressure (ADB screencap calls) that contributes to instability, especially under parallel test execution.

**Fix:** Add a `navigation-screenshots` feature flag (key: `"navigation-screenshots"`, default: `true`) that gates screenshot capture in both the Android and iOS `CtrlProxyClient` navigation callbacks. When disabled:
- Navigation *events* are still recorded (the navigation graph updates normally)
- Navigation *screenshots* are skipped (no ADB screencap calls)
- iOS telemetry navigation events are still published (previously they were gated behind screenshot availability)

**Wiring:**
- **CLI:** `--no-navigation-screenshots` flag parsed in `index.ts` → calls `featureFlagService.setFlag("navigation-screenshots", false)`
- **Daemon options:** `DaemonOptions.noNavigationScreenshots` → `daemon.ts` constructor applies `serverConfig.setNavigationScreenshotsEnabled(false)`
- **Daemon spawn:** `--no-navigation-screenshots` arg threading in `manager.ts` (both `buildDaemonCommand` and `parseDaemonArgs`)
- **Gradle/CI:** JVM system property `automobile.daemon.no.navigation.screenshots=true` or env var `AUTOMOBILE_DAEMON_NO_NAVIGATION_SCREENSHOTS=1` (read in Kotlin `DaemonSocketClient`)
- **Feature flag system:** `FeatureFlagDefinitions` + `FeatureFlagApplier` → `serverConfig.setNavigationScreenshotsEnabled(enabled)`
- **Guard points:** Two in Android `CtrlProxyClient` (hierarchy-change callback and navigation-event callback), one in iOS `CtrlProxyClient` (navigation-event handler - restructured to always record the event first, then conditionally capture the screenshot)

**Design note:** The flag follows the existing positive-capability pattern (flag represents the capability, `defaultValue: true`, CLI uses `--no-X` to disable). This is consistent with `ui-perf-mode`, `ai-recovery`, and `accessibility-auto-detect`.

**Kotlin scope:** The `DaemonSocketClient.buildDaemonCommand` change is limited to adding `withNoNavigationScreenshots`. The `withDismissKeyboardAfterInput` / `withNoUiPerfMode` command refactoring is deferred to Group 8 (JUnit Runner Ergonomics).

**Risk:** Low. Default is `true` (current behavior). Only CI environments that explicitly opt out are affected. Navigation graph data is still collected; only screenshots are gated.

**Files:** `src/features/featureFlags/FeatureFlagDefinitions.ts`, `src/features/featureFlags/FeatureFlagApplier.ts`, `src/features/observe/android/CtrlProxyClient.ts`, `src/features/observe/ios/CtrlProxyClient.ts`, `src/utils/ServerConfig.ts`, `src/index.ts`, `src/daemon/daemon.ts`, `src/daemon/manager.ts`, `src/daemon/types.ts`, `android/.../DaemonSocketClient.kt`

---

## Testing

- All existing tests pass (same 9 pre-existing CtrlProxyManager failures from `main`)
- New tests: `sessionManager.test.ts` (3 cases for `getSessionForDevice`), `socketServerMcpForwardSerialization.test.ts` (queue timeout budget test)
- Build and lint clean

## How these interact with PR #1917

PR #1917 introduced the MCP forward serialization queue. This PR builds on that:
- Change B accounts for time spent waiting in that queue
- Change A provides a second fail-fast mechanism (device loss) that bypasses the queue entirely by cancelling at the session level

```
Emulator crash
  → Device disconnect monitor (A)
    → cancelAndReleaseSession
      → executionTracker.cancelSessionExecutions
        → abort signal fires → plan fails immediately

Queued request with exhausted budget
  → runExclusiveMcpForward (B)
    → remainingTimeoutMs <= 0
      → throws "MCP forward timeout: spent Xms waiting in queue"
        → socket error response → JUnit client fails fast
```

### DIFF
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/WebSocketServer.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/WebSocketServer.kt
@@ -262,11 +262,11 @@ class WebSocketServer(
                     val connectionId = connectionCount.incrementAndGet()
                     Log.d(TAG, "Client #$connectionId connected")
 
-                    synchronized(connections) { connections.add(this) }
-
                     try {
-                      // Send initial connection message
+                      // Send connection greeting before registering for broadcasts
                       send(Frame.Text("""{"type":"connected","id":$connectionId}"""))
+
+                      synchronized(connections) { connections.add(this) }
 
                       // Listen for incoming messages
                       for (frame in incoming) {
--- a/android/junit-runner/src/main/kotlin/dev/jasonpearson/automobile/junit/DaemonSocketClient.kt
+++ b/android/junit-runner/src/main/kotlin/dev/jasonpearson/automobile/junit/DaemonSocketClient.kt
@@ -287,21 +287,45 @@ internal object DaemonSocketPaths {
       val entryPoint = File(localPath, "dist/src/index.js")
       if (entryPoint.exists()) {
         val runtime = resolveRuntimePath()
-        return listOf(runtime, entryPoint.absolutePath, "--daemon", subCommand)
+        return withNoNavigationScreenshots(
+          listOf(runtime, entryPoint.absolutePath, "--daemon", subCommand),
+        )
       }
     }
 
-    // Prefer bunx, fall back to npx, then the global binary.
     val runner = resolvePackageRunner()
-    return if (runner != null) {
-      if (runner.endsWith("npx")) {
-        listOf(runner, "-y", "@kaeawc/auto-mobile@latest", "--daemon", subCommand)
-      } else {
-        listOf(runner, "@kaeawc/auto-mobile@latest", "--daemon", subCommand)
-      }
-    } else {
-      listOf("auto-mobile", "--daemon", subCommand)
+    val command =
+        if (runner != null) {
+          if (runner.endsWith("npx")) {
+            listOf(runner, "-y", "@kaeawc/auto-mobile@latest", "--daemon", subCommand)
+          } else {
+            listOf(runner, "@kaeawc/auto-mobile@latest", "--daemon", subCommand)
+          }
+        } else {
+          listOf("auto-mobile", "--daemon", subCommand)
+        }
+
+    return withNoNavigationScreenshots(command)
+  }
+
+  /**
+   * When true, the spawned daemon passes `--no-navigation-screenshots` to disable screenshot
+   * capture on navigation events. Reduces emulator resource usage in CI where navigation graph
+   * thumbnails are not needed.
+   */
+  private fun noNavigationScreenshotsRequested(): Boolean {
+    if (SystemPropertyCache.getBoolean("automobile.daemon.no.navigation.screenshots", false)) {
+      return true
     }
+    val env = System.getenv("AUTOMOBILE_DAEMON_NO_NAVIGATION_SCREENSHOTS")?.trim()?.lowercase().orEmpty()
+    return env == "1" || env == "true" || env == "yes"
+  }
+
+  private fun withNoNavigationScreenshots(command: List<String>): List<String> {
+    if (!noNavigationScreenshotsRequested()) {
+      return command
+    }
+    return command + "--no-navigation-screenshots"
   }
 
   private fun resolveLocalProjectPath(): String? {
--- a/src/daemon/daemon.ts
+++ b/src/daemon/daemon.ts
@@ -112,6 +112,9 @@ export class Daemon {
     if (options.noUiPerfMode) {
       serverConfig.setUiPerfMode(false);
     }
+    if (options.noNavigationScreenshots) {
+      serverConfig.setNavigationScreenshotsEnabled(false);
+    }
     if (options.memPerfAudit) {
       serverConfig.setMemPerfAuditMode(true);
     }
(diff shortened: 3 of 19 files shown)


kaeawc (approved):
Verified locally: build, lint, and 3569 tests pass. Exercised observe, launchApp, tapOn, pressButton (back/home), and terminateApp on Android emulator — all working correctly on top of #1917.

LGTM.

PR #1919, opened 2026-04-14
feat(tapOn): opt-in preTapStability for Android tap reliability

### LINKED ISSUE #1948: tapOn silently taps stale coordinates when accessibility tree changes between observe and tap
Description: When observe + waitFor confirms a target element (e.g., a search result row), the accessibility tree can change before tapOn executes - due to loading overlays, list refreshes, IME dismiss, or other UI churn. The current tapOn implementation uses the coordinates from the pre-observe snapshot, so the gesture lands on stale coordinates. The tap reports transport-level "success" even though the intended element has moved or disappeared, causing the UI to not advance. The next plan step then times out.

Reproduction: Use executePlan with a flow that searches for a contact, waits for results, and taps a specific result row. The row appears in observe, but by the time tapOn fires, a loading spinner or list refresh has shifted the row. The tap lands on the wrong element or empty space; the next observe step times out.

Expected: tapOn should be able to verify that the target element is still present at stable coordinates before dispatching the gesture, and fail with a clear error if the element has moved or disappeared.

Impact: CI plans that tap dynamically loaded list items (search results, contact pickers, etc.) fail intermittently. Failures are hard to diagnose because the tap reports success.

RyeBread1026:
Closes #1948, #1949, #1950

## Summary

Android `tapOn` was flaky in CI - most visibly in cases where a search result list would silently hit stale coordinates. The row existed when `observe` + `waitFor` passed, then shifted or disappeared (loading overlay, list refresh, IME dismiss) before the tap executed. The old code used pre-observe coordinates, so the UI never advanced and the next plan step timed out.

This PR adds an **opt-in** `preTapStability` parameter to `tapOn`. When enabled, it refreshes the accessibility hierarchy and re-finds the target with stable bounds before dispatching the gesture. Existing `tapOn` calls are unaffected.

### A. Pre-tap stable re-find (core fix)

**Problem:** Between `observe` confirming a target and `tapOn` executing, the accessibility tree can change — loading overlays, list refreshes, keyboard dismiss. The old path used cached bounds, reporting transport-level "success" even when the intended element was gone.

**Fix:** New function `resolveAndroidStableTapTargetAfterRefreshes` in `TapOnElement.ts`:
- Refreshes the hierarchy up to 8 times (32 when loading detected)
- Re-finds the target element each time
- Requires consecutive re-finds with bounds within 3px epsilon (`boundsNearlyEqual`) for churn-prone selectors
- Aborts with a descriptive error if stability never converges

**New utility:** `boundsNearlyEqual(a, b, epsilonPx)` in `src/utils/bounds.ts`

**Risk:** Low. Opt-in only via `preTapStability: true`. When enabled, it can only cause taps to *fail* that would have previously tapped the wrong thing silently. The 3px epsilon accommodates minor rendering jitter without passing through large layout shifts.

**Files:** `src/features/action/TapOnElement.ts`, `src/utils/bounds.ts`, `test/utils/bounds.test.ts`

### B. Loading-aware patience

**Problem:** When a search result list shows a loading spinner between observe and tap, the fixed 8-attempt refresh budget can expire before the list reappears.

**Fix:** `androidViewHierarchyIndicatesLikelyBlockingLoading` scans the hierarchy for common loading indicators (`progress_bar`, `ProgressBar`, `ShimmerFrameLayout`, `ContentLoadingProgressBar`, etc.) and extends pre-tap attempts to 32 when detected.

**Risk:** Low. If the heuristic falsely detects loading, the only effect is more refresh attempts (slightly slower). If it misses real loading, behavior unchanged from base 8-attempt budget.

**Files:** `src/utils/androidTransientLoading.ts`, `test/utils/androidTransientLoading.test.ts`

### C. Selector-aware stability policy

**Problem:** Requiring 2 consecutive ε-stable matches for every tap is overly strict for simple `text` or `elementId`-only taps where the target is typically static.

**Fix:** `androidPreTapConsecutiveStableMatchesRequired(options)` returns:
- **2** consecutive stable matches for churn-prone selectors: `tapClickableParent`, `siblingOfText`, `clickable`, `scrollableContainer`
- **1** for plain `text` or `elementId`-only selectors

Static buttons get a live re-find but don't need double-stability, while search-list rows and dynamically resolved clickable parents still require two consecutive stable matches.

**Risk:** None. Relaxes the guard for simple selectors while keeping it strict for the exact patterns that caused the original flake.

**Files:** `src/features/action/androidPreTapStablePolicy.ts`, `test/features/action/androidPreTapStablePolicy.test.ts`

### D. `preTapStability` tool parameter

**Problem:** Need to expose the new behavior through the MCP tool schema.

**Fix:** `preTapStability: boolean` (opt-in) on `tapOn` Zod schema, args interface, and options model.

**Risk:** None. Optional, additive parameter. Default behavior unchanged.

**Files:** `src/models/TapOnElementOptions.ts`, `src/server/interactionToolTypes.ts`, `src/server/interactionTools.ts`

## File list

| File | Change |
|------|--------|
| `src/features/action/TapOnElement.ts` | +184 — stability loop, `refreshViewHierarchy`, constants |
| `src/features/action/androidPreTapStablePolicy.ts` | New (27 lines) — selector-aware policy |
| `src/utils/androidTransientLoading.ts` | New (45 lines) — loading heuristic |
| `src/utils/bounds.ts` | +17 — `boundsNearlyEqual` |
| `src/models/TapOnElementOptions.ts` | +10 — `preTapStability` param |
| `src/server/interactionToolTypes.ts` | +1 — args interface |
| `src/server/interactionTools.ts` | +7 — Zod schema + mapping |
| `test/features/action/androidPreTapStablePolicy.test.ts` | New (67 lines) |
| `test/utils/androidTransientLoading.test.ts` | New (41 lines) |
| `test/utils/bounds.test.ts` | New (17 lines) |

**Total:** ~426 lines across 10 files (229 modified + 197 new)

## Test results

- **Build:** Clean
- **Lint:** Clean
- **Tests:** 10 Group 3 tests pass; 601 total (only pre-existing `CtrlProxyManager` failures on `main`)


kaeawc (changes requested):
CodeQL flagged an unused assignment to `viewHierarchy` in `src/features/action/TapOnElement.ts` — the value is assigned but never read. Please remove the unused variable or use it as intended.

https://github.com/kaeawc/auto-mobile/security/code-scanning/182

RyeBread1026:
> CodeQL flagged an unused assignment to `viewHierarchy` in `src/features/action/TapOnElement.ts` — the value is assigned but never read. Please remove the unused variable or use it as intended.
> 
> https://github.com/kaeawc/auto-mobile/security/code-scanning/182

Addressed last night 😎

kaeawc (approved):

PR #1956, opened 2026-04-17
executePlan Diagnostics & Video Reliability

### LINKED ISSUE #1957: No screen state captured when executePlan step fails - blind debugging in CI
Description: When an executePlan step fails, the result contains only the error message and step index. There is no snapshot of what was on screen at the time of failure. In CI, debugging requires manually cross-referencing screenshots, daemon logs, and error text to reconstruct the device state - a slow and error-prone process.

Reproduction: Run an executePlan that fails on any step (e.g., observe timeout, tapOn element not found). The step result has success: false and an error message, but no information about what was actually visible on screen.

Expected: On step failure, the executor should capture a best-effort observe snapshot and attach it to the result as a failureObservation, including the active window, visible texts, and resource IDs.

Impact: CI failures require significant manual effort to diagnose. Common issues like "app crashed to home screen" or "unexpected dialog appeared" are invisible without a separate screenshot artifact.

RyeBread1026:
Closes #1957, #1958, #1959, #1960, #1961.

## Summary

Improves `executePlan` post-mortem debugging and fixes Android video recording reliability. Four features: failure observation snapshots, opt-in per-step observe capture, segmented video for long Android plans, and a device-side screenrecord finalization fix.

### A. Failure observation on plan step errors

**Problem:** When an `executePlan` step fails, the result contains only the error message and step index. In CI, debugging requires cross-referencing screenshots, daemon logs, and error text to reconstruct what was on screen at failure time.

**Fix:** After any step failure, the executor runs a best-effort `observe` call and attaches a `failureObservation` to the result. This includes the active window, visible texts, resource IDs, and optionally the full view hierarchy. If the failed step *was* `observe`, the observation is extracted from the failed response directly (no extra call). If the observe itself fails (device gone, timeout), it degrades gracefully to `{ observeError: "..." }`.

**Design choice:** The failure observation runs without an abort signal - even if the failure was caused by abort/cancellation, the snapshot still attempts to capture screen state. The outer `try/catch` ensures it never throws.

**Risk:** Low. One extra observe call only on the failure path. No impact on happy-path execution. Graceful degradation on all error paths.

**Files:** `src/utils/plan/PlanExecutor.ts`, `src/models/FailureObservation.ts` (new), `src/utils/plan/summarizeFailureObservation.ts` (new), `src/models/ExecutePlanResult.ts`, `src/models/Plan.ts`, `src/models/index.ts`, `src/server/planTools.ts`

---

### B. Opt-in `captureObserveSteps` for CI debug

**Problem:** When diagnosing flaky test plans, you often need to see what the screen looked like at each `observe` step, not just at the failure point.

**Fix:** A new `captureObserveSteps` parameter on `executePlan` (`"summary"` or `"full"`) attaches each successful `observe` step's snapshot to `debug.steps[n].details.stepObservation`. In `summary` mode, the full view hierarchy is stripped to keep payloads small. In `full` mode, everything is included.

**Wiring:** The Kotlin JUnit runner reads system property `automobile.junit.executePlan.captureObserveSteps` (or env `AUTOMOBILE_EXECUTE_PLAN_CAPTURE_OBSERVE_STEPS`) and passes it to the daemon. Multi-device plans log a warning and ignore the flag (parallel tracks don't emit unified debug steps).

**Gating:** The `debug` object (step-level timing, params, status) is only included in the MCP tool response when `captureObserveSteps` is set. This avoids inflating the response payload for callers who didn't opt in.

**Also in this commit:** `appendExecutePlanCleanupArgs` - Kotlin-side wiring for `cleanupAppId` / `cleanupClearAppData` system properties, allowing CI to configure post-plan app cleanup without changing test code. The schema fields already existed on `main`; this just teaches the runner to send them.

**Risk:** None for existing callers. Strictly opt-in. No new device calls on the happy path.

**Files:** `src/utils/plan/PlanExecutor.ts`, `src/server/planTools.ts`, `src/models/ExecutePlanResult.ts`, `src/utils/plan/summarizeFailureObservation.ts`, `android/.../AutoMobilePlanExecutor.kt`

---

### B2. `tapDebug` diagnostics in `debug.steps`

**Problem:** When `tapOn` returns diagnostic info about what happened during the tap (coordinates used, injection method, clickable-parent resolution), that data is lost - `executePlan` only surfaces `success`/`error` from tool results.

**Fix:** `mergeToolDiagnosticsIntoStepDetails` copies `tapDebug` from `tapOn` tool results into `debug.steps[n].details.tapDebug` on both success and failure paths. Unwraps `structuredContent` envelope when present (since `createStructuredToolResponse` nests the payload there).

**Risk:** None. Only fires for `tapOn` steps. No extra device calls. Forward-compatible - works today if any `tapOn` handler returns `tapDebug`, and ready for when `TapOnElement` adds emission.

**Files:** `src/utils/plan/PlanExecutor.ts`, `test/plan/planExecutorTapOnDebug.test.ts` (new)

---

### C. Segmented Android video for long `executePlan` runs

**Problem:** Android `adb shell screenrecord` has a hard 180-second cap. Plans running longer than 3 minutes lose video coverage.

**Fix:** Introduces `AndroidSegmentedPlanVideoSession`, which chains multiple screenrecord sessions. The session hooks into `PlanExecutionOptions.onBeforePlanStep` and checks elapsed time before each step. When approaching the 180s cap (rotates at 170s), it stops the current segment and starts a new one. On finalize, all segment file paths and recording IDs are returned.

**Wiring:** `planTools.ts` now branches on platform: Android gets the segmented session, iOS keeps the existing single-recording path. The segmented session's `onBeforePlanStep` hook is wired into `PlanExecutionOptions` alongside `captureObserveSteps`.

**Default behavior change:** All Android `executePlan` runs now use segmented video. This is transparent — short plans get one segment, long plans get multiple. The result now includes `videoFilePaths[]` and `videoRecordingIds[]` instead of a single path.

**Known rough edge:** Each segment lands in its own subdirectory (`archiveRoot/<recordingId>/`) because `VideoRecorderService` creates a per-recording folder. Ideally all segments would be siblings. A TODO comment documents this for future cleanup.

**Risk:** Low. The segmented session gracefully handles start/stop failures with `logger.warn` and continues execution. The `onBeforePlanStep` hook is only wired for Android. iOS path is unchanged.

**Files:** `src/server/androidSegmentedPlanVideoSession.ts` (new), `src/features/video/androidScreenrecord.ts` (new), `src/server/planTools.ts`, `src/utils/plan/PlanExecutor.ts`, `src/models/ExecutePlanResult.ts`

---

### D. Android screenrecord finalization fix

**Problem:** Stopping `adb shell screenrecord` by sending SIGINT to the *host* `adb` process can cause ADB to drop the session before the device-side process writes the MP4 moov atom. This produces tiny/corrupt video files showing a single frozen frame.

**Fix:** Send `pkill -2 screenrecord` on the device first to let the device-side process finalize the file, then wait for the host `adb` process to exit. Falls back to host-side SIGINT (then SIGKILL) if the device-side signal fails. Added an optional ffmpeg pipe escape hatch via `AUTOMOBILE_ANDROID_VIDEO_USE_FFMPEG_PIPE=1`.

**Note:** `pkill -2 screenrecord` signals all screenrecord processes on the device. This is fine for single-recording usage but would interfere with concurrent recordings on the same device.

**Risk:** Low. Strictly more robust shutdown sequence with graceful fallback at every stage. Only affects Android video stop path.

**Files:** `src/features/video/PlatformVideoCaptureBackend.ts`, `src/features/video/HybridVideoCaptureBackend.ts`

---

## Testing

- 14 new/updated tests across 5 files, all passing
- `summarizeFailureObservation.test.ts` - element extraction, empty elements
- `planExecutorCaptureObserveSteps.test.ts` - summary vs full mode, non-observe steps excluded
- `planExecutorOnBeforePlanStep.test.ts` - hook invocation per step, startStep offset
- `planExecutorTapOnDebug.test.ts` - tapDebug copied on success and failure paths
- `androidSegmentedPlanVideoSession.test.ts` - finalize, rotation after elapsed time
- `HybridVideoCaptureBackend.test.ts` - ffmpeg pipe env var routing
- Build passes cleanly, no lint errors

## Stats

20 files changed, +1,371 / -52 lines


kaeawc (approved):
I think this is fine for now. Overall I want to expose view hierarchy history, network calls, etc to the test runners and therefore this implementation will eventually be deprecated. Think its good to try this for now and learn from it.

PR #2078, opened 2026-05-06
fix: improve tap reliability and hierarchy delivery for Android

### LINKED ISSUE #2079: WebSocket timeout cascade causes daemon blindness after slow hierarchy extraction
After v0.0.24's accessibility flag changes increased hierarchy tree size and extraction time, the daemon's `waitForFreshData` call frequently exceeds its 100ms timeout. Each timeout triggers a 5-second cooldown during which all subsequent hierarchy requests skip the WebSocket wait entirely and return stale cached data. Under load, this cascades - the daemon can be blind to UI changes for extended periods, causing tapOn failures. The event-driven broadcast path (async via SharedFlow) is also unreliable under the increased load from larger hierarchies, which compounds the problem.

RyeBread1026:
Closes #2079, #2080, #2081

Addresses three failure modes introduced by accessibility flag changes in 0.0.24 that caused intermittent tapOn failures:

1. WebSocket timeout cascade (#2079): reduce cooldown from 5s to 500ms, use synchronous broadcasts for explicit hierarchy requests, increase default search window from 500ms to 1500ms.

2. ADB pipe congestion (#2080): throttle event-driven hierarchy broadcasts to max once per 250ms, add off-screen element bounds check to reject bad getBoundsInScreen data.

3. Ghost taps (#2081): use dispatchGesture instead of adb shell input tap to reduce base ghost-tap rate, add retryIfNoChange post-tap verification that detects unchanged hierarchy and retries once. Add ensureTap convenience flag combining preTapStability + retryIfNoChange.

### DIFF
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
@@ -4,7 +4,6 @@ import android.accessibilityservice.AccessibilityService
 import android.accessibilityservice.AccessibilityServiceInfo
 import android.accessibilityservice.GestureDescription
 import android.annotation.SuppressLint
-import android.app.ActivityManager
 import android.app.admin.DevicePolicyManager
 import android.content.BroadcastReceiver
 import android.content.ClipData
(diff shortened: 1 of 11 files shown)


android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
```diff
@@ -735,15 +742,22 @@ class CtrlProxy : AccessibilityService() {
                         "Hierarchy changed (hash=${result.hash}, extraction=${result.extractionTimeMs}ms)",
                     )
                     writeHierarchyToFile(result.hierarchy)
-                    broadcastHierarchyUpdate(result.hierarchy)
+                    throttledBroadcast(
```
kaeawc: Definitely need tests for this change

android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
```diff
@@ -1722,6 +1744,25 @@ class CtrlProxy : AccessibilityService() {
...
+   * Rate-limit event-driven hierarchy broadcasts so unsolicited pushes don't saturate the
+   * ADB port-forwarding pipe.
+   */
+  private suspend fun throttledBroadcast(
+      hierarchy: ViewHierarchy,
+      lastBroadcastMs: Long,
+      minIntervalMs: Long,
+      updateTimestamp: (Long) -> Unit,
+  ) {
+    val now = System.currentTimeMillis()
```
kaeawc: Should be using time provider so tests can use fake here

src/features/action/TapOnElement.ts
```diff
@@ -504,6 +518,17 @@ export class TapOnElement extends BaseVisualChange {
     let element = selection.element;
     let containerFoundEver = initialSearch.containerFound;
 
+    if (element && this.isElementCenterOffScreen(element, observeResult.screenSize)) {
```
kaeawc: Seems like this could be refactored into the while loop

kaeawc (approved):
Good changes, some changes needed before merge, otherwise approved

android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
RyeBread1026: Just addressed conflicts from main - looks like this is already addressed with `BroadcastThrottler`.

android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
RyeBread1026: Same here - should be addressed by the rebase.

PR #2115, opened 2026-05-14
fix: SSE keepalive + disconnect monitor hardening

### LINKED ISSUE #2097: Stability failures with remote emulators and long-running tool calls
A few related stability issues cause failures during remote emulator usage and long-running tool execution:

**SSE stream dies during long tool calls**: During `executePlan` runs (6-10 minutes), the Streamable HTTP response stream goes idle. The client-side `fetch()` stream silently dies - the server writes the result to a dead pipe and the client times out, even though the tool completed successfully.

**False device disconnect on remote emulators**: The disconnect monitor polls `getBootedDevices` every 5s with a miss threshold of 2. Remote emulators with intermittently slow ADB cause transient empty responses that don't indicate actual disconnection. When ADB is unreachable (returns 0 devices but devices are tracked), every tracked device gets a miss - causing mass false disconnection.

**Slow ADB calls amplify instability**: `adb devices` times out at 5s, which is too short for remote emulators where round-trips can exceed 10s. `getprop ro.product.model` calls are uncached and were observed taking 27s under ADB contention, blocking other operations. Progress heartbeat errors during long tool calls are silently swallowed, masking delivery failures.

**Accessibility service setup fails on transient errors**: `ensureAccessibilityServiceReady` makes a single attempt to set up the accessibility service. Transient errors like "Operation aborted" or "Command timed out" cause immediate failure, requiring the entire session to be re-established.

Expected behavior: The system should be resilient to transient ADB failures, keep response streams alive during long operations, cache expensive ADB queries, and retry recoverable errors before declaring failure.

RyeBread1026:
Closes #2097

- Add SSE keepalive timer (30s) to Streamable HTTP handler to prevent response stream from going idle during long plan execution
- Add progress heartbeat to executePlan tool with error logging
- Increase disconnect miss threshold from 2 to 3
- Skip miss counting when ADB returns 0 booted devices (ADB unreachable)
- Add miss count logging for disconnect monitor
- Cache ro.product.model results in AndroidEmulatorClient
- Increase ADB device list timeout from 5s to 10s

### DIFF
--- a/src/daemon/daemon.ts
+++ b/src/daemon/daemon.ts
@@ -510,8 +510,6 @@ export class Daemon {
           ? defaultTimer.setInterval(() => {
             if (res.headersSent && !res.writableEnded && !res.destroyed) {
               res.write(":keepalive\n\n");
-            } else if (res.writableEnded || res.destroyed) {
-              defaultTimer.clearInterval(keepaliveTimer!);
             }
           }, 30_000)
           : undefined;
@@ -826,7 +824,7 @@ export class Daemon {
           }
           const misses = (this.deviceDisconnectMisses.get(deviceId) ?? 0) + 1;
           this.deviceDisconnectMisses.set(deviceId, misses);
-          logger.warn(
+          logger.info(
             `[DisconnectMonitor] Device ${deviceId} not in booted list (miss ${misses}/${DEVICE_DISCONNECT_MISS_THRESHOLD}, booted=${bootedDeviceIds.size})`
           );
           if (misses < DEVICE_DISCONNECT_MISS_THRESHOLD) {
--- a/src/daemon/daemon.ts
+++ b/src/daemon/daemon.ts
@@ -49,7 +49,7 @@ import { FeatureFlagService } from "../features/featureFlags/FeatureFlagService"
 import { serverConfig } from "../utils/ServerConfig";
 
 const DEVICE_DISCONNECT_POLL_INTERVAL_MS = 5000;
-const DEVICE_DISCONNECT_MISS_THRESHOLD = 2;
+const DEVICE_DISCONNECT_MISS_THRESHOLD = 3;
 
 /**
  * Main daemon process
@@ -501,12 +501,35 @@ export class Daemon {
           return;
         }
 
+        // SSE keepalive: prevent the fetch() response stream from going idle
+        // during long-running tool calls (e.g. executePlan at ~6-10 min).
+        // Without traffic the client-side stream silently dies; the server
+        // writes the result to a dead pipe and the client eventually times out.
+        // SSE comment lines (`:`) are ignored by EventSourceParserStream.
+        const keepaliveTimer = req.method === "POST"
+          ? defaultTimer.setInterval(() => {
+            if (res.headersSent && !res.writableEnded && !res.destroyed) {
+              res.write(":keepalive\n\n");
+            } else if (res.writableEnded || res.destroyed) {
+              defaultTimer.clearInterval(keepaliveTimer!);
+            }
+          }, 30_000)
+          : undefined;
+
+        const clearKeepalive = () => {
+          if (keepaliveTimer) {defaultTimer.clearInterval(keepaliveTimer);}
+        };
+        res.on("close", clearKeepalive);
+        res.on("finish", clearKeepalive);
+
         // Let the transport handle the request
         try {
           await streamableTransport.handleRequest(req, res, parsedBody);
         } catch (error) {
           logger.error("Streamable HTTP request handling failed:", error);
           sendJsonRpcError("Server error", error);
+        } finally {
+          clearKeepalive();
         }
       } else {
         // 404 for unknown paths
@@ -785,6 +808,17 @@ export class Daemon {
           candidateDeviceIds.add(deviceId);
         }
 
+        // When getBootedDevices returns empty but we have tracked devices, ADB
+        // itself likely failed (timeout, connection error). Counting misses in
+        // this state causes false disconnect detections on remote emulators
+        // where ADB can be intermittently slow.
+        if (bootedDeviceIds.size === 0 && candidateDeviceIds.size > 0) {
+          logger.warn(
+            `[DisconnectMonitor] ADB returned 0 devices but ${candidateDeviceIds.size} tracked — skipping miss count (ADB likely unreachable)`
+          );
+          return;
+        }
+
         for (const deviceId of candidateDeviceIds) {
           if (bootedDeviceIds.has(deviceId)) {
             this.deviceDisconnectMisses.delete(deviceId);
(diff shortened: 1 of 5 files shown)


src/daemon/daemon.ts
```diff
@@ -489,12 +489,35 @@ export class Daemon {
...
+        // during long-running tool calls (e.g. executePlan at ~6-10 min).
+        // Without traffic the client-side stream silently dies; the server
+        // writes the result to a dead pipe and the client eventually times out.
+        // SSE comment lines (`:`) are ignored by EventSourceParserStream.
+        const keepaliveTimer = req.method === "POST"
+          ? defaultTimer.setInterval(() => {
+            if (res.headersSent && !res.writableEnded && !res.destroyed) {
+              res.write(":keepalive\n\n");
+            } else if (res.writableEnded || res.destroyed) {
+              defaultTimer.clearInterval(keepaliveTimer!);
```
kaeawc: Can remove this logic branch, `res.on('close'|'finish')` and the outer `finally` already handle it, dropping it lets us remove the !

src/daemon/daemon.ts
```diff
@@ -768,13 +791,27 @@ export class Daemon {
...
+
         for (const deviceId of candidateDeviceIds) {
           if (bootedDeviceIds.has(deviceId)) {
             this.deviceDisconnectMisses.delete(deviceId);
             continue;
           }
           const misses = (this.deviceDisconnectMisses.get(deviceId) ?? 0) + 1;
           this.deviceDisconnectMisses.set(deviceId, misses);
+          logger.warn(
+            `[DisconnectMonitor] Device ${deviceId} not in booted list (miss ${misses}/${DEVICE_DISCONNECT_MISS_THRESHOLD}, booted=${bootedDeviceIds.size})`
```
kaeawc, line 829: Suggest info, or only warn on the final miss instead of warn for every poll - but this is mainly serving your use case so I leave that up to you

src/utils/android-cmdline-tools/AndroidEmulatorClient.ts
```diff
@@ -11,6 +11,8 @@ import type { AvdConfigReader } from "./AvdConfigReader";
 import { FileAvdConfigReader } from "./AvdConfigReader";
 import type { FormFactor } from "../../models/DeviceMatchCriteria";
 
+const modelNameCache = new Map<string, string>();
```
kaeawc: Could and probably should be private rather than public module level so we don't mutate it later for tests - that would definitely cause flakiness

src/server/planTools.ts
```diff
@@ -157,8 +157,26 @@ const executePlanTool = async (device: BootedDevice, params: {
...
   const startTime = defaultTimer.now();
+
+  // Send periodic progress heartbeats to keep the SSE stream alive during
+  // long-running plan execution. Without these, the Streamable HTTP response
+  // stream can go idle and be silently dropped, causing the MCP client to
+  // time out even though the tool completed successfully.
+  let progressHeartbeatCount = 0;
+  const progressHeartbeat = progress
+    ? defaultTimer.setInterval(() => {
+      progressHeartbeatCount++;
```
kaeawc, line 172: Make sure cancelled tools don't fire one stray heartbeat before finally runs

src/utils/android-cmdline-tools/AdbClient.ts
```diff
@@ -77,7 +77,7 @@ export class AdbClient implements AdbExecutor {
   private readonly retryExecutor: RetryExecutor;
   private readonly timer: Timer;
 
-  private static readonly DEVICE_LIST_TIMEOUT_MS = 5000;
+  private static readonly DEVICE_LIST_TIMEOUT_MS = 10000;
```
kaeawc, line 80: Likely want to plan some simulation testing to tighten the worst failure case more than 30 seconds. I'm going to think more about what that might look like and how to perhaps setup specific environment configurations that optimize where AutoMobile is used in high latency use cases

src/daemon/daemon.ts
```diff
@@ -489,12 +489,35 @@ export class Daemon {
...
+        // writes the result to a dead pipe and the client eventually times out.
+        // SSE comment lines (`:`) are ignored by EventSourceParserStream.
+        const keepaliveTimer = req.method === "POST"
+          ? defaultTimer.setInterval(() => {
+            if (res.headersSent && !res.writableEnded && !res.destroyed) {
+              res.write(":keepalive\n\n");
+            } else if (res.writableEnded || res.destroyed) {
+              defaultTimer.clearInterval(keepaliveTimer!);
+            }
+          }, 30_000)
```
kaeawc: Should probably be a named constant

src/daemon/daemon.ts
```diff
@@ -489,12 +489,35 @@ export class Daemon {
           return;
         }
 
+        // SSE keepalive: prevent the fetch() response stream from going idle
+        // during long-running tool calls (e.g. executePlan at ~6-10 min).
+        // Without traffic the client-side stream silently dies; the server
+        // writes the result to a dead pipe and the client eventually times out.
+        // SSE comment lines (`:`) are ignored by EventSourceParserStream.
+        const keepaliveTimer = req.method === "POST"
+          ? defaultTimer.setInterval(() => {
```
kaeawc, line 511: Should have some unit tests that use FakeTimer to execute these new behaviors

kaeawc (changes requested):

src/daemon/daemon.ts
RyeBread1026: Added.

src/daemon/daemon.ts
RyeBread1026: Updated to `SSE_KEEPALIVE_INTERVAL_MS`.

src/daemon/daemon.ts
RyeBread1026: That makes sense for our use case - updated to info and warn on the final miss.

src/daemon/daemon.ts
RyeBread1026: Right, on, removed the `else if` branch.

src/utils/android-cmdline-tools/AndroidEmulatorClient.ts
RyeBread1026: Moved from module-level `const` to private `modelNameCache` instance member on `AndroidEmulatorClient`.

src/server/planTools.ts
RyeBread1026: Added a `progressHeartbeatStopped` boolean flag. The interval callback checks it before firing, and `clearProgressHeartbeat()` sets it to `true` before clearing the interval. This prevents any stray heartbeat from sneaking through between cancellation and finally.

src/utils/android-cmdline-tools/AdbClient.ts
RyeBread1026: Makes sense - definitely found a lot of different needs in our CI environment. Let me know how I can help with that.

This bump was the quick fix for remote emulators where adb devices was intermittently taking 7-8s. Marking this comment as resolved for now - let me know if there's something we want to do in this PR.

kaeawc (approved):

Issue #2099, opened 2026-05-11
No way to configure accessibility service flags at runtime


RyeBread1026:
The CtrlProxy accessibility service always runs with `FLAG_INCLUDE_NOT_IMPORTANT_VIEWS`, `FLAG_REPORT_VIEW_IDS`, and `FLAG_RETRIEVE_INTERACTIVE_WINDOWS` enabled. There's no way to disable these at runtime for debugging or to reduce hierarchy noise in specific scenarios.

**Expected behavior**: CLI options to toggle individual accessibility flags, synced to the CtrlProxy service on connection.

kaeawc:
Landed in d8f5a1f8 / 52e5ba8c — runtime accessibility service flags configurable via daemon CLI (--no-a11y-*) and synced to CtrlProxy via WebSocket set_accessibility_flags.

Issue #2098, opened 2026-05-11
Reduce ADB overhead during plan execution


RyeBread1026:
When `observe` is called with `waitFor` during plan execution, each polling iteration collects screenshots and the full back stack via ADB. On remote emulators (emulator.wtf), ADB round-trips are significantly slower than local - `adb devices` was observed hanging for 13s, screenshots taking 25s, and orphaned screenshot jobs running for 264+ seconds. This saturates ADB and causes cascading `waitFor` timeouts.

- Screenshots and back stack auto-skipped when `isPlanExecutionActive()` is true
- Orphaned screenshot jobs cancelled via `ScreenshotJobTracker.cancelJob` in PlanExecutor
- Disconnect monitor paused during plan execution
- `noWaitForPollingOverhead` CLI flag
- `skipBackStack` parameter on `observe` tool for manual control

Expected behavior: Plan execution should minimize ADB traffic to only what's needed for UI polling, preventing contention that causes test timeouts.

kaeawc:
Closed by #2116 (merge commit 52e5ba8c, merged 2026-05-16). All bullets verified on main:
- Screenshots/back stack auto-skipped during plan execution
- Orphaned screenshot jobs cancelled via `ScreenshotJobTracker.cancelJob` in PlanExecutor
- Disconnect monitor paused during plan execution
- `--no-waitfor-polling-overhead` CLI flag
- `skipBackStack` parameter on `observe`

PR #2116, opened 2026-05-14
feat: ADB contention reduction, tap resilience, and accessibility config

### LINKED ISSUE #2096: Tapping notifications inside collapsed Android groups opens app generically instead of deep-linking
When an Android app posts 2+ notifications, the system collapses them into a group header. Calling `systemTray` tap on a notification inside a collapsed group opens the app generically instead of triggering the notification's specific deep-link intent.

Root cause: Android routes taps on collapsed group children to the group header rather than the individual notification. The notification's tap target is not actionable until the group is visually expanded.

Approaches that failed:

- Tapping text nodes directly inside collapsed groups
- Using `adb shell input swipe` to expand the group
- Tapping the expand button without first identifying the correct group node
- Matching text without expansion (tap target not clickable)

Additional blocker: `ViewHierarchyExtractor.kt` filters out `isVisibleToUser=false` nodes, but Android marks collapsed group children as not visible even though they're present in the shade. This prevents text matching entirely.

Expected behavior: `systemTray` tap should automatically expand collapsed notification groups and tap the specific notification to trigger its deep-link intent.

RyeBread1026:
Closes #2096, #2097, #2098, #2099, #2118

- Skip screenshots and back stack fetches during plan execution to eliminate ADB contention causing waitFor timeouts in CI
- Cancel orphaned screenshot jobs in PlanExecutor
- Pause disconnect monitor during plan execution
- Add skipBackStack parameter to observe tool
- Add noWaitForPollingOverhead daemon flag
- Add retryIfNoChange and ensureTap to tapOn for ghost tap detection
- Increase TapOnElement no-hierarchy resilience with longer backoff
- Add accessibility setup retry in ToolExecutionContext (max 2 attempts)
- Add configurable accessibility flags (noA11yIncludeNotImportantViews, noA11yReportViewIds, noA11yRetrieveInteractiveWindows)
- Sync a11y flags to device via CtrlProxy WebSocket
- Bypass isVisibleToUser filter for system UI nodes in ViewHierarchyExtractor (enables collapsed notification group matching)
- Fix occlusion filter and ctrlProxyIncomplete false positives
- Add 276 lines of TapOnElement tests (preTapStability + retryIfNoChange)


test/features/action/TapOnElement.retryIfNoChange.test.ts
```diff
@@ -0,0 +1,183 @@
...
+      false,
+      { width: 1080, height: 1920 },
+    );
+
+    expect(sleepDurations).toEqual([150, 100]);
+  });
+});
+
+describe("ensureTap flag expansion", () => {
+  test("ensureTap enables both preTapStability and retryIfNoChange", () => {
```
kaeawc: This would still pass if the real expansion at TapOnElement.ts were deleted - the test is copying some of the implementation its meant to test. Replace with a test with Fakes and still does the assertions needed here.

test/features/action/TapOnElement.retryIfNoChange.test.ts
```diff
@@ -0,0 +1,183 @@
...
+      Object.assign(result, { preTapStability: true, retryIfNoChange: true });
+    }
+
+    expect(result.preTapStability).toBe(true);
+    expect(result.retryIfNoChange).toBe(true);
+
+    expect(tap).toBeDefined();
+  });
+
+  test("ensureTap does not override explicitly set flags", () => {
```
kaeawc: Same problem here

src/features/observe/ObserveScreen.ts
```diff
@@ -1359,13 +1366,15 @@ export class RealObserveScreen implements ObserveScreen {
 
       // Collect all data components with parallelization
       // Note: collectAllData tracks its phases internally, so we just call it directly
-      await this.collectAllData(result, queryOptions, perf, skipWaitForFresh, minTimestamp, signal);
+      const planActive = serverConfig.isPlanExecutionActive();
```
kaeawc: I think the config should be `skipBackstack`. I get why we want to do that when plan execution is active - at least some of the time, but back stack might be necessary for some tests. I also might be able to get the backstack another way that is way more performant, and then we would be able to revert this once I figure it out.

src/features/observe/interfaces/ObserveScreen.ts
```diff
@@ -21,7 +21,9 @@ export interface ObserveScreen {
     perf?: PerformanceTracker,
     skipWaitForFresh?: boolean,
     minTimestamp?: number,
-    signal?: AbortSignal
+    signal?: AbortSignal,
+    skipBackStack?: boolean,
```
kaeawc: Should probably refactor this method signature into its own object since its getting so long

src/utils/ServerConfig.ts
```diff
@@ -155,6 +160,54 @@ class ServerConfig {
...
+
+  setPlanExecutionActive(active: boolean): void {
+    this._planExecutionActive = active;
+  }
+
+  isPlanExecutionActive(): boolean {
+    return this._planExecutionActive;
+  }
+
+  getAccessibilityFlagsConfig(): { includeNotImportantViews: boolean; reportViewIds: boolean; retrieveInteractiveWindows: boolean } {
```
kaeawc: I'd pick one shape for getters instead of both getAccessibilityFlagsConfig and the individual ones

src/index.ts
```diff
@@ -55,8 +55,12 @@ function parseArgs(): {
   dismissKeyboardAfterInput: boolean;
   mcpRecording: boolean;
   navigationScreenshots: boolean;
+  noWaitForPollingOverhead: boolean;
```
kaeawc, line 58: Since observeTools.ts applies this flag for all waitFor observations regardless of plan context but I think the intention was only for plan execution? Just want to check what the intent is and get that aligned

kaeawc (changes requested):
Couple changes needed on this

test/features/action/TapOnElement.retryIfNoChange.test.ts
RyeBread1026: Good call - updated.

test/features/action/TapOnElement.retryIfNoChange.test.ts
RyeBread1026: Done.

src/features/observe/ObserveScreen.ts
RyeBread1026: Makes sense - removed `isPlanExecutionActive()`

src/features/observe/interfaces/ObserveScreen.ts
RyeBread1026: Refactored into `ObserveScreenExecuteOptions`

src/utils/ServerConfig.ts
RyeBread1026: Went with the bundled shape and removed the individual getters

src/index.ts
RyeBread1026: It was intentionally global. Practically, the `--no-waitfor-polling-overhead` is intended for CI, where I expect we'd only be doing plan execution, so I don't think folks will stumble into using this by accident. If there is some need to have this optimization for direct MCP tool calls, I figure it wouldn't hurt to have this available if they want to manually start the deamon with that flag.

kaeawc (approved):

PR #2117, opened 2026-05-14
feat: system tray notification improvements

### LINKED ISSUE #2096: Tapping notifications inside collapsed Android groups opens app generically instead of deep-linking

RyeBread1026:
Closes #2096, #2120

- Auto-expand collapsed notification groups before tapping
- Reorder deadline check in waitForNotificationMatch to avoid rejecting valid matches at the boundary
- Migrate notification.timeout to awaitTimeout via PlanMigrator
- Add top-Y tiebreaker for multiple notification matches
- Migrate bare setInterval to timer.setInterval for FakeTimer compat
- Add systemTray tests and PlanMigrator tests
- Document notification group expansion in design docs and CLAUDE.md

### DIFF
--- a/src/server/systemTrayHelpers.ts
+++ b/src/server/systemTrayHelpers.ts
@@ -12,6 +12,7 @@ import {
   Platform,
   ViewHierarchyResult
 } from "../models";
+import type { ObserveScreenExecuteOptions } from "../features/observe/interfaces/ObserveScreen";
 import { RealObserveScreen } from "../features/observe/ObserveScreen";
 import { defaultAdbClientFactory } from "../utils/android-cmdline-tools/AdbClientFactory";
 import { CtrlProxyClient as IOSCtrlProxyClient } from "../features/observe/ios";
@@ -29,15 +30,7 @@ import { logger } from "../utils/logger";
 // ============================================================================
 
 export interface SystemTrayObserver {
-  execute(
-    queryOptions?: unknown,
-    perf?: unknown,
-    skipWaitForFresh?: boolean,
-    minTimestamp?: number,
-    signal?: AbortSignal,
-    skipBackStack?: boolean,
-    skipScreenshot?: boolean
-  ): Promise<ObserveResult>;
+  execute(options?: ObserveScreenExecuteOptions): Promise<ObserveResult>;
 }
 
 export interface SystemTrayAdb {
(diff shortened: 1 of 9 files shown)


src/server/systemTrayHelpers.ts
```diff
@@ -491,29 +500,131 @@ const nodeHasNotificationRowHint = (node: any): boolean => {
...
+): Promise<boolean> => {
+  const groupNode = match.candidate.groupNode;
+  if (!groupNode) {
+    return false;
+  }
+
+  const expandButton = findExpandButtonInGroup(groupNode);
+  if (!expandButton) {
+    logger.warn("[systemTray] Collapsed group detected but no expand button found in group node");
+    return false;
```
kaeawc: I'd suggest throwing an `ActionableError` here, or at minimum surfacing this state in the result since this will cause the child to be tapped which is what this PR is trying to fix

src/server/systemTrayHelpers.ts
```diff
@@ -33,7 +34,9 @@ export interface SystemTrayObserver {
     perf?: unknown,
     skipWaitForFresh?: boolean,
     minTimestamp?: number,
-    signal?: AbortSignal
+    signal?: AbortSignal,
+    skipBackStack?: boolean,
```
kaeawc: Consider rebasing this PR against latest main and refactoring to use ObserveScreenExecuteOptions in more places

kaeawc (changes requested):

src/server/systemTrayHelpers.ts
RyeBread1026: Done - `expandNotificationGroup` now throws `ActionableError("Collapsed notification group detected but no expand button found. Cannot tap individual notifications inside a collapsed group.")`.

src/server/systemTrayHelpers.ts
RyeBread1026: Done.

kaeawc (approved):

PR #2158, opened 2026-05-18
refactor(ctrl-proxy): replace iOS clearText char-by-char deletion with Cmd+A + Delete

kaeawc:
## Summary
- Replace O(n) character-by-character deletion loop (up to 20×50 = 1000 keystrokes) with O(1) Cmd+A + Delete (2 key events) in iOS `clearText` and `setText`
- Remove ~190 lines of dead code: `clearViaDeletes`, burst-size constants, and `resolveFocusedTextElement`
- Add regression tests verifying iOS sends `request_clear_text` (not `request_set_text`), forwards `resourceId`, and handles disconnection

## Test plan
- [x] TypeScript unit tests pass (17 tests, 140ms)
- [ ] Manual verification on iOS simulator: clearText on populated field clears instantly
- [ ] Manual verification: clearText on empty field is a no-op (no crash/error)
- [ ] Manual verification: setText replaces existing content correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ios/control-proxy/Sources/CtrlProxy/GesturePerformer.swift
+++ b/ios/control-proxy/Sources/CtrlProxy/GesturePerformer.swift
@@ -100,8 +100,7 @@ public class GesturePerformer: GesturePerforming {
         /// Heuristic for treating an `.other`-typed snapshot as a text-input
         /// wrapper. Text fields expose a non-empty `value` (current content)
         /// or a `placeholderValue` (hint text); plain buttons / containers
-        /// do not. False positives would cause `clearViaDeletes` to send
-        /// delete keys to a non-text element.
+        /// do not.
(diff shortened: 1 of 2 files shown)


kaeawc:
Addressed the RN compat feedback — force-pushed with element-type detection:

- **Native UIKit inputs** (`.textField`, `.textView`, `.secureTextField`, `.searchField`): Cmd+A + Delete (O(1))
- **React Native wrappers** (`.other`): per-character delete loop via `element.typeText` so `onChangeText` fires

Detection: `element.elementType == .other` → RN path, everything else → native path. The `clearViaDeletes` loop is preserved for RN (simplified from 80→15 lines, same behavior).

The bare `clearText()` path (no resourceId) still uses Cmd+A + Delete since it can't resolve a specific element — same limitation as before.

Issue #2286, opened 2026-05-21
Only 2 of 7 daemon CLI flags log a startup confirmation


RyeBread1026:
The daemon supports 7 CLI flags relevant to CI tuning, but only 2 emit a startup log line confirming they took effect. The other 5 are silently applied, making it hard to verify from CI logs that the flags are wired through correctly. This bit us - we spent time investigating "why are screenshots still happening?" before discovering that the flags themselves *were* working and the issue was a code path not covered by any flag (see Issue 3).

Issue #2285, opened 2026-05-21
CtrlProxy `getLatestHierarchy` 100ms fresh-data wait too short for contended environments


RyeBread1026:
`CtrlProxyHierarchy.getLatestHierarchy` defaults to a 100ms wait for fresh WebSocket-pushed hierarchy data. When the ADB pipe is busy (e.g. concurrent screenshots, dumpsys, or a transitioning emulator), normal push delivery routinely exceeds 100ms and the call silently falls back to stale cached data. Observed 29 fresh-data timeouts (1:1 with stale-cache returns) in a single 7-minute CI run - a 31% stale-data rate.

Issue #2283, opened 2026-05-21
`retryIfNoChange` fires false-positive retries during activity transitions


RyeBread1026:
The ghost-tap mitigation added in PR #2078 races against CtrlProxy WebSocket push delivery and fires destructive false-positive retries whenever a tap triggers an activity transition. The retry lands on the transitioning destination screen (or an in-flight back affordance) and bounces the user off the intended flow.

Issue #2282, opened 2026-05-21
`retryTapIfNoChange` throws `ReferenceError: isTalkBackEnabled is not defined`


RyeBread1026:
Any `tapOn` with `retryIfNoChange: true` (or `ensureTap: true`, which expands to include `retryIfNoChange`) crashes with a `ReferenceError` at the moment the ghost-tap retry path tries to fire. The element is correctly located and the initial tap completes; the crash happens inside the post-tap evaluation when re-invoking the platform-specific tap implementation.

kaeawc:
Akready fixed https://github.com/kaeawc/auto-mobile/pull/2279

PR #2287, opened 2026-05-21
fix(perf): gate SelectionStateTracker screenshots on --no-ui-perf-mode

### LINKED ISSUE #2284: Issue 3: `SelectionStateTracker` ignores `--no-ui-perf-mode` and dominates CI ADB overhead
Every `tapOn` on a clickable element with an identifier captures two full screenshots (before + after) for visual diff selection-state detection, regardless of CI perf flags. On emulator.wtf this accounts for ~60 seconds of overhead in a 7-minute run (76 screenshots @ ~800ms each), and the resulting ADB pipe pressure stalls CtrlProxy WebSocket pushes enough to false-positive ghost-tap detection (see Issue 2).

The flags `--no-navigation-screenshots` and `--no-waitfor-polling-overhead` correctly suppress two other screenshot paths, but `SelectionStateTracker` is a third path that wasn't wired up. The flag `--no-ui-perf-mode` exists and would have been a natural place to gate this, but the tracker never checked it.

kaeawc:
## Summary
- `SelectionStateTracker.prepare()` now checks `serverConfig.isUiPerfModeEnabled()` and skips before/after screenshot capture when `--no-ui-perf-mode` is set.
- Eliminates ~60s of overhead per emulator.wtf run (76 screenshots @ ~800ms) and removes the ADB pipe pressure that was stalling CtrlProxy WebSocket pushes.
- Matches the gating pattern already used by `PerformanceAudit` and `LaunchApp`.

Closes #2284.

## Test plan
- [x] `bun test test/features/navigation/SelectionStateTracker.test.ts` — new test covers `--no-ui-perf-mode` path
- [x] `bun test test/features/action/TapOnElement.visionFallback.test.ts`
- [x] `turbo run lint build`

### DIFF
--- a/src/features/navigation/SelectionStateTracker.ts
+++ b/src/features/navigation/SelectionStateTracker.ts
@@ -1,6 +1,7 @@
 import { BootedDevice, Element, ObserveResult } from "../../models";
 import { AdbClientFactory, defaultAdbClientFactory } from "../../utils/android-cmdline-tools/AdbClientFactory";
 import { logger } from "../../utils/logger";
+import { serverConfig } from "../../utils/ServerConfig";
 import { SelectedElement } from "../../utils/interfaces/NavigationGraph";
 import { TakeScreenshot } from "../observe/TakeScreenshot";
 import { SelectionDetectionContext, SelectionStateDetector, SelectionStateDetectorLike } from "./SelectionStateDetector";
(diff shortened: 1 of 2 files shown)


kaeawc:
Superseded by #2290, which lands the same SelectionStateTracker gate without an unnecessary DI seam and bundles the related fixes for #2282, #2283, #2285, #2286 that address the full stall-cascade.

PR #2289, opened 2026-05-21
fix(android): raise CtrlProxy fresh-hierarchy wait default from 100ms to 1s

### LINKED ISSUE #2285: CtrlProxy `getLatestHierarchy` 100ms fresh-data wait too short for contended environments

kaeawc:
## Summary

- Closes #2285. Raises `CtrlProxyHierarchy.getLatestHierarchy` default fresh-data wait from 100ms to 1000ms, aligning it with the existing 1s cache-freshness TTL.
- Under ADB pipe contention (concurrent screenshots, dumpsys, emulator transitions), WebSocket-pushed hierarchy data routinely lands after 100ms, causing the call to silently fall back to stale cache — observed ~31% stale-data rate in a 7-minute CI run.
- Centralized the default in `CtrlProxyHierarchy` so the `AndroidCtrlProxyClient` wrapper stays in sync automatically (was previously duplicating `100`).

## Test plan

- [x] New regression test `should return fresh data when WebSocket push arrives after 100ms under contention (regression #2285)` simulates a 300ms-delayed push and asserts `fresh: true`. Fails on old default, passes on new.
- [x] `bun test test/features/observe/CtrlProxyClient.test.ts` — 17/18 pass (1 unrelated pre-existing failure on `main`).
- [x] `bun test test/features/observe/` — 498/499 pass (same pre-existing failure).
- [x] `turbo run lint build` clean.

### DIFF
--- a/src/features/observe/android/AndroidCtrlProxyClient.ts
+++ b/src/features/observe/android/AndroidCtrlProxyClient.ts
@@ -1113,13 +1113,13 @@ export class AndroidCtrlProxyClient extends DeviceServiceClient implements Andro
 
   async getLatestHierarchy(
     waitForFresh: boolean = false,
-    timeout: number = 100,
+    timeout?: number,
     perf: PerformanceTracker = new NoOpPerformanceTracker(),
     skipWaitForFresh: boolean = false,
     minTimestamp: number = 0,
     signal?: AbortSignal
   ): Promise<AccessibilityHierarchyResponse> {
-    return this.hierarchy.getLatestHierarchy(waitForFresh, timeout, perf, skipWaitForFresh, minTimestamp, signal);
+    return this.hierarchy.getLatestHierarchy(waitForFresh, timeout!, perf, skipWaitForFresh, minTimestamp, signal);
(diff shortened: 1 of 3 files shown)


kaeawc:
Closing in favor of #2290, which addresses the root cause (ADB pipe pressure from `SelectionStateTracker` screenshots: 76 → 3 per 7-min run) in addition to the CtrlProxy timeout bump. With the underlying contention removed, the more conservative 500ms wait there is sufficient, and #2290 also handles the ghost-tap-retry cascade that a bare timeout bump could otherwise aggravate.

PR #2290, opened 2026-05-21
make tapOn + observe resilient on high-latency emulators

### LINKED ISSUE #2282: `retryTapIfNoChange` throws `ReferenceError: isTalkBackEnabled is not defined`

RyeBread1026:
Closes #2282, #2283, #2284, #2286

- **`retryTapIfNoChange` no longer false-triggers on activity transitions.** Bumped budgets (150→300ms settle, 500→1500ms refresh) and treat null post-tap hierarchy as inconclusive (bail, don't retry). Constants extracted to `POST_TAP_*_MS`.
- **`SelectionStateTracker` honors `--no-ui-perf-mode`.** Was the dominant ADB-pipe pressure source: 76 → 3 screenshots in a 7-min run.
- **`PlanMigrator`** auto-wraps legacy `tapOn { text|elementId }` under `selector: { ... }` for the 0.0.30 schema. Idempotent.
- **`index.ts`** logs 5 previously-silent CI flags at startup.

### DIFF
--- a/src/features/action/TapOnElement.ts
+++ b/src/features/action/TapOnElement.ts
@@ -73,7 +73,8 @@ interface TapOnElementDependencies {
 }
 
 /**
- * Budgets used by retryTapIfNoChange to decide whether a tap registered.
+ * Post-tap observation budgets used by retryTapIfNoChange to decide whether
+ * a tap registered.
  *
  * Activity transitions commonly take 1-2s on contended emulators (emulator.wtf),
  * and the CtrlProxy WebSocket push for the new hierarchy doesn't arrive until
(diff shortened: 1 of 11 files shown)


kaeawc:
Heads up — #2289 (a narrower CtrlProxy-only bump from 100ms → 1000ms) automerged before I realized this PR existed and addresses the same area more completely. This PR is now `DIRTY` against `main` because of it.

Your approach here is the right one — fixing the ADB pipe pressure at the source (SelectionStateTracker honoring `--no-ui-perf-mode`) plus a more measured 500ms timeout is better than my blunt 1000ms bump in isolation. Happy to rebase this for you over my change if it'd help, or feel free to do it however suits you. Sorry for the churn.

kaeawc:
Needs some test fixes

PR #2291, opened 2026-05-21
fix(action): avoid retryIfNoChange false-positives across activity transitions

### LINKED ISSUE #2283: `retryIfNoChange` fires false-positive retries during activity transitions

kaeawc:
## Summary
- `retryIfNoChange` only compared a content hash of the pre/post-tap `ViewHierarchyResult`, which races against CtrlProxy's debouncer + broadcast throttler. During an activity transition, the next pushed hierarchy can still reflect the pre-transition tree, so hashes match and a destructive retry fires on the destination screen.
- Capture `foregroundActivity`, `packageName`, and `updatedAt` alongside the hash. Before retrying: if either identifier changed → the tap registered, exit. If the post-tap snapshot is not newer than the pre-tap one → refetch once to clear the throttler/debouncer window before deciding.
- Falls back to hash-only behavior when these signals are absent (iOS, older accessibility runners) — no regression on platforms that don't populate them.

Fixes #2283.

## Test plan
- [x] `bun test test/features/action/TapOnElement.retryIfNoChange.test.ts` — 13 pass (4 existing + 5 new branches covered).
- [x] `bun test test/features/action/` — 598 pass.
- [x] `turbo run lint build` — clean.
- New tests cover: activity transition with matching hash, package transition with matching hash, stale-snapshot refetch path, retry when fresh refetch still shows no transition, fallback when signals absent.

### DIFF
--- a/src/features/action/TapOnElement.ts
+++ b/src/features/action/TapOnElement.ts
@@ -50,6 +50,13 @@ import { createTapStrategy } from "./strategies/createTapStrategy";
 
 type SearchUntilStats = NonNullable<TapOnElementResult["searchUntil"]>;
 
+interface PreTapState {
+  hash: string;
+  foregroundActivity?: string;
+  packageName?: string;
+  updatedAt?: number;
+}
+
 /**
  * Dependencies for TapOnElement that can be injected for testing.
  */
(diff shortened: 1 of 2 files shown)


kaeawc:
Superseded by #2290, which fixes #2283 at the root causes (CtrlProxyHierarchy fresh-data wait 100→500ms, SelectionStateTracker honoring `--no-ui-perf-mode` to relieve ADB-pipe pressure stalling CtrlProxy pushes, plus bigger settle/refresh budgets and a null-bail in retryTapIfNoChange) rather than patching the retry symptom.

The semantic `foregroundActivity` / `packageName` early-exit here is defense-in-depth and could come back as a small follow-up after #2290 lands if a content-identical-hash false retry is ever reproduced — not worth the merge conflict otherwise.

PR #2294, opened 2026-05-21
feat: AUTOMOBILE_CTRL_PROXY_VERSION env var to pin CtrlProxy release

### LINKED ISSUE #2293: Can't pin the on-device CtrlProxy release independently of the daemon
The daemon always installs the newest CtrlProxy from `RELEASE_CHECKSUM_REGISTRY`. There's no way to keep the on-device binary at a known-good version while still picking up daemon JS updates from `bunx`. That means a bad CtrlProxy release breaks every consumer at once, CI isn't deterministic across daemon upgrades, and bisecting on-device regressions is awkward.

Wiring is also fragile: `APK_URL` derives from `resolveLatestVersion()` while `APK_SHA256_CHECKSUM` is keyed off `LATEST_RELEASE_VERSION` literally. They agree today but can drift.

RyeBread1026:
Closes #2293

Adds `AUTOMOBILE_CTRL_PROXY_VERSION`, an opt-in env var that pins the on-device CtrlProxy APK/IPA to a specific historical release independent of the daemon JS version. Default behavior (unset) is unchanged - resolves to the newest entry in `RELEASE_CHECKSUM_REGISTRY`.

When set, both the download URL and the expected SHA256 checksum resolve to the pinned version. Unknown values throw at module load with the list of known versions, so CI typos fail loud rather than silently falling back to latest.

Also:

- Unifies URL + checksum behind `RELEASE_VERSION` (previously the URL came from `resolveLatestVersion()` and the checksum was keyed off `LATEST_RELEASE_VERSION` literally - they agree today but could drift).
- iOS bundle metadata now records the actually-installed version via `resolveAssetVersion(IOS_CTRL_PROXY_RELEASE_VERSION)` instead of always writing `resolveLatestVersion()`. Pre-existing bug the pin would have made visible.
- Daemon startup log is unconditional now, annotating `(default, latest in registry)` vs `(pinned via AUTOMOBILE_CTRL_PROXY_VERSION)`. Makes "did the env var propagate?" trivial to diagnose from CI logs.
- `--cli doctor` splits the old conflated "AutoMobile Version" line into AutoMobile Daemon Version (from `package.json`) and `CtrlProxy Release Version` (registry/pinned), since they can intentionally diverge now.

Documented next to the existing `AUTOMOBILE_CTRL_PROXY_APK_PATH` override in `docs/design-docs/plat/android/junit-runner/project-setup.md`.


kaeawc:
While I get the desire to fix the bug of having the right CtrlProxy with right Daemon version, the proposal in this PR seems to allow for CtrlProxy pinning independent of Daemon version. I'd need to understand what this provides instead of just pinning a Daemon version and having the CtrlProxy version associated with that release. Pinning to any CtrlProxy is not something I want to maintain and have backward compatibility for.

RyeBread1026:
> While I get the desire to fix the bug of having the right CtrlProxy with right Daemon version, the proposal in this PR seems to allow for CtrlProxy pinning independent of Daemon version. I'd need to understand what this provides instead of just pinning a Daemon version and having the CtrlProxy version associated with that release. Pinning to any CtrlProxy is not something I want to maintain and have backward compatibility for.

Makes sense to me. I don't have in mind picking an arbitraty CtrlProxy version - just a matching one / not the latest, which I wasn't able to do. Updated.

Issue #2540, opened 2026-06-25
fix(ios): SDK in-app hierarchy (:8766) merged into observes of unrelated foreground apps — ~2.5MB hierarchies


kaeawc:
## Summary

When an iOS app embedding the AutoMobile SDK is running (its in-app hierarchy server bound on port 8766), the SDK hierarchy gets merged into **every** `observe` — including observes whose **foreground app is a different, non-SDK app**. The result is a ~**2.5 MB / ~4000-node** view hierarchy returned for a screen that should be small (e.g. iOS Settings), an **~80× size blowup**. Terminating the SDK app drops the same observe back to ~71 KB.

This is both a **correctness** concern (hierarchy of a backgrounded SDK app merged into the foreground non-SDK app's tree) and a **performance/payload** concern (multi-MB responses that overflow downstream limits).

Reproduced live on `main` (Xcode 26.3, iOS 18.6 sim).

## Impact

- iOS `observe` / `tapOn` / `pressButton` (anything returning an observation) can return multi-MB hierarchies, overflowing MCP tool-result size limits and forcing results to spill to files; slow to serialize/transmit/parse.
- The merged tree includes nodes from an app that is **not** the current foreground app, so element matching can hit stale/incorrect nodes from the backgrounded SDK app.

## Reproduction

1. Launch an iOS app that embeds the AutoMobile SDK (e.g. the Playground app) so its `SdkHierarchyServer` is listening on `localhost:8766`.
2. Foreground a **different**, non-SDK app (e.g. `com.apple.Preferences` / Settings).
3. `observe` → returns ~2.5 MB / ~4000 nodes (Settings should be a few dozen nodes).
4. Terminate the SDK app (Playground) → `observe` of Settings → ~71 KB. The delta is entirely the merged SDK hierarchy.

## Observed evidence (live)

- `launchApp(com.apple.Preferences)` while Playground (SDK) ran → tool result **2,579,810 chars / 39,993 lines**, `className` occurrences ≈ 3948, `success:true`.
- `launchApp(dev.jasonpearson.automobile.Playground)` → **2,579,843 chars** — essentially identical size, i.e. the bulk is the SDK tree, independent of the actual foreground app.
- After `terminateApp(Playground)` → `observe` of Settings → **~71 KB** (normal XCUITest-only tree).

## Likely root cause (with references)

In the runner's hierarchy handler, the SDK hierarchy is fetched and merged unconditionally onto the XCUITest base:

[...]

PR #2549, opened 2026-06-25
[codex] fix iOS SDK hierarchy over-merge

kaeawc:
## Summary

- Scope iOS SDK hierarchy enrichment through the same foreground-bundle guard for explicit observes and debounced broadcasts
- Clear or replace stale SDK hierarchy cache entries when the cached SDK bundle does not match the foreground app
- Prevent SDK-only subtree injection from repeating under containment-only XCUITest matches
- Add regressions for foreign cached SDK nodes, fresh cache replacement, and bounded 3-tab hierarchy duplication

## Root Cause

The explicit hierarchy request path had a foreground-bundle guard, but the debounced broadcast path still merged `sdkHierarchyCache.latest` directly. The merge injection pass also used containment matches as SDK parent placement, so broad SDK containers could cause SDK-only children, such as tab bar labels, to be attached repeatedly under unrelated XCUITest descendants.

## Validation

- `swift test -Xswiftc -warnings-as-errors` from `ios/control-proxy` passed: 214 tests
- `git diff --check` passed

## Notes

- Addresses the reopened feedback in #2540.
- Broader `bash scripts/ios/swift-test.sh` was run earlier and failed only in the live `XCTestRunner` Reminders integration because the simulator could not tap `Done`; `control-proxy` passed in that broader run.

### DIFF
--- a/ios/control-proxy/Sources/CtrlProxy/CommandHandler.swift
+++ b/ios/control-proxy/Sources/CtrlProxy/CommandHandler.swift
@@ -220,6 +220,22 @@ public class CommandHandler: CommandHandling {
         HierarchyMerger.merge(xcuitest: hierarchy, sdk: matchingSdkHierarchy(for: hierarchy))
     }
 
+    func enrichWithCachedSdkHierarchy(_ hierarchy: ViewHierarchy) -> ViewHierarchy {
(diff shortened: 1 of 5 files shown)


kaeawc:
Fixes #2540

PR #2587, opened 2026-06-29
Fix SQLite parallel client safety

kaeawc:
## Summary

- implement real Bun SQLite transaction begin/commit/rollback and gate early app queries until startup migrations finish
- serialize shared SQLite driver operations on Android and iOS so parallel inspector clients cannot close or reuse handles concurrently
- fix Android database path prefix validation and make busy-timeout setup compatible with Robolectric
- add parallel read/write coverage for TypeScript, Android, and iOS SQLite paths
- isolate IOSCtrlProxyManager port allocation tests from local adb-owned ports

## Validation

- `bun run lint`
- `bun run build`
- `bun test`
- `bun test test/db/database.test.ts test/db/databaseLazyPath.test.ts`
- `swift test --filter SQLiteDatabaseDriverConcurrencyTests`
- `swift test` in `ios/auto-mobile-sdk`
- `bash scripts/ios/swift-build.sh`
- `./gradlew :auto-mobile-sdk:testDebugUnitTest`
- `ONLY_TOUCHED_FILES=true bash scripts/ktfmt/validate_ktfmt.sh`
- `ONLY_TOUCHED_FILES=true bash scripts/swiftformat/validate_swiftformat.sh`
- `ONLY_TOUCHED_FILES=true bash scripts/swiftlint/validate_swiftlint.sh`

## Notes

- `bash scripts/validate-bun-test-timings.sh` still reports pre-existing slow Bun tests outside this SQLite change.

### DIFF
--- a/android/auto-mobile-sdk/src/main/kotlin/dev/jasonpearson/automobile/sdk/database/SQLiteDatabaseDriver.kt
+++ b/android/auto-mobile-sdk/src/main/kotlin/dev/jasonpearson/automobile/sdk/database/SQLiteDatabaseDriver.kt
@@ -14,6 +14,7 @@ import java.util.concurrent.ConcurrentHashMap
  */
 class SQLiteDatabaseDriver(private val context: Context) : DatabaseDriver {
 
+  private val databaseLock = Any()
   private val openDatabases = ConcurrentHashMap<String, SQLiteDatabase>()
 
   override fun getDatabases(): List<DatabaseDescriptor> {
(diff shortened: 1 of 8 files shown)


kaeawc (commented):
Reviewed against the SQLite-safety findings from the v0.0.38 manual-test pass (no-op transactions on the single shared Bun connection, the migration race, and the in-app inspector-driver concurrency). **This addresses all three directly and is well-tested — nice work.** A couple of behavior/edge points worth a look before merge.

## ✅ What it fixes (and confirms)
- **No-op transactions → real, isolated transactions.** `BunSqliteDriver` now issues real `begin/commit/rollback` and serializes them on the shared connection via the `#transactionOwner` mutex (`#reserveTransaction` drains in-flight queries and blocks concurrent transactions; `#enterQuery` blocks non-owner queries while a transaction is open). This is exactly the gap that left `installedAppsRepository.replaceInstalledApps` (delete-all-then-insert) non-atomic under #2565's parallel forwards. The tests nail it — especially *"waits to run non-transaction queries while a transaction is open"*, which asserts a concurrent writer doesn't observe mid-transaction state and the rollback test asserts `rows === []`. That's the precise interleaving I was worried about.
- **Migration race.** `waitForMigrationsBeforeQuery` wired as `beforeQuery` gates every query behind `migrationsPromise`, so a `recordToolCall`/`tool_calls` insert that arrives before startup migrations finish now waits instead of racing (and silently dropping). Good.
- **In-app inspector driver concurrency.** Android `synchronized(databaseLock)` and iOS `NSLock operationLock` serialize the SDK SQLite drivers, closing the concurrent-handle-close/reuse hazard for parallel `sqlQuery` clients. Both get concurrency tests.

## 🔶 Main point: migration *failure* now hard-fails all queries (and can't recover without a restart)
Previously the migration promise swallowed errors (`.catch(log)`) and queries ran regardless — best-effort. Now `startMigrations` **re-throws** on failure:

```ts
} catch (error) {
  logger.error("Failed to run migrations on database initialization:", error);
  throw error;            // ← migrationsPromise now rejects
} finally {
  await migrationDb.destroy();
}
```

Because `waitForMigrationsBeforeQuery` does `await migrationsPromise`, a rejected promise makes **every** subsequent `executeQuery` throw — and `migrationsPromise` is never reset, so there's no retry; the DB is unusable for the life of the process. That's arguably *more* correct (fail-safe: don't run against a half-migrated schema), but it's a meaningful semantic change from "log and continue" to "total DB outage until daemon restart." Worth making deliberate:
- Surface it as a clear, actionable daemon error (not just a per-query `Query failed: ...` wrapped string), since today a bad migration would manifest as every tool call failing with a confusing SQL error.
- Consider whether some paths should stay best-effort. `recordToolCall` already swallows errors, so analytics would silently vanish while *functional* queries hard-fail — an asymmetry worth a conscious call.
- A test for both the gate (query issued pre-migration waits) and the failure mode (migration rejects → query rejects) would lock the intended behavior in.

## 🔸 Secondary: possible transaction starvation under continuous autocommit load
`#reserveTransaction` waits for `#activeQueries === 0`, but new non-transaction queries can still start while a transaction is *waiting* to reserve — `#enterQuery` only blocks once `#transactionOwner` is set, and during the reserve-wait it's still `null`. So a steady stream of autocommit reads can keep `#activeQueries > 0` and starve a pending write transaction. Low real-world risk given the daemon's bursty load (it'll hit 0 between bursts), but if you want writer-preference, set a "transaction pending" flag during the reserve so incoming queries also queue behind it.

## 🔹 Minor
- `beforeQuery` is awaited on every `executeQuery` for the life of the process. After startup it's a cheap early-return (`if (migrationsRun || !migrationsPromise) return`), so ~1 extra microtask/query — negligible, just noting it's hot-path.
- Migrations run on a **separate** `migrationDb` connection (`createSqliteKysely(dbPath)`) — fine since app queries are gated until it's destroyed, but worth confirming that connection also gets `configureSqliteDatabase` (busy_timeout/WAL/foreign_keys) so the two connections don't briefly disagree on journal mode on first creation.
- Scope: the Android path-prefix validation, Robolectric busy-timeout compat, and IOSCtrlProxyManager port-test isolation are unrelated to SQLite parallel safety. Reasonable to bundle, but they widen the blast radius of this PR.
- `#notifyWaiters` waking all waiters to re-check is a thundering-herd pattern; correct, just not FIFO-fair (ties into the starvation note).

Overall this is the right shape of fix for the shared-connection model (in-process serialization rather than a pool), and the test coverage is exactly where it should be. The migration-failure semantics are the one thing I'd want an explicit decision on.

Issue #2561, opened 2026-06-28
Multi-daemon detection false-positive: a single daemon's own shell wrapper is counted as an "other" daemon


kaeawc:
## Summary

`--daemon status` (and the start-time conflict check) can warn `Found N other auto-mobile daemon process(es) running … These may be from other worktrees and can cause device pool conflicts` when there is actually only **one** daemon. The detection counts the daemon's own `sh -c` parent wrapper as a separate "other" daemon, because the daemon is launched with `shell: true` (so it is a parent shell + child process tree, two matching PIDs) while the dedup only excludes the current `process.pid`.

This produces spurious device-pool-conflict warnings and makes it harder to tell a real multi-daemon situation from a benign single-daemon one.

## Root cause

The daemon is spawned with `shell: true` (`src/daemon/manager.ts` ~line 437), so each daemon is represented by **two** processes that both match the detection grep: the `/bin/sh -c "<bun> … --daemon-mode"` wrapper and the actual `bun … --daemon-mode` child.

`findAllDaemonProcesses()` (`src/daemon/manager.ts` ~line 212) matches both and only filters out `process.pid`:

```ts
const psOutput = execSync(
  `ps aux | grep -E "auto-mobile.*--daemon-mode|dist/src/index.js --daemon-mode" | grep -v grep`,
  { encoding: "utf-8" }
);
// ...
if (!isNaN(pid) && pid !== process.pid) {
  pids.push(pid);   // the daemon's own sh-wrapper (or child) survives this filter
}
```

Because the wrapper/child PID is not `process.pid`, the daemon's own process tree contributes a PID that is then reported as an "other" daemon.

## Reproduction

1. Start a single daemon (launched via the normal `shell: true` path).
2. Run `--daemon status`.
3. Observe a warning about "N other auto-mobile daemon process(es) … from other worktrees", with the reported PID belonging to the same daemon's shell wrapper (its PPID/child), not a distinct daemon.

`ps aux | grep -E "auto-mobile.*--daemon-mode|dist/src/index.js --daemon-mode" | grep -v grep` shows two lines (parent `sh -c` + child) for one daemon.

## Suggested fixes

- **Exclude the daemon's own process tree, not just `process.pid`** — also filter the parent/child of the current process (PPID and direct children), or
- **Avoid `shell: true`** when spawning the daemon (resolve the absolute command path instead) so each daemon is a single process, or
[...]

PR #2589, opened 2026-06-29
fix: confirm daemon process candidates

kaeawc:
## Summary

- Re-check all daemon process-table candidates for liveness before status warnings or startup takeover signaling.
- Preserve live non-PID-file daemon candidates so genuine cross-worktree conflicts still surface.
- Escalate startup takeover from `SIGTERM` to `SIGKILL` for daemon candidates that survive the grace period.
- Fail closed when process-table inspection fails instead of treating the failure as "no daemons."

## Root Cause

`--daemon status` scanned the full process table and treated every normalized `--daemon-mode` match except the active daemon PID as another daemon. Short-lived startup or health-probe processes could appear in that snapshot and then exit before inspection, producing false multi-daemon warnings.

## User Impact

Single-daemon setups should no longer warn about probe PIDs that are gone by report time, while live rogue daemons from other worktrees remain visible and are still stopped during startup takeover.

## Testing

- `./node_modules/.bin/eslint src/daemon/manager.ts test/daemon/manager.test.ts --fix`
- `bun test test/daemon/manager.test.ts test/daemon/daemonManagerLaunch.test.ts test/daemon/daemonManagerReadiness.test.ts test/daemon/daemonManagerLock.test.ts`
- `bun run build`

### DIFF
--- a/src/daemon/manager.ts
+++ b/src/daemon/manager.ts
@@ -25,7 +25,10 @@ import {
 import { DaemonClient, type DaemonClientFactory, type DaemonClientLike } from "./client";
 import { DaemonState, type DaemonStateLike } from "./daemonState";
 import { Timer, defaultTimer } from "../utils/SystemTimer";
-import { cleanupDaemonFiles } from "./daemonFiles";
+import {
+  cleanupDaemonFiles,
+  isProcessRunning as isDaemonProcessRunning,
+} from "./daemonFiles";
 import {
   DAEMON_LAUNCH_CWD_ENV,
   resolveDaemonLaunchWorkingDirectory,
@@ -110,7 +113,11 @@ export function parseDaemonProcessTable(psOutput: string): DaemonProcessRecord[]
   return records;
 }
 
-export class PsDaemonProcessFinder implements DaemonProcessFinder {
+export interface DaemonProcessLivenessChecker {
+  isProcessRunning(pid: number): boolean;
+}
+
+export class PsDaemonProcessFinder implements DaemonProcessFinder, DaemonProcessLivenessChecker {
   constructor(private readonly runCommand: ProcessTableCommandRunner = execSync) {}
 
   findDaemonProcesses(): DaemonProcessRecord[] {
@@ -119,6 +126,10 @@ export class PsDaemonProcessFinder implements DaemonProcessFinder {
       maxBuffer: DAEMON_PROCESS_TABLE_MAX_BUFFER_BYTES,
     });
     return parseDaemonProcessTable(psOutput);
+  }
+
+  isProcessRunning(pid: number): boolean {
+    return isDaemonProcessRunning(pid);
   }
 }
 
@@ -129,6 +140,10 @@ export interface DaemonProcessSpawner {
 const nodeDaemonProcessSpawner: DaemonProcessSpawner = {
   spawn: nodeSpawn,
 };
+
+function hasProcessLivenessChecker(value: unknown): value is DaemonProcessLivenessChecker {
+  return typeof (value as Partial<DaemonProcessLivenessChecker> | undefined)?.isProcessRunning === "function";
+}
 
 const MAX_DAEMON_STARTUP_LOG_BYTES = 4000;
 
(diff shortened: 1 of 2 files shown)


kaeawc (commented):
Reviewed against #2561 (the false-positive this targets). **It fixes the reported symptom, but I think it over-corrects** — converting a false-positive into a false-negative. Details below.

## ✅ What it fixes
The transient-probe false-positive is genuinely addressed: `findConfirmedDaemonProcesses()` now requires a candidate to be the PID-file-backed daemon **and** pass a liveness re-check, so short-lived `--daemon-mode` startup/health probes that exit before inspection are no longer reported. Tests for the transient case and the dead-pid-file case are good additions.

## ⚠️ Main concern: it silently disables genuine multi-worktree conflict detection
`findConfirmedDaemonProcesses()` can only ever return **0 or 1 PID — the one in the PID file**. Everything else is dropped, including a *live* daemon that isn't the PID-file owner. The new test encodes exactly this:

```ts
// "ignores transient daemon-mode candidates that are not backed by the PID file"
withPidFile(201, pidFilePath => {
  // candidates include 401, and livePids = {201, 401}  ← 401 is ALIVE
  expect(manager.findOtherDaemonProcesses(201)).toEqual([]);  // 401 ignored
});
```

A live PID 401 not in the PID file is treated identically to a dead transient probe. But a live non-PID-file daemon is precisely the thing the warning + takeover existed to catch: a rogue daemon from **another worktree**. Since the PID file is **per-UID and shared** (`/tmp/auto-mobile-daemon-${uid}.pid`, `constants.ts`), every worktree overwrites it, so only the most-recently-started daemon is ever "confirmed" — any earlier still-running daemon becomes an orphan that this code can no longer see.

Concrete consequences:
- `--daemon status` will essentially **never** emit the "Found N other daemon process(es) ... from other worktrees" warning again (confirmed set is the active daemon itself → filtered to `[]`). The PR's "User Impact" frames this warning as only about phantom probe PIDs, but its original job was real cross-worktree conflict detection — that's now gone, not just de-noised.
- `startUnlocked()` takeover now only signals the PID-file daemon, so it **won't stop orphaned daemons from other worktrees** — the exact "stop all other daemons before starting" guarantee it was written for. We hit this very situation live this week: multiple coexisting daemons sharing one socket (made worse because they ignore SIGTERM — see note below).

So the net change trades a **false-positive** (cosmetic, annoying) for a **false-negative** (silent — a real conflict goes unreported and un-stopped). For a device-pool-conflict guard, the false-negative is arguably the more dangerous direction.

## 💡 Suggested approach: re-check liveness of *all* candidates, don't narrow to the PID file
The root issue in #2561 was that probes are *transient* — they exit before inspection. That's distinguishable from a rogue daemon by **liveness at report time**, not by PID-file membership. You already added a `DaemonProcessLivenessChecker` — applying it to every candidate (not just the pid-file one) keeps both properties:

```ts
findConfirmedDaemonProcesses(): number[] {
  return this.findAllDaemonProcesses()
    .filter(pid => this.isProcessRunning(pid)); // drop probes that already exited
}
```

That filters the transient probes from #2561 **and** still surfaces a live rogue daemon from another worktree. (If probe/daemon races at the exact snapshot instant are a concern, a short re-stat delay or a socket/pid-file confirmation per candidate hardens it further — but it should confirm liveness, not collapse to a single PID.)

If narrowing to the PID file *is* intentional, then the warning text ("from other worktrees") and the takeover semantics should be updated to match, and the lost cross-worktree detection called out explicitly — right now the behavior change is broader than the PR description implies.

## Minor / related
- **SIGTERM-only takeover:** this PR touches the takeover path, so worth flagging that even a *confirmed* daemon may not die — daemons ignore SIGTERM in practice (observed live; `--daemon restart` already escalates to SIGKILL, but `startUnlocked`'s loop only sends SIGTERM with no SIGKILL fallback). Confirming the candidate and then failing to actually stop it is a related gap.
- **ENOBUFS still swallowed:** `findAllDaemonProcesses()` returning `[]` on a >16 MiB `ps` buffer (catch → `[]`) still reads as "no daemons" and silently disables detection on large hosts. Out of scope here, but it's the other half of making candidate detection trustworthy.
- **Constructor positional bloat:** `processLivenessChecker` is now positional param #9, alongside the `processFinderOrSpawner` union discriminated by an `in` check. This is well past the point where an options object would be safer against arg-order mistakes.
- **Naming:** `findConfirmedDaemonProcesses()` reads as "may return several," but structurally returns at most one PID. If the single-PID semantics are intended, the name could say so.

Happy to be wrong on the multi-worktree read if there's a code path I'm missing — but the new test asserting a *live* non-PID-file PID is ignored is what convinces me this is a deliberate-but-under-discussed trade-off.

PR #2592, opened 2026-06-29
Record tool call durations

### LINKED ISSUE #2588: tool_calls.duration_ms is never populated (NULL for all rows) — per-call timing not recorded
### Summary
The `tool_calls` table defines a `duration_ms INTEGER` column, but it is **NULL for every row** — the daemon never records how long any tool call took.

### Evidence
On a live daemon DB (`~/.auto-mobile/auto-mobile.db`):

```sql
SELECT count(*) total,
       sum(duration_ms IS NULL) null_dur,
       sum(duration_ms IS NOT NULL) nonnull_dur
FROM tool_calls;
-- total=10378, null_dur=10378, nonnull_dur=0
```

Every one of 10,378 rows (across `observe`, `tapOn`, `sqlQuery`, `pressButton`, etc.) has `duration_ms = NULL`. `SELECT DISTINCT duration_ms` returns only NULL.

### Root cause
- The column exists in the schema (`PRAGMA table_info(tool_calls)` → `duration_ms INTEGER`).
- `ToolCallRepository.recordToolCall()` (`src/db/toolCallRepository.ts:26`, insert at `:35`) builds the `entry` from the `ToolCallRecord` and never sets `duration_ms` (no reference to it in the file).
- The callers (`src/server/toolRegistry.ts:369` and `:459`) don't measure/pass an elapsed duration into the record.

So nothing ever writes the field.

### Impact
- No per-tool latency telemetry: can't answer "how long does sqlQuery/observe/tapOn take?" from the DB, can't spot slow tools or regressions, and `debugSearch`/analytics over `tool_calls` can't surface timing.
- (Found while investigating perceived sqlQuery latency — having this populated would have made that trivial.)

### Suggested fix
Measure elapsed time around the tool handler in the registry wrapper (`toolRegistry.ts`) — `const t0 = timer.now()` before invoking, `duration_ms = timer.now() - t0` after — and thread it through `ToolCallRecord` → `recordToolCall` so the insert populates `duration_ms`. Add a test asserting a recorded call has a non-null, plausible `duration_ms`.

### Environment
Reproduced on `main` @ post-#2585, daemon v0.0.38, macOS.

kaeawc:
## Summary

- Populate `tool_calls.duration_ms` from the `ToolRegistry` elapsed timer.
- Thread `durationMs` through `ToolCallRepository.recordToolCall()` and persist it as integer milliseconds.
- Add focused regression tests for successful and throwing registry calls plus repository persistence.

Fixes #2588

## Testing

- `bun test test/db/toolCallRepository.test.ts test/server/toolRegistry.duration.test.ts`
- `bun run lint`
- `bun run build`
- `bun test --bail` *(fails in existing `test/utils/IOSCtrlProxyManager.test.ts`: expected default/service port `8765` or `8767`, received `8768`; reproduces when running that test file in isolation and is outside this telemetry change)*

### DIFF
--- a/.github/actions/setup-auto-mobile-npm-package/action.yml
+++ b/.github/actions/setup-auto-mobile-npm-package/action.yml
@@ -6,6 +6,10 @@ inputs:
     description: "The shell to use for any steps that use shells"
     default: "bash"
     required: "true"
+  install-global:
+    description: "Whether to globally install the checked-out AutoMobile package"
+    default: "true"
+    required: "false"
 
 runs:
   using: "composite"
@@ -33,6 +37,7 @@ runs:
         bunx turbo run build --output-logs=errors-only
 
     - name: "Globally Install AutoMobile"
+      if: ${{ inputs.install-global == 'true' }}
       shell: ${{ inputs.shell }}
       run: |
         if [[ "${RUNNER_OS:-}" == "Windows" ]]; then
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -773,6 +773,8 @@ jobs:
       - name: "Setup Auto Mobile"
         if: env.RUN_MCP_BUILD_TEST == 'true'
         uses: ./.github/actions/setup-auto-mobile-npm-package
+        with:
+          install-global: "false"
 
       - name: "Run Lint"
         if: env.RUN_MCP_BUILD_TEST == 'true'
--- a/src/db/toolCallRepository.ts
+++ b/src/db/toolCallRepository.ts
@@ -7,6 +7,7 @@ interface ToolCallRecord {
   toolName: string;
   timestamp: string;
   sessionUuid?: string | null;
+  durationMs?: number | null;
 }
 
 export class ToolCallRepository {
@@ -30,6 +31,9 @@ export class ToolCallRepository {
         tool_name: record.toolName,
         timestamp: record.timestamp,
         session_uuid: record.sessionUuid ?? null,
+        duration_ms: record.durationMs === undefined || record.durationMs === null
+          ? null
+          : Math.max(0, Math.round(record.durationMs)),
       };
 
       await db.insertInto("tool_calls").values(entry).execute();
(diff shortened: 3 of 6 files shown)


kaeawc (commented):
Reviewed as the fix for #2588 (`tool_calls.duration_ms` never populated). **Correct and well-tested — looks good to merge.**

## ✅ Fixes #2588
- The DB `recordToolCall(...)` is now invoked in a **`finally`** with `durationMs: this.timer.now() - toolStartMs`, so the duration is recorded for **both** the success and the throwing path — verified by the two new tests (`records elapsed duration for a completed tool call` = 37ms; `records elapsed duration when a tool call throws` = 19ms), driven by the injectable `this.timer` so they're deterministic.
- Persistence is defensive: `duration_ms = record.durationMs == null ? null : Math.max(0, Math.round(durationMs))` — rounds to integer ms, clamps negatives, preserves null. Good.
- Repository-level persistence test covers the round-trip.

This directly closes the gap I reported (10,378/10,378 rows NULL).

## 🔹 Minor / optional
- **Duration is computed twice.** The telemetry path already does `const toolDurationMs = this.timer.now() - toolStartMs` in the try body, and the `finally` recomputes `this.timer.now() - toolStartMs` for the DB record. The two values differ slightly (the `finally` runs a bit later, after the telemetry call / post-processing), so the DB `duration_ms` will be marginally larger than the telemetry `durationMs` for the same call. Not wrong, but reusing one computed value for both would keep them consistent and make "what window does duration_ms measure" unambiguous.
- **What's included in the window:** because it's measured at the `finally`, `duration_ms` includes the registry wrapper's post-handler work (telemetry, session-cache writes, response post-processing), not just the tool handler. That's a reasonable definition of "tool call duration," just worth being explicit about since consumers will read it as latency.
- **Pre-existing flake:** the PR notes the `IOSCtrlProxyManager` port test (`expected 8765/8767, received 8768`) failing in `bun test --bail`. #2587 and #2591 both include a fix that isolates that test from local ports — worth coordinating merge order (or cherry-picking that isolation) so this PR's CI isn't blocked by the same flake.

Net: small, focused, correct, with the right success+throw test coverage. 👍

PR #2615, opened 2026-06-29
fix: ignore transient daemon process matches

### LINKED ISSUE #2561: Multi-daemon detection false-positive: a single daemon's own shell wrapper is counted as an "other" daemon

kaeawc:
## Summary
- Preserve liveness-filtered process-table detection so status warnings and takeover cleanup still catch real cross-worktree daemons.
- Cover transient `--daemon-mode` startup/probe candidates by asserting they are ignored once gone by the liveness re-check.
- Add regression coverage that a live daemon not owning the PID file is still reported and stopped during takeover.

Fixes #2561

## Validation
- `bun test test/daemon/manager.test.ts`
- `bun run build`
- `bun run lint`
- `bun test`

## Notes
- No `.github` pull request template is present in this repo, so this body follows the repo's common Summary/Validation format.

### DIFF
--- a/test/daemon/manager.test.ts
+++ b/test/daemon/manager.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, spyOn, test } from "bun:test";
-import { mkdtempSync, rmSync } from "node:fs";
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { tmpdir } from "node:os";
 import type { ChildProcess, SpawnOptions } from "node:child_process";
@@ -152,18 +152,28 @@ describe("Daemon manager process detection", () => {
 
   function managerWithProcesses(
     records: DaemonProcessRecord[],
-    options: { livePids?: Set<number> } = {}
+    options: { livePids?: Set<number>; pidFilePath?: string } = {}
   ): DaemonManager {
     return new DaemonManager(
       undefined,
       undefined,
       undefined,
       undefined,
-      undefined,
+      options.pidFilePath,
       undefined,
       new FakeDaemonProcessFinder(records, options.livePids),
       undefined
     );
+  }
+
+  function writeDaemonPidFile(pidFilePath: string, pid: number): void {
+    writeFileSync(pidFilePath, JSON.stringify({
+      pid,
+      socketPath: "/tmp/auto-mobile-test.sock",
+      port: 3000,
+      startedAt: 1,
+      version: "test",
+    }));
   }
 
   test("reports a shell-launched daemon once using the long-lived daemon child PID", () => {
@@ -200,7 +210,10 @@ describe("Daemon manager process detection", () => {
     expect(manager.findAllDaemonProcesses()).toEqual([]);
   });
 
-  test("filters the active daemon PID while preserving distinct daemons", () => {
+  test("filters the active daemon PID while preserving distinct live daemons", () => {
+    const dir = mkdtempSync(join(tmpdir(), "daemon-manager-pid-file-test-"));
+    const pidFilePath = join(dir, "daemon.pid");
+    writeDaemonPidFile(pidFilePath, 201);
     const manager = managerWithProcesses([
       {
         pid: 200,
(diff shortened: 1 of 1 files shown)


src/daemon/manager.ts
```diff
@@ -338,6 +339,20 @@ export class DaemonManager implements DaemonManagerLike {
...
   }
 
+  private findPidFileBackedLiveDaemonProcesses(): number[] {
+    const pidData = readPidFileDataSync(this.pidFilePath);
+    if (!pidData || typeof pidData.pid !== "number") {
+      return [];
+    }
+
+    const processTablePids = new Set(this.findAllDaemonProcesses());
+    if (!processTablePids.has(pidData.pid) || !this.isProcessRunning(pidData.pid)) {
```
kaeawc: Narrowing live-daemon detection to **only** the PID recorded in the shared per-UID pid file (`/tmp/auto-mobile-daemon-${uid}.pid`, `constants.ts:51`) means a genuine **live** rogue daemon started from another worktree — one that isnt (or is no longer) the pid-file owner — is no longer reported here. And because `startUnlocked()` takeover also calls this for its SIGTERM/SIGKILL list, that rogue is no longer stopped before a new daemon starts. That cross-worktree conflict is exactly what the warning + the stop-all-other-daemons block exist to catch.

#2589 already solved the transient-probe problem the safe way: scan the full `ps` table for `--daemon-mode` matches and filter by **liveness** — probes exit so they drop out, while a live rogue is still caught. Collapsing to the single pid-file PID loses that. Suggest keeping the liveness-filtered breadth, and adding a test that a live `--daemon-mode` PID **not** in the pid file is still detected (the current tests only assert the suppression direction).

kaeawc (commented):
Reviewed against #2561. This correctly eliminates the transient-probe false-positive (and the new test covers that direction), but I think it over-corrects into a cross-worktree **false-negative** — detail inline.

PR #2616, opened 2026-06-29
Fix iOS video recording FFmpeg finalize

### LINKED ISSUE #2606: videoRecording stop fails on iOS simulator: mov→mp4 FFmpeg finalize produces no output and the error hides the cause
### Summary
On an iOS simulator, `videoRecording` **start** works and captures a raw `.mov` (~3.2 MB observed), but **stop** fails with `FFmpeg output file missing at .../<id>.mp4`. FFmpeg is installed (8.1.2), so the mov→mp4 post-processing step itself is failing — and the error message swallows *why*.

### Repro (iOS sim)
1. `videoRecording { action:"start", platform:"ios", deviceId:<sim> }` → PASS (returns archive mp4 path; raw `.mov` written under `~/.auto-mobile/video-archive/<id>/`).
2. `videoRecording { action:"stop", ... }` → `Error: FFmpeg output file missing at ~/.auto-mobile/video-archive/<id>/<id>-<ts>.mp4`.

### Root cause / diagnostic gap
`src/features/video/FfmpegVideoProcessingBackend.ts` (stop/finalize, ~lines 355-374):
```ts
await waitForExit(ffmpegProcess, ffmpegTracker.exitPromise);

if (ffmpegTracker.exitState.exitCode && ffmpegTracker.exitState.exitCode !== 0) {
  throw new ActionableError(`FFmpeg post-processing failed with code ... : ${ffmpegTracker.stderr.join("")}`);
}

const outputExists = await pathExists(backendHandle.config.outputPath);
if (!outputExists) {
  throw new ActionableError(`FFmpeg output file missing at ${backendHandle.config.outputPath}`);   // <- what we hit
}
```
- The stderr-surfacing branch only fires when `exitCode` is **truthy and non-zero**. If FFmpeg's `exitCode` is `null`/`0`/`undefined` (process killed, never spawned, or exited 0 without producing the file), that branch is skipped and we fall through to the opaque `"FFmpeg output file missing"` — **the FFmpeg command and its stderr are never surfaced**, so we can't see why the transcode produced nothing.
- So there are two problems: (1) a real finalize failure (mov→mp4 produces no output on iOS sim despite FFmpeg being present), and (2) the error reporting hides the cause (ties into #2567 "surface subprocess context").

### Suggested fix
- Always log/surface the exact FFmpeg argv + full stderr on the failure path (and handle `exitCode === null`), so the underlying transcode error is visible instead of "output file missing".
- Then investigate the actual finalize failure on the iOS-sim path (input `.mov` not finalized yet? wrong `-i`/codec/output args? path mismatch between where FFmpeg writes and `config.outputPath`?). The raw `.mov` exists, so input is present — likely a bad arg or a path/timing issue.
- Add a test that `videoRecording start`→`stop` yields a non-empty `.mp4` on a sim.

### Environment
`main` @ post-#2585, daemon v0.0.38, iOS sim iPhone 16 Pro (iOS 18.6), ffmpeg 8.1.2 (Homebrew). Found during an iOS tool sweep.

kaeawc:
## Summary
- Fixes #2606 by remuxing iOS simulator `.mov` captures to `.mp4` with FFmpeg stream copy when no scaling is requested, avoiding unnecessary H.264 re-encode constraints.
- Surfaces actionable FFmpeg post-processing diagnostics, including command, exit code, signal, stderr, and output path for non-zero, null-exit, missing-output, and empty-output failures.
- Adds focused TDD coverage for iOS file-input FFmpeg args and diagnostic failure paths.

## Plan and evaluation criteria
- iOS simulator `videoRecording` stop should produce a non-empty `.mp4` when FFmpeg succeeds.
- FFmpeg failures should include enough subprocess context to identify the underlying cause.
- Existing scaled-output behavior should still transcode with the configured encoder.

## Testing
- `bun test test/features/video/FfmpegVideoProcessingBackend.test.ts`
- `bun run lint`
- `bun run build`
- `bun test`

## Notes
- This repo does not currently include a pull request template file under `.github/`; this body follows the repo's summary/testing convention.
- Initial dependency install with lifecycle scripts failed because `heapdump` does not compile under the local Node 26 headers; `bun install --ignore-scripts` completed and validation passed afterward.

### DIFF
--- a/src/features/video/FfmpegVideoProcessingBackend.ts
+++ b/src/features/video/FfmpegVideoProcessingBackend.ts
@@ -41,6 +41,12 @@ type FfmpegInput =
   | { type: "pipe" }
   | { type: "file"; path: string };
 
+type FfmpegDiagnosticsTracker = Pick<ProcessTracker, "exitState" | "stderr">;
+
+function isFailedExitCode(exitCode: number | null | undefined): boolean {
+  return exitCode !== undefined && exitCode !== 0;
+}
+
 interface FfmpegBackendHandle {
   kind: "ffmpeg";
   platform: "android" | "ios";
@@ -363,16 +369,21 @@ export class FfmpegVideoProcessingBackend implements VideoCaptureBackend {
 
     await waitForExit(ffmpegProcess, ffmpegTracker.exitPromise);
 
-    if (ffmpegTracker.exitState.exitCode && ffmpegTracker.exitState.exitCode !== 0) {
+    if (isFailedExitCode(ffmpegTracker.exitState.exitCode)) {
       throw new ActionableError(
-        `FFmpeg post-processing failed with code ${ffmpegTracker.exitState.exitCode}: ${ffmpegTracker.stderr.join("")}`
+        this.buildFfmpegFailureMessage(
+          "FFmpeg post-processing failed",
+          ffmpegArgs,
+          ffmpegTracker
+        )
       );
     }
 
-    const outputExists = await pathExists(backendHandle.config.outputPath);
-    if (!outputExists) {
-      throw new ActionableError(`FFmpeg output file missing at ${backendHandle.config.outputPath}`);
-    }
+    await this.assertFfmpegOutputReady(
+      backendHandle.config.outputPath,
+      ffmpegArgs,
+      ffmpegTracker
+    );
 
     try {
       await fsPromises.rm(capturePath, { recursive: true, force: true });
@@ -392,6 +403,14 @@ export class FfmpegVideoProcessingBackend implements VideoCaptureBackend {
       args.push("-f", "mp4", "-i", "pipe:0");
     } else {
       args.push("-i", input.path);
+
+      if (!config.resolution) {
+        args.push("-c", "copy");
+        args.push("-movflags", "+faststart");
+        args.push("-y");
+        args.push(config.outputPath);
+        return args;
+      }
     }
 
     args.push("-r", String(config.fps));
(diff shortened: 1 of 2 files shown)


src/features/video/FfmpegVideoProcessingBackend.ts
```diff
@@ -363,16 +369,21 @@ export class FfmpegVideoProcessingBackend implements VideoCaptureBackend {
 
     await waitForExit(ffmpegProcess, ffmpegTracker.exitPromise);
 
-    if (ffmpegTracker.exitState.exitCode && ffmpegTracker.exitState.exitCode !== 0) {
+    if (isFailedExitCode(ffmpegTracker.exitState.exitCode)) {
```
kaeawc, line 372: Good — `isFailedExitCode` now treats `exitCode === null` as a failure and the message surfaces the full ffmpeg argv + stderr (+ an explicit empty-output check). That closes the core diagnostics gap from #2606 where stderr was only surfaced when exitCode was truthy-nonzero.

Two follow-ups: (1) #2606 asked for an end-to-end `start`→`stop` test asserting a non-empty playable `.mp4`; the unit tests assert argv strings + the size check but never run a real transcode, so theres no regression guard against `-c copy` producing empty output. (2) the `-c copy` branch skips `-t`, so `maxDuration` trimming no longer applies for an unscaled remux — worth a comment so its intentional.

kaeawc (commented):
Reviewed against #2606 — correctly fixes both the finalize failure (stream-copy remux for an already-encoded `.mov`) and the swallowed-stderr diagnostics. Two small follow-ups inline.

kaeawc:
Addressed the follow-ups from the review:

- Created #2628 for the real iOS simulator start→stop integration test that asserts a non-empty playable MP4.
- Created #2627 for deciding and testing/documenting maxDurationSeconds behavior in the stream-copy remux path.
- Added an inline comment in the remux branch noting that stream-copy preserves the full capture duration and links the trimming follow-up.

Validation for the update:
- bun test test/features/video/FfmpegVideoProcessingBackend.test.ts
- bun run lint

PR #2617, opened 2026-06-29
Fix iOS highlight runner overlay default

### LINKED ISSUE #2607: highlight on iOS has no working path for non-SDK apps: runner-overlay fallback disabled by default (AUTOMOBILE_IOS_LIVE_HIGHLIGHTS)
### Summary
`highlight` on an iOS device/sim returns failure for any app that doesn't embed the AutoMobile SDK, because the only non-SDK path (the CtrlProxy runner overlay) is **off by default**. So in the common automation case (highlighting elements in an arbitrary/third-party/system app), iOS `highlight` can't render anything — it's not at parity with Android (where highlight works without an app SDK).

### Repro (iOS sim, any system app)
```
highlight { platform:"ios", deviceId:<sim>, shape:{type:"box", bounds:{x:80,y:300,width:240,height:120}, style:{strokeColor:"#FF0000", strokeWidth:6}} }
→ { success:false, error:"Unable to render highlight on iOS: target app SDK highlight bridge unavailable and runner overlay fallback disabled." }
```
Tried against `com.apple.Maps` and Settings — both fail.

### Root cause
`#2582` made the in-app SDK highlight bridge the **primary** path and demoted the runner overlay to a **default-OFF** fallback.
- `ios/control-proxy/Sources/CtrlProxy/CommandHandler.swift:990` — `guard highlightOverlayManager.show(...) else { return error("…SDK highlight bridge unavailable and runner overlay fallback disabled.") }`.
- `HighlightOverlayManager.swift:214` — `render()` does `guard liveOverlayEnabled() else { return false }`.
- `HighlightOverlayManager.swift:248-253` — `defaultLiveOverlayEnabled()` reads `AUTOMOBILE_IOS_LIVE_HIGHLIGHTS` and returns **false** unless it's `1/true/yes`.

So without (a) the target app embedding the AutoMobile SDK *or* (b) launching the runner with `AUTOMOBILE_IOS_LIVE_HIGHLIGHTS=true`, `show()` returns false and `highlight` fails. Neither is the default, so highlight is effectively non-functional on iOS out of the box.

### Impact
- `highlight` is presented as a cross-platform tool, but on iOS it silently has **no working path** for the most common case (any app the tester didn't build with the SDK). This is a parity gap vs Android and surprising for a "visual highlight for debugging" tool.
- Related (worth checking in the same area): the `highlight_response` success default — the TS client uses `success: message.success ?? true`, so a malformed/omitted-`success` native response could be reported as a successful highlight even when nothing drew. (Originally flagged in the #2582 review.)

### Suggested fix
- Enable the runner overlay by default on iOS (so highlight renders without requiring the app to embed the SDK), matching Android — and keep `AUTOMOBILE_IOS_LIVE_HIGHLIGHTS` only as an opt-out; **or**
- If keeping it opt-in is intentional, make the tool description / docs state clearly that iOS highlight requires the embedded SDK or `AUTOMOBILE_IOS_LIVE_HIGHLIGHTS=true`, and surface that requirement in the error (it does name it, but the default-off behavior isn't discoverable).
- Either way, fix the `success ?? true` default so a non-render is never reported as success.

### Environment
`main` @ post-#2585, daemon v0.0.38, iOS sim iPhone 16 Pro (iOS 18.6), local CtrlProxy runner built from current source (advertises `add_highlight`). Found during an iOS tool sweep.

kaeawc:
## Summary
- Enable the iOS CtrlProxy runner highlight overlay by default, with `AUTOMOBILE_IOS_LIVE_HIGHLIGHTS=false` as the opt-out.
- Render iOS runner fallback highlights with Android-style hand-drawn segments: jittered ellipses, variable stroke widths, smoothed/tapered paths, and draw/fade animation.
- Treat malformed iOS `highlight_response` messages without `success` as failures instead of successes.
- Update highlight docs to describe iOS SDK-first rendering with default runner fallback.

Fixes #2607

## Tests
- `bun test test/features/observe/ios/IOSCtrlProxyClient.test.ts`
- `(cd ios/control-proxy && swift test --filter HighlightOverlayManagerTests)`
- `(cd ios/control-proxy && swift test --filter CommandHandlerTests)`
- `bun run build`

## Validation Notes
- `bash scripts/ios/swift-test.sh ios/control-proxy` passes `control-proxy`, but the broader script fails in `XCTestRunner` on a live Reminders plan because `tapOn` could not find text `Done`.
- `bun install` fails in this environment while building the existing `heapdump` native dependency against Node 26.

## Self Review
- Reviewed the final diff for reuse, contracts, docs, and edge cases.
- Hardened the environment parser to trim whitespace before checking opt-out values.
- Added platform-neutral Swift tests for the hand-drawn segment builder so the Android-style visual behavior is covered without UIKit.

### DIFF
--- a/ios/control-proxy/Sources/CtrlProxy/HighlightOverlayManager.swift
+++ b/ios/control-proxy/Sources/CtrlProxy/HighlightOverlayManager.swift
@@ -39,6 +39,8 @@ public struct HighlightOverlayRenderCommand {
     public let strokeColor: String
     public let strokeWidth: Float
     public let dashPattern: [Float]?
+    public let smoothing: String?
+    public let tension: Float?
     public let capStyle: String
     public let joinStyle: String
 }
@@ -96,6 +98,8 @@ public enum HighlightOverlayCommandScaler {
             strokeColor: command.strokeColor,
             strokeWidth: command.strokeWidth,
             dashPattern: command.dashPattern,
+            smoothing: command.smoothing,
+            tension: command.tension,
             capStyle: command.capStyle,
             joinStyle: command.joinStyle
         )
@@ -136,6 +140,9 @@ public struct HighlightOverlayColorComponents: Equatable {
 }
 
 public enum HighlightOverlayCommandBuilder {
+    private static let defaultStrokeWidth: Float = 8
+    private static let defaultPathTension: Float = 0.5
+
     public static func command(for shape: HighlightShape) throws -> HighlightOverlayRenderCommand {
         switch shape.type {
         case "box", "circle":
@@ -147,8 +154,10 @@ public enum HighlightOverlayCommandBuilder {
                 bounds: bounds,
                 points: [],
                 strokeColor: shape.style?.strokeColor ?? "#FF0000",
-                strokeWidth: shape.style?.strokeWidth ?? 3,
+                strokeWidth: shape.style?.strokeWidth ?? defaultStrokeWidth,
                 dashPattern: shape.style?.dashPattern,
+                smoothing: shape.style?.smoothing,
+                tension: shape.style?.tension ?? defaultPathTension,
                 capStyle: shape.style?.capStyle ?? "round",
                 joinStyle: shape.style?.joinStyle ?? "round"
             )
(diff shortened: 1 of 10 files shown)


ios/control-proxy/Sources/CtrlProxy/HighlightOverlayManager.swift
```diff
@@ -200,6 +200,23 @@ public final class HighlightOverlayManager: HighlightOverlayManaging {
...
+
+    public static func liveOverlayEnabled(environment: [String: String]) -> Bool {
+        let flag = environment["AUTOMOBILE_IOS_LIVE_HIGHLIGHTS"]?
+            .trimmingCharacters(in: .whitespacesAndNewlines)
+            .lowercased()
+        if flag == "0" || flag == "false" || flag == "no" {
+            return false
+        }
+        if flag == "1" || flag == "true" || flag == "yes" {
+            return true
```
kaeawc, line 610: `AUTOMOBILE_IOS_LIVE_HIGHLIGHTS` is now a pure opt-**out** (defaults on), so non-SDK apps finally get a working highlight path at Android parity — and `IOSCtrlProxyClient` flips `success ?? true` → `?? false`, so an omitted/malformed `success` isnt reported as a render. Both are exactly what #2607 needed.

The one thing the unit tests cant prove (per the docs own caveat) is that the runner overlay window actually composites over the app-under-test / simulator video capture on every config. Worth a manual on-device confirm: highlight a box on Settings/Maps with the flag unset → visibly drawn + `success:true`.

kaeawc (commented):
Reviewed against #2607 — clean fix that hits all three points: overlay defaults on, `success` no longer defaults true, and Swift + TS tests added.

PR #2618, opened 2026-06-29
Fix critical section plan tool lookup

### LINKED ISSUE #2614: Keep plan-executable tools available inside criticalSection
## Problem

PR #2597 gates `setUIState` behind `debugOnly` to reduce the default MCP tool surface. A follow-up fix kept `setUIState` executable for normal `executePlan` steps by adding a `ToolRegistry.getToolForPlan()` path and marking the tool `planExecutable: true`.

A later review comment pointed out that this does not cover `criticalSection` nested steps. `criticalSection` still resolves each nested step with `ToolRegistry.getTool(step.tool)` in `src/server/criticalSectionTools.ts`, so a non-debug daemon running a critical section step like this can fail even though the same step is valid in a normal plan:

```yaml
- tool: criticalSection
  params:
    lock: form-fill
    deviceCount: 2
    steps:
      - tool: setUIState
        params:
          fields:
            - selector:
                text: Email
              value: user@example.com
```

In non-debug mode, `ToolRegistry.getTool(\"setUIState\")` returns `undefined`, causing `Tool \"setUIState\" not found in registry` from the critical-section executor.

## Context

This came from Codex review feedback on PR #2597 after `setUIState` was moved out of the default MCP-exposed tool list. The first fix handled `executePlan`, but critical sections have a separate nested execution path and did not use the new plan-aware lookup.

Review comment: https://github.com/kaeawc/auto-mobile/pull/2597#discussion_r3489853965

## Expected behavior

Tools that are intentionally hidden from MCP discovery but marked as plan-executable should remain executable anywhere plan steps are executed, including nested `criticalSection` steps.

## Possible fix

Use the same plan-aware lookup path in `criticalSectionTools.ts` that `executePlan` uses, or centralize nested plan-step execution so `executePlan` and `criticalSection` cannot drift on tool availability rules.

Add regression coverage for a non-debug criticalSection containing a `setUIState` nested step.

kaeawc:
## Summary

- Fix `criticalSection` nested step lookup to use the same plan-aware tool resolution as `executePlan`
- Allow explicitly `planExecutable` tools to run from plan paths even when normal MCP discovery/calls are gated by flags such as `--debug` or embedded SDK mode
- Log a warning when plan execution uses a gated `planExecutable` tool, and keep non-plan-executable gated tools blocked

Fixes #2614

## Testing

- `bun test test/server/criticalSectionTools.test.ts`
- `bun test test/server/tools/daemonModeTools.test.ts`
- `bun run lint`
- `bun run build`

### DIFF
--- a/src/server/criticalSectionTools.ts
+++ b/src/server/criticalSectionTools.ts
@@ -132,8 +132,9 @@ const criticalSectionHandler = async (
       );
 
       try {
-        // Get the tool from registry
-        const tool = ToolRegistry.getTool(step.tool);
+        // Critical-section steps are plan steps, so use the same lookup rules
+        // as executePlan for tools hidden from MCP discovery.
+        const tool = ToolRegistry.getToolForPlan(step.tool);
         if (!tool) {
           throw new ActionableError(`Tool "${step.tool}" not found in registry`);
         }
--- a/src/server/toolRegistry.ts
+++ b/src/server/toolRegistry.ts
@@ -252,14 +252,19 @@ class ToolRegistryClass {
     this.timer = timer;
   }
 
-  private isToolAvailable(tool: RegisteredTool): boolean {
+  private getToolAvailabilityGateReasons(tool: RegisteredTool): string[] {
+    const reasons: string[] = [];
     if (tool.debugOnly && !isDebugModeEnabled()) {
-      return false;
+      reasons.push("--debug is disabled");
     }
     if (tool.embeddedSdkOnly && !serverConfig.isEmbeddedSdkEnabled()) {
-      return false;
+      reasons.push("embedded SDK mode is disabled");
     }
-    return true;
+    return reasons;
+  }
+
+  private isToolAvailable(tool: RegisteredTool): boolean {
+    return this.getToolAvailabilityGateReasons(tool).length === 0;
   }
 
   // Register a new tool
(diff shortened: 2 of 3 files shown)


src/server/criticalSectionTools.ts
```diff
@@ -132,8 +132,9 @@ const criticalSectionHandler = async (
       );
 
       try {
-        // Get the tool from registry
-        const tool = ToolRegistry.getTool(step.tool);
+        // Critical-section steps are plan steps, so use the same lookup rules
+        // as executePlan for tools hidden from MCP discovery.
+        const tool = ToolRegistry.getToolForPlan(step.tool);
```
kaeawc, line 137: Right fix — `criticalSection` step resolution now uses `getToolForPlan()` (what `PlanExecutor` uses) instead of `getTool()` (the full MCP-visibility gate), so `planExecutable`-but-hidden tools (from #2597) resolve inside a `criticalSection` block just like they do in `executePlan`. Scoped to plan paths, `getTool()` (MCP discovery / index.ts dispatch) is untouched, and a test asserts hidden tools stay hidden from discovery.

Behavior note: the generalized `getToolForPlan` now also lets `embeddedSdkOnly + planExecutable` tools run from plan paths (previously blocked outright). Intentional and logged, just slightly broader than the literal criticalSection fix — worth calling out in the PR description.

kaeawc (commented):
Reviewed — correct, well-scoped fix for the criticalSection plan-tool-lookup parity gap. One behavior note inline.

PR #2619, opened 2026-06-29
Detect Compose interop hidden regions

### LINKED ISSUE #2374: Detect Compose interop content hiding and signal OCR fallback
## Problem

When an Android app uses Compose `AndroidView` interop to host traditional View-based content (Fragments, RecyclerViews), the Compose interop layer sets `IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS` on the `ViewFactoryHolder` wrapper. This completely hides all child views from the accessibility API — no accessibility service or `uiautomator` can see those nodes.

**Result:** The observe tool returns a near-empty hierarchy for large portions of the screen. The user sees a full channel list, but AutoMobile sees nothing.

### Observed with Slack (Compose + Fragment interop)

The Slack home screen hosts its channel list via:
```
AndroidComposeView
  └── AndroidViewsHandler (IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS)
       └── ViewFactoryHolder (IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS)
            └── FragmentContainerView
                 └── ConstraintLayout (nav_channels_container)
                      └── RecyclerView (recycler_view) — all channel items here
```

The accessibility tree only exposes the top bar and bottom nav. The entire scrollable content area (`[0,368][1440,2752]` — 78% of the screen) has **0 text descendants**.

### What doesn't work

- `FLAG_INCLUDE_NOT_IMPORTANT_VIEWS` — does not override `NO_HIDE_DESCENDANTS`
- `uiautomator dump` — same framework limitation, same empty result
- `findAccessibilityNodeInfosByText()` — respects the same hiding
- Any accessibility service API — the framework enforces the block

### What CtrlProxy CAN detect

From the accessibility tree, the following signals are available at the blocked node:

| Signal | Value |
|--------|-------|
| Ancestor className | `androidx.compose.ui.platform.ComposeView` |
| Node area | 3.4M pixels (78% of screen) |
| Text descendants | 0 |
| Direct children | 1 (covers only ~3% of parent area) |
| className | null (Compose semantic node) |
| Actions | `[accessibility_focus]` only — not interactive |
| Child coverage | Only `[0,290][1440,458]` — vast empty gap below |

### 1. CtrlProxy: Detect and signal content hiding

Add a detection pass after hierarchy extraction that identifies likely Compose interop content-hiding boundaries using the heuristic:

- Node is a descendant of a `ComposeView`
- Node covers >25% of the screen area
- Node has 0 text-bearing descendants
- Node is not itself interactive (no `click`/`scroll` actions)
- Gap between node bounds and child bounds suggests missing content

When detected, include metadata in the hierarchy response:

```json
{
  "contentHiddenRegions": [
    {
      "bounds": { "left": 0, "top": 368, "right": 1440, "bottom": 2752 },
      "reason": "compose-interop-no-hide-descendants",
      "areaPercent": 78
    }
  ]
}
```

### 2. TypeScript MCP server: Screenshot-based fallback

When `contentHiddenRegions` is present in the hierarchy response:

- **For observe:** Take a screenshot, run OCR on the hidden region(s), and merge the OCR-extracted elements into the hierarchy response. Each OCR element gets synthetic bounds from text detection.
- **For tapOn/swipeOn:** When the target text/element is not found in the accessibility hierarchy, check if the target coordinates fall within a `contentHiddenRegion`. If so, use coordinate-based tap/swipe on the OCR-detected element position.
- **For scrolling:** The hidden region likely contains a scrollable container. Coordinate-based swipe gestures within the region bounds should work since `adb input swipe` doesn't rely on the accessibility tree.

### 3. Interaction strategy for hidden content

Since accessibility-based element targeting won't work in hidden regions:

| Action | Strategy |
|--------|----------|
| Tap | OCR detects text + bounds → coordinate tap via `adb input tap` |
| Swipe/Scroll | Coordinate swipe within hidden region bounds |
| Text input | Tap target via coordinates → standard `inputText` flow |
| waitFor | Screenshot + OCR polling instead of hierarchy polling |

## Impact

This affects any app using Compose `AndroidView` interop without properly bridging accessibility semantics. Known affected patterns:
- Compose hosting Fragment-based content (Slack, likely other large apps migrating incrementally to Compose)
- Compose hosting RecyclerView/ListView via `AndroidView`
- Any `AndroidView` wrapped by a parent with `clearAndSetSemantics {}`

## References

- `ViewFactoryHolder` source: `androidx.compose.ui.viewinterop.ViewFactoryHolder`
- Android accessibility flag: `View.IMPORTANT_FOR_ACCESSIBILITY_NO_HIDE_DESCENDANTS` (value 4)
- Verified via `dumpsys activity top` — Views exist with text content but are inaccessible to a11y APIs

kaeawc:
## Summary
- Detect likely Compose AndroidView interop hidden-content regions in CtrlProxy hierarchy extraction.
- Return `contentHiddenRegions` metadata with bounds, reason, and screen area percent.
- Preserve the metadata through Android CtrlProxy TypeScript conversion for observe consumers.

Fixes #2374

## Testing
- `(cd android && ./gradlew :control-proxy:testDebugUnitTest --tests dev.jasonpearson.automobile.ctrlproxy.ViewHierarchyExtractorTest)`
- `bun test test/features/observe/CtrlProxyClient.test.ts -t "should convert accessibility hierarchy to ViewHierarchyResult format"`
- `bun run build`

## Notes
- Full `bun test test/features/observe/CtrlProxyClient.test.ts` still fails in the pre-existing navigation screen-name test: expected `SdkHome`, received `null`.
- `bun install --frozen-lockfile` failed building existing `heapdump` against Node 26, but populated enough dependencies for the focused Bun test and build.


android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/ViewHierarchyExtractor.kt
```diff
@@ -80,6 +85,11 @@ class ViewHierarchyExtractor(private val recompositionStore: RecompositionStore?
               dedupeTextContentDesc,
               accessibilityFocusedNode,
           )
+      val contentHiddenRegions =
+          rootElement?.let {
+            val (screenWidth, screenHeight) = resolveScreenDimensions(it, screenDimensions)
+            detectContentHiddenRegions(it, screenWidth, screenHeight)
```
kaeawc: This only **appends** an optional `contentHiddenRegions` array (gated behind `ComposeView` descent, omitted when empty) and never mutates node visibility/bounds, so non-Compose `observe` output stays byte-identical — nice and safe on the hot path.

Two nits: `countTextBearingNodes` re-walks subtrees per candidate (theoretical O(n^2) on deep Compose trees, bounded in practice by the 25%-area gate); and `directChildCoverage` sums child areas without de-overlapping (only causes false-**negatives**, the safe direction). Consider a test for the multi-window aggregation / `distinctBy { bounds }` dedup path.

kaeawc (commented):
Reviewed — low-risk, purely additive Compose-interop hidden-region detection. Two minor nits inline.

PR #2667, opened 2026-06-29
refactor: consolidate device-vs-session scope resolution in SocketServer

### LINKED ISSUE #2664: Consolidate duplicated device-vs-session resolution in SocketServer.getRequestArgumentScopeKey
## Summary

`SocketServer.getRequestArgumentScopeKey` resolves a forwarding/serialization key through several branches that repeat the same `assignedDevice ? "device:…" : "session:…"` shape. The logic is correct but duplicated, and — more importantly — it accreted branch-by-branch in response to six separate review comments, which is exactly the history that produces a tangle worth consolidating before the next edge case lands on top of it.

**Severity:** Medium (copy-paste in concurrency-critical routing; high future-maintenance risk)

## Where

[`src/daemon/socketServer.ts:414`](https://github.com/kaeawc/auto-mobile/blob/main/src/daemon/socketServer.ts#L414):

```ts
private getRequestArgumentScopeKey(args: unknown): string | undefined {
  if (!args || typeof args !== "object" || Array.isArray(args)) return undefined;
  const record = args as Record<string, unknown>;
  const hasSessionUuid = typeof record.sessionUuid === "string" && record.sessionUuid.length > 0;
  const hasDeviceLabel = typeof record.device === "string" && record.device.length > 0;

  if (hasSessionUuid && hasDeviceLabel) {
    const sessionUuid = this.resolveDeviceLabelSession(record.sessionUuid as string, record.device);
    const assignedDevice = this.getAssignedDeviceForSession(sessionUuid);
    if (assignedDevice) return `device:${assignedDevice}`;   // <-- shape A
    return `session:${sessionUuid}`;                          // <-- shape A
  }

  if (typeof record.deviceId === "string" && record.deviceId.length > 0) {
    return `device:${record.deviceId}`;
  }
  if (hasSessionUuid) {
    const sessionUuid = record.sessionUuid as string;
    const assignedDevice = this.getAssignedDeviceForSession(sessionUuid);
    if (assignedDevice) return `device:${assignedDevice}`;    // <-- shape A again
    return `session:${sessionUuid}`;                          // <-- shape A again
  }
  if (typeof record.__mcpSessionId === "string" && record.__mcpSessionId.length > 0) {
    return this.getImplicitAutolockScopeKey(record.__mcpSessionId, args) ?? `mcp-session:${record.__mcpSessionId}`;
  }
  return undefined;
}
```

The two branches differ only in whether they first run `resolveDeviceLabelSession`; both then collapse to the same "assigned device → `device:`, else `session:`" resolution. `getImplicitAutolockScopeKey` (next method) carries a third copy of the same resolution.

## Why this got in

Introduced in **#2565 "Fix SQLite busy timeout and support parallel daemon requests"** (commit `3dac4045`, closes #2559). This method is the heart of the new keyed-forwarding design that serializes same-`sessionUuid`/`deviceId` calls while letting independent devices run concurrently. The branch structure grew **directly out of review**: the PR received **six P2 comments** all targeting this exact resolution, e.g.:

- *"Key explicit device targets before session IDs"*
- *"Serialize session-bound calls by the resolved device"*
- *"Resolve device-label sessions before keying forwards"* (line 429)
- *"Re-key newly bound session calls by device"*
- *"Serialize implicit autolock calls with their device"* (line 400)
- *"Honor device labels before deviceId when keying"* (line 423)

Each fix added or reordered a branch. The end state is correct (and well-reviewed!), but the per-comment patching left three copies of the device-vs-session resolution. This is the right moment to consolidate — the correctness is now pinned down by review and tests, so a refactor is low-risk.

Introducing PR: https://github.com/kaeawc/auto-mobile/pull/2565
Related issue: #2559

## Relevant prior review feedback

See the six P2 comments above (all on `src/daemon/socketServer.ts` in #2565). They define the **invariants any refactor must preserve**, in priority order: explicit `deviceId`/`device` label > resolved-device-for-session > raw session > implicit autolock/mcp-session. Use them as the test matrix.

## Remediation

Extract the shared resolution into one helper and have each branch feed it:

```ts
private sessionToScopeKey(sessionUuid: string): string {
  const assignedDevice = this.getAssignedDeviceForSession(sessionUuid);
  return assignedDevice ? `device:${assignedDevice}` : `session:${sessionUuid}`;
}
```

Then the two `hasSessionUuid` branches become one-liners (`return this.sessionToScopeKey(resolvedUuid)`), and `getImplicitAutolockScopeKey` reuses it too. Keep the branch *ordering* exactly as the six review comments dictate.

**Effort:** Quick fix (pure extraction).

**Validate:** the keyed-forwarding behavior is covered by `test/daemon/socketServerMcpForwardSerialization.test.ts` and `test/daemon/manager.test.ts`. Run `bun test test/daemon && bun run lint && bun run build`. Add a table-driven test enumerating the six precedence cases above if one doesn't already exist — that pins the invariants so the next edge case can't silently reorder them.

## How to make this area greener

- **When a method earns 3+ review comments about its branch precedence, that's the signal to extract a single resolver + a table-driven precedence test** rather than leaving the branches as the spec. The precedence rules are domain logic; encode them as data (an ordered list of predicates → key builders), not as nested `if`s.
- A precedence table makes the *next* edge case (e.g. a third device-targeting field) a one-row addition with an obvious insertion point, instead of another copy of shape A.
- The surrounding daemon work is otherwise a green flag (e.g. `withRecoverableReconnect` in `daemonMcpProxy.ts` collapsed three copy-pasted retry blocks into one helper) — this is the same move applied to scope-keying.

kaeawc:
## Summary

Consolidates the three duplicated `assignedDevice ? "device:…" : "session:…"` resolutions in `SocketServer.getRequestArgumentScopeKey` / `getImplicitAutolockScopeKey` into a single `sessionToScopeKey()` helper. The logic accreted branch-by-branch across six P2 review comments on #2565, leaving three identical copies of the device-vs-session collapse. This is a pure extraction — correctness is already pinned by review + tests, so the refactor is low-risk.

Closes #2664.

## What changed

- Extracted `private sessionToScopeKey(sessionUuid)` — the one resolver every session-keyed branch feeds (label branch, raw-session branch, and implicit-autolock resolution).
- The two `hasSessionUuid` branches and `getImplicitAutolockScopeKey` are now one-liners that call it.
- **Branch ordering is unchanged**, exactly as the six review comments dictate: device label > explicit `deviceId` > resolved-device-for-session > raw session > implicit autolock / mcp-session > `undefined`.

## Tests

Added `test/daemon/socketServerScopeKey.test.ts` — a table-driven precedence test enumerating all the invariants the issue named (explicit device beats raw session, label honored over stale `deviceId`, unmapped-label fallbacks, autolock resolution, non-object → `undefined`). This pins the precedence as data so the next device-targeting field is a one-row addition with an obvious insertion point instead of another copy of the same shape.

Interface + Fake + FakeTimer, no real device/network, 14 tests in ~146ms.

## Validation

- `bun test test/daemon` — 440 pass, 0 fail
- `bun run lint` — clean
- `bun run build` — success

### DIFF
--- a/src/daemon/socketServer.ts
+++ b/src/daemon/socketServer.ts
@@ -420,31 +420,33 @@ export class UnixSocketServer {
     const record = args as Record<string, unknown>;
     const hasSessionUuid = typeof record.sessionUuid === "string" && record.sessionUuid.length > 0;
     const hasDeviceLabel = typeof record.device === "string" && record.device.length > 0;
-    if (hasSessionUuid && hasDeviceLabel) {
-      const sessionUuid = this.resolveDeviceLabelSession(record.sessionUuid as string, record.device);
-      const assignedDevice = this.getAssignedDeviceForSession(sessionUuid);
-      if (assignedDevice) {
-        return `device:${assignedDevice}`;
-      }
-      return `session:${sessionUuid}`;
-    }
 
+    // Precedence (pinned by #2565 review): a device label resolves the mapped session before
+    // an explicit deviceId, which in turn beats a raw session, which beats implicit autolock.
+    if (hasSessionUuid && hasDeviceLabel) {
+      return this.sessionToScopeKey(this.resolveDeviceLabelSession(record.sessionUuid as string, record.device));
+    }
     // An explicit target device must serialize by physical device even when a session is present.
     if (typeof record.deviceId === "string" && record.deviceId.length > 0) {
       return `device:${record.deviceId}`;
     }
     if (hasSessionUuid) {
-      const sessionUuid = record.sessionUuid as string;
-      const assignedDevice = this.getAssignedDeviceForSession(sessionUuid);
-      if (assignedDevice) {
-        return `device:${assignedDevice}`;
-      }
-      return `session:${sessionUuid}`;
+      return this.sessionToScopeKey(record.sessionUuid as string);
     }
     if (typeof record.__mcpSessionId === "string" && record.__mcpSessionId.length > 0) {
       return this.getImplicitAutolockScopeKey(record.__mcpSessionId, args) ?? `mcp-session:${record.__mcpSessionId}`;
     }
     return undefined;
+  }
+
+  /**
+   * Resolve a session UUID to its forwarding scope key: the bound physical device when one is
+   * assigned (so independent devices serialize together), otherwise the raw session. This is the
+   * single resolver every session-keyed branch feeds, including implicit autolock resolution.
+   */
+  private sessionToScopeKey(sessionUuid: string): string {
+    const assignedDevice = this.getAssignedDeviceForSession(sessionUuid);
+    return assignedDevice ? `device:${assignedDevice}` : `session:${sessionUuid}`;
   }
 
   private getImplicitAutolockScopeKey(mcpSessionId: string, args: unknown): string | undefined {
@@ -459,8 +461,7 @@ export class UnixSocketServer {
       if (!autolockSession) {
         return undefined;
       }
-      const assignedDevice = this.getAssignedDeviceForSession(autolockSession);
-      return assignedDevice ? `device:${assignedDevice}` : `session:${autolockSession}`;
+      return this.sessionToScopeKey(autolockSession);
     } catch (error) {
       logger.debug(`Unable to resolve autolock session for MCP session ${mcpSessionId}: ${error}`);
       return undefined;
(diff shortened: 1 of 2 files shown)


kaeawc:
Note on red CI checks: the failing **iOS** / **XCTestRunner Simulator Tests** jobs are a pre-existing, repo-wide CI failure — the `ObjCExceptionCatcher.m` test build fails to compile under Xcode 26.3 before any test runs (`No .xctestrun file found`). This PR changes only two TypeScript files (`src/daemon/socketServer.ts` + a new test) and touches **zero** native/iOS sources, so it cannot be the cause. The same two checks are red on recently-merged iOS PR #2653. A re-run reproduced the failure identically (not flaky). All TypeScript checks — Node Build & Test on macOS/Ubuntu/Windows, Fast Validation, benchmarks, CodeQL — pass.

_🤖 Posted by [Claude Code](https://claude.com/claude-code)_

PR #2668, opened 2026-06-29
refactor: extract shared network mock-rule serializer (#2662)

### LINKED ISSUE #2662: iOS network mock-rule sync is copy-pasted from Android (require() in ESM, any-typed, debug-swallowed)
## Summary

`IOSCtrlProxyClient.syncNetworkMockRulesToDevice` is a near-verbatim clone of `AndroidCtrlProxyClient.syncNetworkStateToDevice`'s mock-rule mapping. The duplication carried two smells into a second file: a CommonJS `require()` inside an otherwise-ESM module, and an untyped `(r: any)` map that erases the `NetworkState` mock shape. The iOS copy also swallows sync failures at `debug` level.

**Severity:** Medium (copy-paste + type erasure + silent failure on a feature path)

## Where

iOS — [`src/features/observe/ios/IOSCtrlProxyClient.ts:699`](https://github.com/kaeawc/auto-mobile/blob/main/src/features/observe/ios/IOSCtrlProxyClient.ts#L699):

```ts
private syncNetworkMockRulesToDevice(): void {
  if (!serverConfig.isNetworkMockableEnabled()) return;
  try {
    const { NetworkState } = require("../../../server/NetworkState");   // CommonJS in an ESM file
    const state = NetworkState.getInstance();
    const rules = Array.from(state.getMocks().values()).map((r: any) => ({  // any erases the typed shape
      mockId: r.mockId, host: r.host, path: r.path, method: r.method,
      limit: r.limit, remaining: r.limit, statusCode: r.statusCode,
      responseHeaders: r.responseHeaders, responseBody: r.responseBody, contentType: r.contentType,
    }));
    this.sendMessage(JSON.stringify({ type: "set_network_mock_rules", rules }));
  } catch (e) {
    logger.debug(`[IOSCtrlProxyClient] Failed to sync network mock rules on reconnect: ${e}`);  // swallowed at debug
  }
}
```

Android — [`src/features/observe/android/AndroidCtrlProxyClient.ts:1059`](https://github.com/kaeawc/auto-mobile/blob/main/src/features/observe/android/AndroidCtrlProxyClient.ts#L1059) — the identical 10-field rule object, the identical `require()`, the identical `(r: any)`:

```ts
private syncNetworkStateToDevice(): void {
  try {
    const { NetworkState } = require("../../../server/NetworkState");
    const state = NetworkState.getInstance();
    const mocks = state.getMocks();
    const rules = Array.from(mocks.values()).map((r: any) => ({
      mockId: r.mockId, host: r.host, path: r.path, method: r.method,
      limit: r.limit, remaining: r.limit, statusCode: r.statusCode,
      responseHeaders: r.responseHeaders, responseBody: r.responseBody, contentType: r.contentType,
    }));
    this.sendMessage(JSON.stringify({ type: "set_network_mock_rules", rules }));
    // ... Android additionally re-syncs error-simulation state
  }
}
```

## Why this got in

The iOS copy was added in **#2580 "feat(ios): support mockNetwork rules"** (commit `85a7d4d0`, closes #2500). The PR description states the goal was to reach parity with Android "by syncing rules over the existing host → iOS CtrlProxy WebSocket path" — so the Android implementation was the reference, and the mapping was copied wholesale rather than extracted. The `require()` and `(r: any)` were inherited from the Android original (pre-existing debt) and propagated.

Introducing PR: https://github.com/kaeawc/auto-mobile/pull/2580
Related issue: #2500

## Relevant prior review feedback

#2580 review raised a **P2** on the SDK side — *"Read the full POST body before decoding mock rules"* (`SdkHierarchyServer.swift`), about large mock payloads arriving in chunks. That's the device-side counterpart to this host-side serializer; both halves of the mockNetwork path would benefit from being touched together. No reviewer flagged the host-side duplication, which is why it landed.

## Remediation

1. **Extract the shared mapping** into one helper consumed by both clients (e.g. `buildNetworkMockRules(state): NetworkMockRule[]` in a shared module next to `NetworkState`), so the 10-field object lives once.
2. **Replace `require()` with a top-level `import`** — `NetworkState` is already importable; the dynamic require is a copy-paste artifact, not a cycle-breaker (verify no import cycle; if one exists, use a typed `import type` + lazy getter rather than untyped `require`).
3. **Type the map** — `state.getMocks()` returns a typed map; drop `(r: any)` and let the rule interface flow through.
4. **Raise the catch to `warn`** (or narrow it) — a failure to sync mock rules silently breaks the feature for the user; `debug` hides it.

**Effort:** Moderate.

**Validate:** `bun test test/features/observe/ios/ test/features/observe/android/ && bun run lint && bun run build`. The mockNetwork reconnect-sync behavior is covered by the #2580 tests.

## How to make this area greener

- **Android and iOS CtrlProxy clients share a large duplicated surface** (network state, reconnect, message dispatch). Each new cross-platform feature currently means copying a method into the sibling file. A shared `CtrlProxyNetworkSync` collaborator (injected, testable) would let both clients delegate instead of clone — the same interface+fake pattern the repo already applies well elsewhere (e.g. `DeviceConnectionLostNotifier`, which #2580-era work introduced as a green flag).
- **Ban `require()` in `src/` via lint** (`@typescript-eslint/no-require-imports`) so CommonJS can't sneak into the ESM codebase through copy-paste.
- **Treat feature-path failures as `warn`+ by default**; reserve `debug` for genuinely-expected, recoverable conditions.

kaeawc:
## What

Extracts the host → device mock-rule serializer into a single typed helper,
`buildNetworkMockRules(state)` in [`src/server/networkMockRules.ts`](src/server/networkMockRules.ts),
and routes all three call sites through it:

- `IOSCtrlProxyClient.syncNetworkMockRulesToDevice`
- `AndroidCtrlProxyClient.syncNetworkStateToDevice` (keeps its extra error-simulation re-sync)
- `networkTools.syncMockRulesToDevice` (a third clone of the same 10-field map)

Each was an independent copy of the same 10-field object mapping. The helper is the lone owner now.

## Why

Resolves #2662. The iOS copy (added in #2580 to reach Android parity) carried three smells, two of them inherited from the Android original:

1. **CommonJS `require()` in an ESM module** — `const { NetworkState } = require("../../../server/NetworkState")`. Verified this was a copy-paste artifact, not a cycle-breaker: `NetworkState`'s import closure (`resourceRegistry`, `logger`, `SystemTimer`) never reaches the observe clients, so a top-level `import` is safe. Replaced in both clients.
2. **`(r: any)` type erasure** — dropped; the typed `MockRule` shape now flows through `buildNetworkMockRules`.
3. **Silent failure** — the iOS reconnect-sync catch logged at `debug`, hiding a broken feature path. Raised to `warn`.

## Validation

- `bun test test/server/networkMockRules.test.ts` — 5 new unit tests (TDD; empty state, full field map, `remaining` reinit from `limit`, null-limit passthrough, count). All pass.
- `bun test test/features/observe/ios/ test/features/observe/android/` — 84 pass.
- `bun test test/server/networkTools.test.ts` — passes.
- `bun run lint` — clean.
- `bun run build` — succeeds.
- Verified no `require(` remains in either CtrlProxy client.

Closes #2662

### DIFF
--- a/src/features/observe/android/AndroidCtrlProxyClient.ts
+++ b/src/features/observe/android/AndroidCtrlProxyClient.ts
@@ -60,6 +60,8 @@ import { serverConfig } from "../../../utils/ServerConfig";
 import { TelemetryRecorder } from "../../telemetry/TelemetryRecorder";
 import { getPerformanceMonitor } from "../../performance/PerformanceMonitor";
 import type { StackTraceElement } from "../../../server/failuresResources";
+import { NetworkState } from "../../../server/NetworkState";
+import { buildNetworkMockRules } from "../../../server/networkMockRules";
 import type {
   PreferenceFile,
   KeyValueEntry,
@@ -1058,25 +1060,13 @@ export class AndroidCtrlProxyClient extends DeviceServiceClient implements Andro
 
   private syncNetworkStateToDevice(): void {
     try {
-      const { NetworkState } = require("../../../server/NetworkState");
       const state = NetworkState.getInstance();
 
-      // Always sync mock rules on reconnect — use limit (not remaining) so
-      // the device-side store reinitializes fresh counts. Sending an empty
-      // list clears stale rules that may linger from a previous connection.
-      const mocks = state.getMocks();
-      const rules = Array.from(mocks.values()).map((r: any) => ({
-        mockId: r.mockId,
-        host: r.host,
-        path: r.path,
-        method: r.method,
-        limit: r.limit,
-        remaining: r.limit,
-        statusCode: r.statusCode,
-        responseHeaders: r.responseHeaders,
-        responseBody: r.responseBody,
-        contentType: r.contentType,
-      }));
+      // Always sync mock rules on reconnect — buildNetworkMockRules uses limit
+      // (not remaining) so the device-side store reinitializes fresh counts.
+      // Sending an empty list clears stale rules that may linger from a
+      // previous connection.
+      const rules = buildNetworkMockRules(state);
       this.sendMessage(JSON.stringify({ type: "set_network_mock_rules", rules }));
 
       // Always re-sync error simulation state (including disabled) so the
--- a/src/features/observe/ios/IOSCtrlProxyClient.ts
+++ b/src/features/observe/ios/IOSCtrlProxyClient.ts
@@ -34,6 +34,8 @@ import { IOSCtrlProxyManager, CtrlProxyIosManager } from "../../../utils/IOSCtrl
 import { PlatformDeviceManagerFactory } from "../../../utils/factories/PlatformDeviceManagerFactory";
 import { NavigationGraphManager } from "../../navigation/NavigationGraphManager";
 import { serverConfig } from "../../../utils/ServerConfig";
+import { NetworkState } from "../../../server/NetworkState";
+import { buildNetworkMockRules } from "../../../server/networkMockRules";
 import { NavigationScreenshotManager } from "../../navigation/NavigationScreenshotManager";
 import {
   HierarchyNavigationDetector,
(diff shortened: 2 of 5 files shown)


kaeawc:
The two red iOS checks (**XCTestRunner Simulator Tests** → **iOS** aggregator) are a **pre-existing, repo-wide CI failure**, not caused by this PR:

- This PR is **TypeScript-only** — it touches no Swift/ObjC. The failure is a `TEST BUILD FAILED` while compiling `ios/control-proxy/Sources/ObjCExceptionCatcher/ObjCExceptionCatcher.m` under Xcode 26.3 (last edited in #2161, untouched here), surfacing as `No .xctestrun file found`.
- The same **XCTestRunner Simulator Tests** check is currently failing on every other open PR (e.g. #2671, #2672, #2673), confirming it is environmental/infra rather than change-specific.

All change-relevant gates are green: **Node TypeScript Build and Test** passes on macOS/Ubuntu/Windows, plus Fast Validation, MCP Benchmarks, Memory Leak Detection, CodeQL, and Bun Security Audit.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #2734, opened 2026-06-30
fix(ios): deliver allocated CtrlProxy port to the simulator/device runner via xctestrun (#2731)

### LINKED ISSUE #2731: fix(ios): simulator runner ignores allocated port (binds 8765) when Android is connected → client/runner port mismatch, all iOS ops fail
## Summary

When an **Android device and an iOS simulator are connected at the same time**, all iOS CtrlProxy operations (observe, gestures, highlight, sqlQuery, etc.) fail. The daemon's iOS client connects to an allocated non-default port (e.g. `8767`) but the simulator runner always binds its hardcoded `defaultPort = 8765`, because the allocated port never reaches the runner process inside the simulator. The runner is healthy the whole time — it's just on the wrong port from the daemon's perspective.

This defeats the intent of the recent mixed-device port-allocation work (#2716 / #2712 / #2718): the daemon allocates a collision-free port, but the runner is never told to use it.

## Environment

- macOS 14.6 (darwin 24.6.0), Apple Silicon
- Xcode 26.3, iOS 26.2 simulator (iPhone 16 Pro)
- Daemon built from current `main` (post-v0.0.39, 0.0.40 RC), `bun` runtime
- One Android emulator (`emulator-5554`, API 35) **and** one booted iOS simulator

## Reproduction

1. Boot an Android emulator and let the daemon attach its ctrlproxy (this creates an `adb forward tcp:8765`).
2. Boot an iOS simulator.
3. `observe { platform: "ios", deviceId: <sim> }`.

**Expected:** iOS hierarchy returned.
**Actual:** `Failed to retrieve iOS view hierarchy from CtrlProxy iOS`, `screenSize: 0x0`. Daemon log shows an endless reconnect loop to `ws://localhost:8767/ws` ("Failed to connect", failure count climbing). `doctor --ios` reports:

```
iOS Observe Round Trip: FAIL
  runnerPort=8767; clientPort=8767; connected=false; screenSize=0x0;
  hierarchyError=iOS CtrlProxy runner is not running
```

## Evidence the runner is actually healthy — just on the wrong port

```
$ lsof -nP -iTCP:8765-8770 -sTCP:LISTEN
adb        ...  TCP 127.0.0.1:8765 (LISTEN)          # Android ctrlproxy forward (IPv4)
CtrlProxy  ...  TCP *:8765 (LISTEN)                   # iOS runner bound default 8765 (IPv6)
```

Probing the iOS runner directly on **8765** returns a complete, healthy handshake (full 0.0.40 command set incl. `add_highlight`, `execute_sql`, `request_shake`, `request_press_button`, `request_multi_finger_swipe`, `set_network_mock_rules`) plus live `performance_update` frames. The runner works perfectly — the daemon's client is just pointed at 8767.

## Root cause

`src/utils/IOSCtrlProxyManager.ts` launches the simulator runner via `xcodebuild test-without-building` and passes the allocated port through process env:

```ts
// IOSCtrlProxyManager.ts ~1024-1057
// "SIMCTL_CHILD_* prefixed vars are forwarded by simctl (which xcodebuild uses internally
//  on simulators) to the XCUITest runner process after stripping the prefix."   <-- INVALID for `xcodebuild test`
const childEnv = {
  CTRL_PROXY_IOS_PORT: String(this.servicePort),               // set on the host xcodebuild process — does NOT cross into the simulated runner
  SIMCTL_CHILD_CTRL_PROXY_IOS_PORT: String(this.servicePort),   // only honored by `simctl spawn`, NOT `xcodebuild test`
  ...
};
this.processExecutor.spawn("xcodebuild test-without-building -xctestrun ... ", [], { env: { ...process.env, ...childEnv } });
```

The runner reads the **unprefixed** var from its own (simulator) environment:

```swift
// ios/control-proxy/Tests/CtrlProxyUITests/CtrlProxyUITests.swift:100
if let portString = ProcessInfo.processInfo.environment["CTRL_PROXY_IOS_PORT"], ... else default 8765
```

But neither var reaches it:
- `CTRL_PROXY_IOS_PORT` is set on the **host** `xcodebuild` process; host env does not propagate into the **simulated** test-runner process.
- `SIMCTL_CHILD_*` stripping/forwarding is a `simctl spawn`/`launch` feature — `xcodebuild test-without-building` does not route the test runner through `simctl spawn`, so it is ignored.

Result: the runner finds no `CTRL_PROXY_IOS_PORT` and falls back to `CtrlProxy.defaultPort = 8765` (`ios/control-proxy/Sources/CtrlProxy/CtrlProxy.swift:9`).

The env-passing block was last modified in #2718 (0.0.40 window). The `xcodebuild test-without-building` simulator path predates it, but combined with the new non-default port allocation this is now a live breakage.

## Why pure-iOS (no Android) works

With no Android device there is no `adb forward tcp:8765`, so `PortManager.allocate` returns `8765` for the iOS service. The runner's hardcoded default `8765` then coincidentally matches the client — so single-platform iOS works and this stays hidden until an Android device is also connected.

## Suggested fix (one of)

- Inject `CTRL_PROXY_IOS_PORT` into the **xctestrun's** `EnvironmentVariables` / `TestingEnvironmentVariables` (the daemon already owns/writes the xctestrun), so it reaches the simulated runner; **or**
- Pre-seed the simulator env before `xcodebuild test`: `xcrun simctl spawn <udid> launchctl setenv CTRL_PROXY_IOS_PORT <port>`; **or**
- Have the daemon discover the runner's actual listening port (it already probes `DEFAULT_PORT` in `findExternalDirectCtrlProxyProcess`) and bind the client to it instead of assuming `servicePort`.

Also remove/correct the misleading comment at `IOSCtrlProxyManager.ts:1025-1026`.

## Severity

**High.** iOS automation is completely non-functional in the common multi-platform setup (Android + iOS connected together). It fails closed (clear doctor error, no data loss), but there is no workaround short of disconnecting all Android devices.

kaeawc:
## Summary
Fixes #2731: when an Android device and an iOS simulator are connected at the same time, all iOS CtrlProxy operations fail. The daemon allocates a collision-free non-default port (e.g. `8767`) but the in-simulator runner always binds its hardcoded `defaultPort = 8765`, so the client and runner disagree on the port and every iOS op fails closed (`doctor` shows `runnerPort=8767; clientPort=8767; connected=false`, while `lsof` shows the runner healthy on `8765`).

## Root cause
The simulator runner is launched via `xcodebuild test-without-building` and the allocated port was passed through the **host xcodebuild process env** (`CTRL_PROXY_IOS_PORT`) and `SIMCTL_CHILD_*` vars. Neither reaches the in-simulator runner:

- Host process env does **not** cross into the simulated test-runner process.
- `SIMCTL_CHILD_*` stripping/forwarding is a `simctl spawn`/`launch` feature — `xcodebuild test-without-building` does not route the runner through `simctl spawn`, so it is ignored.

The runner reads the **unprefixed** `CTRL_PROXY_IOS_PORT` from its own `ProcessInfo.environment` (`CtrlProxyUITests.swift:100`), finds nothing, and falls back to `CtrlProxy.defaultPort = 8765`. Pure-iOS setups stayed green only because with no Android device the iOS port allocates to `8765`, coincidentally matching the runner's default. The mixed-device port-allocation work (#2716/#2712/#2718) made the iOS port reliably ≠ 8765, activating the latent break.

The only channel that reaches the in-simulator runner is the xctestrun's per-target **`EnvironmentVariables`** dict, which `xcodebuild` injects into the test host (the runner app) — per `xcodebuild.xctestrun(5)`.

## Fix
Option A from the issue.

- New `src/utils/ios-cmdline-tools/XctestrunPlist.ts`: a minimal ordered XML-plist reader/writer (`parsePlist`/`buildPlist`) plus `injectUITestEnvironment`, which merges env vars into the `EnvironmentVariables` dict of every UI-test bundle (`IsUITestBundle`).
- New `IOSCtrlProxyBuilder.writeRunnerEnvironment(xctestrunPath, env, deviceId)`: injects the allocated port (+ timeout/bundleId/deviceId) and writes a **per-launch copy** named `automobile-runner-<deviceId>.xctestrun` in the **same** `Build/Products` dir (so `__TESTROOT__` still resolves). The copy is keyed by deviceId to avoid multi-device races and is excluded from the source-discovery/cleanup globs so it can never be re-selected as the source xctestrun.
- `IOSCtrlProxyManager` simulator + physical paths now point `xcodebuild` at the per-launch xctestrun. Dropped the dead `SIMCTL_CHILD_*` vars (simulator) and the ineffective bare `CTRL_PROXY_IOS_PORT=` build-setting token (physical), and rewrote the misleading comment.
- **Kept** the unprefixed `CTRL_PROXY_IOS_PORT` + `AUTOMOBILE_DEVICE_ID` on the host xcodebuild process env — they are *not* dead: the daemon reads them via `ps eww` for process discovery/ownership/recovery (`findExternalXcodebuildCtrlProxyProcess`, `isDaemonManagedSimulatorXcodebuildProcess`). The physical path now also sets them on the host env for the same reason.

No test was relaxed to pass; the previous `spawn.options.env` assertions that encoded the wrong "host env reaches the runner" assumption were replaced with assertions about the channel that actually reaches the runner (the xctestrun).

## Test plan
- New `test/utils/ios-cmdline-tools/XctestrunPlist.test.ts` — parse/build round-trip + `injectUITestEnvironment` (UI-target-only, preserves/overwrites entries, creates missing dict, ignores non-UI targets).
- New `IOSCtrlProxyBuilder.writeRunnerEnvironment` tests — writes a per-launch copy carrying the injected port without mutating the source; excluded from `getXctestrunPath` (incl. the platform-agnostic glob) and `cleanStaleXctestrunFiles`; device-id filename sanitized; throws an actionable error when the xctestrun has no UI-test bundle.
- `IOSCtrlProxyManager` tests:
  - simulator: spawn points at the per-launch xctestrun, `writeRunnerEnvironment` receives the allocated `8767`, host env keeps the identity vars, no `SIMCTL_CHILD_*`.
  - **boundary**: using the real `writeRunnerEnvironment`, the port the runner *would* read from the spawned `-xctestrun`'s UI-test `EnvironmentVariables` equals `manager.getServicePort()` — the assertion that would have caught this bug (runner `8765` vs client `8767`).
  - physical: spawn points at the per-launch xctestrun, no bare `CTRL_PROXY_IOS_PORT=` build-setting token, host env keeps the identity vars.
- `bun test` → 5226 pass / 0 fail (410 files); `eslint --fix` and `bun build.ts` → green.

### Not in scope / follow-up
- The `doctor` iOS round-trip check (`src/doctor/checks/ios.ts`) still derives both `runnerPort` and `clientPort` from `getServicePort()`, so it structurally cannot detect a runner/client port mismatch (issue's Option C). Making it meaningful requires the runner to report its actual bound port back; left as a follow-up.
- **Verify on hardware first:** with one Android emulator + one iOS simulator connected, `observe { platform: "ios" }` returns a hierarchy (not `screenSize: 0x0`), confirming `xcodebuild` forwards the xctestrun `EnvironmentVariables` into the runner so it binds the allocated port.

Refs #2731

### DIFF
--- a/src/utils/IOSCtrlProxyBuilder.ts
+++ b/src/utils/IOSCtrlProxyBuilder.ts
@@ -67,6 +67,12 @@ type IOSCtrlProxyBundleMetadata = {
  * Handles release bundle download and extraction for CtrlProxy
  */
 export class IOSCtrlProxyBuilder {
+  /**
+   * Filename prefix for the per-launch xctestrun copies written by
+   * {@link writeRunnerEnvironment}. Distinct from the build-products xctestrun so
+   * the copies are excluded from source discovery/cleanup globs.
+   */
+  private static readonly RUNNER_XCTESTRUN_PREFIX = "automobile-runner-";
   private static readonly DEFAULT_PROJECT_ROOT = process.cwd();
   private static readonly DEFAULT_DERIVED_DATA_PATH = "/tmp/automobile-ctrl-proxy";
   private static readonly DEFAULT_SCHEME = "CtrlProxyApp";
@@ -209,7 +215,10 @@ export class IOSCtrlProxyBuilder {
     const productsDir = path.join(this.config.derivedDataPath, "Build", "Products");
     try {
       const files = await fs.readdir(productsDir);
-      const xctestrunFiles = files.filter(file => file.endsWith(".xctestrun"));
+      const xctestrunFiles = files.filter(
+        file => file.endsWith(".xctestrun") &&
+          !file.startsWith(IOSCtrlProxyBuilder.RUNNER_XCTESTRUN_PREFIX)
+      );
       if (xctestrunFiles.length === 0) {
         return null;
       }
@@ -289,7 +298,7 @@ export class IOSCtrlProxyBuilder {
       const safeDeviceId = deviceId.replace(/[^A-Za-z0-9._-]/g, "_") || "device";
       const outputPath = path.join(
         path.dirname(xctestrunPath),
-        `automobile-runner-${safeDeviceId}.xctestrun`
+        `${IOSCtrlProxyBuilder.RUNNER_XCTESTRUN_PREFIX}${safeDeviceId}.xctestrun`
       );
       await fs.writeFile(outputPath, buildPlist(root), "utf-8");
       logger.info(
@@ -311,7 +320,10 @@ export class IOSCtrlProxyBuilder {
     const productsDir = path.join(this.config.derivedDataPath, "Build", "Products");
     try {
       const files = await fs.readdir(productsDir);
-      const xctestrunFiles = files.filter(file => file.endsWith(".xctestrun"));
+      const xctestrunFiles = files.filter(
+        file => file.endsWith(".xctestrun") &&
+          !file.startsWith(IOSCtrlProxyBuilder.RUNNER_XCTESTRUN_PREFIX)
+      );
       if (xctestrunFiles.length <= 1) {
         return;
       }
--- a/src/utils/IOSCtrlProxyBuilder.ts
+++ b/src/utils/IOSCtrlProxyBuilder.ts
@@ -19,6 +19,13 @@ import {
 } from "./IOSCtrlProxyBundleDownloader";
 import { hashAppBundle } from "./ios-cmdline-tools/AppBundleHasher";
 import { resolvePathFromDaemonLaunchWorkingDirectory } from "./workingDirectory";
+import {
+  buildPlist,
+  injectUITestEnvironment,
+  parsePlist,
+  type PlistValue
+} from "./ios-cmdline-tools/XctestrunPlist";
+import { toActionableError } from "../models/ActionableError";
 
 /**
  * Result of CtrlProxy download/install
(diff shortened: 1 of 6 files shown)


kaeawc:
Filed the doctor follow-up from the description as #2735 (make the iOS round-trip check compare the runner's *reported* bound port against the client port). The hardware-verification note stays a merge gate for this PR rather than a tracked issue.

Issue #2732, opened 2026-06-30
Multiple checkouts share one per-uid daemon socket → wrong-build daemon serves frontend (version skew, opaque "Unknown tool" -32603)


kaeawc:
## Summary

On a machine with **two checkouts** of AutoMobile (e.g. a worktree RC build + the main-repo build), they share **one per-uid daemon socket**, so a daemon built from one checkout can silently serve an MCP frontend from the other. After a daemon restart the backend can swap to a **different build** than the frontend, producing version skew. The most visible symptom: a tool the frontend advertises fails at the daemon with `MCP error -32603: Unknown tool: <name>` — and, more dangerously, tool calls execute against the *wrong build's code* with no warning.

## How it manifested (real session)

1. Frontend (MCP stdio) connected to a daemon built from the **worktree** dist (had the `getPreference`/`setPreference` tools, #2717). `setPreference` worked.
2. The daemon was killed (unrelated recovery).
3. On the next call, the session auto-recovered — but the respawned daemon came from the **main-repo** dist (older, no #2717).
4. `setPreference` then returned `MCP error -32603: Unknown tool: setPreference`, because the frontend still advertised the tool (cached) but the main-dist daemon has no such handler.

Both daemons report `Version: 0.0.39` (package.json isn't bumped pre-release), so nothing detects the swap.

## Root cause (all confirmed in source)

1. **One socket per uid, machine-wide.** `src/daemon/constants.ts:36-42` derives socket/pid/lock as `/tmp/auto-mobile-daemon-${uid}.*`. Both checkouts fall through to the same default. The first daemon to bind serves **all** frontends from **all** checkouts.
2. **Daemon spawns from the spawner's own build.** `resolveDaemonLaunchCommand()` (`src/daemon/manager.ts:160-183`) uses `process.execPath` + `process.argv[1]`. Correct in isolation — but it means *whichever frontend wins the restart race* respawns the daemon from *its own* build. The live frontend here was the main-repo build, so recovery produced a main-dist daemon.
3. **Checkouts evict each other.** `startUnlocked` (`manager.ts:415-453`) SIGKILLs every other `--daemon-mode` process via a build-agnostic match (`manager.ts:83-86`).
[...]

Issue #2599, opened 2026-06-29
Daemon restart desyncs connected MCP clients: all tool calls + resource reads return opaque 'Session not found' with no auto-recovery


kaeawc:
### Summary
When the daemon process is restarted out-of-band — which the daemon is *designed* to allow (takeover on `--daemon start`, version changes, env changes, crash recovery) — every already-connected MCP client (the stdio proxy / IDE) gets wedged: **all** forwarded tool calls **and** resource reads fail with `Session not found`, persistently, until the client's MCP connection is manually restarted. The daemon itself is healthy and serving; only the existing client's session is stale, and nothing re-establishes it.

### Reproduced 3× in one session
Each time I restarted the daemon (e.g. `--daemon restart`, or kill + `--daemon start` to change env), the existing MCP frontend immediately began returning:
```
Error: Session not found
Error: MCP error -32603: Failed to read resource from daemon: Session not found
```
for every `tools/call` (observe, getDeviceState, sqlQuery, …) and every resource read. A daemon restart alone did **not** fix it; only reconnecting the MCP server / restarting the client recovered it. (Confirmed the daemon was alive and owned the socket the whole time.)

### Mechanism
- The daemon's `UnixSocketServer` creates a per-connection socket session: `const sessionId = randomUUID()` + `this.sessions.set(sessionId, session)` on connect (`src/daemon/socketServer.ts:141,149`), and `this.sessions.delete(sessionId)` on close/error (`:193-199`). `handleRequest()` returns `Session not found` when the session isn't in the map (`:232-238`).
- There's also an MCP-streamable session (`__mcpSessionId`) between the daemon's forwarding MCP client and its own MCP HTTP server, with a "Session not found → reset + reconnect MCP client + retry" path (`:295-298`).
- When the daemon process is replaced, the frontend (`daemonMcpProxy` / `DaemonClient`) is left with a connection/session that the **new** daemon never registered. The existing reconnect/retry paths (`daemonMcpProxy` reconnect at `:336/340/402/406`, and the `:295` MCP-client retry) do **not** re-establish a working session in this scenario, so the wedge persists and the raw `Session not found` is surfaced to the caller.

### Impact
- The daemon is *built* to be restarted (multi-worktree takeover in #2566, version/env changes, crash recovery), but any restart silently breaks every connected client with an unactionable error and **no auto-recovery** — the user just sees tools fail until they figure out they must reconnect.
- Resources are affected too (not just tools), so even read-only inspection wedges.
[...]

PR #2739, opened 2026-06-30
fix(daemon): type transport failures so sibling MCP sessions recover after daemon restart (#2737)

### LINKED ISSUE #2737: Verify build-mismatch daemon restart doesn't worsen the #2599 session-wedge for sibling MCP sessions
## Summary

A build-mismatch daemon restart introduced in #2733 (`DaemonMcpProxy.ensureBuildMatches()`) replaces a wrong-build daemon from the connecting frontend's entrypoint. The connecting session recovers fine (the restart happens in `doConnect()` before its client is created), and the new `Unknown tool:` self-heal recovers that one symptom — but **other** sibling MCP sessions still attached to the old daemon get wedged with `Session not found` and no auto-recovery (the #2599 session-wedge), for any non-tool-call path.

## Why separate from #2733

#2733 deliberately scopes to detecting/replacing the wrong build and self-healing the `Unknown tool:` symptom. General daemon-restart auto-recovery for *all* connected sessions is the broader #2599 work. This issue is to (a) verify #2733's restart doesn't make the #2599 wedge more frequent in multi-session setups, and (b) extend self-heal/auto-recovery beyond the `Unknown tool:` path if needed.

## Tasks

- Reproduce: two MCP sessions on one daemon; trigger a build-mismatch restart from one; confirm what the other session sees on its next call (expected: `Session not found` until it reconnects).
- Decide whether the `Session not found` recovery in `isRecoverableDaemonSessionError` / `withRecoverableReconnect` already covers the sibling session's next call, and if not, extend it.
- Confirm #2733's restart does not increase wedge frequency vs the existing version-mismatch restart path.

Refs #2599, #2732, #2733

kaeawc:
## Summary
Verifies issue #2737 and closes the sibling-session gap it identified. A daemon restart — the build-mismatch restart from #2733, the version-mismatch restart, an env-change restart, or crash recovery — recovers the *connecting* session, but a **sibling** MCP session still attached to the old daemon wedged on its next call with no auto-recovery (the #2599 wedge).

The fix is **restart-cause-agnostic** and lives at the **transport layer**, so it covers every restart path (including #2733's build-mismatch restart) identically.

> Rebased onto post-#2733 `main` — #2733 merged (`42060e44`) and edits the same `isRecoverableDaemonSessionError` method. This branch reconciles cleanly: it deliberately makes **no** change to that predicate.

## Root cause
When a sibling's daemon is killed/restarted, its forwarded call fails differently depending on timing, and only the typed cases were recoverable:
1. `Session not found` — the frontend reconnects and the new daemon rejects the unknown session → already recoverable.
2. `DaemonUnavailableError("Socket connection lost")` — the *next* call finds `!this.socket` (`client.ts`) → already recoverable.
3. **The in-flight request at the moment of death** → not recoverable. Over a **Unix domain socket** (the daemon's transport — no TCP RST) a dying daemon delivers **EOF → `close`**, not an `error` event, so the in-flight request was never rejected and **hung until its request timeout** (then surfaced a non-recoverable `McpTimeoutError`). A raw `error` event (`EPIPE` on a later write) was surfaced verbatim and also not matched.

## Fix
Type the failure at the layer that owns it — `DaemonClient` — instead of pattern-matching error strings up in the proxy:
- **`toDaemonTransportError()`** wraps the raw socket `error` into a `DaemonUnavailableError` (preserving the transport detail), used by the `fail()` path.
- The **`close` handler** now nulls the socket and rejects any in-flight requests with a recoverable `DaemonUnavailableError`, so the EOF/`close` path (the common Unix-socket daemon-death case) recovers **promptly** instead of hanging until timeout. Graceful `client.close()` clears pending first, so this is a no-op there.
- The proxy's existing **`instanceof DaemonUnavailableError`** branch reconnects and retries **once**. **No predicate broadening.**

### Why this layer (addressing review)
Putting the recoverability signal in the transport keeps **one source of truth** and — critically — avoids a false positive: a *daemon-returned application error* that merely mentions a transport code stays an `ActionableError` and is **not** retried. Concretely, `EmulatorConsoleClient` surfaces a downstream `connect ECONNREFUSED 127.0.0.1:5554` for `sendSms`; a substring-matching predicate would have wrongly reconnected the daemon and **re-run the non-idempotent send**. A regression test locks this out.

### Scope / honest caveats
- **Retry-once can re-execute a non-idempotent in-flight call** if the daemon happened to process it before dying. This is the genuinely new surface vs. the pre-existing `Session not found` (a daemon *response*, side effect never ran) and `Socket connection lost` (thrown *before* the write) paths — not an "identical" policy. On a restart the in-flight request almost always did not complete, the retry is capped at one, and daemon-side request dedup is out of scope (broader #2599 work).
- **Reactive recovery only.** Proactive daemon-identity-change detection (#2599's alternative; #2733's `ensureBuildMatches` already has the primitive) and cache-hit staleness are intentionally **not** in scope here.

## Test plan
TDD. New `test/daemon/daemonTransportError.test.ts`:
- `toDaemonTransportError`: wraps raw `ECONNRESET`/`EPIPE`/`socket hang up` as `DaemonUnavailableError` (detail preserved); passes a typed error through unchanged.
- **Real-socket** in-flight test: server drops the connection mid-request → the pending call rejects **promptly** with `DaemonUnavailableError` (deterministic, ~34ms; no fakes).

New `sibling-session recovery` block in `test/daemon/daemonMcpProxy.test.ts`:
- End-to-end sibling recovery for `callTool` and `readResource` (reconnect + retry once → success).
- Recovery **capped at one retry**.
- **Regression guard**: a daemon-returned tool error containing `connect ECONNREFUSED` is **not** retried (single call, no reconnect).

- `bun test` → **5253 pass**, 0 fail; `bun test test/daemon/` → 480 pass (~0.6s), each unit test <100ms.
- `turbo run lint build` → green. `scripts/all_fast_validate_checks.sh` → 10/10.

Refs #2599, #2732, #2733
Closes #2737

### DIFF
--- a/test/daemon/daemonMcpProxy.test.ts
+++ b/test/daemon/daemonMcpProxy.test.ts
@@ -6,6 +6,7 @@ import {
   DaemonToolUnavailableError,
 } from "../../src/daemon/daemonMcpProxy";
 import { DaemonClient, DaemonUnavailableError, type DaemonClientLike } from "../../src/daemon/client";
+import { ActionableError } from "../../src/models";
 import { DAEMON_VERSION, DAEMON_VERSION_RESTART_COOLDOWN_MS } from "../../src/daemon/constants";
 import { logger } from "../../src/utils/logger";
 import { FakeDaemonManager } from "../fakes/FakeDaemonManager";
@@ -763,6 +764,128 @@ describe("DaemonMcpProxy", () => {
         expect(staleClient.readResourceCalls).toEqual(["automobile:devices/booted"]);
         expect(freshClient.connectCallCount).toBe(1);
         expect(freshClient.readResourceCalls).toEqual(["automobile:devices/booted"]);
+      } finally {
+        isAvailableSpy.mockRestore();
+        await proxy.close();
+      }
+    });
+  });
+
+  describe("sibling-session recovery after daemon restart (#2737)", () => {
+    // A daemon restart (build-mismatch #2733, version-mismatch, env change, or
+    // crash recovery) tears down every other connected session's live socket.
+    // DaemonClient now types that transport failure as DaemonUnavailableError
+    // (see daemonTransportError.test.ts), so the sibling's next forwarded call
+    // is recoverable here and reconnects+retries once instead of wedging (#2599).
+    // Critically, a *daemon-returned application error* that merely mentions a
+    // transport code (e.g. a tool reporting a downstream `connect ECONNREFUSED`)
+    // stays an ActionableError and must NOT be retried.
+
+    test("callTool recovers when a sibling's socket dropped (DaemonUnavailableError)", async () => {
+      const recoveredResult = { content: [{ type: "text", text: "sibling recovered" }] };
+      const staleClient = new ScriptedDaemonClient({
+        toolError: new DaemonUnavailableError("Daemon socket connection lost: connection closed"),
+      });
+      const freshClient = new ScriptedDaemonClient({ toolResult: recoveredResult });
+      const clients = [staleClient, freshClient];
+      const isAvailableSpy = spyOn(DaemonClient, "isAvailable").mockResolvedValue(true);
+
+      const proxy = new DaemonMcpProxy({
+        clientFactory: () => clients.shift()!,
+        autoStartDaemon: false,
+      });
+
+      try {
+        const result = await proxy.callTool("observe", { deviceId: "device-1" });
+
+        expect(result).toEqual(recoveredResult);
+        expect(staleClient.closeCallCount).toBe(1);
+        expect(staleClient.callToolCalls).toHaveLength(1);
+        expect(freshClient.connectCallCount).toBe(1);
+        expect(freshClient.callToolCalls).toEqual([
+          { toolName: "observe", params: { deviceId: "device-1" } },
+        ]);
+      } finally {
+        isAvailableSpy.mockRestore();
+        await proxy.close();
+      }
+    });
+
+    test("readResource recovers when a sibling's socket dropped", async () => {
+      const recoveredResult = { contents: [{ uri: "automobile:devices/booted", text: "[]" }] };
+      const staleClient = new ScriptedDaemonClient({
+        resourceError: new DaemonUnavailableError("Daemon socket connection lost: connection closed"),
+      });
+      const freshClient = new ScriptedDaemonClient({ resourceResult: recoveredResult });
+      const clients = [staleClient, freshClient];
+      const isAvailableSpy = spyOn(DaemonClient, "isAvailable").mockResolvedValue(true);
+
+      const proxy = new DaemonMcpProxy({
+        clientFactory: () => clients.shift()!,
+        autoStartDaemon: false,
+      });
+
+      try {
+        const result = await proxy.readResource("automobile:devices/booted");
+
+        expect(result).toEqual(recoveredResult);
+        expect(staleClient.closeCallCount).toBe(1);
+        expect(freshClient.readResourceCalls).toEqual(["automobile:devices/booted"]);
+      } finally {
+        isAvailableSpy.mockRestore();
+        await proxy.close();
+      }
+    });
+
+    test("caps sibling recovery at one retry", async () => {
+      const firstClient = new ScriptedDaemonClient({
+        toolError: new DaemonUnavailableError("Daemon socket connection lost: connection closed"),
+      });
+      const secondClient = new ScriptedDaemonClient({
+        toolError: new DaemonUnavailableError("Daemon socket connection lost: connection closed"),
+      });
+      const clients = [firstClient, secondClient];
+      const isAvailableSpy = spyOn(DaemonClient, "isAvailable").mockResolvedValue(true);
+
+      const proxy = new DaemonMcpProxy({
+        clientFactory: () => clients.shift()!,
+        autoStartDaemon: false,
+      });
+
+      try {
+        await expect(proxy.callTool("observe", {})).rejects.toBeInstanceOf(DaemonUnavailableError);
+        expect(firstClient.closeCallCount).toBe(1);
+        expect(firstClient.callToolCalls).toHaveLength(1);
+        expect(secondClient.callToolCalls).toHaveLength(1);
+        expect(clients).toHaveLength(0);
+      } finally {
+        isAvailableSpy.mockRestore();
+        await proxy.close();
+      }
+    });
+
+    test("does NOT retry a daemon-returned tool error that mentions a transport code", async () => {
+      // Regression guard: a tool whose *downstream* connection fails surfaces the
+      // raw errno in its message (e.g. EmulatorConsoleClient: sendSms over the
+      // emulator console). That ActionableError reaches the proxy as a normal tool
+      // failure — retrying it would re-run a non-idempotent action and mask the
+      // real cause. It must be surfaced, not classified as a recoverable session.
+      const client = new ScriptedDaemonClient({
+        toolError: new ActionableError(
+          "Emulator console connection to 127.0.0.1:5554 failed: connect ECONNREFUSED 127.0.0.1:5554"
+        ),
+      });
+      const isAvailableSpy = spyOn(DaemonClient, "isAvailable").mockResolvedValue(true);
+
+      const proxy = new DaemonMcpProxy({
+        clientFactory: () => client,
+        autoStartDaemon: false,
+      });
+
+      try {
+        await expect(proxy.callTool("sendSms", {})).rejects.toThrow("ECONNREFUSED");
+        expect(client.callToolCalls).toHaveLength(1);
+        expect(client.closeCallCount).toBe(0);
       } finally {
         isAvailableSpy.mockRestore();
         await proxy.close();
--- a/src/daemon/client.ts
+++ b/src/daemon/client.ts
@@ -28,6 +28,28 @@ export class DaemonUnavailableError extends Error {
   }
 }
 
+/**
+ * Normalize a socket-level transport failure into a `DaemonUnavailableError`.
+ *
+ * When a daemon restart/crash tears down the connection, an in-flight request's
+ * socket emits a raw transport error (`ECONNRESET` / `EPIPE` / "socket hang up")
+ * with no daemon-level meaning. Surfacing that raw error verbatim wedges every
+ * other connected session (#2599): the proxy only treats `DaemonUnavailableError`
+ * and the daemon's own `Session not found` as recoverable, so a raw `ECONNRESET`
+ * is not retried (#2737). Typing it here — at the layer that knows it is a socket
+ * failure — keeps the recoverability signal in one place: the proxy reconnects via
+ * its existing `instanceof DaemonUnavailableError` branch, and a *daemon-returned*
+ * application error that merely mentions a transport code (e.g. a tool reporting a
+ * downstream `connect ECONNREFUSED`) stays an `ActionableError` and is correctly
+ * not retried. Already-typed `DaemonUnavailableError`s pass through unchanged.
+ */
+export function toDaemonTransportError(error: Error): DaemonUnavailableError {
+  if (error instanceof DaemonUnavailableError) {
+    return error;
+  }
+  return new DaemonUnavailableError(`Daemon socket connection lost: ${error.message}`);
+}
+
 export interface DaemonClientRecoveryOptions extends StaleDaemonFileCleanupOptions {
   /**
    * When true, `isAvailable()` performs an observation-only probe and never
@@ -181,10 +203,14 @@ export class DaemonClient {
           this.socket.destroy();
           this.socket = null;
         }
-        rejectPendingRequests(error);
+        // Type the transport failure so the proxy can recover sibling sessions
+        // wedged by a daemon restart (#2599/#2737) instead of surfacing a raw
+        // ECONNRESET/EPIPE/"socket hang up" that its recovery does not match.
+        const failure = toDaemonTransportError(error);
+        rejectPendingRequests(failure);
         if (!settled) {
           settled = true;
-          reject(error);
+          reject(failure);
         }
       };
 
@@ -217,7 +243,20 @@ export class DaemonClient {
 
       this.socket.on("close", () => {
         this.connected = false;
+        this.socket = null;
         logger.info("Daemon socket connection closed");
+        // A daemon restart/crash closes the socket. Over a Unix domain socket
+        // (no TCP RST) a dying daemon delivers EOF -> "close" with no "error"
+        // event, so any in-flight request would otherwise hang until its request
+        // timeout. Reject pending requests with a recoverable transport error so
+        // the proxy reconnects to the restarted daemon and retries (#2599/#2737).
+        // After fail() the pending map is already cleared, so this is a no-op on
+        // the error path.
+        if (this.pendingRequests.size > 0) {
+          rejectPendingRequests(
+            new DaemonUnavailableError("Daemon socket connection lost: connection closed")
+          );
+        }
       });
     });
   }
--- a/test/daemon/daemonTransportError.test.ts
+++ b/test/daemon/daemonTransportError.test.ts
@@ -1,0 +1,87 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import { createServer, type Server, type Socket } from "node:net";
+import { mkdtempSync, rmSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir, platform } from "node:os";
+import {
+  DaemonClient,
+  DaemonUnavailableError,
+  toDaemonTransportError,
+} from "../../src/daemon/client";
+
+const isWindows = platform() === "win32";
+
+describe("toDaemonTransportError", () => {
+  test("wraps a raw transport error as a recoverable DaemonUnavailableError", () => {
+    const raw = new Error("read ECONNRESET");
+    (raw as NodeJS.ErrnoException).code = "ECONNRESET";
+
+    const wrapped = toDaemonTransportError(raw);
+
+    expect(wrapped).toBeInstanceOf(DaemonUnavailableError);
+    // The original transport detail is preserved for diagnostics.
+    expect(wrapped.message).toContain("ECONNRESET");
+  });
+
+  test("wraps EPIPE and 'socket hang up' the same way", () => {
+    for (const message of ["write EPIPE", "socket hang up"]) {
+      const wrapped = toDaemonTransportError(new Error(message));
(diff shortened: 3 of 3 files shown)


src/daemon/daemonMcpProxy.ts
```diff
@@ -364,8 +364,39 @@ export class DaemonMcpProxy {
       return true;
     }
 
+    // A daemon restart (build-mismatch #2733, version-mismatch, env change, or
+    // crash recovery) tears down the socket of every other connected session.
+    // A sibling session's in-flight request is rejected with the raw transport
+    // error from the dying socket — not "Session not found" — so it must be
+    // treated as recoverable too, or the sibling wedges (#2599 / #2737).
+    const code = (error as NodeJS.ErrnoException | null)?.code;
+    if (typeof code === "string" && DaemonMcpProxy.RECOVERABLE_TRANSPORT_CODES.has(code)) {
```
kaeawc: **[Vera — Risky (regression), non-blocking]** The PR description says adding these transport codes is "not new... identical policy" as the pre-existing `Session not found` retry. That isn't quite true, and the difference matters for non-idempotent tools (`tapOn`/`inputText`/`swipeOn`). Trace the two pre-existing recoverable signatures for an in-flight `callTool`:

- `Session not found` arrives as a daemon **response** error (`client.ts:277`, wrapped in `ActionableError`). For a response to come back at all, a *new* daemon received the request and rejected the unknown session — so the side effect was **not** executed. Safe to retry.
- `Socket connection lost` is thrown when `!this.socket` *before the write* (`client.ts:338-345`) — the request was **never sent**. Safe to retry.

The newly-added `ECONNRESET`/`EPIPE`/`socket hang up` arrive via `fail(error)` rejecting an **already-written, in-flight** request (`client.ts:184`, `213-216`). This is the one disconnect shape where the daemon **may have already executed the side effect** before the socket reset raced the response. There is no idempotency/dedup guard on the daemon (`grep` for `requestId` in `src/daemon/` shows none), and the retry mints a **fresh** `randomUUID()` request id (`client.ts:306`) — so even a hypothetical dedup would be defeated. Net: this change genuinely widens the double-execution window for non-idempotent tools relative to the two pre-existing signatures, which were both side-effect-safe.

Not a blocker (the wedge is worse than an occasional double-tap, and a killed daemon usually didn't finish), but the "identical policy" claim should be softened — a tap can fire twice on a real restart where it previously could not. Worth a follow-up to scope the in-flight transport-error retry to read-only/idempotent calls (`observe`, `tools/list`, `readResource`), or to thread the original request id through the retry so the daemon side could dedup later.

test/daemon/daemonMcpProxy.test.ts
```diff
@@ -762,6 +762,193 @@ describe("DaemonMcpProxy", () => {
...
+        await expect(proxy.callTool("observe", {})).rejects.toThrow("Invalid arguments");
+        expect(client.callToolCalls).toHaveLength(1);
+        expect(client.closeCallCount).toBe(0);
+      } finally {
+        isAvailableSpy.mockRestore();
+        await proxy.close();
+      }
+    });
+
+    test("sibling recovers independently while another session triggers a version-mismatch restart", async () => {
```
kaeawc: **[Vera — Incomplete]** This "two-session reproduction" doesn't reproduce the cross-session wedge it claims to. `proxyA` and `proxyB` are fully independent instances with independent fake clients and no shared daemon/socket — `staleB` is scripted to throw `ECONNRESET` *unconditionally*, with zero causal link to `managerA.restart()`. The assertion `expect(managerA.restartCalled).toBe(true)` and B's recovery are two unrelated facts; B would recover identically if A's restart line were deleted. So this proves the predicate works for B in isolation (already covered by the per-code tests above) but proves nothing about ordering/timing of a real restart tearing down a sibling's live socket — the exact thing #2737 asks to verify. The comment's "proving the restart does not worsen the #2599 wedge" overstates what's tested; consider a single shared FakeDaemon whose restart actually drives the sibling's disconnect, or label this as a predicate-coverage test rather than a reproduction.

kaeawc (commented):
Concurrency/failure-mode review (Vera): The recovery direction is sound and the retry stays correctly capped at one (verified `caps transport-disconnect recovery at one retry`, `bun test test/daemon/daemonMcpProxy.test.ts` = 37 pass). One substantive caveat on the double-execution framing, plus a note that the two-session test does not actually exercise a shared daemon.

src/daemon/daemonMcpProxy.ts
```diff
@@ -364,8 +364,39 @@ export class DaemonMcpProxy {
...
+      return true;
+    }
+
     const message = error instanceof Error ? error.message : String(error);
-    return message.includes("Session not found");
+    return (
+      message.includes("Session not found") ||
+      message.includes("Socket connection lost") ||
+      message.includes("socket hang up") ||
+      DaemonMcpProxy.messageHasTransportCode(message)
```
kaeawc: **[Priya — Direction (wrong layer)]** The recoverability signal is being reconstructed one layer too high. `client.ts` is the layer that actually knows a socket died, and it *already* models that as a typed error: the no-socket path throws `DaemonUnavailableError("Socket connection lost")` (client.ts:342/395), which this predicate already catches via the `instanceof DaemonUnavailableError` branch (line 363). The only gap is the *in-flight* path — `socket.on("error", e => fail(e))` (client.ts:213-216) rejects pending requests with the **raw** socket error, untyped. So we now have two sources of truth for 'transport died': a typed class on one path and `error.code`/message-substring re-parsing on the other.

The more durable primitive: wrap the raw socket error in `DaemonUnavailableError` (or a new `DaemonTransportError extends DaemonUnavailableError`) at client.ts:215, where `.code` is still reliably present. Then this whole block — the `RECOVERABLE_TRANSPORT_CODES` set, `messageHasTransportCode`, and the `"socket hang up"`/`"Socket connection lost"` substrings — collapses back into the existing `instanceof` check, and the boundary you're defending with the 'application error that mentions socket' test stops being string-shaped. Message-substring matching for `ECONNRESET`/`EPIPE` is the fragile part: it's the fallback for errors forwarded through the daemon's MCP layer that get re-wrapped as `ActionableError` (client.ts:277) and lose `.code` — but that wrapping is itself in our code, so the right fix is to preserve the code/type across that boundary, not pattern-match the prose downstream.

src/daemon/daemonMcpProxy.ts
```diff
@@ -364,8 +364,39 @@ export class DaemonMcpProxy {
...
+    );
+  }
```
kaeawc, line 549: **[Priya — Drift risk]** This is a real, *active* convergence problem, not a hypothetical. #2733 merged today (mergeCommit 42060e4) and it edits **this same method** (`isRecoverableDaemonSessionError`) — adding an `Unknown tool:` branch and an `invalidateCache()` on the version-restart path — plus the same `test/daemon/daemonMcpProxy.test.ts`. Neither change is in this branch (`buildIdentity.ts`, `ensureBuildMatches`, the `Unknown tool:` branch are all absent), and `gh pr view 2739` reports the PR as **CONFLICTING**. So the merge conflict the brief asks about is already here, dead-center in the predicate this PR grows.

The deeper issue is that we now have two ad-hoc recovery extensions bolted onto the same predicate/`withRecoverableReconnect` from two different PRs — one keying on `Unknown tool:`, one on transport codes — with no shared contract. That's the signal a unifying abstraction is missing: recovery classification wants to be one typed decision (`is this error a daemon-replacement signal?`) sourced where the error originates, not an accreting list of substrings two PRs append to independently. At minimum this PR should rebase onto post-#2733 main and reconcile the predicate as one coherent function; ideally the transport-disconnect and unknown-tool signals land as typed errors from their respective origins rather than both as string matches here.

src/daemon/daemonMcpProxy.ts
```diff
@@ -364,8 +364,39 @@ export class DaemonMcpProxy {
       return true;
     }
 
+    // A daemon restart (build-mismatch #2733, version-mismatch, env change, or
+    // crash recovery) tears down the socket of every other connected session.
+    // A sibling session's in-flight request is rejected with the raw transport
+    // error from the dying socket — not "Session not found" — so it must be
+    // treated as recoverable too, or the sibling wedges (#2599 / #2737).
```
kaeawc: **[Priya — Direction (leaves the wedge reachable)]** This closes the wedge only on paths that *throw a matched error*. #2599's own 'Suggested fix' offered a strictly more robust alternative: detect a daemon **identity change** (PID/`startedAt` from `--daemon status`) and proactively reconnect, or have the daemon lazily create a session for an unknown connection. The reactive approach here can't cover a sibling whose next call is a cache hit (e.g. `listTools`/`listResources` short-circuit on `cachedTools` at line 424 and never reach the socket) or any path that returns a stale-but-successful response from a half-open socket — those never surface `ECONNRESET`/`Session not found`, so the sibling silently talks to nothing or serves stale data. `ensureBuildMatches` from #2733 already proves the proactive primitive exists (it diffs identity in `doConnect()` and restarts) — a sibling-side identity check on reconnect would subsume both this and #2733's self-heal. I'm not asking you to build it here, but the PR should explicitly note *why* the proactive identity-change check is deferred, since #2599 named it as the alternative and it's the design this keeps approximating.

test/daemon/daemonMcpProxy.test.ts
```diff
@@ -762,6 +762,193 @@ describe("DaemonMcpProxy", () => {
...
+
+      // Session A: triggers the restart path.
+      const managerA = new FakeDaemonManager();
+      const initialStatus = {
+        running: true,
+        pid: 1234,
+        port: 3000,
+        socketPath: "/tmp/test.sock",
+        version: OLDER_VERSION,
+        startedAt: ANCIENT_TIMESTAMP,
```
kaeawc: **[Priya — Incomplete (issue requirement)]** #2737 task (c) is 'confirm #2733's restart does not increase wedge frequency vs the existing version-mismatch restart path.' This test demonstrates the **version-mismatch** path (session A forces a version restart, sibling B recovers), and the PR body asserts the build-mismatch path is 'identical' because both route through `daemonManager.restart()`. But #2733's `ensureBuildMatches()` isn't in this branch at all, so nothing here actually exercises it — the 'identical' claim is asserted, not shown. The sibling disconnect is also simulated by a `ScriptedDaemonClient` returning a canned `ECONNRESET`, not produced by a real restart tearing down a real socket, so the test validates the predicate, not the end-to-end wedge behavior the issue wants confirmed. To genuinely discharge task (c) after rebasing onto #2733: drive the sibling's failure through a build-mismatch restart and assert the same recovery, so the 'does not increase wedge frequency' conclusion is observed rather than reasoned about.

kaeawc (commented):
Architecture/direction review (Priya). The reactive fix is reasonable as a stopgap, but I have three substantive direction concerns and one completeness gap. (1) LAYERING: the recoverability signal is being reconstructed in the wrong layer — client.ts already knows it's a transport error and already wraps one transport case in a typed DaemonUnavailableError; the in-flight-disconnect case should be wrapped the same way rather than re-parsed by string in the proxy. (2) DRIFT: #2733 merged today and edits the exact same predicate + test file this PR extends — the PR is already CONFLICTING, and #2733's code isn't in this branch, so the description's 'covers #2733 identically' is unverifiable here. (3) COMPLETENESS: #2737 task (c) is asserted, not demonstrated. Details inline.

src/daemon/daemonMcpProxy.ts
kaeawc: **[Marcus — Risky (false positive)]** `messageHasTransportCode(message)` substring-matches the bare errno tokens anywhere in the message. This fires on genuine application errors that embed a *downstream* socket failure verbatim. Confirmed paths:

- `src/utils/EmulatorConsoleClient.ts:93` throws `ActionableError("Emulator console connection to ${host}:${port} failed: ${err.message}")`. When the emulator console refuses, `err.message` is `"connect ECONNREFUSED 127.0.0.1:5554"` → this message contains `ECONNREFUSED`. This is a tool path (sendSms / emulator-console commands), so it surfaces through the daemon as `response.error` → `ActionableError` (client.ts:277).
- `src/utils/hostControlClient.ts:180` returns `{success:false, error:"Connection failed: ${err.message}"}` → `"Connection failed: connect ECONNREFUSED 127.0.0.1:PORT"`.

Verified Node behavior: a raw `createConnection` error's `.message` is literally `"connect ECONNREFUSED 127.0.0.1:PORT"` (the token is in the message, not just `.code`). The repo has several `net`/`createConnection` paths (hostControlClient, EmulatorConsoleClient, IOSCtrlProxyManager) whose messages can reach a surfaced error string.

Consequence: "emulator console is down" (a sticky, legitimate failure that reconnecting the AutoMobile daemon will NOT fix) gets treated as recoverable. The proxy resets the connection, retries once, and on the second identical failure the user sees the error prefixed by a misleading `"Daemon session is stale, reconnecting and retrying once"` warning — the true cause is masked. For non-idempotent tools (sendSms) the operation is also re-attempted. The `error.code` check at L372-375 already covers the legitimate sibling-disconnect case (real transport errors carry `.code`); this substring widening only adds false positives. Suggest removing it.

src/daemon/daemonMcpProxy.ts
```diff
@@ -364,8 +364,39 @@ export class DaemonMcpProxy {
...
+    const code = (error as NodeJS.ErrnoException | null)?.code;
+    if (typeof code === "string" && DaemonMcpProxy.RECOVERABLE_TRANSPORT_CODES.has(code)) {
+      return true;
+    }
+
     const message = error instanceof Error ? error.message : String(error);
-    return message.includes("Session not found");
+    return (
+      message.includes("Session not found") ||
+      message.includes("Socket connection lost") ||
```
kaeawc: **[Marcus — Direction]** The two explicit phrase matches here (`"Socket connection lost"`, `"socket hang up"`) are defensible: `"Socket connection lost"` is produced by the transport layer itself (client.ts:342/395, a `DaemonUnavailableError`), and `"socket hang up"` is Node's standard `ECONNRESET`-class message for a peer that closed mid-response. These are tied to the actual daemon socket, not interpolated from downstream tool errors, so they don't share the false-positive risk of the bare-errno `messageHasTransportCode` path. No change needed to these two lines specifically — flagging only to scope the finding above precisely.

test/daemon/daemonMcpProxy.test.ts
```diff
@@ -762,6 +762,193 @@ describe("DaemonMcpProxy", () => {
...
+        expect(firstClient.callToolCalls).toHaveLength(1);
+        expect(secondClient.callToolCalls).toHaveLength(1);
+        expect(clients).toHaveLength(0);
+      } finally {
+        isAvailableSpy.mockRestore();
+        await proxy.close();
+      }
+    });
+
+    test("does not retry genuine application errors that merely mention a socket", async () => {
```
kaeawc: **[Marcus — Incomplete coverage]** This boundary test only proves the predicate ignores the bare word `"socket"` (`"Invalid arguments: socket must be a number"` — no errno token, no `"socket hang up"`/`"Socket connection lost"` phrase). It does NOT exercise the contract boundary that actually matters after this PR: a genuine *application* error whose message embeds a downstream errno token. The missing case — e.g. `toolError: new Error("Emulator console connection to 127.0.0.1:5554 failed: connect ECONNREFUSED 127.0.0.1:5554")` asserting `callToolCalls` has length 1 and `closeCallCount` is 0 — would currently FAIL, because `messageHasTransportCode` matches `ECONNREFUSED` and wrongly retries. That failing test is precisely the proof the new contract is under-specified. Please add it (and decide whether the desired behavior is no-retry, which argues for dropping the bare-errno-in-message matching).

kaeawc (commented):
Correctness/false-positive audit (Marcus): The `error.code` half of this predicate is sound and correctly type-safe. The problem is the `messageHasTransportCode(message)` substring half. The daemon returns application failures as `{success:false, error:"<string>"}`, which client.ts:277 wraps verbatim in `ActionableError(response.error)` — and several tools embed a *downstream* socket error's `err.message` into that string. Node's raw socket errors carry the literal errno token in `.message` (verified: `createConnection` to a dead port → `"connect ECONNREFUSED 127.0.0.1:PORT"`). So a genuine, sticky downstream failure now gets mis-classified as a recoverable AutoMobile-daemon transport blip, triggering a futile reconnect+retry that masks the real cause behind a "stale session" warning. The PR's "negligible, only IOSCtrlProxyManager.ts:134" claim is incomplete — that line is a `code ===` compare, not a message; the message-substring exposure is wider. Two concrete paths below, plus a coverage gap. The `error.code` path alone would have covered the legitimate sibling-disconnect case (those errors DO carry `.code`); the message-substring widening is what introduces the false positives. Recommend dropping `messageHasTransportCode` (and the bare errno-in-message matching) and relying on `error.code` + the explicit `"socket hang up"`/`"Socket connection lost"`/`"Session not found"` phrases, which are produced by the transport layer itself rather than interpolated from arbitrary downstream errors.

kaeawc:
## Addressed all review feedback — redesigned + rebased onto post-#2733 `main`

Thanks all. The first revision used `error.code`/substring matching in the proxy predicate. The reviews converged on a better design, and surfaced that **#2733 had merged** (same method, conflicting). Force-pushed a redesign that resolves every point:

**@Priya (wrong layer / drift):** Moved the recoverability signal into the transport layer (`DaemonClient`) as a typed `DaemonUnavailableError`, collapsing the string-matching back into the proxy's existing `instanceof` branch — one source of truth. Rebased onto post-#2733 `main`; this branch now makes **no** change to `isRecoverableDaemonSessionError`, so there's nothing to drift against #2733's `Unknown tool:` extension.

**@Marcus (false positive — confirmed real bug):** You were right that `messageHasTransportCode` would mis-retry a daemon-returned `connect ECONNREFUSED` (e.g. `EmulatorConsoleClient` → `sendSms`). The typed-at-transport approach eliminates the class entirely: that error stays an `ActionableError`, never a `DaemonUnavailableError`. Added the exact regression test you specified (errno-bearing tool error is not retried).

**@Vera (in-flight double-execution framing + contrived test):** Dropped the "identical policy" claim — the PR now states plainly that the in-flight retry is the genuinely new surface and why retry-once is still correct (capped; in-flight request almost never completed on a restart). Replaced the contrived two-session fake-swap with (a) a **real-socket** test proving an in-flight request rejects promptly with `DaemonUnavailableError` when the daemon drops the connection, and (b) honest proxy-level recovery tests.

Investigating the real socket also exposed a deeper gap your critique pointed at: over a **Unix domain socket there is no TCP RST**, so a dying daemon delivers **EOF → `close`** (no `error`) and the in-flight request *hung until timeout*. The fix now also rejects pending requests in the `close` handler, so the common daemon-death case recovers promptly.

`bun test` → 5253 pass / 0 fail; lint+build + fast-validate green.

PR #2740, opened 2026-06-30
feat(doctor): surface daemon build identity (entryScript/buildId) + skew (#2736)

### LINKED ISSUE #2736: doctor: surface daemon build identity (entryScript / buildId) to spot wrong-build skew
## Summary

After #2733, the daemon PID file records build identity (`entryScript` + `buildId`, a content hash of the daemon's entry script) and `manager.status()` surfaces it — but nothing **displays** it. Surfacing it in `doctor` (and any daemon-status output) would let a developer see *which build* the running daemon is and spot wrong-build skew before it bites, instead of discovering it only when a tool fails.

## Background

On a machine with two checkouts sharing one per-uid daemon socket, the daemon serving a frontend can be from a *different* build (see #2732). #2733 added build identity to `PidFileData` / `DaemonStatus` and made the proxy detect + self-heal the skew, but provides no way to simply *observe* the current daemon's identity.

## Proposal

- Add `entryScript` / `buildId` to the `doctor` daemon section and to whatever surfaces `DaemonStatus` to the user.
- When the connecting client's build identity differs from the running daemon's, show both (the proxy already computes the client identity via `getCurrentBuildIdentity()` in `src/daemon/buildIdentity.ts`).

## Acceptance

- `doctor` output includes the daemon's `buildId` (and `entryScript`).
- A skew between client and daemon build is visible in `doctor` output without having to trigger a failing tool call.

Refs #2732, #2733

kaeawc:
## Summary
Closes #2736. Surfaces the running daemon's **build identity** (`buildId` + `entryScript`, recorded in the PID file by #2733) so a developer can see *which build* the daemon is and spot **wrong-build skew** before a tool call fails — instead of discovering it only when a tool returns `Unknown tool`.

Two checkouts on one machine share a single per-uid daemon socket, so the daemon serving a frontend can be from a *different* build (#2732). #2733 added build identity to `PidFileData` / `DaemonStatus` and made the proxy self-heal the skew, but nothing **displayed** it. This PR adds the missing display in the two user-facing surfaces of `DaemonStatus`.

## Changes
- **`describeBuildIdentity(identity)`** (`src/daemon/buildIdentity.ts`) — shared renderer producing `"<buildId> (<entryScript>)"`, falling back to `unknown` for legacy daemons, so both surfaces format identically.
- **`doctor` — new `Daemon Build Identity` check** (`src/doctor/checks/automobile.ts`, wired into `runAutoMobileChecks`):
  - `pass` showing the daemon's `buildId` + `entryScript` when the client and daemon builds match;
  - `warn` showing **both** identities + a restart recommendation when they differ (skew visible without a failing tool call);
  - `skip` when the daemon is not running;
  - reuses `buildIdentitiesMatch`, so a legacy daemon without build identity does **not** raise a false skew.
- **`--daemon status` CLI** (`runDaemonCommand` in `src/daemon/manager.ts`) — adds `Build ID` / `Entry Script` lines and a skew warning, built by the pure, unit-tested `daemonBuildIdentityStatusLines(status, client)`.

A review pass caught that the doctor console formatter renders `CheckResult.value` *instead of* `message`; setting `value=buildId` would have hidden the `entryScript`. Fixed by carrying both in `message` (no `value`).

## Acceptance (from #2736)
- ✅ `doctor` output includes the daemon's `buildId` **and** `entryScript`.
- ✅ Client/daemon build skew is visible in `doctor` (and `--daemon status`) output without triggering a failing tool call.

## Test plan
- TDD: tests written first (red), then implementation (green).
- New/updated fast unit tests (injected fakes, no real fs/daemon, <100ms each):
  - `test/daemon/buildIdentity.test.ts` — `describeBuildIdentity` rendering + `unknown` fallback.
  - `test/doctor/automobile.test.ts` — pass/warn/skip + no-false-skew-for-legacy-daemon + presence in `runAutoMobileChecks`.
  - `test/daemon/manager.test.ts` — `daemonBuildIdentityStatusLines` always reports Build ID/Entry Script, falls back to `unknown` without warning, and warns showing both builds on skew.
- `bun test test/doctor/ test/daemon/` → 609 pass.
- `bun run lint` + `bun run build` → green.

Refs #2732, #2733

### DIFF
--- a/src/daemon/buildIdentity.ts
+++ b/src/daemon/buildIdentity.ts
@@ -88,3 +88,14 @@ export function buildIdentitiesMatch(a: BuildIdentity, b: BuildIdentity): boolea
   }
   return true;
 }
+
+/**
+ * Render a {@link BuildIdentity} as a single human-readable token:
+ * `"<buildId> (<entryScript>)"`. Falls back to `unknown` for an empty entry
+ * script so the rendering stays stable for legacy daemons that predate build
+ * identity. Shared by `doctor` and the `--daemon status` CLI so both surface the
+ * daemon's build in the same format.
+ */
+export function describeBuildIdentity(identity: BuildIdentity): string {
+  return `${identity.buildId} (${identity.entryScript || "unknown"})`;
+}
--- a/src/daemon/buildIdentity.ts
+++ b/src/daemon/buildIdentity.ts
@@ -99,3 +99,24 @@ export function buildIdentitiesMatch(a: BuildIdentity, b: BuildIdentity): boolea
 export function describeBuildIdentity(identity: BuildIdentity): string {
   return `${identity.buildId} (${identity.entryScript || "unknown"})`;
 }
+
+/**
+ * Project the build fields carried on a daemon status / PID-file record into a
+ * {@link BuildIdentity}, normalizing the optional wire fields (a missing
+ * `entryScript` becomes `""`, a missing `buildId` becomes the unknown sentinel).
+ *
+ * Centralizes the `DaemonStatus`/`PidFileData` → `BuildIdentity` mapping that
+ * `doctor`, the `--daemon status` CLI, and the MCP proxy all need, so they agree
+ * on how an unidentified (legacy) daemon is represented. Accepts a structural
+ * shape rather than importing `DaemonStatus` to avoid a module cycle and to keep
+ * the surface narrow.
+ */
+export function buildIdentityFromStatus(record: {
+  entryScript?: string;
+  buildId?: string;
+}): BuildIdentity {
+  return {
+    entryScript: record.entryScript ?? "",
+    buildId: record.buildId ?? UNKNOWN_BUILD_ID,
+  };
+}
--- a/src/doctor/checks/automobile.ts
+++ b/src/doctor/checks/automobile.ts
@@ -188,11 +188,13 @@ export async function checkDaemonBuildIdentity(
     const client = (dependencies.getClientBuildIdentity ?? getCurrentBuildIdentity)();
 
     if (buildIdentitiesMatch(client, daemon)) {
+      // No `value`: the console formatter renders `value` *instead of* `message`,
+      // and we want both the buildId and the entryScript visible — so carry them
+      // in the message.
       return {
         name: "Daemon Build Identity",
         status: "pass",
         message: `Build ${describeBuildIdentity(daemon)}`,
-        value: daemon.buildId,
       };
     }
 
--- a/src/doctor/checks/automobile.ts
+++ b/src/doctor/checks/automobile.ts
@@ -201,8 +201,10 @@ export async function checkDaemonBuildIdentity(
       status: "warn",
       message: `Build skew: daemon ${describeBuildIdentity(daemon)}, client ${describeBuildIdentity(client)}`,
       recommendation:
-        "The running daemon is a different build than this checkout. Restart it from here: " +
-        "bunx @kaeawc/auto-mobile@latest --daemon restart",
+        "The running daemon is a different build than this checkout. Restart the daemon " +
+        "from THIS checkout so it matches — run `--daemon restart` with the same auto-mobile " +
+        "CLI you invoked here. Avoid `@latest`, which starts the published build and may not " +
+        "match this checkout.",
     };
   } catch (error) {
     // Diagnostic path: log the underlying error before returning a typed failure
--- a/src/doctor/checks/automobile.ts
+++ b/src/doctor/checks/automobile.ts
@@ -9,6 +9,12 @@ import { DaemonManager } from "../../daemon/manager";
 import { getDaemonHealthReport } from "../../daemon/debugTools";
 import type { DaemonHealthReport } from "../../daemon/debugTools";
 import type { DaemonStatus } from "../../daemon/types";
+import {
+  buildIdentitiesMatch,
+  describeBuildIdentity,
+  getCurrentBuildIdentity,
+} from "../../daemon/buildIdentity";
+import type { BuildIdentity } from "../../daemon/buildIdentity";
 import { LATEST_RELEASE_VERSION, RELEASE_VERSION, resolveAssetVersion } from "../../constants/release";
 import { getMcpServerVersion } from "../../utils/mcpVersion";
 import { defaultAdbClientFactory } from "../../utils/android-cmdline-tools/AdbClientFactory";
@@ -27,9 +33,15 @@ export interface DaemonStatusDependencies {
   getDaemonHealthReport?: () => Promise<DaemonHealthReport>;
 }
 
+export interface DaemonBuildIdentityDependencies {
+  daemonManager?: DaemonStatusManager;
+  getClientBuildIdentity?: () => BuildIdentity;
+}
+
 export interface AutoMobileCheckDependencies {
   checkDaemonStatus?: () => Promise<CheckResult>;
   checkDaemonConnectivity?: () => Promise<CheckResult>;
+  checkDaemonBuildIdentity?: () => Promise<CheckResult>;
 }
 
 /**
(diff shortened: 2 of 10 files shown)


kaeawc (commented):
Architecture review: two ownership/duplication findings. The new doctor check and CLI status block are well-shaped and follow the per-concern check + injectable-deps conventions — keeping them as separate functions is fine. But the new code re-introduces a DaemonStatus→BuildIdentity projection that should be owned by buildIdentity.ts, and it stops short of unifying the format with the pre-existing user-facing skew surfaces in daemonMcpProxy.ts.

src/doctor/checks/automobile.ts
```diff
@@ -145,6 +157,64 @@ export async function checkDaemonConnectivity(
...
+      return {
+        name: "Daemon Build Identity",
+        status: "skip",
+        message: "Daemon is not running",
+      };
+    }
+
+    const daemon: BuildIdentity = {
+      entryScript: status.entryScript ?? "",
+      buildId: status.buildId ?? "unknown",
```
kaeawc: The `{ entryScript: status.entryScript ?? "", buildId: status.buildId ?? "unknown" }` projection from `DaemonStatus` to `BuildIdentity` now exists in **five** places: here, manager.ts:1079-1080 (also new in this PR), and daemonMcpProxy.ts:388, 430, 621. The PR adds two more copies of the exact same mapping without factoring it. Translating the daemon’s wire `DaemonStatus` into a `BuildIdentity` is the buildIdentity module’s responsibility — it owns those two fields. Recommend adding `export function buildIdentityFromStatus(status: Pick<DaemonStatus, "entryScript" | "buildId">): BuildIdentity` to `src/daemon/buildIdentity.ts` (type-only import of `DaemonStatus` from types.ts — no import cycle, types.ts has no imports) and calling it here, in manager.ts, and ideally the three proxy sites. This is the right PR to do it: rule-of-three is well exceeded.

src/daemon/buildIdentity.ts
```diff
@@ -88,3 +88,14 @@ export function buildIdentitiesMatch(a: BuildIdentity, b: BuildIdentity): boolea
...
 }
+
+/**
+ * Render a {@link BuildIdentity} as a single human-readable token:
+ * `"<buildId> (<entryScript>)"`. Falls back to `unknown` for an empty entry
+ * script so the rendering stays stable for legacy daemons that predate build
+ * identity. Shared by `doctor` and the `--daemon status` CLI so both surface the
+ * daemon's build in the same format.
+ */
+export function describeBuildIdentity(identity: BuildIdentity): string {
```
kaeawc, line 99: `describeBuildIdentity` documents itself as "Shared by `doctor` and the `--daemon status` CLI so both surface the daemon’s build in the same format" — good. But the **already-shipped user-facing surfaces** of build skew, the MCP error messages in daemonMcpProxy.ts (`DaemonBuildMismatchError`, lines 88-90, and `DaemonToolUnavailableError`, lines 123-124), still hand-roll the identical `build=${buildId} (${entryScript || "unknown"})` token inline rather than calling this helper. Those errors are what the MCP client actually sees at tool-call time, so they are the primary skew surface. As written, the format can drift between the three surfaces. Recommend having those two error constructors also render via `describeBuildIdentity(...)` so all three surfaces stay in lockstep — that is the whole point of extracting the helper.

src/doctor/checks/automobile.ts
```diff
@@ -145,6 +157,64 @@ export async function checkDaemonConnectivity(
...
+
+    return {
+      name: "Daemon Build Identity",
+      status: "warn",
+      message: `Build skew: daemon ${describeBuildIdentity(daemon)}, client ${describeBuildIdentity(client)}`,
+      recommendation:
+        "The running daemon is a different build than this checkout. Restart it from here: " +
+        "bunx @kaeawc/auto-mobile@latest --daemon restart",
+    };
+  } catch (error) {
```
kaeawc, line 209: This `catch` returns a typed failure without logging the underlying error, violating error-handling convention #2 in `CLAUDE.md` — which names `doctor` checks as the example and requires logging *before* returning so there is a trace when the user only sees the summarized message. The user sees `Could not check daemon build identity: <msg>` but no stack/trace is recorded. Add a log before returning:

```ts
} catch (error) {
  logger.warn(`daemon build identity check failed: ${error instanceof Error ? error.message : String(error)}`, error);
  return {
    name: \"Daemon Build Identity\",
    status: \"warn\",
    message: `Could not check daemon build identity: ${error instanceof Error ? error.message : String(error)}`,
  };
}
```

`logger` is already imported (line 23). (The pre-existing `checkDaemonStatus` catch at lines 110-117 has the same gap, but this new check should follow the convention.)

kaeawc (commented):
One finding on the new doctor check's error path; the build-identity skew semantics, undefined-vs-empty handling, and CLI status wiring all check out and are well-covered by the new tests.

src/doctor/checks/automobile.ts
kaeawc: Done in 619ef8384ff23feeef340f710fcfea3b53d34ce0 — extracted `buildIdentityFromStatus()` in `buildIdentity.ts` and replaced all 5 inlined projections (doctor, `--daemon status`, and the 3 proxy sites). Used a narrow structural param instead of importing `DaemonStatus` to avoid the cycle.

src/daemon/buildIdentity.ts
kaeawc: Done in 619ef8384ff23feeef340f710fcfea3b53d34ce0 — `DaemonBuildMismatchError` and `DaemonToolUnavailableError` now format both identities via `describeBuildIdentity`, so all three skew surfaces share one format (message text is byte-identical, so existing proxy tests still pass).

src/doctor/checks/automobile.ts
kaeawc: Done in 619ef8384ff23feeef340f710fcfea3b53d34ce0 — `checkDaemonBuildIdentity`'s catch now `logger.warn(...)`s the underlying error before returning the typed `warn` (convention #2). Added a test covering the status-read-failure path.

src/doctor/checks/automobile.ts
```diff
@@ -145,6 +157,64 @@ export async function checkDaemonConnectivity(
...
+        status: "skip",
+        message: "Daemon is not running",
+      };
+    }
+
+    const daemon: BuildIdentity = {
+      entryScript: status.entryScript ?? "",
+      buildId: status.buildId ?? "unknown",
+    };
+    const client = (dependencies.getClientBuildIdentity ?? getCurrentBuildIdentity)();
```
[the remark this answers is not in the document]
kaeawc: Fixed in 3a08166999659d9eee11e2e58a7a071f25b9b063. You're right — `--cli doctor` is daemon-first, so `runDoctor` executes in the daemon process and the check compared the daemon to itself (always pass). The invoking process is the only one that knows the real client build, so the CLI now reconciles the check client-side: `handleDoctorResult` calls the new `applyClientBuildIdentity()`, which re-runs `checkDaemonBuildIdentity()` in the CLI process (its own `getCurrentBuildIdentity()` vs the daemon's PID-file identity), splices the authoritative verdict into the report, and recomputes summary/recommendations so a real skew is surfaced and counted. (The pure-MCP `doctor` tool path still self-compares in the daemon, but the proxy already detects that skew at connection time via `DaemonBuildMismatchError`.)

src/doctor/checks/automobile.ts
```diff
@@ -145,6 +158,68 @@ export async function checkDaemonConnectivity(
...
+      };
+    }
+
+    return {
+      name: "Daemon Build Identity",
+      status: "warn",
+      message: `Build skew: daemon ${describeBuildIdentity(daemon)}, client ${describeBuildIdentity(client)}`,
+      recommendation:
+        "The running daemon is a different build than this checkout. Restart it from here: " +
+        "bunx @kaeawc/auto-mobile@latest --daemon restart",
```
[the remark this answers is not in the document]
kaeawc: Fixed in 3a08166999659d9eee11e2e58a7a071f25b9b063. Dropped the hard-coded `@latest` from the skew remediation in both `doctor` and `--daemon status` — it now tells the user to restart from THIS checkout with the same CLI they invoked (which `resolveDaemonLaunchCommand()` already honors via the current entry script), and explicitly warns that `@latest` would start the published build instead.

PR #2741, opened 2026-06-30
fix(doctor): detect iOS runner/client port mismatch via runner-reported port (#2735)

### LINKED ISSUE #2735: doctor: iOS round-trip check can't detect runner/client port mismatch
The iOS observe round-trip check derives **both** `runnerPort` and `clientPort` from `manager.getServicePort()` (`src/doctor/checks/ios.ts:253-254`), so `hasPortMismatch` (`:913`) always compares two identical numbers and is structurally incapable of catching a runner-binds-wrong-port mismatch — exactly the failure mode of #2731.

Follow-up to #2734, which fixed the underlying bug (the simulator/device runner now receives the allocated port via the xctestrun `EnvironmentVariables`) but left the diagnostic blind: `doctor --ios` reported `runnerPort=8767; clientPort=8767; connected=false` even though the runner was actually bound to `8765`.

## Fix direction
Make the runner report its **actual bound port** back to the daemon (e.g. in the `/health` payload), then have the check compare the *reported* runner port against the client port so a genuine mismatch surfaces. This mirrors the host-control self-heal pattern at `IOSCtrlProxyManager.ts:1742-1749` (adopt/re-tunnel to the port the process reports).

Refs #2731, #2734.

kaeawc:
## Summary
Fixes #2735. The iOS observe round-trip check derived **both** `runnerPort` and `clientPort` from `manager.getServicePort()` (`src/doctor/checks/ios.ts:253-254`), so `hasPortMismatch` (`:913`) always compared two identical numbers and was structurally incapable of catching a runner-binds-wrong-port mismatch — the #2731 failure mode (`doctor --ios` reported `runnerPort=8767; clientPort=8767; connected=false` while the runner was actually healthy on `8765`).

Follow-up to #2734, which fixed the underlying bug (the runner now receives the allocated port via the xctestrun `EnvironmentVariables`) but left the diagnostic blind.

## Root cause
`createIosObserveRoundTripInspector` set `runnerPort = manager.getServicePort()` and `clientPort = runnerPort`. The client's connection port is also derived from the daemon's allocated service port, so both sides of the mismatch comparison were the *daemon's* port — never the runner's **actual bound** port.

## Fix
The issue's stated direction: make the runner report its actual bound port back to the daemon via `/health`, then compare the *reported* runner port against the client port.

- **`ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift`**: thread the server's bound port into `WebSocketConnection` and echo it in the `/health` payload (`{"status":"ok","deviceId":...,"port":<boundPort>}`).
- **`src/utils/IOSCtrlProxyManager.ts`**: new `getReportedRunnerPort(): Promise<number | null>` — probes `/health` on the candidate ports (allocated service port + hardcoded default, mirroring `findExternalDirectCtrlProxyProcess`) and returns the reported `port`, **filtered by `deviceId`** so a sibling/Android runner on the shared default port is never mistaken for ours. Returns null when no matching runner answers or the runner is too old to report a port. Added to the `CtrlProxyIosManager` interface (+ `NoOpIOSCtrlProxyManager`, `FakeIOSCtrlProxyManager`).
- **`src/doctor/checks/ios.ts`**: `runnerPort` now comes from `getReportedRunnerPort()` (falling back to the service port when unavailable, so a healthy runner is never falsely flagged); `clientPort` stays the client connection port. When the runner is unreachable on the client port but reports a **different** bound port, the check surfaces an actionable `iOS CtrlProxy runner is bound to port X but the client expects port Y` instead of the misleading "not running". Added the method to the `IosRunnerManager` interface.

No false-negative introduced: when no runner answers, `runnerPort` falls back to the service port (no spurious mismatch) while `!connected` still flags a genuinely-down runner.

## Test plan
- `test/utils/IOSCtrlProxyManager.test.ts` — `getReportedRunnerPort`: reports the bound port on the service port; discovers the runner on the default port when the service port is silent (#2731); ignores a health response from a different device on a shared port; returns null for an older runner that omits `port`; returns null when no runner answers.
- `test/doctor/ios.test.ts` — round-trip inspector: reports the runner's reported bound port decoupled from the client port; surfaces the actionable mismatch error; falls back to the service port when the runner reports no port.
- `bun test` → 5220 pass / 0 fail (409 files); `eslint --fix` and `bun build.ts` → green; `swift build`/`swift test` for `ios/control-proxy` → green. No new `tsc` errors introduced.

### Verify on hardware
With one Android emulator + one iOS simulator connected (the #2731 setup), run `auto-mobile --doctor --ios`: a runner bound to the default port while the client expects the allocated port should now report a port mismatch with both ports and the actionable message — not "runner is not running".

### Follow-up required: re-cut the iOS runner release
This change adds the `port` field to the runner's `/health` protocol. Normal installs run the **downloaded pinned runner** (`IOS_CTRL_PROXY_IPA_URL` → `src/constants/release.ts`, currently v0.0.39), which does not yet report `port`. Until the runner release is re-cut and the checksum registry bumped (a coordinated `chore:` release PR — the `ipaSha256` must match a published artifact and cannot be hand-authored here), the deployed runner won't advertise its port.

This is intentionally **fail-safe**: when no `port` is reported, `getReportedRunnerPort()` returns null and the check falls back to the service port — no false mismatch, identical to today's behavior. The new diagnostic activates automatically once the updated runner ships. Locally it can be exercised now via `AUTOMOBILE_CTRL_PROXY_IOS_BUNDLE_PATH` pointing at a source build.

Refs #2731, #2734.

### DIFF
--- a/ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift
+++ b/ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift
@@ -115,6 +115,7 @@ public class WebSocketServer: WebSocketServing {
             id: connectionId,
             connection: nwConnection,
             queue: queue,
+            boundPort: port,
             sdkHierarchyCache: sdkHierarchyCache
         ) { [weak self] message in
             self?.handleMessage(message, connectionId: connectionId)
@@ -329,12 +330,16 @@ class WebSocketConnection {
     private let onMessage: (Data) -> Void
     private let onClose: () -> Void
     private let sdkHierarchyCache: (any SdkHierarchyCaching)?
+    /// The port the server is actually bound to; echoed in /health so the daemon
+    /// can detect a runner/client port mismatch (issue #2735).
+    private let boundPort: UInt16
     private var isWebSocketUpgraded = false
 
     init(
         id: Int,
         connection: NWConnection,
         queue: DispatchQueue,
+        boundPort: UInt16,
         sdkHierarchyCache: (any SdkHierarchyCaching)? = nil,
         onMessage: @escaping (Data) -> Void,
         onClose: @escaping () -> Void
@@ -342,6 +347,7 @@ class WebSocketConnection {
         self.id = id
         self.connection = connection
         self.queue = queue
+        self.boundPort = boundPort
         self.sdkHierarchyCache = sdkHierarchyCache
         self.onMessage = onMessage
         self.onClose = onClose
@@ -413,7 +419,7 @@ class WebSocketConnection {
     }
 
     private func handleHealthCheck() {
-        var payload: [String: String] = ["status": "ok"]
+        var payload: [String: Any] = ["status": "ok", "port": Int(boundPort)]
         if let deviceId = Self.currentDeviceId() {
             payload["deviceId"] = deviceId
         }
--- a/src/utils/IOSCtrlProxyManager.ts
+++ b/src/utils/IOSCtrlProxyManager.ts
@@ -1380,6 +1380,11 @@ export class IOSCtrlProxyManager implements CtrlProxyIosManager {
     return null;
   }
 
+  /** Single source of truth for "is this a usable TCP port number". */
+  private static isValidPort(value: unknown): value is number {
+    return typeof value === "number" && Number.isInteger(value) && value > 0 && value <= 65535;
+  }
+
   private parseCtrlProxyPortFromProcessArgs(args: string): number | null {
     const match = args.match(/(?:^|\s)(?:SIMCTL_CHILD_)?CTRL_PROXY_IOS_PORT=(\d+)(?:\s|$)/);
     if (!match) {
@@ -1387,7 +1392,7 @@ export class IOSCtrlProxyManager implements CtrlProxyIosManager {
     }
 
     const port = Number.parseInt(match[1], 10);
-    return Number.isNaN(port) || port <= 0 || port > 65535 ? null : port;
+    return IOSCtrlProxyManager.isValidPort(port) ? port : null;
   }
 
   private async ensureServicePortReadyForLaunch(): Promise<void> {
@@ -1920,11 +1925,7 @@ export class IOSCtrlProxyManager implements CtrlProxyIosManager {
       if (health.deviceId !== this.device.deviceId) {
         return null;
       }
-      if (typeof health.port !== "number" || !Number.isInteger(health.port) ||
-          health.port <= 0 || health.port > 65535) {
-        return null;
-      }
-      return health.port;
+      return IOSCtrlProxyManager.isValidPort(health.port) ? health.port : null;
     } catch {
       return null;
     }
--- a/src/utils/IOSCtrlProxyManager.ts
+++ b/src/utils/IOSCtrlProxyManager.ts
@@ -47,6 +47,7 @@ export interface CtrlProxyIosManager extends ProxyManager {
   start(): Promise<void>;
   stop(): Promise<void>;
   getServicePort(): number;
+  getReportedRunnerPort(): Promise<number | null>;
   setAutoRestart(enabled: boolean): void;
   isAutoRestartEnabled(): boolean;
   forceRestart(): Promise<void>;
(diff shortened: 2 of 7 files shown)


kaeawc (commented):
No findings — reviewed for the TS-correctness / test-rigor lens (Priya). The fix is sound and the regression is genuinely caught.

Verified:
- No false-positive mismatch risk: both manager.getServicePort() and the detached client port derive from the same deterministic PortManager.allocate(deviceId), so when getReportedRunnerPort() returns null, runnerPort === servicePort === clientPort (no spurious fail). In the healthy running case the Swift runner echoes its real bound port (== servicePort), so still equal.
- The new "bound to X but client expects Y" branch is reachable and correct for the #2731 mode: isRunning() probes only servicePort while getReportedRunnerPort() probes {servicePort, DEFAULT_PORT}, so a runner bound to the default while the service port is silent yields running=false + reportedRunnerPort!=servicePort and hits the explicit mismatch message rather than the misleading "not running".
- Unconditional getReportedRunnerPort() per booted sim has no correctness impact: the !installed branch still wins the message.
- Test coverage is adequate end-to-end: ios.test.ts:757-766 already proves checkIosObserveRoundTrip returns status=fail with both ports in the message when runnerPort!=clientPort, and the new inspector tests prove the inspector now emits decoupled ports (runnerPort=8765, clientPort=8767) for the #2731 scenario; the halves compose.
- bun test test/doctor/ios.test.ts test/utils/IOSCtrlProxyManager.test.ts: 138 pass, 0 fail.

One non-blocking note (not posted inline, defensive-only): readReportedPortFromHealth accepts any numeric health.port, unlike parseCtrlProxyPortFromProcessArgs which rejects <=0/>65535. The Swift runner echoes Int(boundPort) (a UInt16 == the actual bound port), so a bogus value is unreachable in practice and changes no real outcome — flagging only for parity if you want symmetric validation.

src/utils/IOSCtrlProxyManager.ts
```diff
@@ -1869,6 +1870,50 @@ export class IOSCtrlProxyManager implements CtrlProxyIosManager {
...
+      return null;
+    }
+    try {
+      const health = JSON.parse(body) as { status?: unknown; deviceId?: unknown; port?: unknown };
+      if (health.status !== "ok") {
+        return null;
+      }
+      // A runner that reports a deviceId must match ours; one that omits it
+      // (older runner) is accepted on the ports we already scope to this device.
+      if (health.deviceId !== undefined && health.deviceId !== this.device.deviceId) {
```
kaeawc: Real (low-severity) scoping gap: the `deviceId === undefined` accept branch can adopt *another iOS device's* runner under exactly the failure mode this PR targets.

The runner only omits `deviceId` from /health when `currentDeviceId()` returns nil, i.e. neither `AUTOMOBILE_DEVICE_ID` nor `SIMULATOR_UDID` reached the runner (ios WebSocketServer.swift:437-439). But `SIMCTL_CHILD_AUTOMOBILE_DEVICE_ID` and `SIMCTL_CHILD_CTRL_PROXY_IOS_PORT` are forwarded on the *same* simctl channel (IOSCtrlProxyManager.ts:1033 + 1035). The #2731 root cause is precisely that `CTRL_PROXY_IOS_PORT` failed to propagate (runner fell back to 8765). When that propagation fails, the deviceId var very likely fails too -> the runner reports `{"status":"ok","port":8765}` with **no deviceId** -> this guard's `undefined` branch accepts it for *any* probing device.

In a two-simulator setup, manager-for-sim-A (servicePort 8767) probes 8765, finds sim-B's deviceId-less runner, and adopts port 8765 as "its" reported port. Note `port` and `deviceId` are NOT coupled the way the doc comment implies: `deviceId` predates this PR (it exists on HEAD~1), only `port` is new -- so a runner can genuinely report `port` while omitting `deviceId`.

Impact is bounded: this only feeds a misleading doctor message ("bound to 8765 but client expects 8767" attributed to the wrong simulator), not a wrong client connection -- the wrong-device adoption only occurs when `isRunning()` on servicePort is false, which routes to the error branch (ios.ts:280) rather than `createClient(device, runnerPort)`. Still, the deviceId scoping silently degrades in the exact scenario it's meant to protect.

Suggest: drop the `undefined` acceptance for the *port-reporting* path. Since any runner new enough to emit `port` is also new enough to emit `deviceId` (both controlled by the same currentDeviceId()/env), require a positive deviceId match here: `if (health.deviceId !== this.device.deviceId) return null;`. That removes the cross-device window with no loss of genuine-runner coverage.

kaeawc (commented):
Reviewed through a discovery / false-adoption lens (Dmitri). The fix is well-targeted and the #2731 detection path works end-to-end. I verified the four scenarios that worried me; one is a real (low-severity) scoping gap, the rest are sound.

**Does it observe the broken state? Yes.** In the #2731 case (runner on 8765, servicePort 8767): `isRunning()` -> `checkHealthEndpoint()` -> `checkHealthEndpointOnPort(this.servicePort=8767)` finds nothing and returns false, while `getReportedRunnerPort()` probes the `Set([8767, 8765])` and finds 8765. The doctor then hits the new mismatch branch (ios.ts:280) and emits the actionable error. Confirmed by the new tests (138 pass, 120ms).

**Candidate-port set is complete given the runner's design.** The runner derives its port from `CTRL_PROXY_IOS_PORT` env, falling back to `CtrlProxy.defaultPort` = 8765 (CtrlProxyUITests.swift:99-106, CtrlProxy.swift:9). TS sets that env to `this.servicePort` (IOSCtrlProxyManager.ts:1030/1033). So the runner can only ever bind servicePort or 8765 -> `{servicePort, DEFAULT_PORT}` covers all real cases. The new `boundPort` echoed in /health (WebSocketServer.swift:118,422) is the genuine bind target, so the reported value is trustworthy.

**Reported==null + runner down still fails (not silently passes).** Both null and `isRunning()`==false -> the `else` at ios.ts:286-287 keeps "runner is not running". Good.

**Caching/TOCTOU: benign.** `getReportedRunnerPort()` does fresh curl probes (no cached port); `isRunning()` only caches a boolean. This is a diagnostic path, so the probe->connect window is acceptable.

**Cross-PLATFORM false adoption (Android adb-forward on 8765): mitigated, but not by the mechanism the code comments claim.** Android `/health` returns plain text `"OK"` (android/.../WebSocketServer.kt:327), not JSON. `JSON.parse` either throws or yields a non-object, so `health.status === "ok"` is false and the Android response is rejected. The deviceId guard the PR comment leans on never even runs for Android. Fine today, but it means the safety actually rests on the case-sensitive `status === "ok"` object check, not on deviceId scoping -- worth knowing if anyone later changes Android's payload to JSON.

See the inline comment for the one real gap (deviceId-less cross-device adoption).

src/utils/IOSCtrlProxyManager.ts
kaeawc: Fixed in 7fb629e1. `readReportedPortFromHealth` now requires an exact device match (`health.deviceId !== this.device.deviceId → null`) before trusting a reported port, closing the sibling-simulator false-adoption edge. This loses no genuine coverage: a runner old enough to omit its device id is also too old to report a `port` (rejected by the numeric check anyway). Also added out-of-range port rejection and corrected the comment to credit the real cross-platform defense (Android /health returns plain text `OK`, rejected by the strict JSON `status === "ok"` check). New tests: deviceId-less rejection + out-of-range rejection.

ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift
```diff
@@ -413,7 +419,7 @@ class WebSocketConnection {
     }
 
     private func handleHealthCheck() {
-        var payload: [String: String] = ["status": "ok"]
+        var payload: [String: Any] = ["status": "ok", "port": Int(boundPort)]
```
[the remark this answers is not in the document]
kaeawc: Correct, and noted as a required follow-up. Two points:

1. **Fails safe, no regression.** When the deployed runner doesn't report `port`, `getReportedRunnerPort()` returns null and the check falls back to `servicePort` (`runnerPort = reportedRunnerPort ?? servicePort`), so it never produces a *false* mismatch — the diagnostic simply stays dormant until the runner ships, exactly as it does today.

2. **The release cut can't ride in this PR.** `ipaSha256` in `src/constants/release.ts` must match an actually-published artifact; hand-authoring a checksum here would break downloads. This repo cuts the runner release via dedicated `chore:` PRs (e.g. #2679, #1992) off the release workflow. This PR is the protocol + diagnostic change; the runner re-cut + registry bump is the coordinated follow-up. I've called this out explicitly in the PR description.

PR #2742, opened 2026-06-30
feat(daemon): stamp git SHA into pre-release version for dev builds (#2738)

### LINKED ISSUE #2738: Stamp git SHA into pre-release daemon version so the version gate distinguishes dev builds
## Summary

The underlying enabler of the wrong-build daemon skew (#2732) is that pre-release builds all report the same `package.json` version (e.g. `0.0.39`), so `DaemonMcpProxy.ensureVersionMatches()` early-returns and the version gate is blind. #2733 makes the version string irrelevant for *skew detection* by adding a content-hash build identity, but the version gate itself is still blind pre-release.

## Proposal (belt-and-suspenders)

Stamp a git short SHA (and/or `-dev`/dirty marker) into the version string for non-release/dev builds, so two checkouts at different commits report different versions and the existing version gate also distinguishes them. Lower priority now that build identity (`src/daemon/buildIdentity.ts`) exists, but it would make the version-based diagnostics (`doctor`, logs, `DaemonVersionMismatchError`) meaningful for dev builds.

## Acceptance

- Dev/non-release builds report a version that varies with the git commit.
- Release builds are unchanged.

Refs #2732, #2733

kaeawc:
## Summary
Pre-release builds all report the same `package.json` version (`0.0.39`), so the daemon version gate (`DaemonMcpProxy.ensureVersionMatches()`) and every version-based diagnostic (`doctor`, logs, `DaemonVersionMismatchError`) are blind to dev-build skew between two checkouts (#2732). This stamps a git short SHA (+ dirty marker) into the version string for **source checkouts** so the reported version varies with the commit, while **release installs are unchanged**.

Belt-and-suspenders to the content-hash build identity from #2733: that gate already detects + self-heals wrong-build daemons; this makes the *version string itself* meaningful for dev builds.

## Changes
- **`src/utils/mcpVersion.ts`** — `getMcpServerVersion()` resolves a base version (explicit `MCP_SERVER_VERSION` override > `npm_package_version` > `package.json`) and, for a source checkout, appends `+g<shortSha>[.dirty]`. Refactored into pure, injectable pieces:
  - `formatMcpServerVersion(base, git)` — pure formatter (no git → unchanged).
  - `resolveMcpServerVersion(deps)` — pure resolution with injected env/package/git readers.
  - `readGitVersion(cwd)` — one-time `git rev-parse --short=12 HEAD` + tracked-only dirty check; returns `null` (→ unstamped) when git is absent. **Skips any `node_modules` location** so a published package nested inside a host git repo is never stamped with the host's commit.
- **`src/daemon/daemonMcpProxy.ts`**
  - `ensureVersionMatches()` compares only the **release portion** (`baseVersion()` strips `+…` metadata) and **defers** same-release dev-skew (`0.0.39+gAAA` vs `0.0.39+gBBB`) to the build-identity gate (`ensureBuildMatches`, #2733) instead of hard-throwing `nonNumeric` — preserving #2733's restart/self-heal.
  - `DaemonVersionMismatchError` restart hint strips build metadata (a `+`-stamped version is not an installable npm tag): `bunx @kaeawc/auto-mobile@0.0.39 --daemon restart`.
  - Added injectable `clientVersion` to `DaemonMcpProxyConfig` (decouples the gate tests from the ambient stamped `DAEMON_VERSION`).

## Acceptance check
- Dev build (this checkout): `getMcpServerVersion()` → `0.0.39+gec6f9cee8444.dirty` ✓ (varies with commit)
- Release build (no git / `node_modules`): `0.0.39` unchanged ✓
- `MCP_SERVER_VERSION` override returned verbatim ✓

## Test plan
- New `test/utils/mcpVersion.test.ts` — pure unit tests for formatter, resolver, release detection (no git/network).
- `test/daemon/daemonMcpProxy.test.ts` — restart-hint metadata stripping; dev-skew **defers** to the build gate (does not restart/throw); existing version-gate suite pinned to a plain injected `clientVersion`.
- `test/daemon/integration/versionMismatchRestart.test.ts` — pinned client version for the numeric-comparison path.
- `bun test` → 5264 pass, 0 fail. `turbo run lint build` → green.

Refs #2732, #2733
Closes #2738

### DIFF
--- a/src/daemon/daemonMcpProxy.ts
+++ b/src/daemon/daemonMcpProxy.ts
@@ -18,6 +18,18 @@ export type VersionMismatchReason =
   | "nonNumeric"
   | "restartMismatch";
 
+/**
+ * Strip semver build metadata (everything after the first `+`) from a version.
+ *
+ * Dev/non-release builds stamp a git short SHA as build metadata (e.g.
+ * `0.0.39+g1a2b3c4.dirty`, see `getMcpServerVersion`). The numeric version gate
+ * compares only the release portion; reconciling two builds that share a release
+ * version but differ by commit is the build-identity gate's job, not the version
+ * gate's. The metadata is also not an installable npm tag, so the restart hint
+ * uses the stripped version too.
+ */
+export const baseVersion = (version: string): string => version.split("+")[0];
+
 export type BuildMismatchReason =
   | "autoStartDisabled"
   | "cooldown"
@@ -40,8 +52,11 @@ export class DaemonVersionMismatchError extends DaemonUnavailableError {
     detail: string;
     retryAfterMs?: number;
   }) {
-    const restartCommand = params.clientVersion.length > 0 && params.clientVersion !== "unknown"
-      ? `bunx @kaeawc/auto-mobile@${params.clientVersion} --daemon restart`
+    // The git-SHA build metadata on dev builds is not an installable npm tag,
+    // so the bunx hint must point at the plain published version.
+    const installableVersion = baseVersion(params.clientVersion);
+    const restartCommand = installableVersion.length > 0 && installableVersion !== "unknown"
+      ? `bunx @kaeawc/auto-mobile@${installableVersion} --daemon restart`
       : "the same installed auto-mobile package";
     const retryGuidance = params.retryAfterMs !== undefined
       ? ` Retry after ${params.retryAfterMs}ms or restart the daemon with: ${restartCommand}`
@@ -151,6 +166,8 @@ export interface DaemonMcpProxyConfig {
   timer?: Pick<Timer, "now">;
   /** This client's build identity (for testing; defaults to the current process build) */
   buildIdentity?: BuildIdentity;
+  /** This client's version for the daemon version gate (defaults to DAEMON_VERSION; injectable for testing) */
+  clientVersion?: string;
 }
 
 /**
@@ -199,6 +216,7 @@ export class DaemonMcpProxy {
   private clientFactory: DaemonClientFactory;
   private readonly timer: Pick<Timer, "now">;
   private readonly buildIdentity: BuildIdentity;
+  private readonly clientVersion: string;
   private connecting: Promise<void> | null = null;
   private connected: boolean = false;
 
@@ -221,6 +239,7 @@ export class DaemonMcpProxy {
     ));
     this.timer = config.timer ?? defaultTimer;
     this.buildIdentity = config.buildIdentity ?? getCurrentBuildIdentity();
+    this.clientVersion = config.clientVersion ?? DAEMON_VERSION;
   }
 
   /**
@@ -285,19 +304,30 @@ export class DaemonMcpProxy {
     }
 
     const runningVersion = status.version?.trim() ?? "";
-    if (runningVersion === DAEMON_VERSION) {
+    if (runningVersion === this.clientVersion) {
       return;
     }
 
-    const cmp = runningVersion.length > 0
-      ? compareVersions(DAEMON_VERSION, runningVersion)
+    // Compare only the release portion. Two dev builds that share a release
+    // version but differ by git commit (`0.0.39+gAAA` vs `0.0.39+gBBB`) are
+    // reconciled by the build-identity gate (ensureBuildMatches, which restarts
+    // to the correct build on a content-hash mismatch). The version gate must
+    // defer here rather than hard-throw, or it would defeat that self-heal.
+    const runningBase = baseVersion(runningVersion);
+    const clientBase = baseVersion(this.clientVersion);
+    if (runningBase === clientBase) {
+      return;
+    }
+
+    const cmp = runningBase.length > 0
+      ? compareVersions(clientBase, runningBase)
       : Number.POSITIVE_INFINITY;
 
     if (!this.config.autoStartDaemon) {
       throw this.versionMismatchError(runningVersion, "autoStartDisabled", "auto-start is disabled");
     }
 
-    if (runningVersion.length > 0 && !Number.isFinite(cmp)) {
+    if (runningBase.length > 0 && !Number.isFinite(cmp)) {
       throw this.versionMismatchError(
         runningVersion,
         "nonNumeric",
@@ -317,7 +347,7 @@ export class DaemonMcpProxy {
       const daemonAgeMs = this.timer.now() - status.startedAt;
       if (daemonAgeMs < DAEMON_VERSION_RESTART_COOLDOWN_MS) {
         logger.warn(
-          `[DaemonMcpProxy] Skipping version-mismatch restart due to cooldown: daemon ${runningVersion || "unknown"} is ${daemonAgeMs}ms old, client version is ${DAEMON_VERSION}`
+          `[DaemonMcpProxy] Skipping version-mismatch restart due to cooldown: daemon ${runningVersion || "unknown"} is ${daemonAgeMs}ms old, client version is ${this.clientVersion}`
         );
         throw this.versionMismatchError(
           runningVersion,
@@ -329,7 +359,7 @@ export class DaemonMcpProxy {
     }
 
     logger.info(
-      `[DaemonMcpProxy] Daemon version ${runningVersion || "unknown"} differs from MCP server ${DAEMON_VERSION}, restarting daemon`
+      `[DaemonMcpProxy] Daemon version ${runningVersion || "unknown"} differs from MCP server ${this.clientVersion}, restarting daemon`
     );
     await this.daemonManager.restart(this.config.daemonOptions ?? {});
     // The replacement daemon may expose a different tool set; drop the cache so we
@@ -344,7 +374,7 @@ export class DaemonMcpProxy {
 
     const restartedStatus = await this.daemonManager.status();
     const restartedVersion = restartedStatus.version?.trim() ?? "";
-    if (!restartedStatus.running || restartedVersion !== DAEMON_VERSION) {
+    if (!restartedStatus.running || restartedVersion !== this.clientVersion) {
       throw this.versionMismatchError(
         restartedVersion,
         "restartMismatch",
@@ -360,7 +390,7 @@ export class DaemonMcpProxy {
     retryAfterMs?: number,
   ): DaemonVersionMismatchError {
     const daemonVersion = runningVersion.length > 0 ? runningVersion : "unknown";
-    const clientVersion = DAEMON_VERSION.trim();
+    const clientVersion = this.clientVersion.trim();
     return new DaemonVersionMismatchError({
       clientVersion,
       daemonVersion,
--- a/src/daemon/daemonMcpProxy.ts
+++ b/src/daemon/daemonMcpProxy.ts
@@ -297,40 +297,46 @@ export class DaemonMcpProxy {
       return;
     }
 
-    // Compare only the release portion. Two dev builds that share a release
-    // version but differ by git commit (`0.0.39+gAAA` vs `0.0.39+gBBB`) are
-    // reconciled by the build-identity gate (ensureBuildMatches, which restarts
-    // to the correct build on a content-hash mismatch). The version gate must
-    // defer here rather than hard-throw, or it would defeat that self-heal.
+    // The release portions (before the `+g<sha>` dev stamp) drive the
+    // newer/older decision. Equal release + differing full strings means two
+    // source checkouts at the same release but different commits (dev-skew).
     const runningBase = releaseVersion(runningVersion);
     const clientBase = releaseVersion(this.clientVersion);
-    if (runningBase === clientBase) {
-      return;
-    }
-
-    const cmp = runningBase.length > 0
-      ? compareVersions(clientBase, runningBase)
-      : Number.POSITIVE_INFINITY;
+    const sameRelease = runningBase === clientBase;
 
     if (!this.config.autoStartDaemon) {
       throw this.versionMismatchError(runningVersion, "autoStartDisabled", "auto-start is disabled");
     }
 
-    if (runningBase.length > 0 && !Number.isFinite(cmp)) {
-      throw this.versionMismatchError(
-        runningVersion,
-        "nonNumeric",
-        "version comparison is not numeric"
-      );
+    if (!sameRelease) {
+      const cmp = runningBase.length > 0
+        ? compareVersions(clientBase, runningBase)
+        : Number.POSITIVE_INFINITY;
+
+      if (runningBase.length > 0 && !Number.isFinite(cmp)) {
+        throw this.versionMismatchError(
+          runningVersion,
+          "nonNumeric",
+          "version comparison is not numeric"
+        );
+      }
+
+      if (cmp <= 0) {
+        throw this.versionMismatchError(
+          runningVersion,
+          "daemonNewer",
+          "the running daemon is newer than this client"
+        );
+      }
     }
 
-    if (cmp <= 0) {
-      throw this.versionMismatchError(
-        runningVersion,
-        "daemonNewer",
-        "the running daemon is newer than this client"
-      );
-    }
+    // Reach here when the client is strictly newer OR the daemon is a same-release
+    // dev-skew (different git stamp). Both reconcile by restarting the daemon from
+    // this client's build. Same-release dev-skew must be handled HERE rather than
+    // deferred to ensureBuildMatches: the build-identity hash covers only the entry
+    // script (process.argv[1]), so in unbundled source-mode runs (`bun src/index.ts`)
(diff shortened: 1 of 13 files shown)


kaeawc (commented):
Reviewed as the distributed-systems skeptic, hunting for restart storms, version-gate masking, and #2733 self-heal regressions. Traced every branch of ensureVersionMatches (exact-match, same-base defer, autoStartDisabled, nonNumeric, daemonNewer, cooldown, restart, restartMismatch) and walked the two-checkout (both 0.0.39, different SHA) connect sequence: the version gate correctly steps aside (returns) on same-release dev-skew so it no longer competes with the build gate's restart, content-hash backstops stale version caches, and each doConnect performs at most one restart per gate (no livelock). The releaseVersion strip is applied at both numeric-comparison sites (the gate and PlanMigrator) and the only stamped consumers that matter are diagnostics. Even the autoStartDisabled-with-dev-skew path that the version gate now bypasses still gets blocked downstream by ensureBuildMatches. No blocking correctness or architecture issues found from my perspective.

src/utils/mcpVersion.ts
```diff
@@ -20,32 +40,121 @@ const findPackageJson = (startDir: string): string | null => {
...
+ * null when git is unavailable or the package is not inside a working tree
+ * (i.e. a published/release install), which keeps release versions unstamped.
+ *
+ * A published package lives under `node_modules`; `git rev-parse` searches
+ * upward and would otherwise report the *host project's* commit, wrongly
+ * stamping a release build. Treat any node_modules location as a release
+ * install and skip git entirely.
+ */
+export const readGitVersion = (cwd: string = moduleDir()): GitVersionInfo | null => {
+  if (cwd.split(path.sep).includes("node_modules")) {
```
kaeawc: The `node_modules` segment check is the only structural release guard, but it does NOT cover the two most common `bunx @kaeawc/auto-mobile` layouts (the exact command the new restart hint recommends). I verified the real paths on this machine:

- bunx temp extraction: `.../T/bunx-501-@kaeawc/auto-mobile@0.0.38/dist/src/index.js` — `node_modules` is a *sibling* dep dir; the running module's own path has no `node_modules` segment.
- bun global cache: `~/.bun/install/cache/@kaeawc/auto-mobile@0.0.38@@@1/dist/src` — no `node_modules` segment either.

Both pass this guard and fall through to `runGit`, relying solely on git being absent above them. But `git rev-parse --short=12 HEAD` searches *upward*. I reproduced a release install getting falsely stamped:

```
# release-style dir with NO node_modules segment, nested in a host git repo
mkdir -p reltest/.bun-cache/auto-mobile-0.0.39/dist && cd reltest && git init -q && git commit -q --allow-empty -m host
readGitVersion(".../reltest/.bun-cache/auto-mobile-0.0.39/dist")
  => { shortSha: "d768d0e150d6", dirty: false }
formatMcpServerVersion("0.0.39", g) => "0.0.39+gd768d0e150d6"   # the HOST repo's commit
```

This violates acceptance #2 (release builds unchanged) AND reports a version that points at an unrelated project's SHA — actively misleading for diagnostics. The `node_modules` heuristic is a proxy for "is this a release install"; a more robust signal is needed. Suggested fix: positively detect a release install (e.g. `git rev-parse --show-toplevel` and confirm the package.json we resolved is actually *tracked* in that repo / the toplevel is an ancestor of moduleDir AND owns our package.json), or gate stamping behind an explicit dev signal (e.g. `AUTOMOBILE_DEV_BUILD=1` set by `bun run`, or stamp at build.ts time into the artifact) rather than 'git happens to resolve from cwd'.

src/utils/mcpVersion.ts
```diff
@@ -20,32 +40,121 @@ const findPackageJson = (startDir: string): string | null => {
...
-      const parsed = JSON.parse(raw) as { version?: string };
-      if (parsed.version) {
-        cachedVersion = parsed.version;
-        return parsed.version;
-      }
-    } catch {
-      // Fall through to unknown
+const runGit = (cwd: string, args: string[]): string | null => {
+  try {
+    const result = spawnSync("git", args, { cwd, encoding: "utf-8", timeout: 2000 });
```
kaeawc, line 83: Timeout stacking + startup cost. `readGitVersion` makes up to two sequential `spawnSync` calls, each with `timeout: 2000`, so the worst case is ~4s of blocking — and this runs at module load via `DAEMON_VERSION = getMcpServerVersion()` in `daemon/constants.ts:136`, i.e. on EVERY daemon and CLI process start (it's only cached within a single process). On a healthy repo it's ~27ms here (measured), but on a slow/network filesystem or a pathological repo, a daemon/CLI startup can stall up to 4s synchronously before anything else runs. Two mitigations worth considering: (1) skip the second `git status` spawn and the whole git path entirely once a release signal is established (see the node_modules-guard comment) so release installs never spawn git at all; (2) better direction — stamp the SHA at build time in `build.ts` (write it into the bundled artifact as a constant or define), so the published/dev artifact carries its identity with zero runtime git spawn and zero ambient-cwd ambiguity. Build-time stamping also closes the host-repo false-stamp hole structurally.

kaeawc (commented):
Release-engineering review focused on whether the reported version is correct across every install path. The version-string refactor and the daemon-gate split (compare release portion, defer dev-skew to the build-identity gate) are sound. But the release-detection heuristic has a real hole: the two most common `bunx @kaeawc/auto-mobile` paths are NOT under a `node_modules` segment, so they pass the node_modules guard and rely entirely on git being absent above them — and `git rev-parse` walks upward, so a release install that happens to sit inside a host git tree gets stamped with the HOST repo's SHA, violating acceptance #2 and reporting a misleading version. I reproduced this. Separately, the synchronous git spawn runs at module load (DAEMON_VERSION in daemon/constants.ts) on every daemon/CLI start with two stacked 2s timeouts (up to 4s) — worth gating on a release signal and/or build-time stamping. Details inline.

test/daemon/daemonMcpProxy.test.ts
```diff
@@ -227,7 +234,26 @@ describe("DaemonMcpProxy", () => {
...
+        try {
+          await proxy.listTools();
+          expect(fakeManager.restartCalled).toBe(false);
+        } finally {
+          isAvailableSpy.mockRestore();
+          await proxy.close();
+        }
+      });
+
+      test("defers dev-build skew (same release version, different git stamp) to the build-identity gate", async () => {
```
kaeawc: This is the only test that drives a *stamped* string through `ensureVersionMatches`, but it only proves the version gate *defers* — the build identity here is absent (`makeProxy` sets no `buildId`/`entryScript`), so `buildIdentitiesMatch` returns true and the proxy just connects; the build gate never restarts. The combined behavior the issue is actually about — same release version + a *different* build identity → version gate defers AND build gate self-heals by restarting — is covered nowhere. The `makeBuildProxy` suite can't cover it either: it sets `version: DAEMON_VERSION` on both sides, so `ensureVersionMatches` exits at the `runningVersion === this.clientVersion` equality check (line ~296) before the new `releaseVersion` deferral (line ~305) is ever reached. I verified the implementation does self-heal in this combined case (proxy with `clientVersion: "0.0.39+gAAA"`, `runningVersion: "0.0.39+gBBB"`, differing buildId → `restartCalled === true`). Add a test that pins differing stamped versions sharing a release AND differing build identities and asserts `restartCalled === true`, so the deferral-into-build-gate path is pinned rather than only exercised incidentally.

test/utils/mcpVersion.test.ts
```diff
@@ -0,0 +1,132 @@
...
+});
+
+describe("readGitVersion (release-install detection)", () => {
+  test("returns null for a node_modules location without consulting git (release install)", () => {
+    // A published package nested under a host project's git repo must not be
+    // stamped with the host's commit.
+    expect(readGitVersion("/some/host/project/node_modules/@kaeawc/auto-mobile/dist")).toBeNull();
+  });
+
+  test("returns null when the cwd is not a git checkout", () => {
```
kaeawc: Both `readGitVersion` tests only hit early-return null branches (node_modules guard, non-checkout). The success path — spawning git, returning `{ shortSha, dirty }`, and specifically the dirty computation `status !== null && status.length > 0` with `--untracked-files=no` — has zero coverage. `runGit` (a real `spawnSync`) is baked into `readGitVersion` with no injection seam, unlike `readPackageVersion`/`readGitVersion` which are cleanly injected into `resolveMcpServerVersion`. Suggest extracting the git runner as an injectable dep (mirroring `McpVersionDeps`) so the shortSha-success and clean-vs-dirty branches get pure, real-IO-free coverage — right now the most behavior-bearing part of the new module is untested and not testable without spawning git.

kaeawc (commented):
The Test-Quality & API-Design Purist: the core acceptance is implemented (DAEMON_VERSION is stamped via getMcpServerVersion, releaseVersion is the single canonical stamp-stripper, clientVersion is a clean injectable seam, not a backdoor) and I verified the PlanMigrator regression tests genuinely fail against the old parseVersion (both the SHA-hex corruption and the .dirty→NaN paths). Two real coverage gaps remain: the combined self-heal path that is the whole point of #2738 is never proven, and readGitVersion's success/dirty branch has no seam and no test. Both are coverage holes against working code, not bugs.

src/utils/mcpVersion.ts
kaeawc: Fixed in 57a01636. `readGitVersion` now stamps only when the enclosing repo's top-level `package.json` is this package (`@kaeawc/auto-mobile`) — a release copy vendored inside an unrelated git repo (bunx temp / global cache with no `node_modules` segment) no longer picks up the host repo's SHA. Verified with a foreign-repo repro (returns null). Pure unit tests cover the ownership branch.

src/utils/mcpVersion.ts
kaeawc: Addressed in 57a01636. The result is cached for the process lifetime (one resolution), and release installs now short-circuit before the SHA/status spawns (node_modules segment, or a `--show-toplevel` whose package.json isn't ours), so they pay at most one git call. Kept runtime resolution rather than build-time stamping because `bun src/index.ts` dev runs never execute build.ts — runtime git is the only path that covers source-mode, which is exactly where this signal matters most (see the Codex thread below).

src/daemon/daemonMcpProxy.ts
```diff
@@ -285,19 +293,30 @@ export class DaemonMcpProxy {
...
-      ? compareVersions(DAEMON_VERSION, runningVersion)
+    // Compare only the release portion. Two dev builds that share a release
+    // version but differ by git commit (`0.0.39+gAAA` vs `0.0.39+gBBB`) are
+    // reconciled by the build-identity gate (ensureBuildMatches, which restarts
+    // to the correct build on a content-hash mismatch). The version gate must
+    // defer here rather than hard-throw, or it would defeat that self-heal.
+    const runningBase = releaseVersion(runningVersion);
+    const clientBase = releaseVersion(this.clientVersion);
+    if (runningBase === clientBase) {
+      return;
```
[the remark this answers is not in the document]
kaeawc: Good catch — fixed in 57a01636. Rather than defer same-release dev-skew to `ensureBuildMatches` (blind to non-entry-file changes in source mode, as you note), `ensureVersionMatches` now treats it as a reconcilable mismatch and restarts the daemon to this client's build, cooldown-bounded. Added a test that pins the restart on differing stamps at the same release.

test/daemon/daemonMcpProxy.test.ts
kaeawc: Addressed in 57a01636. The version gate now restarts (not defers) on same-release dev-skew, and a `makeProxy`-based test pins `restartCalled === true` for differing stamped versions sharing a release (plus a cooldown-window variant). Removed the obsolete defer-to-build-gate test.

test/utils/mcpVersion.test.ts
kaeawc: Addressed in 57a01636. `readGitVersion` now takes an injectable git runner (`GitRunner`) and package-name reader, so the shortSha-success path, the clean-vs-dirty `status` computation, and the repo-ownership branch all get pure, spawn-free coverage via a fake runner.

kaeawc:
## Review round addressed (57a01636)

Thanks for the thorough reviews. Summary of changes since the initial push:

**Correctness (Release-Engineering + Codex P1)**
- **Release-detection hole** — `readGitVersion` no longer relies on a `node_modules` path segment alone. It stamps only when the enclosing git repo's top-level `package.json` is `@kaeawc/auto-mobile`, so a release copy vendored inside a host git repo (bunx temp / global cache layouts) can't be stamped with the host's SHA. Verified with a foreign-repo repro.
- **Dev-skew must not be silently deferred** — `ensureVersionMatches` previously deferred same-release/different-stamp to `ensureBuildMatches`, but the build-identity hash only covers the entry script and is blind to non-entry-file changes in source mode (`bun src/index.ts`). It now reconciles dev-skew by restarting to this client's build, cooldown-bounded — self-heals in both source and built modes.

**Test coverage (Test-Quality)**
- `readGitVersion` git runner is now injectable; the shortSha-success, clean-vs-dirty, and repo-ownership branches get pure, spawn-free coverage.
- The version gate's dev-skew self-heal (restart) and its cooldown variant are now pinned; obsolete defer-test removed.
- PlanMigrator regression: stamped `mcpVersion` (incl. `.dirty`) is stripped before numeric parse, with tests that fail against the old `parseVersion`.

**Direction note (build-time stamping)** — kept runtime resolution because `bun src/index.ts` dev runs never execute `build.ts`, and source-mode is exactly where the stamp matters most (the build hash is blind there). The result is cached per process, and release installs short-circuit before any SHA/status spawn.

Distributed-systems review found no correctness/regression issues. Full suite: 5275 pass, 0 fail; lint + build green.

src/utils/mcpVersion.ts
```diff
@@ -20,32 +40,149 @@ const findPackageJson = (startDir: string): string | null => {
...
+ * comparisons still work once {@link releaseVersion} strips the metadata, while
+ * the full string still varies per commit for diagnostics (doctor, logs,
+ * `DaemonVersionMismatchError`).
+ */
+export const formatMcpServerVersion = (baseVersion: string, git: GitVersionInfo | null): string => {
+  if (!git || git.shortSha.length === 0) {
+    return baseVersion;
+  }
+  const dirty = git.dirty ? ".dirty" : "";
+  return `${baseVersion}+g${git.shortSha}${dirty}`;
```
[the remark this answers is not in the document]
kaeawc: Fixed in 037f141c. `readGitVersion` now hashes the tracked diff (`git diff HEAD`) into the stamp — `0.0.39+g<sha>.dirty.<diffhash>` — so two checkouts at the same commit with different uncommitted edits report different versions and the gate restarts instead of short-circuiting on string equality. Falls back to a bare `.dirty` when the diff is unavailable; covered by new pure unit tests.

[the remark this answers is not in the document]
kaeawc: Fixed in 037f141c. `PlanSerializer` now persists `releaseVersion(getMcpServerVersion())` into a plan's mcpVersion, so a dev build's `+g<sha>` stamp never reaches the schema gate (`^\d+\.\d+\.\d+$`) that runs before migration. The stamp's purpose (daemon skew detection + diagnostics) is unrelated to plan provenance, so storing the release portion is the right semantics; the migrator's metadata-stripping stays as defense for any externally-authored plans. Added a test asserting the persisted value matches the schema pattern.

src/utils/mcpVersion.ts
```diff
@@ -20,32 +45,164 @@ const findPackageJson = (startDir: string): string | null => {
...
+ * `DaemonVersionMismatchError`).
+ */
+export const formatMcpServerVersion = (baseVersion: string, git: GitVersionInfo | null): string => {
+  if (!git || git.shortSha.length === 0) {
+    return baseVersion;
+  }
+  // Dirty suffix: `.dirty` for the human marker, plus the tracked-diff hash when
+  // available so distinct working trees at the same commit don't collide.
+  const dirty = git.dirty ? (git.dirtyHash ? `.dirty.${git.dirtyHash}` : ".dirty") : "";
+  return `${baseVersion}+g${git.shortSha}${dirty}`;
```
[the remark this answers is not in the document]
kaeawc: Good catch — fixed in ee947566. Both recorders (`mcpRecordingManager.ts`, `testRecordingManager.ts`) now persist `releaseVersion(getMcpServerVersion())` into the recorded plan's mcpVersion, so record/stop/replay from a dev checkout produces a schema-valid plan. The `version:` sites at `proxyServer.ts`/`server/index.ts`/`socketServer.ts` keep the stamp intentionally — they're server-info/IDE-status diagnostics, not schema-validated plans. Added a test asserting a recorded plan's mcpVersion matches `^\d+\.\d+\.\d+$` on a dev build.

PR #2747, opened 2026-07-01
fix(release): wire iOS runner checksum + add version-equality release gate

### LINKED ISSUE #2745: Release integrity: iOS runner-binary checksum never wired (verification disabled) + no version-equality release gate
## Summary

Two release-pipeline integrity gaps let versions/artifacts drift without detection: (A) the iOS CtrlProxy **runner-binary checksum is never populated**, so runner integrity verification is permanently disabled; (B) **no release gate asserts version equality** across the manifests + checksum registry.

## Part A — iOS runner-binary integrity check is permanently disabled (HIGH)

- `IOS_CTRL_PROXY_RUNNER_SHA256` and `IOS_CTRL_PROXY_APP_HASH` are hardcoded to `""` (`src/constants/release.ts:187-188`). Empty ⇒ the verification is skipped: `IOSCtrlProxyBuilder.getExpectedAppHash()` returns `""` for the simulator path and the device app-hash check no-ops; the runner's runtime identity falls back to the advertised `IOS_RUNNER_FEATURE_COMMANDS` feature probe only (`src/features/observe/ios/IOSCtrlProxyClient.ts:296-310`, `src/doctor/checks/ios.ts:895-927`).
- The plumbing to set it **exists but is never wired**: `.github/workflows/build-ctrl-proxy-ios-ipa.yml` computes `runner_sha256` and exposes it as a workflow output, and `scripts/.../generate-release-constants.sh:167-170` can write `IOS_CTRL_PROXY_RUNNER_SHA256` — but **no workflow passes the value into the script**, so it stays `""` every release.
- **Effect:** a tampered or stale runner binary inside an otherwise-correctly-checksummed IPA is undetectable by integrity check; stale-runner-same-version is only caught at runtime by the doctor feature-command probe (not at download/build time). (This is the "iOS runner checksum lags features" symptom — it's not lag, the hash is simply never set.)

**Fix:** wire the computed `runner_sha256` (and the app hash) into the `generate-release-constants.sh` invocation in the release / nightly / build-ipa workflows; assert it's non-empty at release.

## Part B — No version-equality gate at release

- `RELEASE_VERSION = "latest"` (`src/constants/release.ts:157`) — **not derived from `package.json`**. The literal `"latest"` resolves *through* `RELEASE_CHECKSUM_REGISTRY[0].version`. The npm version and the asset versions are kept equal **only because `bump-versions.sh` writes them together** (package.json, plugin.json, server.json, marketplace.json, gradle.properties, gradle build files) — nothing verifies they stay equal.
- `release.yml` → `verify-artifact-sha256.sh` gates that the rebuilt artifact's sha256 matches `registry[0]`, but does **not** assert `registry[0].version === git tag === package.json.version` across the manifests.
- `android/gradle.properties` `VERSION_NAME = 0.0.39-SNAPSHOT` (only normalized at Maven publish via `-PVERSION_NAME=<tag>`), so source/local Gradle builds produce a coordinate that diverges from the jar's `Implementation-Version` (which is read from `package.json`, `build.gradle.kts:20-24,128-134`). Stale comment at `build.gradle.kts:19` claims the version comes from gradle.properties.

**Fix:** add a release gate asserting version equality across all manifests + `registry[0].version` + the git tag; make `package.json.version` the canonical source (derive `RELEASE_VERSION`/asset versions from it); resolve the `-SNAPSHOT` divergence; correct the comment.

Refs #2742 (version source-of-truth). Companion to the client-handshake and CI-pinning issues.

kaeawc:
## Summary

Closes #2745. Fixes two release-pipeline integrity gaps that let versions/artifacts drift undetected.

### Part A — iOS runner-binary checksum was never wired (verification disabled)
- `IOS_CTRL_PROXY_RUNNER_SHA256` shipped as `""` every release, so the runner integrity check in `IOSCtrlProxyBuilder` (simulator path, `src/utils/IOSCtrlProxyBuilder.ts:874-889`) was permanently skipped — a tampered/stale runner inside an otherwise-valid IPA was undetectable.
- The plumbing already existed (`build-ctrl-proxy-ios-ipa.yml` exposes `runner_sha256`; `generate-release-constants.sh` accepts the env var) but **no workflow passed the value through**.
- **Fix:** wire `IOS_CTRL_PROXY_RUNNER_SHA256: ${{ needs.build-ctrl-proxy-ios-ipa.outputs.runner_sha256 }}` into the generate-release-constants step of `prepare-release.yml` (the step that writes it when adding a new registry entry), `release.yml`, and `nightly.yml`.

### Part B — no version-equality gate at release
- Nothing asserted that the npm version, every release manifest, the checksum registry, and the git tag name the same version — they stayed equal only because `bump-versions.sh` writes them together.
- **Fix:** new `scripts/ci/verify-release-integrity.sh <version>`, run in `release.yml` (after generating constants) and `prepare-release.yml`. It asserts version equality across `package.json` (canonical), `.claude-plugin/plugin.json`, `.claude-plugin/marketplace.json` (`plugins[0]`), `server.json` (top-level + `packages[]`), `android/gradle.properties` `VERSION_NAME` (`-SNAPSHOT` base), and `release.ts` `registry[0].version` vs the git tag — and that `IOS_CTRL_PROXY_RUNNER_SHA256` is a populated 64-hex sha256.
- Corrected the stale version-source comment at `android/junit-runner/build.gradle.kts:19` (the jar's `Implementation-Version` tracks `package.json`; only the Maven coordinate uses `VERSION_NAME`).

### Design decisions
- **`RELEASE_VERSION="latest"` kept, not rewritten to import `package.json`.** `release.ts` is an auto-generated "do not edit" file, and deriving asset URLs straight from `package.json` would point at not-yet-published versions. `package.json` is instead made canonical by *enforcement*: the release gate + a unit test asserting `resolveLatestVersion() === package.json.version`.
- **`-SNAPSHOT` kept** for Gradle/Maven dev coordinates (intended; normalized via `-PVERSION_NAME` at publish); the gate compares the SNAPSHOT-stripped base so source/local builds can't diverge.
- **`IOS_CTRL_PROXY_APP_HASH` intentionally not required** — it is device-only (empty = skip on simulator) and no CI step produces it. Wiring a device app-hash output is left as a follow-up; the gate does not require it, so releases are not blocked.

## Testing (TDD)
- `test/bats/verify-release-integrity.bats` (new, 12 cases): full pass; a failing case per manifest; SNAPSHOT-base accepted; `v` tag prefix accepted; empty/malformed runner sha rejected.
- `test/bats/release-workflow-wiring.bats` (new, 5 cases): each workflow passes the runner sha; release/prepare-release run the gate.
- `test/bats/generate-release-constants.bats` (extended): runner sha written in release + checksum-only modes; malformed rejected.
- `test/constants/release.test.ts` (extended): `resolveLatestVersion() === package.json.version`; runner sha is `""` or 64-hex.

Verified: `bats test/bats/{verify-release-integrity,release-workflow-wiring,generate-release-constants}.bats` (21/21), `shellcheck` clean, `bun test test/constants/release.test.ts` (16/16), `eslint` clean, `bun run build` succeeds, `actionlint` clean on the new steps. Reproduced the gate catching the real repo's current empty-runner-sha state and passing a simulated post-prepare-release state.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -57,10 +57,13 @@ jobs:
         run: |
           APK_CURRENT=$(grep 'apkSha256' src/constants/release.ts | head -1 | sed 's/.*"\([a-f0-9]\{64\}\)".*/\1/')
           IPA_CURRENT=$(grep 'ipaSha256' src/constants/release.ts | head -1 | sed 's/.*"\([a-f0-9]\{64\}\)".*/\1/')
+          RUNNER_CURRENT=$(grep '^export const IOS_CTRL_PROXY_RUNNER_SHA256' src/constants/release.ts | sed 's/.*= "\([a-f0-9]*\)".*/\1/')
           echo "apk_sha256=$APK_CURRENT" >> $GITHUB_OUTPUT
           echo "ipa_sha256=$IPA_CURRENT" >> $GITHUB_OUTPUT
+          echo "runner_sha256=$RUNNER_CURRENT" >> $GITHUB_OUTPUT
           echo "Current APK SHA256: ${APK_CURRENT:-(empty)}"
           echo "Current IPA SHA256: ${IPA_CURRENT:-(empty)}"
+          echo "Current runner SHA256: ${RUNNER_CURRENT:-(empty)}"
 
       - name: "Check if updates needed"
         id: check
@@ -69,9 +72,12 @@ jobs:
           OLD_APK="${{ steps.current.outputs.apk_sha256 }}"
           NEW_IPA="${{ needs.build-ctrl-proxy-ios-ipa.outputs.sha256 }}"
           OLD_IPA="${{ steps.current.outputs.ipa_sha256 }}"
+          NEW_RUNNER="${{ needs.build-ctrl-proxy-ios-ipa.outputs.runner_sha256 }}"
+          OLD_RUNNER="${{ steps.current.outputs.runner_sha256 }}"
 
           APK_CHANGED="false"
           IPA_CHANGED="false"
+          RUNNER_CHANGED="false"
 
           if [ "$NEW_APK" != "$OLD_APK" ]; then
             APK_CHANGED="true"
@@ -91,10 +97,24 @@ jobs:
             echo "IPA SHA256 unchanged"
           fi
 
+          # The runner binary lives inside the IPA, so a runner change usually
+          # also changes the IPA sha — but detect it independently so an empty
+          # or stale runner sha (e.g. a registry entry cut before the runner-sha
+          # wiring) is backfilled even when the IPA is byte-identical.
+          if [ "$NEW_RUNNER" != "$OLD_RUNNER" ]; then
+            RUNNER_CHANGED="true"
+            echo "Runner SHA256 changed"
+            echo "  Previous: ${OLD_RUNNER:-(empty)}"
+            echo "  New:      $NEW_RUNNER"
+          else
+            echo "Runner SHA256 unchanged"
+          fi
+
           echo "apk_changed=$APK_CHANGED" >> $GITHUB_OUTPUT
           echo "ipa_changed=$IPA_CHANGED" >> $GITHUB_OUTPUT
+          echo "runner_changed=$RUNNER_CHANGED" >> $GITHUB_OUTPUT
 
-          if [ "$APK_CHANGED" = "true" ] || [ "$IPA_CHANGED" = "true" ]; then
+          if [ "$APK_CHANGED" = "true" ] || [ "$IPA_CHANGED" = "true" ] || [ "$RUNNER_CHANGED" = "true" ]; then
             echo "needed=true" >> $GITHUB_OUTPUT
           else
             echo "needed=false" >> $GITHUB_OUTPUT
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -105,6 +105,7 @@ jobs:
         env:
           APK_SHA256_CHECKSUM: ${{ needs.build-control-proxy-apk.outputs.sha256 }}
           IOS_CTRL_PROXY_SHA256_CHECKSUM: ${{ needs.build-ctrl-proxy-ios-ipa.outputs.sha256 }}
+          IOS_CTRL_PROXY_RUNNER_SHA256: ${{ needs.build-ctrl-proxy-ios-ipa.outputs.runner_sha256 }}
         run: bash scripts/generate-release-constants.sh
 
       - name: "Build PR title"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,6 +101,7 @@ jobs:
         run: >
           bash scripts/ci/verify-release-integrity.sh
           "${{ steps.version.outputs.version }}"
+          /tmp/control-proxy.ipa
 
       - name: Run tests
         env:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -93,7 +93,14 @@ jobs:
           APK_SHA256_CHECKSUM: ${{ steps.verify_checksum.outputs.checksum }}
           IOS_CTRL_PROXY_SHA256_CHECKSUM:
             ${{ steps.verify_ipa_checksum.outputs.checksum }}
+          IOS_CTRL_PROXY_RUNNER_SHA256:
+            ${{ needs.build-ctrl-proxy-ios-ipa.outputs.runner_sha256 }}
         run: bash scripts/generate-release-constants.sh
+
+      - name: Verify release integrity
+        run: >
+          bash scripts/ci/verify-release-integrity.sh
+          "${{ steps.version.outputs.version }}"
 
       - name: Run tests
         env:
--- a/android/junit-runner/build.gradle.kts
+++ b/android/junit-runner/build.gradle.kts
@@ -16,7 +16,11 @@ java {
   withSourcesJar()
 }
 
-// Version comes from root project's gradle.properties (VERSION_NAME)
+// The jar's Implementation-Version tracks the npm package version (package.json),
+// the canonical release version. The Maven publish coordinate, by contrast, uses
+// the Gradle `version` (VERSION_NAME in gradle.properties, normalized to the tag
+// via -PVERSION_NAME=<version> at release). scripts/ci/verify-release-integrity.sh
+// asserts these stay in lockstep so source/local builds can't diverge.
 val npmPackageVersion =
     providers.fileContents(rootProject.layout.projectDirectory.file("../package.json")).asText.map { packageJson ->
       val parsed = JsonSlurper().parseText(packageJson) as Map<*, *>
--- a/android/junit-runner/build.gradle.kts
+++ b/android/junit-runner/build.gradle.kts
@@ -18,9 +18,12 @@ java {
 
 // The jar's Implementation-Version tracks the npm package version (package.json),
 // the canonical release version. The Maven publish coordinate, by contrast, uses
-// the Gradle `version` (VERSION_NAME in gradle.properties, normalized to the tag
-// via -PVERSION_NAME=<version> at release). scripts/ci/verify-release-integrity.sh
-// asserts these stay in lockstep so source/local builds can't diverge.
+// the Gradle `version` (VERSION_NAME in gradle.properties), which carries a
+// -SNAPSHOT suffix locally and is normalized to the tag via -PVERSION_NAME at
+// release. So a plain local `:junit-runner:jar` still yields Implementation-Version
+// 0.0.40 with a 0.0.40-SNAPSHOT coordinate — that suffix is intentional. At
+// release, scripts/ci/verify-release-integrity.sh asserts the SNAPSHOT-stripped
+// base version here matches package.json and the git tag.
 val npmPackageVersion =
     providers.fileContents(rootProject.layout.projectDirectory.file("../package.json")).asText.map { packageJson ->
       val parsed = JsonSlurper().parseText(packageJson) as Map<*, *>
--- a/scripts/ci/verify-release-integrity.sh
+++ b/scripts/ci/verify-release-integrity.sh
@@ -47,28 +47,45 @@ check() {
   fi
 }
 
+# Extract a JSON field, failing the whole gate loudly (clean message, no python
+# traceback) if the file is malformed or the field is missing — a swallowed
+# extraction would otherwise surface only as a confusing "expected X, got ''".
 json_field() {
   # $1 = file, $2 = python expression evaluated against `d` (parsed JSON)
-  python3 -c 'import json,sys; d=json.load(open(sys.argv[1])); print(eval(sys.argv[2]))' "$1" "$2"
+  python3 -c 'import json,sys
+try:
+    d = json.load(open(sys.argv[1]))
+    print(eval(sys.argv[2]))
+except Exception as exc:
+    sys.stderr.write(f"failed to read {sys.argv[2]} from {sys.argv[1]}: {exc}\n")
+    sys.exit(1)' "$1" "$2"
 }
 
-check "package.json version" "$(json_field "$PACKAGE_JSON" 'd["version"]')"
-check ".claude-plugin/plugin.json version" "$(json_field "$PLUGIN_JSON" 'd["version"]')"
-check ".claude-plugin/marketplace.json plugins[0].version" \
-  "$(json_field "$MARKETPLACE_JSON" 'd["plugins"][0]["version"]')"
-check "server.json version" "$(json_field "$SERVER_JSON" 'd["version"]')"
+pkg_version="$(json_field "$PACKAGE_JSON" 'd["version"]')" || exit 1
+check "package.json version" "$pkg_version"
+plugin_version="$(json_field "$PLUGIN_JSON" 'd["version"]')" || exit 1
+check ".claude-plugin/plugin.json version" "$plugin_version"
+marketplace_version="$(json_field "$MARKETPLACE_JSON" 'd["plugins"][0]["version"]')" || exit 1
+check ".claude-plugin/marketplace.json plugins[0].version" "$marketplace_version"
+server_version="$(json_field "$SERVER_JSON" 'd["version"]')" || exit 1
+check "server.json version" "$server_version"
 
-while IFS= read -r pkg_version; do
-  check "server.json packages[].version" "$pkg_version"
-done < <(python3 -c 'import json,sys
-d=json.load(open(sys.argv[1]))
-for p in d.get("packages", []):
-    print(p["version"])' "$SERVER_JSON")
+server_packages="$(python3 -c 'import json,sys
+try:
+    d = json.load(open(sys.argv[1]))
+    for p in d.get("packages", []):
+        print(p["version"])
+except Exception as exc:
+    sys.stderr.write(f"failed to read packages from {sys.argv[1]}: {exc}\n")
+    sys.exit(1)' "$SERVER_JSON")" || exit 1
+while IFS= read -r pkg_ver; do
+  [ -n "$pkg_ver" ] && check "server.json packages[].version" "$pkg_ver"
+done <<< "$server_packages"
 
 # gradle.properties keeps a -SNAPSHOT suffix for dev/Maven coordinates; the
 # release coordinate is overridden with -PVERSION_NAME at publish. Only the base
-# version participates in equality.
-gradle_version="$(grep -E '^VERSION_NAME=' "$GRADLE_PROPS" | head -1 | cut -d= -f2)"
+# version participates in equality. (tr strips a stray CR from CRLF checkouts.)
+gradle_version="$(grep -E '^VERSION_NAME=' "$GRADLE_PROPS" | head -1 | cut -d= -f2 | tr -d '\r')"
 check "android/gradle.properties VERSION_NAME (base)" "${gradle_version%-SNAPSHOT}"
 
 registry_version="$(grep -m1 -E '^[[:space:]]+version: "' "$RELEASE_TS" \
--- a/scripts/ci/verify-release-integrity.sh
+++ b/scripts/ci/verify-release-integrity.sh
@@ -1,0 +1,101 @@
+#!/usr/bin/env bash
+# Verify release integrity before publishing.
+#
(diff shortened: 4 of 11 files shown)


.github/workflows/release.yml
```diff
@@ -93,8 +93,15 @@ jobs:
           APK_SHA256_CHECKSUM: ${{ steps.verify_checksum.outputs.checksum }}
           IOS_CTRL_PROXY_SHA256_CHECKSUM:
             ${{ steps.verify_ipa_checksum.outputs.checksum }}
+          IOS_CTRL_PROXY_RUNNER_SHA256:
+            ${{ needs.build-ctrl-proxy-ios-ipa.outputs.runner_sha256 }}
         run: bash scripts/generate-release-constants.sh
 
+      - name: Verify release integrity
```
kaeawc, line 100: Blocker (reproduced): this gate hard-fails on release for any version whose registry entry already has an empty `IOS_CTRL_PROXY_RUNNER_SHA256`, and nothing in `release.yml` can fix it. Trace: the `Generate release constants` step above (env now includes the runner sha) runs `generate-release-constants.sh`, but when the tag version already exists in the registry the script early-exits at `scripts/generate-release-constants.sh:106-109` (`INFO: Version X already exists ... skipping`, `exit 0`) BEFORE reaching the runner-sha write at `scripts/generate-release-constants.sh:167-170`. So the empty `""` in source is never overwritten, then this gate exits 1 on `verify-release-integrity.sh:82-86`.

Reproduced against current `main`: `src/constants/release.ts` ships `IOS_CTRL_PROXY_RUNNER_SHA256 = ""` for the already-registered `0.0.40` entry. Running `bash scripts/ci/verify-release-integrity.sh 0.0.40` at HEAD passes every version check and then fails: `IOS_CTRL_PROXY_RUNNER_SHA256 must be a 64-char hex sha256, got ''`. Tagging `v0.0.40` after this PR merges would dead-end the release job. This is a new hard-fail path that didn't exist before.

Impact: any release of a version whose entry predates the wiring (e.g. the current `0.0.40`), or any hand-pushed tag that didn't go through a runner-sha-writing prepare-release, is blocked with no automated way forward — only a manual source edit. Fixes to consider: (a) backfill the runner sha for the current registry[0] entry in this PR so the gate passes on the next tag; and/or (b) let `generate-release-constants.sh` still update the scalar runner-sha even on the version-already-exists path instead of `exit 0`, so `release.yml`'s own freshly-built runner sha self-heals the constant before the gate runs.

.github/workflows/nightly.yml
```diff
@@ -105,6 +105,7 @@ jobs:
         env:
           APK_SHA256_CHECKSUM: ${{ needs.build-control-proxy-apk.outputs.sha256 }}
           IOS_CTRL_PROXY_SHA256_CHECKSUM: ${{ needs.build-ctrl-proxy-ios-ipa.outputs.sha256 }}
+          IOS_CTRL_PROXY_RUNNER_SHA256: ${{ needs.build-ctrl-proxy-ios-ipa.outputs.runner_sha256 }}
```
kaeawc, line 128: Gap: this correctly passes the runner sha, but the `Update release constants` step it belongs to is gated by `steps.check.outputs.needed == 'true'` (nightly.yml:104), and `needed` only flips when the APK or IPA sha256 changes (nightly.yml:65-101) — the runner sha is never part of the change detection. So if the constant is `""` (current `main`) but the IPA sha is unchanged, nightly never runs this step and never populates the runner sha. Combined with the release.yml early-exit above, there is no automated path that fixes an empty runner sha for a stable IPA — the only self-healing trigger is an IPA-content change, which is exactly the case where it was already fine. Suggest folding a `runner_sha empty-or-changed` condition into the `check` step so nightly can backfill it independently of the IPA sha.

kaeawc (commented):
Release-engineer review: the runner-sha wiring is real end-to-end (create-ipa.sh emits `runner_sha256`, workflows pass it, gate reproduces). But the gate can hard-fail a release with no in-pipeline remediation for any already-registered version, and nightly can't self-heal an empty runner sha unless the IPA also changes. Two blockers below; reproduced against the real repo.

scripts/ci/verify-release-integrity.sh
```diff
@@ -0,0 +1,101 @@
...
+
+registry_version="$(grep -m1 -E '^[[:space:]]+version: "' "$RELEASE_TS" \
+  | sed 's/.*version: "\([^"]*\)".*/\1/')"
+check "src/constants/release.ts registry[0].version" "$registry_version"
+
+# Runner-binary checksum must be populated. Empty ("" default) means integrity
+# verification is silently skipped — exactly the gap this gate exists to catch.
+runner_sha="$(grep -E '^export const IOS_CTRL_PROXY_RUNNER_SHA256' "$RELEASE_TS" \
+  | sed 's/.*= "\([^"]*\)".*/\1/')"
+if [[ "$runner_sha" =~ ^[a-f0-9]{64}$ ]]; then
```
kaeawc: This proves *something* 64-hex was written, but nothing binds it to the runner inside the shipped IPA. The real integrity rests on the IPA hash (IOSCtrlProxyBuilder.ts:754 verifies the whole IPA; the runner is inside it — ctrl-proxy-create-ipa.sh:168,176). Two latent gaps this syntactic check does not close: (a) IOS_CTRL_PROXY_RUNNER_SHA256 is a single global scalar (src/constants/release.ts:193), overwritten in place each release by generate-release-constants.sh:167-169, whereas ipaSha256 is versioned per registry entry — so the runner sha is only ever consistent with registry[0]; if a pinned-older-version download is ever wired (the resolveChecksum(<pinned>) path already exists in release.ts:131-147), verifyPlatformArtifacts (IOSCtrlProxyBuilder.ts:882) would compare the *newest* runner sha against an *older* IPA's runner and reject a valid bundle. (b) In release.yml/prepare-release.yml, runner_sha256 comes from needs.build-ctrl-proxy-ios-ipa.outputs.runner_sha256 (the build job) while ipaSha256 is recomputed from the re-downloaded/rebuilt artifact — same build today, so consistent, but the gate enforces none of that coupling. Consider re-deriving the runner sha from the sha-pinned IPA at gate time (unzip + hash CtrlProxyUITests-Runner) so the check actually binds, or store it per-registry-entry alongside ipaSha256.

kaeawc (commented):
Supply-chain security review

Net-positive: the runner sha is now actually populated, so IOSCtrlProxyBuilder.verifyPlatformArtifacts (simulator path, src/utils/IOSCtrlProxyBuilder.ts:874-889) stops no-op'ing, and the version-equality gate is real and well-tested. Two things to flag, one latent hole and one framing correction, plus one confirmed defensible scope cut.

1. Latent hole: the runner sha is a single global scalar (release.ts:193) while ipaSha256 is per-registry-entry (release.ts:27-123). generate-release-constants.sh:167-169 overwrites it in place each release. The gate only checks it is 64-hex — no binding to the shipped IPA's runner. Today the values stay consistent only because IOS_CTRL_PROXY_RELEASE_VERSION is hardwired to "latest"→registry[0] and both derive from one ctrl-proxy-create-ipa.sh run. If a pinned-older-version download path is ever wired (the resolveChecksum(<pinned>) machinery already exists), the newest runner sha would mismatch the older IPA's runner and reject a valid older bundle. See inline comment.

2. Framing: the actual integrity boundary for the simulator runner is the IPA hash (verifyBundle, IOSCtrlProxyBuilder.ts:754) — the runner lives inside that already-hash-verified, simulator-only IPA (scripts/ios/ctrl-proxy-create-ipa.sh:168,176). The runner-sha check is redundant defense-in-depth, not the primary guarantee. Good to keep, but the gate proving 'a 64-hex value was written' should not be read as proving the runner matches the shipped IPA.

3. Confirmed defensible: IOS_CTRL_PROXY_APP_HASH staying "" is fine — the released artifact is simulator-only (Debug-iphonesimulator products, ctrl-proxy-create-ipa.sh:116,131), so there is no device app bundle for CI to hash. The PR body documents this. No action needed.

kaeawc (commented):
**Build-systems review** — I ran the gate against adversarial fixtures (regex/glob metachars in the version, zero-package `server.json`, second/diverging `packages[]`, CRLF, empty-array, `set -e` interaction, python KeyError), ran `shellcheck` (clean) and all three BATS files (21/21), and traced the runner-sha wiring end-to-end. The shell is mostly robust — notably the `while … done < <(python3 …)` uses process substitution so the loop body runs in the current shell and `errors+=` from a diverging `packages[1]` IS caught (verified: exit 1). The registry grep correctly targets `registry[0]` and skips the `version: string;` interface field and `LATEST_RELEASE_VERSION`. Metachar versions are safe (string `[ != ]`, not regex). Part A is genuinely fixed and the gate is not a permanent false-positive: `ctrl-proxy-create-ipa.sh` emits `runner_sha256`, which flows through `needs.build-ctrl-proxy-ios-ipa.outputs.runner_sha256` → env → `generate-release-constants.sh` (scalar write runs in both modes) → gate. Two substantive items below.

**Design gap vs issue #2745 (non-blocking):** the issue asked to *derive* RELEASE_VERSION/asset versions from package.json and *resolve* the -SNAPSHOT divergence. The PR instead keeps `RELEASE_VERSION="latest"` + `VERSION_NAME=…-SNAPSHOT` and enforces base-equality via gate+test. The "don't derive because it'd point at unpublished versions" argument is legitimate, but the -SNAPSHOT divergence still exists at local/source build time — the gate tolerates it rather than resolving it. Reasonable call, but it's *tolerate*, not *resolve*; see the build.gradle.kts comment below.

android/junit-runner/build.gradle.kts
```diff
@@ -16,7 +16,11 @@ java {
   withSourcesJar()
 }
 
-// Version comes from root project's gradle.properties (VERSION_NAME)
+// The jar's Implementation-Version tracks the npm package version (package.json),
+// the canonical release version. The Maven publish coordinate, by contrast, uses
+// the Gradle `version` (VERSION_NAME in gradle.properties, normalized to the tag
+// via -PVERSION_NAME=<version> at release). scripts/ci/verify-release-integrity.sh
+// asserts these stay in lockstep so source/local builds can't diverge.
```
kaeawc: The comment overstates what the gate does. It claims the gate "asserts these stay in lockstep so source/local builds can't diverge" — but source/local builds still diverge, exactly as issue #2745 describes.

Repro: `VERSION_NAME=0.0.40-SNAPSHOT` (gradle.properties:143) and `Implementation-Version` = `npmPackageVersion` = package.json = `0.0.40`. A local `./gradlew :junit-runner:jar` produces jar `Implementation-Version=0.0.40` alongside a Maven coordinate `version=0.0.40-SNAPSHOT` in the *same* build — those diverge, and nothing in this PR changes that. The gate (a) runs only in CI at release time, and (b) strips `-SNAPSHOT` and compares only the base (verify-release-integrity.sh:71-72), so it *tolerates* the suffix rather than keeping coordinate and manifest "in lockstep." Confirmed a valid release still passes with `VERSION_NAME=0.0.40-SNAPSHOT` (BATS case "accepts VERSION_NAME with a -SNAPSHOT suffix").

Verdict: comment is inaccurate — "source/local builds can't diverge" is false; they still diverge, the gate just accepts the SNAPSHOT-suffixed base. Reword to say the gate asserts base-version equality in CI and that the SNAPSHOT suffix is an accepted dev/Maven convention (or actually resolve the divergence per the issue). Not a code bug, but this is load-bearing documentation and currently misleads.

scripts/ci/verify-release-integrity.sh
```diff
@@ -0,0 +1,101 @@
...
+  if [ "$actual" != "$EXPECTED" ]; then
+    errors+=("${label}: expected '${EXPECTED}', got '${actual}'")
+  else
+    echo "  OK  ${label} = ${actual}"
+  fi
+}
+
+json_field() {
+  # $1 = file, $2 = python expression evaluated against `d` (parsed JSON)
+  python3 -c 'import json,sys; d=json.load(open(sys.argv[1])); print(eval(sys.argv[2]))' "$1" "$2"
```
kaeawc: `json_field`'s python failure is silently swallowed. If a JSON file is malformed or missing the key, `python3 … eval(...)` raises and exits non-zero, but because it's used as `check "…" "$(json_field …)"` the failing command substitution does NOT trip `set -e` (command substitution in a simple command's argument list is exempt), so `check` just receives `""`.

Repro I ran (package.json with no `version` key):
```
Traceback (most recent call last): … KeyError: 'version'
  OK  .claude-plugin/plugin.json version = 0.0.40
  …
  - package.json version: expected '0.0.40', got ''
EXIT: 1
```
It still fails here only because `"" != "0.0.40"`. Verdict: not a live false-negative today (empty always mismatches a real semver), but the gate relies on the equality check to mask a swallowed parser crash rather than failing explicitly, and it dumps a python traceback into release logs. Low severity; consider capturing into a local first (`actual=$(json_field …) || exit 1`) so a genuine parse failure aborts loudly instead of degrading to an empty-string compare.

kaeawc:
## Review feedback addressed (commit f3efcb65)

Thanks for the three reviews. All actionable findings are fixed; here's the mapping.

**Release-engineering — gate could dead-end a release (blocker).** Confirmed. `generate-release-constants.sh` early-exited on a version already in the registry *before* the scalar runner-sha write, so at tag time the runner sha could never be refreshed and the new gate would hard-fail. Fixed: the version-exists path now skips only the registry *prepend* and still refreshes the scalar constants. The release job self-heals its own runner sha from the same build that produces the sha-verified IPA. Reproduced: with `0.0.40` already registered and an empty runner sha, the release-mode run now populates it (no duplicate entry) and the gate passes. New BATS case `refreshes runner sha for an already-registered version` guards it.

**Release-engineering — nightly never healed an empty runner sha.** Fixed: nightly now detects runner-sha drift independently of the IPA sha and folds it into the `needed` gate, so an empty/stale runner sha is backfilled via the nightly PR even when the IPA is byte-identical.

**Build-systems — `json_field` swallowed python failures.** Fixed: extraction now fails loudly with a clean message (`failed to read <field> from <file>: <err>`) and exits non-zero, no traceback in release logs. Also strips a stray CR for CRLF checkouts.

**Build-systems — build.gradle.kts comment overclaimed.** Correct, and fixed. The comment no longer claims the gate prevents local divergence; it now states the `-SNAPSHOT` suffix is intentional and that the gate checks base-version equality *in CI*. To be explicit on the design note: this PR **tolerates** `-SNAPSHOT` and enforces base-version equality at release — it does not eliminate the local coordinate vs `Implementation-Version` difference, which is inherent to SNAPSHOT semantics.

**Security — runner sha is a global scalar, not per-entry.** Agreed it's latent (not exploitable today: the iOS download always resolves to `latest`/registry[0], and the runner lives inside that IPA, so its integrity is transitively covered by `registry[0].ipaSha256`). Rather than a premature schema refactor, I documented the constraint at the declaration in `release.ts` with a pointer to move it to a per-entry `runnerSha256` if pinned-older iOS downloads are ever wired. Framing corrected: the gate's non-empty check re-enables the *simulator* runner verification (was a permanent no-op) — it is defense-in-depth, not a binding of runner↔shipped-IPA; the IPA hash provides that binding.

**Scope note (unchanged):** `IOS_CTRL_PROXY_APP_HASH` stays unwired — CI produces only a `Debug-iphonesimulator` product, so there is no device app bundle to hash. The gate does not require it, so releases aren't blocked. Left as a follow-up for when a device build product exists.

scripts/ci/verify-release-integrity.sh
kaeawc: Addressed in 5dcb0d64 — the gate now actually binds. `verify-release-integrity.sh` takes an optional IPA path; release.yml and prepare-release.yml pass the downloaded `/tmp/control-proxy.ipa`, and the gate unzips `CtrlProxyUITests-Runner` from it, hashes it, and requires it to equal the recorded `IOS_CTRL_PROXY_RUNNER_SHA256`. A stale/substituted runner inside an otherwise-valid bundle now fails at release, closing latent gap (b). Gap (a) (scalar-vs-per-entry) stays documented at the declaration for the pinned-download runtime path; the release-time binding is against registry[0]/latest, which is what ships.

.github/workflows/release.yml
kaeawc: Fixed in f3efcb65 (self-heal) + 5dcb0d64 (binding). `generate-release-constants.sh` no longer `exit 0`s on a version already in the registry — it skips only the registry *prepend* and still refreshes the scalar constants, so this release step repopulates `IOS_CTRL_PROXY_RUNNER_SHA256` from its own build before the gate runs. Reproduced: with 0.0.40 already registered and an empty runner sha, the release-mode run now populates it (no duplicate entry) and the gate passes. Guarded by BATS `refreshes runner sha for an already-registered version`. The gate also now binds the recorded sha to the runner inside `/tmp/control-proxy.ipa`.

.github/workflows/nightly.yml
kaeawc: Fixed in f3efcb65. Nightly now reads the current runner sha from source and detects runner-sha drift independently of the APK/IPA sha, folding it into the `needed` gate (`runner_changed` → `needed=true`). So an empty or stale runner sha is backfilled via the nightly PR even when the IPA is byte-identical. The runner-lives-inside-the-IPA point still holds, but the empty-backfill case is now covered explicitly.

android/junit-runner/build.gradle.kts
kaeawc: Fixed in f3efcb65. The comment no longer claims local builds "can't diverge." It now states the `-SNAPSHOT` suffix is intentional (local coordinate stays `X-SNAPSHOT` while Implementation-Version is the package.json `X`), and that the CI release gate only asserts the SNAPSHOT-stripped *base* matches package.json + tag. Explicitly: the PR *tolerates* the suffix and enforces base equality at release — it does not eliminate the local coordinate/Implementation-Version difference, which is inherent to SNAPSHOT semantics.

scripts/ci/verify-release-integrity.sh
kaeawc: Fixed in f3efcb65. `json_field` now catches the exception, writes a clean `failed to read <field> from <file>: <err>` to stderr, and `sys.exit(1)` — and each caller assigns to a local with `… || exit 1`, so a malformed/missing field fails the gate loudly instead of surfacing as a confusing `expected X, got ''` (and no python traceback in release logs). Also added a `tr -d '\\r'` guard on the gradle line for CRLF checkouts.

PR #2748, opened 2026-07-01
feat(release): hermetic single-version pinning knobs for external CI (#2746)

### LINKED ISSUE #2746: Provide hermetic single-version CI pinning for all clients (daemon + ctrlproxy APK/IPA + junit-runner + XCTestRunner)
## Summary

There is no single knob or documented hermetic recipe for an **external CI consumer** to pin every AutoMobile component — daemon/stdio, Android ctrlproxy APK, iOS CtrlProxy IPA, Android junit-runner, iOS XCTestRunner — to one coherent version. The repo's own CI is hermetic *by construction* (every component comes from one checkout), which masks the gaps that external consumers hit: independent inputs, test-time network fetches, and unverifiable assets.

## How it works today (confirmed)

- **Repo CI (hermetic):** the junit-runner test task sets `automobile.daemon.local.project.path` = repo root (spawns in-tree `dist/`, no npm fetch), `AUTOMOBILE_CTRL_PROXY_APK_PATH` = freshly assembled `:control-proxy` APK, `automobile.daemon.force.restart=true` (`android/junit-runner/build.gradle.kts:105-126`). iOS CI build-for-testing's CtrlProxy then `bun add -g .` for the daemon (`.github/workflows/pull_request.yml:1167-1256`). Alignment is "everything from one commit," not enforced.

## Gaps for external consumers

1. **Daemon fetched over the network at test time.** Kotlin spawns `bunx/npx @kaeawc/auto-mobile@<version>` (`DaemonSocketClient.kt:287-326`) → npm-registry dependency unless a local path / pre-warmed cache is used. iOS XCTestRunner autostart runs a bare PATH `auto-mobile --daemon start` (`AutoMobileEnvironment.swift:132-168`) — no version pin at all.
2. **ctrlproxy APK + iOS IPA have no consumer-facing version knob.** They're runtime-downloaded by the daemon from `src/constants/release.ts` `registry[0]` (GitHub Releases). Pinning the immutable npm daemon version transitively pins them (baked URL + sha256), so it's deterministic *via the daemon* but **not hermetic** (depends on GitHub reachability), and there's no `AUTOMOBILE_ASSET_BASE_URL`-style mirror knob.
3. **Checksum escape hatches.** `AUTOMOBILE_CTRL_PROXY_APK_PATH` override skips checksum entirely (`CtrlProxyManager.ts:1565-1573`); otherwise a SHA mismatch is accepted non-blockingly (`acceptedPreinstalled`, `CtrlProxyManager.ts:646-657`).
4. **iOS XCTestRunner has no version constant** (no field in `Package.swift`, SPM pinned by git tag only) — it can neither pin nor self-identify a daemon version.
5. **Android jar `Implementation-Version` (from package.json) vs Maven coordinate (`-SNAPSHOT`)** are two different strings for the same artifact.
6. **`@latest` advice strings** in doctor/CLI/manager (`src/doctor/checks/automobile.ts:109,116,150,428`; `src/cli/index.ts`; `src/daemon/manager.ts:1145`) — advice only, but a human who starts a `@latest` daemon that a pinned runner then **reuses** (see the shared-socket issue) gets silent drift.

## Recommendations

- **Single `AUTOMOBILE_VERSION` knob** that selects daemon + Android APK + iOS IPA together; document per-platform hermetic CI recipes:
  - *Android:* pin the Maven coordinate + `automobile.daemon.package.version` + a pre-provisioned `AUTOMOBILE_CTRL_PROXY_APK_PATH` + `automobile.daemon.force.restart=true`.
  - *iOS:* pin SPM + daemon version + vendored `AUTOMOBILE_CTRL_PROXY_IOS_IPA_PATH` + `AUTOMOBILE_SKIP_CTRL_PROXY_IOS_BUILD=1`.
- **`AUTOMOBILE_ASSET_BASE_URL`** to mirror APK/IPA for hermetic, offline-capable provisioning; support an offline daemon (local dist path / cached bunx) so no npm fetch at test time.
- **Bake a version into the XCTestRunner SPM package** (mirrors Android's jar version; also unblocks the client-handshake issue).
- **Replace user-facing `@latest`** strings with the resolved concrete version (`resolveLatestVersion()` already exists).
- **Gate CI jobs on `auto-mobile --doctor` (Android) / `--doctor --ios` (iOS)** so the existing stale-runner / observe-round-trip / build-identity checks become hard failures.

Companion to the daemon-socket-handshake and release-integrity issues. Refs #2731, #2732, #2742.

kaeawc:
## Summary
External CI consumers had no single knob or hermetic recipe to pin every AutoMobile component — daemon, Android ctrlproxy APK, iOS ctrlproxy IPA, and their checksums — to one coherent version. The daemon is the source of truth the Android/iOS clients delegate to (it bakes the asset URLs + checksums and reports them via `ide/status`), yet on the TypeScript side those were module-load `const`s derived from the floating `"latest"` literal, with GitHub as the only host and no version knob. `AUTOMOBILE_VERSION` was read by the Kotlin runner but **ignored by TypeScript**, and ~20 user-facing `@latest` advice strings invited silent drift.

This adds the daemon-side pinning layer for issue #2746:
- **`AUTOMOBILE_VERSION`** — one knob selecting daemon + APK + IPA + both checksums.
- **`AUTOMOBILE_ASSET_BASE_URL`** — mirror the APK/IPA host for offline/hermetic CI.
- Concrete-version advice everywhere, never `@latest`.

Native follow-ups (bake a version into the XCTestRunner SPM package, gradle recipe wiring, Swift `startDaemon` pin) are intentionally out of scope — they can't be unit-tested in the TS harness and belong in a focused change.

## Changes
- **`src/constants/release.ts`** — new pure, env-injectable resolvers: `resolvePinnedVersion`, `resolveAssetBaseUrl`, `resolveApkUrl` / `resolveIpaUrl`, `resolveApkChecksum` / `resolveIpaChecksum`, and `resolveDaemonInstallSpecifier` (concrete `@kaeawc/auto-mobile@<version>` for advice). `buildReleaseAssetUrl` gains a base-URL parameter. Module-level `APK_URL` / `IPA_URL` / checksum exports are preserved (now derived via the resolvers) for backward compatibility.
- **Consumers resolve at call time** (so the daemon honors env set at launch): `CtrlProxyManager.ts` (APK download URL + expected checksum), `IOSCtrlProxyBuilder.ts` (`getBundleUrl`/`getExpectedChecksum` fallbacks, hermetic cached-fallback gate, metadata version), `daemon/socketServer.ts` `ide/status` (concrete `releaseVersion` + mirror/version-aware URLs + checksums), `server/bootedDeviceResources.ts` (compat checksums).
- **Concrete-version advice** replacing `@latest` in `doctor/checks/automobile.ts`, `cli/index.ts` (error paths + help/usage text), `daemon/manager.ts`, `daemon/debugTools.ts`. `doctor/index.ts` `report.version` and `doctor/checks/ios.ts` runner diagnostic line now report the pinned version.
- **Docs** — `docs/using/hermetic-ci-pinning.md` (per-platform Android + iOS recipes, mirror layout, `ide/status` verification) wired into `mkdocs.yml`.

## Acceptance check
- `AUTOMOBILE_VERSION=0.0.18` → APK url, IPA url, APK sha, IPA sha all resolve to `0.0.18` coherently; `ide/status` reports `releaseVersion: "0.0.18"`.
- `AUTOMOBILE_ASSET_BASE_URL=https://mirror.test/am` → `https://mirror.test/am/<version>/control-proxy-debug.apk`; both knobs compose.
- Unset env → unchanged default behavior (concrete latest registry version, GitHub host, existing checksums).
- No `@kaeawc/auto-mobile@latest` string remains in `src/`.

## Test plan
- **`test/constants/release.test.ts`** — pure resolver tests: pinning, mirror base (trailing-slash strip), composition, unknown-version → empty checksum, concrete specifier never `@latest`.
- **`test/daemon/socketServerIdeStatus.test.ts`** — `ide/status` reports concrete `releaseVersion` and honors both knobs (mirror + version).
- **`test/doctor/automobile.test.ts`**, **`test/doctor/ios.test.ts`**, **`test/doctor/index.test.ts`** — advice/report strings are concrete and honor `AUTOMOBILE_VERSION`.
- `bun test` → 5338 pass / 0 fail (+2 new). `bun run lint` + `bun run build` green; `tsc --noEmit` clean on touched files.

Refs #2731, #2732, #2742
Closes #2746

### DIFF
--- a/docs/using/hermetic-ci-pinning.md
+++ b/docs/using/hermetic-ci-pinning.md
@@ -102,13 +102,17 @@ mirror can only serve unverified assets if you deliberately opt out of verificat
    itself. Pin it by installing the exact `@kaeawc/auto-mobile@0.0.40` on the runner's
    PATH, or start the daemon yourself before the test job. (Baking the pin into the
    Swift runner is tracked as follow-up work for #2746.)
-2. **Vendor the IPA and skip the build** so nothing is fetched or compiled at test time:
+2. **Vendor the IPA** so nothing is fetched or compiled at test time:
    ```bash
    export AUTOMOBILE_CTRL_PROXY_IOS_IPA_PATH=/opt/automobile/control-proxy.ipa
-   export AUTOMOBILE_SKIP_CTRL_PROXY_IOS_BUILD=1
    ```
-   (or set `AUTOMOBILE_ASSET_BASE_URL` to your mirror for a checksummed download of a
-   registry-known version).
+   The daemon copies (and extracts) this local bundle directly — no network fetch, no
+   Xcode build. **Do not also set `AUTOMOBILE_SKIP_CTRL_PROXY_IOS_BUILD=1`**: that flag
+   short-circuits `needsRebuild()` and returns *before* the vendored IPA is consumed, so
+   on a fresh host CtrlProxy would never be installed. Use `AUTOMOBILE_SKIP_CTRL_PROXY_IOS_BUILD=1`
+   only when the extracted xctestrun artifacts are *already* present on the host.
+   Alternatively, set `AUTOMOBILE_ASSET_BASE_URL` to your mirror for a checksummed
+   download of a registry-known version (leave the IPA path unset).
 3. **Gate the job** on doctor:
    ```bash
    bunx @kaeawc/auto-mobile@0.0.40 --cli doctor --ios
--- a/docs/using/hermetic-ci-pinning.md
+++ b/docs/using/hermetic-ci-pinning.md
@@ -1,0 +1,126 @@
+# Hermetic CI Pinning
+
+AutoMobile is made of several components that must agree on one version:
+
+| Component | What it is |
+|-----------|------------|
+| **Daemon** | The `@kaeawc/auto-mobile` npm package (stdio/MCP + Unix-socket server) |
+| **Android CtrlProxy APK** | On-device accessibility service, downloaded by the daemon |
+| **iOS CtrlProxy IPA** | On-device XCUITest runner bundle, downloaded by the daemon |
+| **Android junit-runner** | Kotlin runner that spawns the daemon during tests |
+| **iOS XCTestRunner** | Swift runner that autostarts the daemon during tests |
+
+The repo's own CI is hermetic *by construction* — every component comes from a single
+checkout — so a version knob is not needed there. **External CI consumers** don't have
+that luxury: their inputs are independent, and without pinning they can silently drift
+(a `@latest` daemon pulling one APK while a runner expects another). This page documents
+how to pin everything to **one coherent version** and, optionally, mirror the release
+assets for offline builds.
+
+## The knobs
+
+### `AUTOMOBILE_VERSION` — the single-version knob
+
+Set `AUTOMOBILE_VERSION` to a released version (e.g. `0.0.40`) to pin the daemon package,
+the Android APK, the iOS IPA, and their expected SHA-256 checksums together. The daemon
+resolves all four from one value, so there is no way for them to disagree.
+
+```bash
+export AUTOMOBILE_VERSION=0.0.40
+```
+
+- Unset (or `latest`) resolves to the newest entry in the daemon's baked release registry
+  — a **concrete** version, never the floating `@latest` tag.
+- An unknown version resolves to an empty checksum (download proceeds unverified only if
+  you have also opted into a checksum-skip override — see below).
+- The Android Kotlin runner already reads `AUTOMOBILE_VERSION` as a fallback for the
+  daemon package version, so the same env var lines up the runner-spawned daemon too.
+
+### `AUTOMOBILE_ASSET_BASE_URL` — the mirror knob
+
+By default the daemon downloads the APK/IPA from GitHub Releases. For hermetic,
+offline-capable CI, mirror the assets and point the daemon at your mirror:
+
+```bash
+export AUTOMOBILE_ASSET_BASE_URL=https://artifacts.internal/automobile
+```
+
+The daemon fetches `${AUTOMOBILE_ASSET_BASE_URL}/${version}/control-proxy-debug.apk`
+(and `.../control-proxy.ipa`). Lay your mirror out with a directory per version:
+
+```
+https://artifacts.internal/automobile/
+  0.0.40/
+    control-proxy-debug.apk
+    control-proxy.ipa
+```
+
+Checksums are still verified against the daemon's baked registry, so a mirror cannot
+serve a tampered asset for a pinned version.
+
+## Android hermetic recipe
+
+1. **Pin the runner + daemon:**
+   ```bash
+   export AUTOMOBILE_VERSION=0.0.40
+   ```
+   In Gradle, pin the Maven coordinate of `junit-runner` to the matching release and set
+   `-Dautomobile.daemon.package.version=0.0.40` (or rely on `AUTOMOBILE_VERSION`).
+2. **Provision the APK out-of-band** so no network fetch happens at test time:
+   ```bash
+   export AUTOMOBILE_CTRL_PROXY_APK_PATH=/opt/automobile/control-proxy-debug.apk
+   ```
+   (or set `AUTOMOBILE_ASSET_BASE_URL` to your mirror for a checksummed download).
+3. **Force a clean daemon** so a stale reused daemon can't serve a different build:
+   ```bash
+   export AUTOMOBILE_DAEMON_PACKAGE_VERSION=0.0.40   # explicit runner-side pin
+   # -Dautomobile.daemon.force.restart=true
+   ```
+4. **Gate the job** on doctor so drift is a hard failure:
+   ```bash
+   bunx @kaeawc/auto-mobile@0.0.40 --cli doctor
+   ```
+
+## iOS hermetic recipe
+
+1. **Pin the runner + daemon:**
+   ```bash
+   export AUTOMOBILE_VERSION=0.0.40
+   ```
+   Pin the XCTestRunner SPM dependency to the matching git tag.
+2. **Vendor the IPA and skip the build** so nothing is fetched or compiled at test time:
+   ```bash
+   export AUTOMOBILE_CTRL_PROXY_IOS_IPA_PATH=/opt/automobile/control-proxy.ipa
+   export AUTOMOBILE_SKIP_CTRL_PROXY_IOS_BUILD=1
+   ```
+   (or set `AUTOMOBILE_ASSET_BASE_URL` to your mirror for a checksummed download).
+3. **Gate the job** on doctor:
+   ```bash
+   bunx @kaeawc/auto-mobile@0.0.40 --cli doctor --ios
+   ```
+
+## Verifying the pin
+
+Ask the running daemon what it will actually fetch — the `ide/status` handler reports the
+**concrete** resolved version plus the exact (mirror-aware) URLs and checksums:
+
+```jsonc
+{
+  "releaseVersion": "0.0.40",
+  "android": {
+    "ctrlProxy": {
+      "url": "https://artifacts.internal/automobile/0.0.40/control-proxy-debug.apk",
+      "expectedSha256": "…"
+    }
+  },
+  "ios": {
+    "xcTestService": {
+      "url": "https://artifacts.internal/automobile/0.0.40/control-proxy.ipa",
+      "expectedSha256": "…"
+    }
+  }
+}
+```
+
+If `releaseVersion` is anything other than the version you pinned, the environment is not
+hermetic — check that `AUTOMOBILE_VERSION` is exported into the daemon's process.
--- a/docs/using/hermetic-ci-pinning.md
+++ b/docs/using/hermetic-ci-pinning.md
@@ -29,12 +29,18 @@ resolves all four from one value, so there is no way for them to disagree.
 export AUTOMOBILE_VERSION=0.0.40
 ```
 
-- Unset (or `latest`) resolves to the newest entry in the daemon's baked release registry
-  — a **concrete** version, never the floating `@latest` tag.
-- An unknown version resolves to an empty checksum (download proceeds unverified only if
-  you have also opted into a checksum-skip override — see below).
-- The Android Kotlin runner already reads `AUTOMOBILE_VERSION` as a fallback for the
-  daemon package version, so the same env var lines up the runner-spawned daemon too.
+- Unset (or `latest`, case-insensitive) resolves to the newest entry in the daemon's
+  baked release registry — a **concrete** version, never the floating `@latest` tag.
+- A version **in** the daemon's baked registry is downloaded and its SHA-256 is verified.
+- A version **not** in the registry (a typo, or a real release newer than the daemon's
+  baked registry) has no checksum to verify against, so the download **fails closed** —
+  the daemon refuses to install an unverifiable asset. To use a not-yet-baked release,
+  either upgrade the daemon to one whose registry includes it, or take the explicit
+  escape hatch: `AUTOMOBILE_SKIP_ACCESSIBILITY_CHECKSUM=1` (Android) / vendor a trusted
+  bundle via `AUTOMOBILE_CTRL_PROXY_IOS_IPA_PATH` (iOS).
+- The Android Kotlin runner reads `AUTOMOBILE_VERSION` as a fallback for the daemon
+  package version, so a daemon **this runner spawns** lines up. A **pre-existing** shared
+  daemon does not re-read it — see [Shared daemon caveat](#shared-daemon-caveat).
 
 ### `AUTOMOBILE_ASSET_BASE_URL` — the mirror knob
 
@@ -55,8 +61,11 @@ https://artifacts.internal/automobile/
     control-proxy.ipa
 ```
 
-Checksums are still verified against the daemon's baked registry, so a mirror cannot
-serve a tampered asset for a pinned version.
+For any version in the daemon's baked registry, the downloaded asset's SHA-256 is
+verified against that registry, so a mirror **cannot** serve a tampered asset for a
+registry-known version. For a version **not** in the registry there is nothing to verify
+against, so the download fails closed (see the `AUTOMOBILE_VERSION` notes above) — a
+mirror can only serve unverified assets if you deliberately opt out of verification.
 
 ## Android hermetic recipe
 
@@ -87,17 +96,38 @@ serve a tampered asset for a pinned version.
    ```bash
    export AUTOMOBILE_VERSION=0.0.40
    ```
-   Pin the XCTestRunner SPM dependency to the matching git tag.
+   Pin the XCTestRunner SPM dependency to the matching git tag. Note: the iOS
+   XCTestRunner autostart currently spawns a bare-PATH `auto-mobile --daemon start`
+   (`AutoMobileEnvironment.swift`) — it does **not** yet pin the daemon package version
+   itself. Pin it by installing the exact `@kaeawc/auto-mobile@0.0.40` on the runner's
+   PATH, or start the daemon yourself before the test job. (Baking the pin into the
+   Swift runner is tracked as follow-up work for #2746.)
 2. **Vendor the IPA and skip the build** so nothing is fetched or compiled at test time:
    ```bash
    export AUTOMOBILE_CTRL_PROXY_IOS_IPA_PATH=/opt/automobile/control-proxy.ipa
    export AUTOMOBILE_SKIP_CTRL_PROXY_IOS_BUILD=1
    ```
-   (or set `AUTOMOBILE_ASSET_BASE_URL` to your mirror for a checksummed download).
+   (or set `AUTOMOBILE_ASSET_BASE_URL` to your mirror for a checksummed download of a
+   registry-known version).
 3. **Gate the job** on doctor:
    ```bash
    bunx @kaeawc/auto-mobile@0.0.40 --cli doctor --ios
    ```
+
+## Shared daemon caveat
+
+The pin is a property of **the daemon's launch environment**, resolved where the download
+happens. AutoMobile runs **one daemon per UID**, shared across worktrees via a single
+socket + PID file, and runners **reuse an already-running daemon** rather than restart it.
+So if a daemon is already up with a different `AUTOMOBILE_VERSION` (or none), a second job
+that exports a new pin will silently be served by the **existing** daemon — the pin is
+ignored until the daemon is restarted.
+
+For hermetic CI, force a clean daemon so the pin actually takes effect:
+
+- **Android:** `-Dautomobile.daemon.force.restart=true` (already in the recipe above).
+- **Any platform:** `bunx @kaeawc/auto-mobile@<version> --daemon restart` before the job,
+  then confirm via `ide/status` (below) that `releaseVersion` matches your pin.
 
 ## Verifying the pin
 
@@ -123,4 +153,6 @@ Ask the running daemon what it will actually fetch — the `ide/status` handler 
 ```
 
 If `releaseVersion` is anything other than the version you pinned, the environment is not
-hermetic — check that `AUTOMOBILE_VERSION` is exported into the daemon's process.
+hermetic — either `AUTOMOBILE_VERSION` was not exported into the daemon's process, or a
+pre-existing daemon with a different pin is still serving (see
+[Shared daemon caveat](#shared-daemon-caveat)). Restart the daemon and re-check.
--- a/src/constants/release.ts
+++ b/src/constants/release.ts
@@ -150,11 +150,13 @@ export function resolveChecksum(
  * Version of the latest validated release in the registry.
  * Used to construct download URLs for pinned releases.
  */
-export function resolveLatestVersion(): string {
-  if (RELEASE_CHECKSUM_REGISTRY.length === 0) {
+export function resolveLatestVersion(
+  registry: ReleaseChecksumEntry[] = RELEASE_CHECKSUM_REGISTRY
+): string {
+  if (registry.length === 0) {
     return LATEST_RELEASE_VERSION;
   }
-  return RELEASE_CHECKSUM_REGISTRY[0].version;
+  return registry[0].version;
 }
 
 // --- Backward-compatible exports derived from RELEASE_VERSION ---
@@ -166,9 +168,12 @@ export const RELEASE_VERSION: string = LATEST_RELEASE_VERSION;
  * (URLs, on-disk metadata, doctor checks) want a concrete version like
  * "0.0.30", never the placeholder "latest".
  */
-export function resolveAssetVersion(version: string): string {
+export function resolveAssetVersion(
+  version: string,
+  registry: ReleaseChecksumEntry[] = RELEASE_CHECKSUM_REGISTRY
+): string {
   if (version === LATEST_RELEASE_VERSION) {
-    return resolveLatestVersion();
+    return resolveLatestVersion(registry);
   }
   return version;
 }
@@ -197,11 +202,44 @@ type EnvLike = Record<string, string | undefined>;
 /**
  * Resolve the pinned version from `AUTOMOBILE_VERSION`. Returns the trimmed value
  * when set, otherwise the `"latest"` placeholder (which downstream resolvers turn
- * into the concrete newest registry entry).
+ * into the concrete newest registry entry). The `latest` sentinel is normalized to
+ * lower-case so every sink agrees — `resolveChecksum` matches `latest`
+ * case-insensitively but `resolveAssetVersion`/URL building use a strict `===`
+ * compare, so an un-normalized `LATEST` would otherwise resolve to a valid checksum
+ * yet a 404 URL (an incoherent triple).
  */
 export function resolvePinnedVersion(env: EnvLike = process.env): string {
   const trimmed = env[AUTOMOBILE_VERSION_ENV]?.trim();
-  return trimmed && trimmed.length > 0 ? trimmed : LATEST_RELEASE_VERSION;
+  if (!trimmed || trimmed.length === 0) {
+    return LATEST_RELEASE_VERSION;
+  }
+  return trimmed.toLowerCase() === LATEST_RELEASE_VERSION ? LATEST_RELEASE_VERSION : trimmed;
+}
+
+/**
+ * True when `AUTOMOBILE_VERSION` names a concrete version (not unset, not the
+ * `latest` sentinel). Used to decide whether an unverifiable download should
+ * fail closed.
+ */
+export function isExplicitPin(env: EnvLike = process.env): boolean {
+  return resolvePinnedVersion(env) !== LATEST_RELEASE_VERSION;
+}
+
+/**
+ * True when the effective pin resolves to a checksum-bearing registry entry.
+ * A `latest` pin is known iff the registry is non-empty; a concrete pin is known
+ * iff the registry contains it. A pinned-but-unknown version cannot be
+ * integrity-verified — see the fail-closed guards in the CtrlProxy managers.
+ */
+export function isPinnedVersionKnown(
+  env: EnvLike = process.env,
+  registry: ReleaseChecksumEntry[] = RELEASE_CHECKSUM_REGISTRY
+): boolean {
+  const pinned = resolvePinnedVersion(env);
+  if (pinned === LATEST_RELEASE_VERSION) {
+    return registry.length > 0;
+  }
+  return registry.some(entry => entry.version === pinned);
 }
 
 /**
@@ -219,9 +257,10 @@ export function resolveAssetBaseUrl(env: EnvLike = process.env): string {
 function buildReleaseAssetUrl(
   filename: string,
   version: string,
-  baseUrl: string = DEFAULT_ASSET_BASE_URL
+  baseUrl: string = DEFAULT_ASSET_BASE_URL,
+  registry: ReleaseChecksumEntry[] = RELEASE_CHECKSUM_REGISTRY
 ): string {
-  const assetVersion = resolveAssetVersion(version);
+  const assetVersion = resolveAssetVersion(version, registry);
   if (assetVersion === LATEST_RELEASE_VERSION) {
     // Degenerate case: empty registry, no concrete version to key off.
     if (baseUrl === DEFAULT_ASSET_BASE_URL) {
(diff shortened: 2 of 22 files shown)


docs/using/hermetic-ci-pinning.md
```diff
@@ -0,0 +1,126 @@
...
+(and `.../control-proxy.ipa`). Lay your mirror out with a directory per version:
+
+```
+https://artifacts.internal/automobile/
+  0.0.40/
+    control-proxy-debug.apk
+    control-proxy.ipa
+```
+
+Checksums are still verified against the daemon's baked registry, so a mirror cannot
```
kaeawc: **[Supply-chain]** This claim is overstated as written. `resolveApkChecksum`/`resolveIpaChecksum` return `""` for any version *not in the baked registry* (`resolveChecksum` → `if (!entry) return ""`). Both download paths gate verification on `expectedChecksum.length > 0` (`CtrlProxyManager.downloadApk` line ~840 and `doPrefetch` line ~251; iOS `verifyBundle` line ~756), so an empty checksum means the fetch proceeds **unverified** with only a `logger.warn`. 

Reproduced: `AUTOMOBILE_VERSION=99.99.99 AUTOMOBILE_ASSET_BASE_URL=https://evil-mirror.test/am` →
```
url:      https://evil-mirror.test/am/99.99.99/control-proxy-debug.apk
checksum: ""   (verification skipped)
```
This is not hypothetical: the registry is capped at 100 entries and older versions are evicted, and a consumer can legitimately pin a real-but-not-yet-baked release (e.g. `0.0.41`) or fat-finger a version. In every one of those cases a mirror **can** serve a tampered asset. Either (a) fail hard when a pinned version resolves to an empty checksum while a non-default `AUTOMOBILE_ASSET_BASE_URL` is set (turn the silent hole into a hard error in the hermetic recipe), or (b) qualify this sentence: verification only holds for versions present in the baked registry; an unknown/aged-out pin downgrades to unverified. The earlier bullet (line ~40) hints at this but ties it only to a checksum-skip override, not to the mirror path. Verified by reading the code + `bun -e` repro. — Sofia (supply-chain)

src/constants/release.ts
```diff
@@ -173,21 +173,102 @@ export function resolveAssetVersion(version: string): string {
...
+}
+
+/** iOS IPA download URL honoring `AUTOMOBILE_VERSION` + `AUTOMOBILE_ASSET_BASE_URL`. */
+export function resolveIpaUrl(env: EnvLike = process.env): string {
+  return buildReleaseAssetUrl("control-proxy.ipa", resolvePinnedVersion(env), resolveAssetBaseUrl(env));
+}
+
+/** Expected APK SHA-256 for the pinned version (empty string if unknown). */
+export function resolveApkChecksum(env: EnvLike = process.env): string {
+  return resolveChecksum(resolvePinnedVersion(env), "android");
```
kaeawc, line 294: **[Supply-chain]** `resolveApkChecksum`/`resolveIpaChecksum` silently return `""` for an unknown pinned version (via `resolveChecksum`'s `if (!entry) return ""`). That empty string is the trust hole: every consumer treats `length === 0` as "skip verification" rather than "fail closed." For a *hermetic pinning* feature this is the wrong default — a pinned-but-unregistered version composed with a mirror URL produces a real download URL and no integrity check. Consider a strict variant (or a flag threaded through the hermetic path) that throws when `AUTOMOBILE_VERSION` is set to a version absent from the registry, instead of degrading to unverified. At minimum the resolver's doc-comment should state that an empty return disables verification downstream, so future callers don't assume it's benign. Verified against `resolveChecksum` (line 143) + the `.length > 0` consumer branches. — Sofia (supply-chain)

src/utils/CtrlProxyManager.ts
```diff
@@ -249,7 +247,7 @@ export class AndroidCtrlProxyManager implements CtrlProxyManager {
       AndroidCtrlProxyManager.verifyApkIntegrityStatic(apkPath);
 
       // Verify checksum if provided
-      const expectedChecksum = AndroidCtrlProxyManager.expectedChecksumOverride ?? APK_SHA256_CHECKSUM;
+      const expectedChecksum = AndroidCtrlProxyManager.expectedChecksumOverride ?? resolveApkChecksum();
```
kaeawc, line 264: **[Supply-chain]** This is one of the two download sites that inherit the empty-checksum bypass. When `resolveApkChecksum()` returns `""` (unknown pinned version), the `if (expectedChecksum.length > 0)` guard below skips SHA verification and the prefetched APK from `resolveApkUrl()` — which honors `AUTOMOBILE_ASSET_BASE_URL` — is installed unverified. Same shape at `downloadApk` (~line 840). Separately, note the pre-existing `acceptedPreinstalled` path (line ~644): even with a *known* checksum, an installed APK whose SHA differs from expected is accepted non-blocking (`acceptedPreinstalled: true`) rather than rejected. Both bypasses stay reachable under the documented recipes, so the PR should not claim end-to-end hermeticity without either closing them or explicitly listing them as accepted gaps in `hermetic-ci-pinning.md`. Verified by reading both download paths + the version-check flow. — Sofia (supply-chain)

src/utils/IOSCtrlProxyBuilder.ts
```diff
@@ -661,7 +661,7 @@ export class IOSCtrlProxyBuilder {
     if (override !== null) {
       return override;
     }
-    return IOS_CTRL_PROXY_SHA256_CHECKSUM;
+    return resolveIpaChecksum();
```
kaeawc, line 680: **[Supply-chain]** iOS side has the identical empty-checksum bypass. `getExpectedChecksum()` falls back to `resolveIpaChecksum()`, which is `""` for an unknown pinned version. Downstream, `isBundleValid` returns `true` for *any* bundle when `expectedChecksum` is empty (line ~741 `if (!expectedChecksum) return true`), and `verifyBundle` skips the SHA check with only a warn (line ~756). So a pinned-but-unregistered version + `AUTOMOBILE_ASSET_BASE_URL` installs an unverified IPA. The hermetic-mode `isLatest` gate you added (line ~708) only disables the *cached-fallback*; it does nothing about the unverified *download* when the checksum is empty. Note also `AUTOMOBILE_CTRL_PROXY_IOS_IPA_PATH` (the vendored-IPA override this PR's iOS recipe recommends) bypasses checksum entirely — consistent with Android's APK path override. These are defensible as scoped-out, but the docs claim hermeticity, so they should be named as explicit trust boundaries rather than left silent. Verified by reading `getExpectedChecksum`, `isBundleValid`, and `verifyBundle`. — Sofia (supply-chain)

kaeawc (commented):
Reviewed through a supply-chain / hermeticity lens. The resolver plumbing is clean and the pinning/mirror composition works as advertised for versions that live in the baked registry. But the feature's central security promise — "a mirror cannot serve a tampered asset for a pinned version" — does not hold for the empty-checksum path, and two pre-existing bypasses remain reachable under the recipes this PR documents. Findings below, each reproduced. — Sofia (supply-chain)

src/daemon/socketServer.ts
```diff
@@ -558,19 +559,22 @@ export class UnixSocketServer {
       }
       case "ide/status": {
         return {
+          // Concrete pinned version (honors AUTOMOBILE_VERSION), never the
+          // floating "latest" tag — external consumers must see exactly what the
+          // daemon will fetch (#2746).
           version: getMcpServerVersion(),
```
kaeawc, line 565: **[Lifecycle]** This reports `resolvePinnedVersion()` from **the daemon's own** `process.env`, not the caller's — which is honest, but it silently masks the reuse-drift hole. Verified path: the Kotlin runner (`DaemonSocketClient.kt:90`) does `if (!forceRestart && isAvailable(socketPath)) return` and the Swift runner (`AutoMobileEnvironment.swift:172`) does `if isDaemonRunning() { return true }`. Per this repo's per-UID / cross-worktree single-daemon model (#2599/#2732), the daemon that's *already running* was launched with whatever `AUTOMOBILE_VERSION` the **first** caller had (or none → `latest`→0.0.40). A second checkout exporting `AUTOMOBILE_VERSION=0.0.18` connects to that pre-existing daemon; `ide/status` then returns `releaseVersion: "0.0.40"` and the daemon downloads the 0.0.40 APK/IPA — the 0.0.18 pin is silently ignored, exactly the drift the PR sets out to kill. The doc even says "If `releaseVersion` is anything other than the version you pinned, the environment is not hermetic" — but nothing enforces it. Fix options: (a) have `ide/status` also echo back the *caller-requested* version (client passes it) and flag a mismatch, or (b) make `ide/status` a hard error when the daemon's pin ≠ a caller-supplied pin, forcing `--daemon force.restart`. Verification: `AUTOMOBILE_VERSION=0.0.18 bunx … --daemon start`, then from a second shell without the pin call `ide/status` → you get 0.0.18, not latest; and vice-versa. The value is daemon-global, never caller-scoped. — Marcus (daemon-lifecycle)

src/constants/release.ts
```diff
@@ -173,21 +173,102 @@ export function resolveAssetVersion(version: string): string {
...
+export const DEFAULT_ASSET_BASE_URL = "https://github.com/kaeawc/auto-mobile/releases/download";
+
+type EnvLike = Record<string, string | undefined>;
+
+/**
+ * Resolve the pinned version from `AUTOMOBILE_VERSION`. Returns the trimmed value
+ * when set, otherwise the `"latest"` placeholder (which downstream resolvers turn
+ * into the concrete newest registry entry).
+ */
+export function resolvePinnedVersion(env: EnvLike = process.env): string {
```
kaeawc, line 211: **[Lifecycle]** `resolvePinnedVersion` reads `process.env` at call time inside the daemon process. That's correct for a freshly-spawned daemon, but the pin therefore lives-and-dies with the daemon's launch environment — it is NOT re-read per request from the caller. Combined with runner-side daemon reuse (Kotlin `DaemonSocketClient.kt:90`, Swift `AutoMobileEnvironment.swift:172`, and the TS single-daemon policy in `manager.ts:468-478` which stops-then-starts only when *this* process runs `--daemon start`), the effective pin is 'whoever launched the surviving daemon wins'. Two worktrees wanting different `AUTOMOBILE_VERSION`s over one shared daemon get the launcher's version, silently. This is a runtime coherence gap, not just docs. Minimum mitigation: on `--daemon start`, stamp the resolved `AUTOMOBILE_VERSION` into the PID file and have the runners (or `ide/status`) compare the caller's env against it and fail loudly on mismatch instead of reusing. — Marcus (daemon-lifecycle)

src/constants/release.ts
```diff
@@ -173,21 +173,102 @@ export function resolveAssetVersion(version: string): string {
...
-    // Degenerate case: empty registry. Fall back to GitHub's redirecting
-    // /latest/download/ endpoint.
-    return `https://github.com/kaeawc/auto-mobile/releases/latest/download/${filename}`;
+    // Degenerate case: empty registry, no concrete version to key off.
+    if (baseUrl === DEFAULT_ASSET_BASE_URL) {
+      // Fall back to GitHub's redirecting /latest/download/ endpoint.
+      return `https://github.com/kaeawc/auto-mobile/releases/latest/download/${filename}`;
+    }
+    // A mirror has no redirecting endpoint; use a conventional /latest/ path.
+    return `${baseUrl}/latest/${filename}`;
```
kaeawc, line 271: **[Lifecycle]** Mirror `/latest/` fallback: when `AUTOMOBILE_ASSET_BASE_URL` is set AND the registry is empty (degenerate `assetVersion === "latest"`), this emits `${baseUrl}/latest/${filename}`. But `resolveApkChecksum`/`resolveIpaChecksum` return `""` for `latest` on an empty registry, so the download proceeds **unverified** against a mirror URL — the one case where a mirror could serve a tampered asset for an unpinnable version. The docs claim 'a mirror cannot serve a tampered asset for a pinned version', which is true, but this degenerate `latest`-on-mirror path is exactly the unpinned hole. Low severity (empty registry is not a shipping state) but worth a guard: refuse the mirror `/latest/` path unless a checksum-skip override is explicitly set, mirroring how the APK download already warns on empty checksum. — Marcus (daemon-lifecycle)

docs/using/hermetic-ci-pinning.md
```diff
@@ -0,0 +1,126 @@
...
+```bash
+export AUTOMOBILE_VERSION=0.0.40
+```
+
+- Unset (or `latest`) resolves to the newest entry in the daemon's baked release registry
+  — a **concrete** version, never the floating `@latest` tag.
+- An unknown version resolves to an empty checksum (download proceeds unverified only if
+  you have also opted into a checksum-skip override — see below).
+- The Android Kotlin runner already reads `AUTOMOBILE_VERSION` as a fallback for the
+  daemon package version, so the same env var lines up the runner-spawned daemon too.
```
kaeawc: **[Lifecycle]** "the same env var lines up the runner-spawned daemon too" — only for a daemon this runner *spawns*. The Kotlin runner reuses an already-running daemon without restart (`DaemonSocketClient.kt:87-92`) unless `-Dautomobile.daemon.force.restart=true`, and the Swift runner (`AutoMobileEnvironment.swift:170-185`) has no force-restart path at all and its `startDaemon` (line 132-168) hardcodes `["--daemon", "start"]` with zero version pin — it only inherits ambient env, so on iOS the daemon package version isn't pinned even on a fresh spawn. For the doc to be truthful, the Android recipe's reliance on `AUTOMOBILE_VERSION` should state that a *pre-existing* daemon must be force-restarted (step 3 mentions `automobile.daemon.force.restart` but frames it as optional), and the iOS recipe currently cannot honor `AUTOMOBILE_VERSION` for the daemon package at all — that's the scoped Swift follow-up, but the doc reads as if it already works. — Marcus (daemon-lifecycle)

kaeawc (commented):
Reviewing through a daemon-lifecycle lens: does the pin actually hold end-to-end at runtime in a long-lived, per-UID, cross-worktree daemon spawned from Kotlin/Swift?

**The TS-internal wiring is coherent.** I grepped `src/` for every remaining importer of the frozen module-level consts (`APK_URL`, `IOS_CTRL_PROXY_IPA_URL`, the checksum exports, `RELEASE_VERSION`) — zero remain outside `release.ts` itself. Every runtime consumer (CtrlProxyManager download + checksum, IOSCtrlProxyBuilder getBundleUrl/getExpectedChecksum/download, `ide/status`, `bootedDeviceResources`) now resolves at call time. I verified with a `bun` snippet that `resolveApkUrl()` reflects `process.env.AUTOMOBILE_VERSION` mutated after import, so the frozen consts and the resolvers can't drift *within one process* — the report path and the download path read the same env. The frozen consts are now effectively dead code at runtime (only backward-compat tests read them), which is fine. Test-env restore in the new tests is airtight (undefined→delete vs restore in `finally`, and the pure resolver tests inject `env` explicitly rather than mutating global) — no leakage.

**The real hole is the process boundary, not the TS.** The pin is honored where the daemon *reads its own* `process.env` — i.e. only if the daemon was launched with `AUTOMOBILE_VERSION` set. But the whole architecture (per the repo's own #2599/#2732 notes) is a single long-lived daemon on a per-UID socket, reused across worktrees, and *both* runners reuse an already-running daemon without restarting it. So a second checkout that pins a different version connects to a daemon still serving the first version's assets/checksums, and `ide/status` faithfully reports the *daemon's* launch version — which silently contradicts the caller's intended pin with no warning. Swift `startDaemon` is additionally pin-less entirely. Details inline. I've separated the genuine runtime gap (reuse drift) from the explicitly-scoped native follow-ups. — Marcus (daemon-lifecycle)

src/constants/release.ts
```diff
@@ -173,21 +173,102 @@ export function resolveAssetVersion(version: string): string {
...
+type EnvLike = Record<string, string | undefined>;
+
+/**
+ * Resolve the pinned version from `AUTOMOBILE_VERSION`. Returns the trimmed value
+ * when set, otherwise the `"latest"` placeholder (which downstream resolvers turn
+ * into the concrete newest registry entry).
+ */
+export function resolvePinnedVersion(env: EnvLike = process.env): string {
+  const trimmed = env[AUTOMOBILE_VERSION_ENV]?.trim();
+  return trimmed && trimmed.length > 0 ? trimmed : LATEST_RELEASE_VERSION;
```
kaeawc: **[Correctness]** Mixed-case `AUTOMOBILE_VERSION=LATEST` (or `Latest`) yields an incoherent asset triple. `resolvePinnedVersion` trims but does not lowercase, so `"LATEST"` flows through unchanged. Downstream the two consumers disagree: `resolveChecksum` lowercases (returns the real latest SHA — there's even a passing test `resolveChecksum("LATEST") === newest`), but `resolveAssetVersion` uses a strict `=== "latest"` compare, so the URL becomes `.../LATEST/control-proxy-debug.apk` (404) and the specifier becomes `@kaeawc/auto-mobile@LATEST`.

Verified (`bun -e`, env `{AUTOMOBILE_VERSION:"LATEST"}`):
- `resolveApkChecksum` -> `8e89...ca87aa` (valid latest)
- `resolveApkUrl` -> `.../download/LATEST/control-proxy-debug.apk` (broken)
- `resolveDaemonInstallSpecifier` -> `@kaeawc/auto-mobile@LATEST`
- and via socketServer.ts:566, `ide/status.releaseVersion` -> `"LATEST"`

Because the SHA is real, this doesn't fail cleanly — the runner fetches a 404 with a valid-looking expected checksum. Fix at the source so the `latest` sentinel is normalized once (lowercase the trimmed value before the length check, mapping a case-insensitive `latest` back to `LATEST_RELEASE_VERSION`), after which every sink agrees. A concrete pinned version like `0.0.18` must stay untouched. — Priya (api/correctness)

src/utils/IOSCtrlProxyBuilder.ts
```diff
@@ -703,7 +703,9 @@ export class IOSCtrlProxyBuilder {
       const bundleReady = await this.isBundleValid(bundlePath, expectedChecksum);
 
       if (!bundleReady) {
-        const isLatest = IOS_CTRL_PROXY_RELEASE_VERSION.trim().toLowerCase() === LATEST_RELEASE_VERSION;
+        // When a version is pinned (AUTOMOBILE_VERSION), hermetic mode disables the
+        // silent cached-bundle fallback so a failed download fails hard (#2746).
+        const isLatest = resolvePinnedVersion().trim().toLowerCase() === LATEST_RELEASE_VERSION;
```
kaeawc: **[Correctness]** This is the smoking gun for the `resolvePinnedVersion` casing bug two files over: here the hermetic-fallback gate defensively re-normalizes with `.trim().toLowerCase()`, so `AUTOMOBILE_VERSION=LATEST` is correctly treated as latest (cached fallback stays enabled). But the sibling calls in the same class — `resolveIpaUrl()`/`resolveIpaChecksum()` (getBundleUrl/getExpectedChecksum) and the metadata `version: resolveAssetVersion(resolvePinnedVersion())` at line ~774 — do NOT normalize, so for the very same env they emit a `.../LATEST/control-proxy.ipa` URL, a valid latest SHA, and `metadata.version: "LATEST"`. Inconsistent normalization across call sites of one resolver is exactly why the fix should live inside `resolvePinnedVersion` (release.ts:204), after which this local `.trim().toLowerCase()` becomes redundant. — Priya (api/correctness)

src/constants/release.ts
```diff
@@ -173,21 +173,102 @@ export function resolveAssetVersion(version: string): string {
...
+  filename: string,
+  version: string,
+  baseUrl: string = DEFAULT_ASSET_BASE_URL
+): string {
   const assetVersion = resolveAssetVersion(version);
   if (assetVersion === LATEST_RELEASE_VERSION) {
-    // Degenerate case: empty registry. Fall back to GitHub's redirecting
-    // /latest/download/ endpoint.
-    return `https://github.com/kaeawc/auto-mobile/releases/latest/download/${filename}`;
+    // Degenerate case: empty registry, no concrete version to key off.
```
kaeawc, line 265: **[Coverage]** This degenerate empty-registry branch (and its sibling mirror `/latest/` path just below) has no test and, as written, can't get one: `buildReleaseAssetUrl` reaches the global `RELEASE_CHECKSUM_REGISTRY` transitively via `resolveAssetVersion`->`resolveLatestVersion` and takes no injectable registry param, unlike `resolveChecksum`. So `resolveAssetVersion("latest")` can only return `latest` when the global registry is empty — unreachable in a unit test. The PR's `test/constants/release.test.ts` covers empty-registry only for `resolveChecksum` (which does accept an injected registry). Either thread a registry param through `resolveAssetVersion`/`buildReleaseAssetUrl` so this branch is exercisable, or drop the dead branch. — Priya (api/correctness)

docs/using/hermetic-ci-pinning.md
```diff
@@ -0,0 +1,126 @@
...
+the Android APK, the iOS IPA, and their expected SHA-256 checksums together. The daemon
+resolves all four from one value, so there is no way for them to disagree.
+
+```bash
+export AUTOMOBILE_VERSION=0.0.40
+```
+
+- Unset (or `latest`) resolves to the newest entry in the daemon's baked release registry
+  — a **concrete** version, never the floating `@latest` tag.
+- An unknown version resolves to an empty checksum (download proceeds unverified only if
```
kaeawc: **[Correctness]** This safety claim is not backed by the code. Verified in `CtrlProxyManager.ts:840,856-859` (and the prefetch path at 251): when the expected checksum is empty, the download path just logs `"checksum verification SKIPPED - no checksum provided (development mode)"` and proceeds unverified — no override required. The `shouldSkipChecksum()` / `AUTOMOBILE_SKIP_ACCESSIBILITY_CHECKSUM` gate (line 1564) is a separate opt-out and does not govern the empty-checksum case. So a typo'd or not-yet-registered `AUTOMOBILE_VERSION` (e.g. `0.0.99`) combined with `AUTOMOBILE_ASSET_BASE_URL` silently fetches whatever the mirror serves with zero integrity checking. This also contradicts the doc's own line 58-59 ("a mirror cannot serve a tampered asset for a pinned version") — true only for a known pinned version. Please correct the parenthetical to state an unknown version downloads unverified, and consider whether an unknown pinned version should hard-fail instead of falling back to skip. — Priya (api/correctness)

kaeawc (commented):
Reviewed for resolver correctness across all inputs and whether the PR truly closes #2746. The default/no-env path is byte-identical to before (verified: APK_URL/APK_SHA/IPA_URL/IPA_SHA/releaseVersion all unchanged), composition of both knobs works, and the `@latest` sweep of `src/` is clean. One real correctness bug and one doc-vs-code mismatch, plus a missing-coverage note — details inline.

**The bug (root cause):** `resolvePinnedVersion` normalizes with `.trim()` but NOT case, while the two sinks it feeds disagree on case-handling. `resolveChecksum` lowercases its input (there's even a test asserting `resolveChecksum("LATEST") === newest`), but `resolveAssetVersion` compares strictly `=== "latest"`. So `AUTOMOBILE_VERSION=LATEST` (or `Latest`) produces an incoherent triple: a valid latest SHA-256, a 404 URL (`.../LATEST/control-proxy-debug.apk`), and specifier `@kaeawc/auto-mobile@LATEST`. Verified via `bun -e` — this propagates into `ide/status` and into doctor's reported version. The fix belongs in one place: lowercase the sentinel in `resolvePinnedVersion` so every downstream resolver agrees. — Priya (api/correctness)

kaeawc:
Addressed all three reviews in 5913b39.

**Mixed-case `latest` bug (correctness):** `resolvePinnedVersion` now normalizes the `latest` sentinel case-insensitively, so `AUTOMOBILE_VERSION=LATEST` resolves to a coherent concrete version/URL/checksum/specifier instead of `valid SHA + /LATEST/ 404`. Removed the now-redundant defensive `.trim().toLowerCase()` in `IOSCtrlProxyBuilder`.

**Fail-open hermeticity hole (supply-chain + correctness + lifecycle):** an explicitly-pinned version absent from the baked registry now **fails closed** at both download paths (`CtrlProxyManager` APK, `IOSCtrlProxyBuilder` IPA) rather than silently installing an unverifiable asset — closing the mirror-tamper vector. The existing overrides (`AUTOMOBILE_SKIP_ACCESSIBILITY_CHECKSUM` / vendored `AUTOMOBILE_CTRL_PROXY_IOS_IPA_PATH`) stay as the deliberate escape hatch for a real-but-not-yet-baked release. New pure predicates `isExplicitPin` / `isPinnedVersionKnown`; guards covered by tests on both platforms (fail-closed + escape-hatch installs).

**Untestable degenerate branch (correctness):** threaded an injectable registry through `resolveAssetVersion` → `buildReleaseAssetUrl` → `resolveApkUrl`/`resolveIpaUrl`; the empty-registry mirror `/latest/` fallback is now unit-tested.

**Doc overstatements (all three):** `hermetic-ci-pinning.md` now documents the fail-closed behavior + escape hatch, drops the absolute "cannot serve a tampered asset" claim (scoped to registry-known versions), adds a **Shared daemon caveat** (the pin is the daemon's launch-env property; a reused cross-worktree daemon must be force-restarted; verify via `ide/status`), and flags the still-pinless iOS Swift `startDaemon` as tracked follow-up.

**Deferred (noted, not in this PR):** Marcus's PID-file version stamp + client-side mismatch enforcement is a genuine improvement but the comparison must happen in the Kotlin/Swift clients (out of the TS surface here); documented the restart requirement + `ide/status` drift check as the interim guard.

`bun test` → 5351 pass / 0 fail (+13); lint + build green.

kaeawc:
Addressed the Codex review's fresh findings on `5913b39` in `b875454` — the earlier fail-closed guard only covered the *download* paths, and Codex correctly found two reuse bypasses:

- **Android readiness path** (`CtrlProxyManager.ts:863`): `ensureCompatibleVersion()` returned `"skipped"` (accepting an already-installed APK) before `downloadApk()` ran. Extracted a shared `assertPinnedVersionVerifiable()` and moved it to the **top of `ensureCompatibleVersion`** (the readiness gate) plus `downloadApk`, so a reused device with an unknown pin now fails closed before any installed service is accepted. Test: installed-but-unknown-pin → throws; `+ AUTOMOBILE_SKIP_ACCESSIBILITY_CHECKSUM=1` → accepted.
- **iOS reuse path** (`IOSCtrlProxyBuilder.ts:769`): `needsRebuild()` returned false for a cached bundle + metadata when the checksum was empty, so `setup()` reused the cached runner without reaching `verifyBundle`. Added the shared guard to `needsRebuild` (after the `SKIP_BUILD` opt-out) as well as `verifyBundle`. Test: cached bundle + unknown pin → throws; vendored `AUTOMOBILE_CTRL_PROXY_IOS_IPA_PATH` → exempt.
- **iOS recipe doc** (`hermetic-ci-pinning.md:108`): corrected — `AUTOMOBILE_CTRL_PROXY_IOS_IPA_PATH` must **not** be paired with `AUTOMOBILE_SKIP_CTRL_PROXY_IOS_BUILD=1`, since the skip flag returns from `needsRebuild()` before the vendored IPA is consumed. The vendored-IPA path already avoids fetch/compile on its own.

`bun test` → 5355 pass / 0 fail (+4); lint + build green.

kaeawc:
Addressed the Codex finding on `e3d03e99` in `6c7756b`: the static startup prefetch (`prefetchApk` → `doPrefetch`) was the one path that still background-downloaded an unverifiable APK for an unknown `AUTOMOBILE_VERSION` pin, ahead of the instance-level guards. `doPrefetch()` now **skips the network fetch** (returns null) for an unknown explicit pin — respecting the same `AUTOMOBILE_SKIP_ACCESSIBILITY_CHECKSUM` / APK-path / checksum-override escape hatches — so no unchecked APK is ever cached. Skipping rather than throwing keeps daemon startup alive; the install path still fails closed. Extracted a shared static `isChecksumSkipConfigured()` so the instance and static paths can't drift. +1 test (prefetch performs zero downloads for an unknown pin). `bun test` 5358 pass / 0 fail; lint + build green.

kaeawc:
Addressed the vendored-IPA finding on `6c7756b` in the latest push: `needsRebuild()` now **forces extraction whenever `AUTOMOBILE_CTRL_PROXY_IOS_IPA_PATH` is set**, so on a reused host with a cached xctestrun + metadata the vendored IPA is actually consumed instead of the stale cached runner being silently reused. Test updated to assert this with a pre-existing cached bundle + metadata present. `bun test` 5358 pass / 0 fail; lint + build green.

kaeawc:
Addressed both findings on `f2288ca` in the latest push:
- **doctor** now returns `fail` (not `skip`) for an unverifiable explicit pin, so the documented `--cli doctor` gate actually blocks in CI.
- **booted-device compat** (Android) now reports `isCompatible: false` for an unverifiable pin instead of treating an empty expected checksum as compatible.

Centralized the whole fail-closed decision into one static `AndroidCtrlProxyManager.isPinnedVersionUnverifiable()` (respecting the skip/override escape hatches), now the single source of truth for the download/readiness guards, the startup prefetch, doctor, and the booted-device check — so they can't drift. +5 tests. `bun test` 5363 pass / 0 fail; lint + build green.

kaeawc:
Addressed the three iOS-symmetric findings on `64f0efd` in the latest push (mirrors the Android fixes):
- **`IOSCtrlProxyManager.setup()`** now fails closed at the top, before the already-running / already-attempted short-circuits — so a reused simulator with a running runner can't serve an unverifiable pin.
- **`doctor --ios`** now returns `fail` for an unverifiable pin (classifyRunner only checks the command set, so it would otherwise stay `pass`).
- **booted-device compat** (iOS) now reports `isCompatible: false` for an unverifiable pin.

Centralized in a static `IOSCtrlProxyBuilder.isPinnedVersionUnverifiable()` (vendored-IPA/bundle-path + checksum-override escape hatches), the single source of truth for the iOS build/reuse guards, `setup()`, doctor, and compat — mirroring the Android `AndroidCtrlProxyManager.isPinnedVersionUnverifiable()`. +6 tests. `bun test` 5369 pass / 0 fail; lint + build green.

kaeawc:
Filed follow-up issues for the deferred work (scoped out of this PR):

- **#2812** — Enforce the version pin across a reused/shared daemon (PID-file stamp + client mismatch fail). *The one deferred item that's a real runtime gap; addressed here only via the "Shared daemon caveat" doc.*
- **#2813** — iOS: honor `AUTOMOBILE_VERSION` when the XCTestRunner autostarts the daemon (Swift `startDaemon`).
- **#2814** — iOS: bake a version constant into the XCTestRunner SPM package (mirrors Android's jar `Implementation-Version`).
- **#2815** — Supply-chain hardening: `acceptedPreinstalled` known-SHA mismatch + `AUTOMOBILE_ASSET_BASE_URL` query-string validation.

The Android gradle hermetic recipe is documented in `docs/using/hermetic-ci-pinning.md` and needs no code follow-up (the Kotlin runner already reads `AUTOMOBILE_VERSION`, and `--cli doctor` now fails closed on an unverifiable pin).

PR #2825, opened 2026-07-01
fix(ios): deterministic daemon startup + video-recording stop in XCTestRunner CI

kaeawc:
Makes the **XCTestRunner Simulator Tests** check deterministic by fixing two independent, pre-existing flakes. The two fixes are separate commits.

## FLAKE 1 — Daemon fails to start (`auto-mobile not found in PATH`)
**Symptom:** `RemindersLaunchPlanTests` skipped with the generic `Failed to start AutoMobile Daemon. Ensure auto-mobile is installed and on PATH.`

**Root cause (from the PerfTimer log):** `runDaemonSubcommand: ERROR - auto-mobile not found in PATH`. The ~19.6s in the log was almost entirely `hasBootedSimulator`/`simctl` (19.3s); the daemon start itself failed in ~300ms. `bun add -g .` installs the CLI to `~/.bun/bin`, which is **not** on `$GITHUB_PATH`, so the `swift test` subprocess couldn't resolve `auto-mobile`. The generic skip message hid the real cause. Other CI scripts (`verify-runtime-deps.sh`, `daemon-lifecycle.sh`) already work around this with `export PATH="${HOME}/.bun/bin:${PATH}"`; the XCTestRunner job never did.

So this flake is **primarily CI-infra**, plus a product-code diagnostic improvement:

- **CI (infra):** new [`scripts/ci/ensure-daemon-ready.sh`](scripts/ci/ensure-daemon-ready.sh) adds `~/.bun/bin` to `$GITHUB_PATH`, verifies `auto-mobile` resolves (fail-fast with a clear `::error::`), then pre-starts the daemon and polls `--daemon health` with bounded backoff until ready. Wired as a new step before the swift-test step; the test then reuses the already-ready daemon. A daemon-start failure now surfaces **loudly in its own step** instead of a silent test skip. Covered by [`test/bats/ensure-daemon-ready.bats`](test/bats/ensure-daemon-ready.bats) (6 tests, all mock `auto-mobile` — no daemon/device needed).
- **Swift (product):** `DaemonManager` now surfaces the real failure via `DaemonStartupResult` (`.executableNotFound` / `.launchFailed` / `.readinessTimeout` / `.versionSkew`). `runDaemonSubcommand` distinguishes a missing CLI from a launched-but-failed process; the XCTest skip message now names the actual cause. New pure-logic unit tests in `XCTestRunnerTests`.

The TS daemon-readiness path (`src/daemon/manager.ts`) was already robust (socket polling + connection verification + retry backoff) and was **not** implicated in the observed failure, so it's untouched.

## FLAKE 2 — iOS video recording file missing on stop
**Symptom:** `Failed to stop video recording: iOS recording file missing at .../<uuid>-raw.mov`.

**Root cause:** PR #2730 gave `simctl` a generous 30s SIGKILL window to flush the moov atom, but `stop` still did a **single `stat`** immediately after the recorder process exited (`postProcessRecording`). On a loaded runner the moov-atom flush + filesystem write can lag behind process exit, so that one glance sees the file missing/0-bytes and hard-fails.

**Fix:** new `waitForRecordingFileReady()` polls (bounded, exponential backoff via the repo `Backoff` primitive, injectable `Timer`) until the raw `.mov` exists, is non-empty, and its size is **stable across two probes** (so a still-flushing/truncated file is never handed to ffmpeg). On timeout it throws a diagnostic naming what was observed (`never appeared` / `stayed empty (0 bytes)` / `never stabilized`) instead of a bare "missing". Confirmed we read the correct `<uuid>-raw.mov` path (before the ffmpeg `-c copy` remux to the final `.mp4`).

New fast unit tests in `test/features/video/FfmpegVideoProcessingBackend.test.ts` use a scripted fake probe + `FakeTimer` (late-appearing file, growing-then-stable file, never-appears, stays-empty) — no device.

## Validation
- `bunx turbo run lint build` → green
- `bun test test/features/video/` → 74 pass, 0 fail (~156ms)
- `swift build` + `swift test --filter XCTestRunnerTests` → 10 pass, 0 fail
- `bats test/bats/ensure-daemon-ready.bats` → 6 pass; `shellcheck` clean

No iOS simulator available locally to exercise the live recording start/stop; the fix is covered by the fake-probe unit tests and the CLI/BATS integration path.

Refs #2730

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1218,6 +1218,12 @@ jobs:
       - name: "Boot iOS Simulator (Xcode 26.2)"
         run: ./scripts/ios/boot-simulator.sh
 
+      # Put `auto-mobile` (installed to ~/.bun/bin by `bun add -g .`) on PATH for
+      # the swift-test step, and pre-start + wait for the daemon so a startup
+      # failure surfaces loudly here instead of silently skipping the test (#2730).
+      - name: "Ensure AutoMobile daemon ready"
+        run: ./scripts/ci/ensure-daemon-ready.sh
+
       - name: "Run Reminders integration tests (Xcode 26.2)"
         timeout-minutes: 10
         env:
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1224,6 +1224,12 @@ jobs:
       - name: "Ensure AutoMobile daemon ready"
         run: ./scripts/ci/ensure-daemon-ready.sh
 
+      # Launch CtrlProxy on the booted sim now (a slow, one-time xcodebuild launch)
+      # so the test's first `observe` doesn't pay that cost inside its short
+      # waitFor window and time out (#2730 follow-up).
+      - name: "Warm up iOS CtrlProxy"
+        run: ./scripts/ci/ensure-ctrl-proxy-ready.sh
+
       - name: "Run Reminders integration tests (Xcode 26.2)"
         timeout-minutes: 10
         env:
@@ -1252,6 +1258,14 @@ jobs:
         run: |
           brew list ffmpeg >/dev/null 2>&1 || brew install ffmpeg
           ./scripts/ios/video-recording-start-stop-integration.sh
+
+      # This is a freshly-booted sim, so re-confirm the daemon and warm CtrlProxy
+      # again before the second Reminders run (same rationale as the 26.2 block).
+      - name: "Ensure AutoMobile daemon ready (Xcode 26.3)"
+        run: ./scripts/ci/ensure-daemon-ready.sh
+
+      - name: "Warm up iOS CtrlProxy (Xcode 26.3)"
+        run: ./scripts/ci/ensure-ctrl-proxy-ready.sh
 
       - name: "Run Reminders integration tests (Xcode 26.3)"
         timeout-minutes: 10
--- a/ios/XCTestRunner/Sources/XCTestRunner/AutoMobileEnvironment.swift
+++ b/ios/XCTestRunner/Sources/XCTestRunner/AutoMobileEnvironment.swift
@@ -184,6 +184,20 @@ public enum DaemonManager {
         return runDaemonSubcommand("start", repoRoot: repoRoot)
     }
 
+    /// Map a launch/restart outcome to the startup failure it represents, or nil when the process
+    /// launched cleanly. Pure so the failure-cause mapping is unit-testable without spawning a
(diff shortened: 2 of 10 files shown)


kaeawc:
---
**Update (root cause corrected after first CI run):** The initial theory was "`~/.bun/bin` just isn't on `$GITHUB_PATH`." The new gate's first run proved deeper: `bun add -g .` does **not** reliably leave a runnable `auto-mobile` bin on the runner — the global copy under `~/.bun/install` can lack `dist/`, leaving a **dangling** `~/.bun/bin/auto-mobile` symlink that `command -v` rejects. So `ensure-daemon-ready.sh` now (a) puts `~/.bun/bin` on PATH/`$GITHUB_PATH`, and (b) if `auto-mobile` still doesn't resolve, symlinks it onto the freshly-built workspace `dist/src/index.js` (a `#!/usr/bin/env bun` executable that version-matches the checkout — the same entry `daemon-lifecycle.sh` runs). The gate still fails loudly if neither exists. Covered by an end-to-end BATS test of the shim-link fallback.

kaeawc:
**Third issue found & fixed (un-masked by the FLAKE 1 fix).** Once the daemon started reliably, the Reminders Swift test finally *ran* and failed on `observe waitFor {text: "Reminders"}` timing out.

Root cause (reproduced locally on a fresh iOS 26.2 sim): the daemon launches CtrlProxy lazily (`xcodebuild test-without-building`) on the first `observe`, and that launch takes longer than the plan's short `waitFor` window — so the first in-test observe times out. The job builds CtrlProxy but never ran the **documented "CtrlProxy pre-installation" warm-up** (`docs/design-docs/plat/ios/xctestrunner/ci-integration.md`). (It was *not* an onboarding issue — CtrlProxy's accessibility snapshot includes the underlying main-UI "Reminders" nodes behind the onboarding sheet.)

Fix: new `scripts/ci/ensure-ctrl-proxy-ready.sh` runs `observe` once in a dedicated, generous step so CtrlProxy is attached before the bounded test; wired into both sim lifecycles (Xcode 26.2 + 26.3). `observe` exits non-zero when it can't produce a hierarchy, so a clean exit is a real gate.

**Validated locally on a fresh iOS 26.2 simulator:** the exact Reminders plan passes end-to-end (3/3 steps) with CtrlProxy warm. The same run also finalized a valid MP4 — so the **FLAKE 2 video-stop fix is confirmed on a real simulator**, not just unit tests. BATS: 6 new tests; shellcheck clean.

Issue #2846, opened 2026-07-02
refactor(ios): migrate control-proxy off the flat WebSocketRequest bag to a typed command model


kaeawc:
## Summary

The iOS `control-proxy` XCUITest runner decodes every inbound WebSocket command into a single **flat `WebSocketRequest` struct with ~43 optional fields**, then dispatches via a **`switch request.type`** over `RequestType.<case>.rawValue`, with each handler re-extracting and unwrapping optionals from the shared bag. This is the same "kitchen-sink DTO + stringly-typed switch" pattern the Android `control-proxy` just retired in #2752 / #2771. This issue tracks migrating the iOS side to a **typed per-command model** for the same benefits: compile-time exhaustiveness, per-command required fields, and no optional-bag unwrapping.

The iOS sibling of #2752; part of the repo-wide stale-code cleanup catalog.

## Background — current state (`main`, `ios/control-proxy/Sources/CtrlProxy`)

- `Models.swift:7` — `public struct WebSocketRequest: Codable` is a flat bag of **~43 optional fields** (`x, y, x1, y1, x2, y2, offset, fingerCount, pressDurationMs, holdTime, centerX, distanceStart, rotationDegrees, text, resourceId, label, action, bundleId, sinceTimestamp, disableAllFiltering, id, shape, fileName, key, value, valueType, appId, databasePath, query, table, limit, …`). The iOS analog of Android's retired `LegacyWebSocketRequest`.
- `Models.swift:1026` — `public enum RequestType: String, CaseIterable` holds the `type` discriminator strings.
- `CommandHandler.swift:68` — `public func handle(_ request: WebSocketRequest) -> Any` does `switch request.type` (`:72`) over `RequestType.<case>.rawValue` (~38 cases). Each `handle*` method pulls optionals off the shared struct and force/optional-unwraps them; unknown types fall to a `default` returning an "Unknown command type" error.
- `Tests/CtrlProxyTests/CommandHandlerTests.swift` exercises the handler.

## Note: the iOS command set is NOT the Android one

[...]

Issue #2752, opened 2026-07-01
refactor(android): finish WebSocketServer typed-handler migration in control-proxy (retire LegacyWebSocketRequest)


kaeawc:
## Summary

The `control-proxy` on-device WebSocket server was mid-migration from an untyped "callback + flat DTO" request model to a typed "sealed request + message handler" model. The migration was started (the target types and handler interface exist, the server has a handler branch, the old type is `@Deprecated`) but never finished and wiring. As a result the **deprecated legacy path is the only live path, and the "preferred" typed path is dead code.** This issue documents the context and the decision to **finish the migration** rather than delete the typed path.

Found during a repo-wide stale-code cleanup audit; tracked as part of that series (see _Related_).

## Background — two request-handling designs

`WebSocketServer` (in `android/control-proxy`, the on-device accessibility runner) receives commands from the MCP server over WebSocket (port 8765) and dispatches them. Its own KDoc describes two modes:

> 1. Handler mode (preferred): Pass a `WebSocketMessageHandler` to receive typed requests
> 2. Callback mode (legacy): Pass individual callback lambdas for each message type

— `android/control-proxy/.../ctrlproxy/WebSocketServer.kt:100-104`

**Handler mode ("preferred", from the `protocol` module):**
- `WebSocketMessageHandler.handleMessage(request: WebSocketRequest): WebSocketResponse?` — `android/protocol/.../protocol/WebSocketMessageHandler.kt`
- sealed `WebSocketRequest` with one data class per request type — `android/protocol/.../protocol/WebSocketRequest.kt`
- sealed `WebSocketResponse` — `android/protocol/.../protocol/WebSocketResponse.kt`
- Server branch: decode into the sealed `ProtocolRequest`, call `messageHandler.handleMessage(...)`, broadcast the response — `WebSocketServer.kt:495-515`

**Callback mode (legacy):**
- ~35 individual callback lambdas passed to the `WebSocketServer` constructor — `WebSocketServer.kt:112-…`
- a kitchen-sink `LegacyWebSocketRequest` data class with ~40 nullable fields — `WebSocketServer.kt:38-90`
- a 43-case `when (request.type)` over the flat `type: String` — `WebSocketServer.kt:521-…`

## Current state — the roles are inverted

The intended-to-be-retired path is the one actually running, and the intended-to-be-canonical path can never execute:

[...]

PR #2854, opened 2026-07-02
refactor(ios): migrate control-proxy to a typed command model (#2846)

kaeawc:
## What

Migrates the on-device iOS `control-proxy` XCUITest runner off the flat
`WebSocketRequest` struct (~43 optional fields) + `switch request.type` dispatch
to a **typed command model**: an `enum WebSocketRequest` with one case per
command, each carrying a small per-command payload struct that declares only the
fields that command uses. `CommandHandler.handle` is now an **exhaustive**
`switch` over that enum — no `default` — so the compiler enforces that every
command is handled.

This is the on-device Swift sibling of the Android sealed-class migration
([#2752](https://github.com/kaeawc/auto-mobile/issues/2752) /
[#2771](https://github.com/kaeawc/auto-mobile/pull/2771)) and the TS-client
migration ([#2835](https://github.com/kaeawc/auto-mobile/pull/2835)). Closes
[#2846](https://github.com/kaeawc/auto-mobile/issues/2846).

## Why

Every inbound command previously decoded into one shared bag of optionals, and
each handler re-extracted and unwrapped optionals from it. A command's required
fields were invisible in the type, unknown types fell to a `default`, and there
was no compile-time guarantee the dispatch table was complete. After this change:

- Each command's required fields are part of its payload type.
- `CommandHandler.handle` is exhaustive — adding a `RequestType` case without a
  handler branch is a **compile error**.
- No handler reads fields off an unrelated command's payload.

## Wire compatibility

**The JSON on the wire is unchanged** — same `type` discriminator strings
(`RequestType` is kept as the single source of those strings) and same field
names. Verified against the TS iOS request builders under
`src/features/observe/ios/` and the shared `SharedTextDelegate`.

Genuinely-required wire fields (tap `x/y`, gesture coords, `set_text` text,
`ime_action`/`keyboard`/`press_button`/`clipboard`/`action` action,
`launch_app` bundleId, `rotate` orientation, `set_preference` key/valueType,
`remove_preference` key, `set_network_mock_rules` rules) are now **non-optional**
and enforced at the decode boundary — a malformed command that omits one is
rejected during `JSONDecoder().decode(...)` and `WebSocketServer` returns a
`requestId`-correlated error, exactly as before. This is safe because the TS
client correlates responses purely by `requestId`
(`src/utils/RequestManager.ts`; the response `type` is log-only) and every one
of those fields is always sent by the corresponding TS builder.

Fields with handler-side defaults, and fields whose absence returns a
**command-specific typed error response** (`add_highlight` shape, storage/
database identifiers), stay optional with their in-handler guards intact, so
those response contracts are preserved byte-for-byte.

## Scope

- `Models.swift` — flat `WebSocketRequest` struct replaced by the typed enum +
  per-command payload structs + a custom `Decodable init(from:)` keyed on
  `RequestType` (unknown type → `DecodingError`). Adds `requestType` /
  `typeString` / `requestId` accessors.
- `CommandHandler.swift` — `handle` is an exhaustive `switch`; each `handle*`
  takes its typed payload; dead optional-bag unwraps removed. Runtime guards
  (nil storage inspector / DB client, SDK bundle match, highlight shape) kept.
- `WebSocketServer.swift` — logging/perf keys use `request.typeString`.
- Tests migrated to the enum API; adds `TypedRequestDecodeTests.swift`.

## Reviewer notes

- Payload structs use `var … = nil` on optionals so they get **both** a
  synthesized `Decodable` (which decodes `var`-with-default) **and** a memberwise
  initializer with defaults for test construction. (`let … = nil` would be
  excluded from decoding.)
- `request_hierarchy` and `request_hierarchy_if_stale` share `RequestHierarchy`;
  `request_two_finger_swipe` and `request_multi_finger_swipe` share
  `RequestMultiFingerSwipe` and route to the same handler (default 2 fingers) —
  behaviorally identical to before.
- Unknown-command handling moved from a `default` branch to a decode rejection;
  `testUnknownCommandFailsToDecode` covers it.
- `CommandError.unknownCommand` is now unreferenced — but it was already
  unconstructed on `main` (pre-existing dead case), so it's left out of scope.

## Testing

- New `Tests/CtrlProxyTests/TypedRequestDecodeTests.swift` — one JSON round-trip
  decode+dispatch test per `RequestType` case (the whole inbound command
  surface), a discriminator-coverage test iterating `RequestType.allCases`, and
  decode-rejection tests for the newly-required fields.
- `swift build` — clean, no warnings.
- `swift test` — **290 tests pass** (~68 ms), up from the 246-test baseline.
- `swiftformat --lint` / `swiftlint` — clean on changed files (only a
  pre-existing `ok` identifier-name warning in unchanged code).

### DIFF
--- a/ios/control-proxy/CtrlProxy.xcodeproj/project.pbxproj
+++ b/ios/control-proxy/CtrlProxy.xcodeproj/project.pbxproj
@@ -11,6 +11,7 @@
 		0A4F48C4211E43B829855704 /* SdkHierarchyClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = E7C65BE687554E71E587B731 /* SdkHierarchyClient.swift */; };
 		0DCD8748EF0E74AD806385A5 /* SdkHierarchyCache.swift in Sources */ = {isa = PBXBuildFile; fileRef = 043D2CA5107B76077BA6301C /* SdkHierarchyCache.swift */; };
 		0E8A9A62BEF9C4E672D09764 /* PerfProviderTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = DEF454F64F7AF11BD79DB97D /* PerfProviderTests.swift */; };
+		0F29D59ABFB52692F62B434F /* TypedRequestDecodeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = EA84FDE08DAFEAD66F399887 /* TypedRequestDecodeTests.swift */; };
 		12D60B8E4D067E91189066AD /* CommandHandlerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 25C10BFA5B6504892F8459D7 /* CommandHandlerTests.swift */; };
 		141E5CD4FAF4509F538E736C /* OSLogReader.swift in Sources */ = {isa = PBXBuildFile; fileRef = DBA97A6C3A1240A055D9876C /* OSLogReader.swift */; };
 		158C6F3B528C02BE92AA6132 /* HierarchyMerger.swift in Sources */ = {isa = PBXBuildFile; fileRef = FAAF8CEA9B9A28B052E57E0A /* HierarchyMerger.swift */; };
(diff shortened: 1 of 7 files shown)


kaeawc:
## Multi-perspective review pass

Ran three independent reviews (Codable-correctness, wire-drift, architecture/completeness). Consensus: wire-compatible, no blockers. One real regression + two cleanups fixed in `8ff32ed`:

- **Regression fixed** — an unknown command `type` was rejected with a generic `DecodingError` (`localizedDescription` = *"The data couldn't be read…"*), which no longer matched the TS client's `rewriteUnknownCommandError` regex `/^Unknown command type: (.+)$/`, silently dropping the *"runner is older than the daemon, rebuild it"* diagnostic. Now throws `CommandError.unknownCommand` with the exact `"Unknown command type: <type>"` text (also revives that previously-dead case).
- **Dead code removed** — the `defaultFingers` parameter on `handleMultiFingerSwipe` (leftover from the removed two-method split).
- **Test rigor** — decode-rejection tests now assert `DecodingError.keyNotFound(<field>)` for the specific missing field, so they prove the required-field contract rather than passing on any thrown error.

### Consciously deferred (out of this refactor's scope)
- **Cross-language drift tripwire** — there's no test locking the TS-emitted iOS command set against the Swift `RequestType` set. Worth adding (the Android/#2835 sibling has one). Filed as a follow-up; it will also surface a **pre-existing** latent bug: `CtrlProxyVoiceOver.ts:98` sends `request_voiceover_action`, which has never existed in `RequestType` (unchanged by this PR — the old flat struct rejected it too).
- **Direction cleanups** (optional) — the `requestId`/`requestType` accessors and `responseType(for:)` are mechanical per-command switches; a shared payload protocol + an enum-derived `responseType` would make them compiler-enforced. Non-blocking; left as potential follow-up.

Verification: `swift build` clean, `swift test` green (290 tests), `swiftformat --lint` + `swiftlint` clean on changed files.

PR #2856, opened 2026-07-02
fix(ios): honestly report DND unsupported on iOS 18+ simulators (#2545)

### LINKED ISSUE #2545: fix(ios): setDeviceState binary Do Not Disturb does not verify on iOS 18.6 simulator (notifyutil read-back unchanged)
## Summary

On an iOS 18.6 simulator, `setDeviceState({ doNotDisturb: { enabled: true } })` posts the DND notification but the **binary toggle never verifies** — `notifyutil` reads back the unchanged state (`0`), so the tool returns `success: false` / `verified: false`. The `getDeviceState` read path works (reads `rawValue: "0"`), but the write/toggle does not take effect on iOS 18.6. The #2508 honest-contract reporting is correct (it refuses to claim success), but binary DND set is effectively **non-functional** on this iOS version.

## Repro

```jsonc
getDeviceState({ platform: "ios", include: ["doNotDisturb"] })
// → { doNotDisturb: { supported: true, capability: "binary", enabled: false, mode: "off",
//      rawValue: "0", method: "ios_simulator_notifyutil", bestEffort: true } }   // read works

setDeviceState({ platform: "ios", doNotDisturb: { enabled: true } })
// → { success: false, doNotDisturb: { ..., requestedMode: "none", verified: false,
//      warning: "iOS simulator Do Not Disturb notification was posted, but the binary DND
//                toggle did not verify: notifyutil did not read back the requested state." } }

setDeviceState({ platform: "ios", doNotDisturb: { mode: "priority" } })
// → { success: false, ..., warning: "iOS DND is binary on the simulator; requested \"priority\"
//      has no per-mode/priority/alarms fidelity (no public Focus API), and the binary DND toggle
//      did not verify: notifyutil did not read back the requested state." }
```

## Impact

- `setDeviceState` DND cannot actually enable DND on iOS 18.6 simulators — every enable attempt reports failure (correctly, but the feature doesn't work).
- The read path (`getDeviceState`, `notifyutil -g`) works, so the regression is specifically the write/toggle: the posted notification (`method: ios_simulator_notifyutil`) does not flip the `notifyutil` key that the verify step reads back. Likely the DND `notifyutil`/Darwin-notification key or simctl mechanism changed in iOS 18.x.

## Suggested investigation / fix

- Confirm the exact `notifyutil` key written vs read on iOS 18.6 (e.g. `com.apple.donotdisturb.*` / `com.apple.springboard.*`) — they may have diverged, so the post doesn't affect the key the verify reads.
- Check whether iOS 18.x requires a different simctl/Darwin mechanism to toggle the binary DND/Focus state on the simulator.
- The honest-contract reporting is good and should stay; the goal is to make the binary toggle actually verify on currently-supported simulators.

## Environment

- macOS (darwin 24.6.0, arm64), Xcode 26.3, iOS 18.6 simulator (iPhone 16 Pro); daemon built from `main` (0.0.37). `method: ios_simulator_notifyutil`.

kaeawc:
## What

On an iOS **18+** simulator, both `setDeviceState` and `getDeviceState` Do Not Disturb now return an honest **`capability: "unsupported"`** result with a precise reason and issue **no** notifyutil command, instead of posting/reading a `donotdisturbd`-owned key that has no effect and reporting a misleading result.

iOS ≤17 simulators keep the existing legacy `notifyutil` best-effort path unchanged. Physical iOS devices keep their existing unsupported early-return.

Closes #2545.

## Why

Reverse-engineered on a real iOS 18.6 simulator (`xcrun simctl`, iPhone 16 Pro):

1. **`donotdisturbd` owns and resets the legacy key.** `notifyutil -s com.apple.donotdisturb.enabled 1 -p` then a fresh-process `-g` reads back `0` **immediately** — the daemon pins the key to its authoritative value. (Control: an *unmanaged* notify key, `com.apple.BiometricKit.enrollmentChanged`, set the exact same way persists at `1` for 8s+, proving notify(3) state itself persists cross-process — the DND revert is daemon reclamation, not registration lifetime.) The same-invocation `-g` in the old combined command sometimes read `1` before the daemon reclaimed, which is exactly the flaky `verified` #2545 saw.
2. **The key is non-authoritative on iOS 18.** DND / Focus moved to the **`com.apple.donotdisturbd`** daemon (private Core Data store + Focus-assertion model). The legacy `com.apple.donotdisturb.enabled` notification neither reflects nor controls real Focus state, and Apple exposes **no public setter or reader**.

So the binary toggle cannot verify — and a read is a confident falsehood (always `0`) — on iOS 18+ through any public/stable mechanism. In keeping with the honest-contract reporting from #2508, the fix reports this precisely rather than posting/reading a dead key.

## How

- New util [`src/utils/ios-cmdline-tools/iosVersion.ts`](src/utils/ios-cmdline-tools/iosVersion.ts): `parseIosMajorVersion` and `iosMajorVersionFromSimctlListDevices` (shape-guarded — a valid JSON scalar yields `null`, never throws).
- [`DeviceState`](src/features/utility/DeviceState.ts): `resolveSimulatorIosMajorVersion` (prefers `BootedDevice.iosVersion`/`osVersion`, falls back to a live `simctl list devices <udid> --json` probe) gates **both** `setIosDoNotDisturb` and `getIosDoNotDisturb`. On iOS major > 17 they return `IOS18_SIM_DND_UNSUPPORTED_ERROR` with no notifyutil command.

## Not broadened (verified)

`BiometricAuth.ts` uses the same `notifyutil -1 -s -g -p` pattern for Touch ID enrollment. Verified on the iOS 18.6 sim that its key **persists** (no daemon reclaims it), so biometrics are **not** affected — intentionally left out of scope.

## Follow-up (out of scope here)

Behavior-based detection (write → independent-process readback) could make the version gate a fast-path hint rather than the sole signal, and would be more future-proof if Apple ever restores a lever. Deferred; the version gate is correct for all shipping runtimes today.

## Testing

- `test/utils/ios-cmdline-tools/iosVersion.test.ts` — version-helper units incl. JSON-scalar/`null`-literal (no-throw) and multi-runtime resolution.
- `test/features/utility/DeviceState.test.ts` — iOS 18+ read **and** write unsupported (asserting **no** notifyutil command), iOS 17 legacy read/write, iOS 26 end-to-end unsupported, `osVersion`-only fallback, live `simctl list` resolution, unknown-version → legacy fallback.
- Full suite: **5658 pass, 0 fail** (`bun test`). `bun run lint` clean, `bun run build` OK. Each unit test < 100ms.
- **Verified end-to-end on a real iOS 18.6 simulator** (incl. the live-version fallback with no `iosVersion` on the device).

## Reviewed

Three independent review passes (iOS-platform / regression / product perspectives), all findings addressed in the second commit: corrected the mechanism claim, gated the read path, hardened the version parser, and updated `docs/design-docs/plat/ios/simctl.md`.

### DIFF
--- a/docs/design-docs/plat/ios/simctl.md
+++ b/docs/design-docs/plat/ios/simctl.md
@@ -286,13 +286,15 @@ per-package resilience of iOS app-data restore.
 
 ## Do Not Disturb (setDeviceState)
 
-<kbd>✅ Implemented</kbd> <kbd>📱 Simulator Only</kbd>
+<kbd>✅ Implemented</kbd> <kbd>📱 Simulator (iOS ≤17) Only</kbd>
 
 The `setDeviceState` / `getDeviceState` MCP tools control Do Not Disturb. On iOS,
 DND is **simulator-only and binary** (on/off) — this is a hard platform
-limitation, not a missing wiring detail.
+limitation, not a missing wiring detail. **On iOS 18+ simulators it is not
+available at all** (see below): DND moved to the private `donotdisturbd` Focus
+daemon, so both read and write report `capability: "unsupported"`.
 
-### How it works
+### How it works (iOS ≤17 simulators)
 
 1. **Binary toggle** — the only lever the simulator exposes is the
    `com.apple.donotdisturb.enabled` Darwin notification, driven via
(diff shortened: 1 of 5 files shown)


kaeawc:
### Follow-up work filed

Deferred items and one un-closed review thread from this PR are now tracked:

- #2861 — Replace the iOS-version gate with **behavior-based** DND detection (write → independent-process readback). Deferred here by design; version gate is correct for all shipping runtimes today.
- #2862 — **Verify the legacy `notifyutil` DND path actually works on iOS ≤17** simulators. `donotdisturbd` has existed since iOS 15, so the preserved `≤17` path may itself be non-functional — it couldn't be tested (no iOS ≤17 runtime was installed). May require lowering the unsupported boundary.
- #2864 — Verify whether the `notifyutil -1` registration is truly required for `notify_set_state`. Review evidence (a biometric key set without `-1` persisted) suggests the shared helper's comment overstates it; only the DND error string was corrected here.

Issue #2760, opened 2026-07-01
Output reduction 5: compact/TOON observe encoding (--observe-result-compact)


kaeawc:
Part of the MCP output-context reduction effort. See the dependency graph across issues in this series.

## Goal
Experimental compact encoding of observe output behind `--observe-result-compact` — no-indent JSON for the nested tree plus TOON (Token-Oriented Object Notation) for the uniform element arrays.

## Rationale / scoping
TOON writes a uniform array of objects as a header row + data rows and uses indentation instead of braces — 30–60% fewer tokens **on uniform tabular data**, much less on ragged/nested data. Applied here:
- **`elements.clickable / text / media`** — near-uniform arrays → TOON's sweet spot. Encode these as TOON.
- **`viewHierarchy.hierarchy`** — ragged, variable-depth, heterogeneous tree → little TOON benefit and the **highest escaping risk** (node `text`/`content-desc` values contain arbitrary commas, quotes, newlines). Keep as compact (no-indent) JSON, not TOON.
- Include a one-line legend so the format is self-describing to the model.

## Dependency coupling
Only meaningful **after** `structuredContent` gating (#2759): `structuredContent` must stay valid JSON, so compression only pays off on the text block once the JSON duplicate is gone.

## Scope
- `ObserveResultOutput.ts` — compact/TOON encoder path, gated by `--observe-result-compact`.
- A TOON writer util with robust escaping.

## Acceptance
- Round-trip / escaping tests on adversarial text values (commas, quotes, newlines in labels).
- Byte delta measured vs the Issue #2755 fixture.
- Format signed off on first real output before finalizing.

## Dependencies
- Finalize hook (#2758).
- `structuredContent` gating (#2759).

kaeawc:
Closing the loop on this one: I prototyped the TOON approach in #2996 (now closed) and measured it against the compact-bounds work that landed in #2976, and **compact-bounds is the better option — TOON doesn't earn its keep here.**

### Measurements (on the #2755 android-home fixture)

`finalizeToolResponse` always runs `sanitizeObserveResult` first, so the honest baseline is *sanitized-pretty*, not raw pretty:

| representation | tokens (cl100k) |
|---|---|
| raw pretty | 21,911 |
| sanitized-pretty (real ship baseline) | 13,789 |
| compact JSON only (de-indent) | ~8,755 |
| compact JSON + TOON (this issue's design) | ~8,549 |

[...]

Issue #2758, opened 2026-07-01
Output reduction 3: finalizeToolResponse hook + lastHierarchy cache repair


kaeawc:
Part of the MCP output-context reduction effort. See the dependency graph across issues in this series.

## Goal
Wire `sanitizeObserveResult` in at the single serialization chokepoint, and repair the dormant session-cache write that the diff baseline depends on.

## Background
Handlers pre-serialize via `createStructuredToolResponse(result)` → an MCP envelope `{ content:[{type:"text",text}], structuredContent }`. The one post-handler return is `src/server/toolRegistry.ts:449`, where both the tool `name` and `sessionUuid` are in scope. An `ObserveResult` appears in exactly two output positions: top-level (`observe`) or `.observation` (action results, attached at `src/features/action/BaseVisualChange.ts:317`).

**Latent bug:** the session-cache write at `src/server/toolRegistry.ts:436` reads `response?.viewHierarchy`, but `response` is the envelope — so `lastHierarchy` is effectively never written. The diff baseline (#2761) needs this to work.

## Scope
- Add `finalizeToolResponse(response, { name, sessionUuid })` called at `toolRegistry.ts:449`:
  - Extract payload: prefer `structuredContent`, else `JSON.parse(content[0].text)` (guarded).
  - Locate the `ObserveResult`: payload itself for `observe`, else `payload.observation`.
  - Apply `sanitizeObserveResult`.
  - Rewrite **both** `structuredContent` and `content[0].text` (via `stringifyToolResponse`).
- Repair the `lastHierarchy` cache write to read from `structuredContent`.

## Acceptance
- Integration test: an `observe` response and a `tapOn` response are both sanitized (text + structuredContent).
- `lastHierarchy` is actually populated after an `observe`.

## Dependencies
- Config plumbing (#2756).
- `sanitizeObserveResult` module (#2757).

kaeawc:
Dependency #2757 (`sanitizeObserveResult`) is implemented in PR #2872 and ready to wire.

Exact export to import at the `toolRegistry.ts:449` chokepoint:

```ts
import { sanitizeObserveResult, SanitizeObserveConfig } from "../features/observe/output/ObserveResultOutput";
// sanitizeObserveResult(obs: ObserveResult, cfg: { dropElements: boolean; trimNodes?: boolean }): ObserveResult
```

Notes for the wiring:
[...]

Issue #2756, opened 2026-07-01
Output reduction 1: config plumbing for output-reduction flags (CLI + env)


kaeawc:
Part of the MCP output-context reduction effort. See the dependency graph across issues in this series.

## Goal
Add the config plumbing for all new output-reduction flags so each parses from CLI **and** an `AUTOMOBILE_*` env var, routes through the existing feature-flag pipeline, and is readable anywhere via `serverConfig`.

## Flags introduced (env equivalents, CLI wins via `||`)
| Flag | ENV | Default |
|---|---|---|
| `--observe-result-drop-elements` | `AUTOMOBILE_OBSERVE_RESULT_DROP_ELEMENTS` | off (elements kept) |
| `--observe-result-compact` | `AUTOMOBILE_OBSERVE_RESULT_COMPACT` | off |
| `--tool-results-no-structured-content` | `AUTOMOBILE_TOOL_RESULTS_NO_STRUCTURED_CONTENT` | off |
| `--actions-diff-observe` | `AUTOMOBILE_ACTIONS_DIFF_OBSERVE` | off |
| `--actions-no-observe` | `AUTOMOBILE_ACTIONS_NO_OBSERVE` | off |

(The per-node trim and perf-audit strip are default-on behaviors, not flags — see their issues.)

## Files / approach
Follow the exact existing pattern (e.g. `predictive-ui`):
- `src/features/featureFlags/FeatureFlagDefinitions.ts` — add `FeatureFlagKey`s + `FEATURE_FLAG_DEFINITIONS` entries.
- `src/features/featureFlags/FeatureFlagApplier.ts` — `apply()` cases calling `serverConfig` setters.
- `src/utils/ServerConfig.ts` — getter/setter pairs (leaf singleton; no circular-import risk).
- `src/index.ts` — parse (`args.includes("--x") || process.env.AUTOMOBILE_X === "1"`), add to `parseArgs` return, and add rows to the `cliOverrides` table + startup log.

## Acceptance
- Each flag flips its `serverConfig.isX()` from both a CLI flag and its env var.
- Unit test covering CLI-takes-precedence-over-env parse.

## Dependencies
None (foundation).

Issue #2755, opened 2026-07-01
Output reduction 0: measurement harness (fixture + byte-breakdown script)


kaeawc:
Part of the MCP output-context reduction effort. See the dependency graph across issues in this series.

## Goal
Establish a repeatable before/after measurement so every later reduction can be quantified against a fixed baseline.

## Context
A single `observe` on the Android home screen currently returns **66,560 characters** — which *exceeds the MCP tool-output token cap*, so the harness spills it to a file instead of returning it inline. Measured byte breakdown of that sample:

| Field | Bytes | % |
|---|---|---|
| `elements` | 24,626 | 37% |
| `performanceAudit` | 20,640 | 31% |
| `viewHierarchy.hierarchy` | 17,403 | 27% |
| `perfTiming` | 2,452 | 4% |
| everything else | ~1,400 | 2% |

Within `performanceAudit`, the raw gfxinfo dump appears **twice**: `metrics.gfxinfoRaw` (9,449 B) and again embedded in the `diagnostics` string (9,737 B).

## Scope
- Commit the captured 66,560-char home-screen observe result as a test fixture (e.g. `test/fixtures/observe/android-home-66k.json`).
- Add a small script (bash + `jq`, per repo convention under `scripts/`) that prints the byte-breakdown by top-level field and by `viewHierarchy` sub-key.

## Acceptance
- Script emits per-field byte counts for any observe-result JSON passed to it.
- Fixture is importable by unit tests in later issues.

## Dependencies
None (foundation).

PR #2872, opened 2026-07-03
feat(observe): sanitizeObserveResult output transform (#2757)

### LINKED ISSUE #2757: Output reduction 2: sanitizeObserveResult (perf-strip + node-trim + elements-drop)

## Goal
A pure, output-only transform that shrinks a single `ObserveResult` for serialization — perf-audit strip, per-node trim (default on), and elements-drop (gated).

## Guiding principle
**Output-only:** operates on a copy destined for the wire, never the in-memory `ObserveResult`. Internal consumers (`BaseVisualChange`, `summarizeFailureObservation`, action selection, predictions) must be unaffected.

## Scope — new `src/features/observe/output/ObserveResultOutput.ts`
`sanitizeObserveResult(obs, cfg)` applying, in order:
1. **Perf-audit strip** (always, when `--debug-perf` populated it): null `metrics.gfxinfoRaw` and `metrics.cpuStatsRaw`; truncate `diagnostics` to the summary lines above `--- GFXINFO DUMP ---`. Keep computed metrics (`p50–p99`, `jankCount`, `cpuUsagePercent`, `threadCount`, `touchLatencyMs`, `anrDetected`) and `violations[]`. Source: `src/features/performance/PerformanceAudit.ts` (`gfxinfoRaw` L46/151, `cpuStatsRaw` L47/152, `diagnostics` L57). ~20k → ~1k when debug-perf on.
2. **Per-node trim** (default on) over `viewHierarchy.hierarchy`: drop `view-id` when `=== resource-id`; omit `"false"`/default booleans; omit empty-string fields. Lossless for decision-making; ~20–40% off the tree. Must **not** touch `rawViewHierarchy` (raw stays raw).
3. **Elements-drop** (gated by `--observe-result-drop-elements`): omit the `elements` block.

## Acceptance
- Pure-function unit tests over the Issue #2755 fixture asserting byte reduction per step; each <100ms.
- Assert the input `ObserveResult` object is not mutated (deep-clone boundary).

## Dependencies
- Config plumbing (#2756) for the gates.
- Measurement fixture (#2755) for tests.

kaeawc:
## What

Adds `sanitizeObserveResult(obs, cfg)` — a **pure, output-only** transform (new `src/features/observe/output/ObserveResultOutput.ts`) that shrinks a single `ObserveResult` for serialization. It applies, in order:

1. **Perf-audit strip** (always, when `performanceAudit` is populated): nulls `metrics.gfxinfoRaw` and `metrics.cpuStatsRaw`, and truncates `diagnostics` to the summary lines above `--- GFXINFO DUMP ---`. Computed metrics (`p50–p99`, `jankCount`, `cpuUsagePercent`, `threadCount`, `touchLatencyMs`, `anrDetected`) and `violations[]` are preserved.
2. **Per-node trim** (default on) over `viewHierarchy.hierarchy`: drops `view-id` when it equals `resource-id`, omits default-false booleans, and omits empty-string fields. **Never touches `rawViewHierarchy`** (raw stays raw).
3. **Elements-drop** (gated by `cfg.dropElements`, i.e. `--observe-result-drop-elements`): omits the flattened `elements` block.

Closes #2757.

## Why

Part of the MCP output-context reduction effort (milestone: Output Size Reduction). An observe result exceeds the MCP tool-output token cap; the heaviest offenders are the perf-audit raw dumps (~17k bytes, gfxinfo inlined twice), the duplicated `elements` tree, and redundant per-node fields. This transform removes that waste **on a copy destined for the wire only** — the in-memory `ObserveResult` is left untouched so internal consumers (`BaseVisualChange`, `summarizeFailureObservation`, action selection, predictions) are unaffected.

Dependencies #2755 (measurement fixture) and #2756 (config plumbing) have already landed.

## How

- **Deep-clone boundary:** clones via `JSON.parse(JSON.stringify(obs))` — matching the repo's existing hierarchy-cloning convention (`ViewHierarchy.ts`) and the wire's own JSON semantics, so it can never throw on a value `structuredClone` would reject.
- **Hierarchy shape variance:** real captures put an **array** in `hierarchy.node` (typed as a single node) as well as on child `node[]`; a `toNodeArray` normalizer collapses both into one traversal path.
- **Default-false booleans:** restricted to a known attribute set (`clickable`, `long-clickable`, `focusable`, `focused`, `scrollable`, `checkable`, `checked`, `selected`, `password`) so a legit text field holding `"false"` is never dropped. `enabled` (default *true*) is intentionally excluded to stay lossless.
- `cfg.trimNodes` defaults on; pass `false` to emit the untrimmed hierarchy.

## Scope note (deliberately not wired in yet)

Per the issue, this PR delivers **only** the pure function + unit tests. Wiring it into the tool-output boundary is the downstream issue **#2758** (`finalizeToolResponse` hook), and the compact/TOON encoding is **#2760** — both intentionally out of scope here to avoid overlapping those PRs.

## Testing

- New `test/features/observe/output/ObserveResultOutput.test.ts` (18 tests, all < 100ms), measured against the frozen #2755 baseline fixture with the production formatter (`stringifyToolResponse`):
  - **Purity:** input `ObserveResult` deep-equal before/after; distinct output reference.
  - **Perf-strip:** raws nulled, diagnostics truncated at the marker, computed metrics + `violations` preserved, ≥10k byte reduction, no-op when audit absent.
  - **Node-trim:** no node has `view-id === resource-id`; synthetic default-false booleans + empty strings dropped; meaningful values survive; `rawViewHierarchy` byte-for-byte unchanged; measurable byte reduction; `trimNodes:false` disables it.
  - **Elements-drop:** gated both ways; measurable byte reduction when on.
  - **Combined:** all three steps shrink the baseline to < 70% of original bytes.
- `bun run lint` clean, `bun run build` OK on the new files.
- Note: one pre-existing unrelated failure (`AndroidCtrlProxyClient > getLatestHierarchy > should preserve SDK screen names…`) reproduces on clean `main` without this change — not introduced here.

## Follow-up (out of scope)

- Wire `sanitizeObserveResult` into the observe/action output path under #2758.

### DIFF
--- a/src/features/observe/output/ObserveResultOutput.ts
+++ b/src/features/observe/output/ObserveResultOutput.ts
@@ -10,6 +10,14 @@ import type { ViewHierarchyNode } from "../../../models/ViewHierarchyResult";
  * `ObserveResult`. Internal consumers (`BaseVisualChange`,
  * `summarizeFailureObservation`, action selection, predictions) read the
  * original object and must be unaffected, so this is a pure function.
+ *
+ * Distinct concern from `ViewHierarchy.cleanNodeProperties`: that is an
+ * allow-list normalizer applied during capture (it *drops* any attribute not
+ * on its list, e.g. `className`). This is a lossless deny-list shrink of the
+ * already-materialized output tree — it only removes redundant/default fields
+ * and must preserve everything else the result carries. The two share the
+ * empty-string / default-boolean / `enabled`-default-true rules on purpose so
+ * their trimmed output agrees; they are not interchangeable.
  */
 
 /** Marker that separates the human-readable perf summary from the raw dump. */
@@ -21,6 +29,10 @@ export const GFXINFO_DUMP_MARKER = "--- GFXINFO DUMP ---";
  * `"false"` / `false` value is lossless for decision-making. Restricting the
  * drop to this known set (rather than any falsy value) avoids nuking a legit
  * text field that happens to hold the string `"false"`.
+ *
+ * `enabled` is deliberately absent: its default is *true*, so it is handled
+ * separately (dropped when true, kept when `"false"`) — the same rule the
+ * repo's canonical node cleaner uses (`ViewHierarchy.cleanNodeProperties`).
  */
 const DEFAULT_FALSE_BOOLEAN_ATTRS: ReadonlySet<string> = new Set([
   "clickable",
@@ -36,8 +48,10 @@ const DEFAULT_FALSE_BOOLEAN_ATTRS: ReadonlySet<string> = new Set([
 
 export interface SanitizeObserveConfig {
   /**
-   * Gate for elements-drop (`--observe-result-drop-elements`). When true the
-   * flattened `elements` block is omitted from the output copy.
+   * Gate for elements-drop. Maps to the `observeResultDropElements` output-
+   * reduction flag (`--observe-result-drop-elements`); the wiring layer (#2758)
+   * supplies it from `ServerConfig.isObserveResultDropElementsEnabled()`. When
+   * true the flattened `elements` block is omitted from the output copy.
    */
   dropElements: boolean;
   /**
@@ -134,6 +148,13 @@ function trimHierarchyNodes(node: ViewHierarchyNode | undefined): void {
   for (const [key, value] of Object.entries(attrs)) {
     // Omit empty-string fields.
     if (value === "") {
+      delete attrs[key];
+      continue;
+    }
+    // Omit `enabled` when true: it defaults to true, so absence reads as
+    // enabled by convention. A disabled control keeps its explicit
+    // `enabled: "false"`. Mirrors ViewHierarchy.cleanNodeProperties.
+    if (key === "enabled" && (value === true || value === "true")) {
       delete attrs[key];
       continue;
     }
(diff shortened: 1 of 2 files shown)


kaeawc:
## Multi-perspective review addressed (3 independent passes)

Ran three independent review agents — **regression hunter**, **architecture/direction**, and **test-quality/completeness**. Summary and resolutions:

**Regression hunter:** No confirmed bugs. Verified purity/mutation isolation, traversal shape variance (`toNodeArray`), perf-strip edge cases (marker@0, null diagnostics, undefined metrics), and JSON-clone data safety against 10 adversarial inputs. Confirmed `gfxMetrics`/`perfTiming` are correctly out of scope per the issue.

**Architecture:** Flagged that node-trim shares rules with `ViewHierarchy.cleanNodeProperties`. Resolved — verified that cleaner is an **allow-list** normalizer that would drop `className` and other real output fields, so it is a *different concern* from this lossless **deny-list** output shrink; added a doc comment making the distinction explicit rather than force-merging them. Kept the narrow `SanitizeObserveConfig` (per CLAUDE.md narrow-interface guidance) and documented the `dropElements → observeResultDropElements` flag mapping for the #2758 wiring.

**Test-quality:** Two substantive gaps fixed:
- **`enabled` handling** — now drops `enabled="true"` (default-true) and preserves `enabled="false"`, matching the repo convention (`ViewHierarchy.ts:672,694`); previously left untouched (bloat). New test asserts both directions.
- **Token reduction** — the issue's real cap metric is tokens, not bytes; added a token-count reduction assertion.
- Added a synthetic multi-node byte-reduction test exercising the false-boolean/empty-string/enabled paths that the frozen fixture cannot (the extractor emits those already-clean).

Tests: **21 pass, 0 fail**, each <100ms. Lint + build clean.

Issue #2790, opened 2026-07-01
fix(db): non-transactional get-or-create / read-modify-write races across coverage, navigation, prediction, and appearance repositories


kaeawc:
## Summary
Four DB repositories implement get-or-create / read-modify-write (RMW) as a `SELECT` followed by a separate `INSERT`/`UPDATE` across two `await`s with no surrounding transaction. Because the single `bun:sqlite` connection is guarded by an async mutex that serializes individual queries but *releases across `await` boundaries*, two overlapping callers can both read "not found" (or read the same counter) and then race — one loses. Depending on the table's constraints this either throws `UNIQUE constraint failed` and drops the event (coverage), or silently loses counter increments (prediction/navigation), corrupting downstream metrics. Found via a DB concurrency/migration audit (2026-07-01), finding R2356, priority P1, verified against source.

## Background & history
- **Coverage** (`src/db/testCoverageRepository.ts` `recordNodeVisit`/`recordEdgeTraversal`): introduced in commit `c95e6ecd` (2026-01-08), [#429](https://github.com/kaeawc/auto-mobile/pull/429) — added navigation-graph test coverage tracking. UNIQUE indexes `idx_test_node_coverage_session_node` / `idx_test_edge_coverage_session_edge` created in migration `src/db/migrations/2026_01_08_000_test_coverage.ts:57-64,93-100`.
- **Navigation** (`src/db/navigationRepository.ts` `getOrCreateNode`, `addOrUpdateSuggestion`, and the other `getOrCreate*` helpers): introduced in commit `1ec6a251` (2025-12-31), [#92](https://github.com/kaeawc/auto-mobile/pull/92) — migrate to DB-backed navigation graph. UNIQUE index `idx_navigation_nodes_app_screen` from `src/db/migrations/2025_12_30_001_navigation_graph.ts:34-41`.
- **Prediction** (`src/db/predictionHistoryRepository.ts` `upsertTransitionStats`): introduced in commit `29286c05` (2026-01-04, PR not identified) — added prediction history tracking. UNIQUE index `idx_prediction_transition_key` from `src/db/migrations/2026_01_02_000_prediction_history.ts:79-85`.
- **Appearance** (`src/db/appearanceConfigRepository.ts` `setConfig`): introduced in commit `aca64883` (2026-01-17, PR not identified) — sync device appearance with host.

[...]

Issue #2785, opened 2026-07-01
fix(db): destructive migration auto-recovery (default-on) can DROP every table on a populated database


kaeawc:
## Summary
On daemon startup, `runMigrations()` auto-recovers from "corrupted migrations" errors, and its last-resort step (`resetDatabaseState`) **DROPs every non-`sqlite_%` table** with no row-count, backup, or explicit-opt-in guard. Recovery is **enabled by default** — it only turns off when `AUTOMOBILE_MIGRATION_RECOVERY` is explicitly falsy. Routine developer actions that Kysely treats as corruption (an out-of-order backport migration, or renaming an already-executed migration) therefore silently destroy a populated local database. The audit reproduced full data loss (an `id=42` row wiped) against the repo's own `runMigrations()`. Found via a DB concurrency/migration audit (2026-07-01), finding M1, priority P0, verified against source.

## Background & history
- Introduced: commit `592be5c9` (2026-01-19), [PR #965](https://github.com/kaeawc/auto-mobile/pull/965) ("Handle corrupted migration history on startup", issue #947). The entire recovery path — `isMigrationRecoveryEnabled`, `rebuildMigrationTable`, `resetDatabaseState`, and `recoverCorruptedMigrations` — landed in this single commit.
- All migrations live in `src/db/migrations/` (also seeded by `592be5c9`). The date-prefix scheme has real duplicates today — `2026_01_03_000`, `2026_01_11_000`, and `2026_01_27_000` each appear twice — which makes lexical ordering fragile and out-of-order insertions more likely.
- Why it's shaped this way: the goal was to make the singleton daemon self-heal from a broken local migration history on startup rather than crash-loop. That is reasonable for an empty/throwaway dev DB, but the reset step never distinguished "empty scratch DB" from "populated DB with user data."

## The problem
Kysely's `migrateToLatest()` throws a `corrupted migrations` error whenever the on-disk migration set diverges from the recorded history — most commonly when a migration whose name sorts **before** an already-executed migration appears (out-of-order backport), or when an executed migration is **renamed** (`allowUnorderedMigrations` defaults to `false`). Both are routine developer actions.

`runMigrations()` catches that error and, by default, enters recovery:

```ts
// src/db/migrator.ts:32
function isMigrationRecoveryEnabled(): boolean {
  const envValue =
    process.env.AUTOMOBILE_MIGRATION_RECOVERY ?? process.env.AUTO_MOBILE_MIGRATION_RECOVERY;
[...]

Issue #2791, opened 2026-07-01
fix(db): make NavigationGraphManager.recordNavigationEvent graph write transactional


kaeawc:
## Summary
`NavigationGraphManager.recordNavigationEvent` writes node + node-visit + modals + edge + edge-traversal + UI elements + edge-modals + scroll position + `touchApp` as a sequence of separately-awaited statements with **no enclosing transaction**. If any statement mid-sequence throws (e.g. a UNIQUE conflict per findings R2/R3) or the daemon is interrupted, the navigation graph is left **partially applied** — a node/visit persists but the corresponding edge/coverage rows are missing, leaving the on-disk graph internally inconsistent. Blast radius: every WebSocket-sourced navigation event during recording/exploration.
Found via a DB concurrency/migration audit (2026-07-01), finding R4, priority P1, verified against source.

## Background & history
- Introduced: commit `1ec6a251` (2025-12-31), [PR #92](https://github.com/kaeawc/auto-mobile/pull/92) — "Migrate to database-backed navigation graph with persistent back stack tracking". This same PR both created `recordNavigationEvent` and the schema in `src/db/migrations/2025_12_30_001_navigation_graph.ts` (nodes/edges/apps tables).
- The method was ported straight from an in-memory graph model where a partial update was harmless (it was all one synchronous mutation). Once the writes became individual awaited SQL round-trips against a persistent store, atomicity was silently lost — nothing in the port wrapped them in a transaction.
- Complements R2/R3 (atomic upserts): those remove the *most likely* mid-sequence throw, but process interruption (crash, daemon shutdown mid-await) can still tear the write, so a transaction is warranted independently.

## The problem
All navigation repositories (`NavigationRepository`, `TestCoverageRepository`) resolve to the **same singleton `getDatabase()`** connection, so a `db.transaction()` can cover all of them — but today none of these writes are transactional. The relevant sequence:

```ts
// src/features/navigation/NavigationGraphManager.ts:319-417
const node = await this.repository.getOrCreateNode(this.currentAppId, screenName, timestamp);
if (this.activeTestSession) {
  await this.testCoverageRepository.recordNodeVisit(this.activeTestSession.id, node.id, timestamp);
}
// ... setNodeModals ...
if (this.currentScreen && this.currentScreen !== screenName) {
  const edge = await this.repository.createEdge(this.currentAppId, this.currentScreen, screenName, toolName, toolArgs, timestamp);
  if (this.activeTestSession) {
[...]

PR #2875, opened 2026-07-03
fix(db): atomic recordFailure upsert + UNIQUE index on failure_groups.signature

### LINKED ISSUE #2789: fix(db): add UNIQUE index on failure_groups.signature and make recordFailure an atomic upsert
## Summary

`FailureAnalyticsRepository.recordFailure()` does a non-atomic get-or-create against `failure_groups` keyed on `signature`: it `SELECT`s the group, then either `UPDATE`s (`total_count + 1`, `unique_sessions`) or `INSERT`s a brand-new group with a fresh `crypto.randomUUID()`. There is no transaction, no `ON CONFLICT`, and the `signature` column has only a **plain** (non-unique) index. Two failures with the same signature that overlap across `await` boundaries either both read "no group" and both `INSERT` (duplicate groups for one signature) or both read `total_count = N` and the later `UPDATE` clobbers (lost count increments). Found via a DB concurrency/migration audit (2026-07-01), finding R1, priority P1, verified against source.

## Background & history

- Introduced: commit `470f69bb` (2026-01-27), "feat: add failures tab debugging features with real-time streaming". PR not identified (commit subject has no `(#NNNN)`). Both the `recordFailure()` get-or-create logic and the `idx_failure_groups_signature` index were added in this single commit; the file was later renamed to `failureAnalyticsRepository.ts`.
- Schema provenance: the `failure_groups` table and its `signature` index were created in `src/db/migrations/2026_01_27_000_failures.ts` (same commit). The index is declared with `.createIndex(...).column("signature")` — **not** `.unique()`, so the DB does not enforce one row per signature.
- Why it looks this way: the get-or-create was written as a straightforward read-then-write, which is correct under a naive "one query at a time" mental model. It predates the recognition that the daemon's single shared connection releases its mutex across `await`s (see below), so interleaving was not considered.

## The problem

The daemon runs a **single `bun:sqlite` connection guarded by a hand-rolled async mutex** (`src/db/bunSqliteDialect.ts`). That mutex serializes each individual query but is **released across every `await`**, so a multi-`await` read-modify-write sequence like `recordFailure()` interleaves with a concurrent call. All of this is inside ONE singleton daemon per uid — this is in-process concurrency, not a cross-process issue.

`recordFailure()` reads the group, then conditionally inserts or updates in separate awaited statements:

```ts
// src/db/failureAnalyticsRepository.ts:136
const existingGroup = await db
  .selectFrom("failure_groups")
  .select(["id", "total_count", "unique_sessions", "first_occurrence", "tool_call_info_json"])
  .where("signature", "=", input.signature)
  .executeTakeFirst();

let groupId: string;

if (existingGroup) {
  groupId = existingGroup.id;
  // ... compute newUniqueSessions, merge tool info ...
  const update: FailureGroupUpdate = {
    last_occurrence: now,
    total_count: existingGroup.total_count + 1,   // read-modify-write on a stale value
    unique_sessions: newUniqueSessions,
    // ...
  };
  await db.updateTable("failure_groups").set(update).where("id", "=", groupId).execute();
} else {
  groupId = crypto.randomUUID();                  // fresh id per caller
  const group: NewFailureGroup = { id: groupId, signature: input.signature, total_count: 1, /* ... */ };
  await db.insertInto("failure_groups").values(group).execute();  // no ON CONFLICT
}
```

Two interleavings, both reachable when two failures share a signature (e.g. an Android and an iOS failure recorded near-simultaneously, or a crash-loop burst emitting the same signature repeatedly):

1. **Duplicate groups.** Both calls `await` the `SELECT`, both see no existing group, both fall into the `else` branch, each generates its own `groupId`, and both `INSERT`. Because the index is not `UNIQUE`, the DB accepts both rows. The signature now maps to two group rows; subsequent reads (`getFailureGroups`, breakdowns) split occurrences and counts across the duplicates, and which duplicate a new occurrence attaches to is nondeterministic.
2. **Lost increment.** Both calls read `existingGroup.total_count = N`. Each computes `N + 1` and writes it. The second `UPDATE` overwrites the first with the same `N + 1`, so two occurrences produce `total_count = N + 1` instead of `N + 2`. The same last-writer-wins loss applies to `unique_sessions`.

The `failure_occurrences` insert at line 219 is unaffected (its `id` is a per-call UUID), so occurrence rows are never lost — only the aggregated group counts drift and duplicate groups accumulate.

## Impact

- **Correctness of the failures dashboard.** `total_count` / `unique_sessions` under-report, and duplicate groups fragment a single signature into multiple cards, so the streaming failures tab shows wrong aggregates. Data is misleading, not lost (occurrences remain).
- **Narrowness.** Requires two same-signature `recordFailure()` calls whose `await` windows overlap. Most failure recording is sporadic, so the common case is fine; the window opens during bursts — crash loops, an ANR storm, or simultaneous multi-device test runs emitting identical signatures. That is precisely when failure analytics matter most.
- No crash, no daemon corruption; severity is data-quality, which is why this is P1 rather than P0.

## Proposed fix

Two coordinated changes:

1. **New migration** adding a `UNIQUE` index on `failure_groups.signature` (replacing or supplementing the plain one). The migration MUST de-dupe any existing duplicate `signature` rows first, or the `CREATE UNIQUE INDEX` will fail on databases that already accumulated duplicates. A safe approach: keep the earliest `first_occurrence` row per signature, re-point `failure_occurrences.group_id` and `failure_notifications.group_id` from the losers to the keeper, sum `total_count`/`unique_sessions` onto the keeper, delete the losers, then drop `idx_failure_groups_signature` and create it `.unique()`. Add as a new file (e.g. `src/db/migrations/2026_07_01_000_failure_groups_signature_unique.ts`); `FileMigrationProvider` in `src/db/migrator.ts` picks up files by name order automatically.

   **Ordering is MUST-ordered, not incidental (per review — Mira, Sam).** The migration connection runs with `PRAGMA foreign_keys = ON` (`configureSqliteDatabase`, `src/db/database.ts:47`), and SQLite migrations here are **not** wrapped in a DDL transaction (`SqliteAdapter.supportsTransactionalDdl === false`). `failure_occurrences.group_id` is a real CASCADE FK (`src/db/migrations/2026_01_27_000_failures.ts:66-67` → `.references("failure_groups.id").onDelete("cascade")`), so you MUST `UPDATE failure_occurrences SET group_id = <keeper> WHERE group_id IN (<losers>)` **before** `DELETE FROM failure_groups WHERE id IN (<losers>)` — delete first and CASCADE wipes the very occurrences you meant to keep (data *loss*, not just data-quality drift). Repoint occurrences via `UPDATE` (do not delete+reinsert) so their children `failure_captures` / `failure_occurrence_screens` (which FK to `failure_occurrences.id`) are unaffected.

   `failure_notifications.group_id` (`:168`) is **NOT a foreign key** — it is a plain `text notNull()` column with no `.references(...)`. So repointing it is a data-integrity nicety, not FK-driven: cascade will neither clean it nor stop you. If you forget to repoint it (or repoint in the wrong order), you silently strand notifications pointing at a deleted group id — nothing throws.

   **Deterministic keeper tiebreak (per review — Mira).** Keying on `signature` and keeping earliest `first_occurrence` is ambiguous when two losers tie on `first_occurrence` (same-ms burst — the exact scenario this bug targets), which can leave two rows and make `CREATE UNIQUE INDEX` throw. Pick the keeper with a total order, e.g. `ORDER BY first_occurrence ASC, id ASC LIMIT 1` (min(id) tiebreak). Sum `total_count`/`unique_sessions` onto the keeper.

   **Wrap the whole de-dup in one transaction and make the index create idempotent (per review — Sam).** `CREATE UNIQUE INDEX` failing halfway after losers are already deleted leaves the DB worse than found — wrap repoint → sum counts → repoint notifications → delete losers → drop old index → create unique index in a single `db.transaction().execute(...)`. The migration must also survive #2785's destructive-recovery replay (`resetDatabaseState` drops all tables and re-runs every migration on an empty DB): make it idempotent / safe on an empty DB and add `.ifNotExists()` to the new index create, or the replay throws "index already exists" and can tip back into the destructive reset path.

2. **Convert `recordFailure()` to an atomic upsert** so the count bump is a single statement, mirroring `installedAppsRepository.upsertInstalledApp` (`src/db/installedAppsRepository.ts:120`) and the atomic increment in `navigationRepository.updateNodeVisit` (`src/db/navigationRepository.ts:1041`, `eb("visit_count", "+", 1)`):

```ts
// sketch — atomic group upsert keyed on signature
const groupId = crypto.randomUUID(); // used only if this is the first row for the signature
await db
  .insertInto("failure_groups")
  .values({ id: groupId, signature: input.signature, total_count: 1, unique_sessions: 1, /* ...first-seen fields... */ })
  .onConflict(oc =>
    oc.column("signature").doUpdateSet(eb => ({
      last_occurrence: now,
      total_count: eb("failure_groups.total_count", "+", 1),
      updated_at: new Date().toISOString(),
      // tool_call_info merge stays; see note below
    }))
  )
  .execute();

// (per review — Mira, Sam) DROP this redundant follow-up SELECT. `.returning("id")`
// works in this hand-rolled dialect (proven: failureEventRepository.ts:116,
// networkEventRepository.ts:61; bunSqliteDialect hasReturning path :167-175 —
// executeQuery returns rows when the SQL includes "returning"). Because the upsert
// uses DO UPDATE (not DO NOTHING), RETURNING yields the row on BOTH insert and
// conflict paths. Get the id atomically in one statement instead:
//   .onConflict(...).returning("id").executeTakeFirstOrThrow()
// This also removes a real race: the separate SELECT could interleave with
// fire-and-forget retention cleanup (cleanupRetention, :850) deleting the group,
// making executeTakeFirstOrThrow() throw and drop the occurrence insert.
```

`unique_sessions` must stay correct: it should increment only when the session is new for that group. The current "did this session already appear" check reads `failure_occurrences`, which is itself racy. Compute it in a follow-up statement after the occurrence insert (e.g. `UPDATE ... SET unique_sessions = (SELECT COUNT(DISTINCT session_id) FROM failure_occurrences WHERE group_id = ?)`), so it is derived from committed occurrence rows rather than an increment on a stale read. This is genuinely race-free (per review — Mira): the `COUNT(DISTINCT session_id)` is evaluated inside SQLite in one statement over committed rows, so two interleaved `recordFailure` recomputes both write the same correct value (idempotent) — state it as *derived, not incremented*. **Cost caveat (per review — Mira):** there is no index on `failure_occurrences(group_id, session_id)`, so this `COUNT(DISTINCT)` scans all occurrences for the group each call — an O(rows-in-group) scan per event under the burst workload this fix targets. Either accept it (bounded by retention) or add the composite index in the same migration. The `mergeToolCallInfo` logic can be preserved by reading the existing group inside the same resolve step and folding it into the `doUpdateSet`, or by a small follow-up update — do not drop tool-call-info merging. (Note: table-qualifying the increment `eb("failure_groups.total_count","+",1)` inside `doUpdateSet` is fine; the unqualified form `eb("total_count","+",1)` compiles too — in-repo precedent `navigationRepository.ts:1044` uses the unqualified form.)

Must NOT: break the existing occurrence/screen/capture/notification inserts or the background retention cleanup; change the shape of `getFailureGroups` output; regress the non-conflict (first-failure) path.

- [ ] N concurrent `recordFailure()` calls for a single signature yield **exactly one** `failure_groups` row for that signature with `total_count === N`.
- [ ] N concurrent calls for one signature spanning M distinct `sessionId`s yield `unique_sessions === M` (assert `unique_sessions` is *derived* via `COUNT(DISTINCT session_id)`, not incremented, so this test is meaningful — per review).
- [ ] The new migration runs cleanly on a DB that already contains duplicate-signature groups (de-dupes, re-points occurrences/notifications, then creates the UNIQUE index) — verified by seeding duplicate-signature groups **plus occurrences plus notifications** before migrating.
- [ ] (per review — Mira, Sam) The de-dup migration MUST repoint `failure_occurrences` **before** deleting loser groups — seed occurrences under a loser group and assert they survive re-pointed (delete-first would cascade-wipe them). Explicit orphan check: after migration, **zero** `failure_notifications.group_id` values reference a non-existent `failure_groups.id`.
- [ ] (per review — Mira) Deterministic keeper tiebreak on a `first_occurrence` tie (`ORDER BY first_occurrence ASC, id ASC`) — seed two duplicate rows sharing `first_occurrence` and assert the migration collapses them to one so `CREATE UNIQUE INDEX` succeeds.
- [ ] (per review — Sam) Migration is idempotent / safe on an empty DB and re-runnable under #2785's destructive-recovery replay (new index created with `.ifNotExists()`); whole de-dup runs inside one `db.transaction()`.
- [ ] (per review — Mira, Sam) Add a **dialect test** asserting the upsert's `.onConflict(...).doUpdateSet(...).returning("id")` returns the *pre-existing* group id on the conflict path — this dialect's RETURNING handling is hand-rolled and untested for the upsert case.
- [ ] The `Promise.all([...])` concurrency test is shown to FAIL against current `main` before the fix (true red→green regression, not a silent no-op) — achievable deterministically against the real in-memory `bun:sqlite` from `createTestDatabase()` (`test/db/testDbHelper.ts`).
- [ ] Tool-call-info merge behavior for `tool_failure` groups is preserved (existing merge assertions still pass).
- [ ] A regression/unit test in `test/db/failureAnalyticsRepository.test.ts` exercises the concurrent path using the repo's interface + fake conventions (inject `Timer`/`FakeTimer` and an injected `Kysely<Database>`), runs in <100ms, and is non-flaky (deterministic interleaving, not sleep-based).
- [ ] `bun test` green for the failures suites; `turbo run lint build test` green.

## References

- `src/db/failureAnalyticsRepository.ts:129` (`recordFailure` entry)
- `src/db/failureAnalyticsRepository.ts:136` (get-or-create SELECT)
- `src/db/failureAnalyticsRepository.ts:179` (INSERT branch, fresh UUID)
- `src/db/migrations/2026_01_27_000_failures.ts:45` (non-unique `idx_failure_groups_signature`)
- `src/db/installedAppsRepository.ts:120` (`onConflict` upsert precedent)
- `src/db/navigationRepository.ts:1041` (atomic increment precedent)
- `src/db/bunSqliteDialect.ts` (single-connection mutex released across awaits)
- `src/db/migrator.ts:178` (`FileMigrationProvider` picks up migrations by filename order)
- Finding R1, audit date 2026-07-01.

- **FK cascade ordering is a data-*loss* trap, not a nicety.** The migration connection runs `PRAGMA foreign_keys = ON` (`src/db/database.ts:47`) and SQLite migrations are not wrapped in a DDL transaction (`SqliteAdapter.supportsTransactionalDdl === false`). `failure_occurrences.group_id` is a real CASCADE FK (`2026_01_27_000_failures.ts:66-67`); you MUST repoint occurrences to the keeper **before** deleting losers or CASCADE wipes them.
- **`failure_notifications.group_id` is NOT an FK** (`:168`, plain `text notNull()`) — repointing it is a data-integrity nicety (orphan avoidance), not FK-driven.
- **Deterministic keeper tiebreak** — a `first_occurrence` tie (same-ms burst) leaves two rows and makes `CREATE UNIQUE INDEX` throw; use `ORDER BY first_occurrence ASC, id ASC` (min(id) tiebreak).
- **Drop the redundant post-upsert id-resolution SELECT** — `.returning("id")` works in this dialect (proven at `failureEventRepository.ts:116`, `networkEventRepository.ts:61`; dialect `hasReturning` at `bunSqliteDialect.ts:167-175`). Use `.onConflict(...).returning("id").executeTakeFirstOrThrow()`; the extra SELECT is also a race against fire-and-forget `cleanupRetention`.
- **`unique_sessions` recompute is race-free** (single-statement `COUNT(DISTINCT session_id)` over committed rows, idempotent under interleave) — but there's no `failure_occurrences(group_id, session_id)` index, so it's an O(rows-in-group) scan per event; accept it or add the composite index in the same migration.

### Direction (no dedicated architecture comment on this issue)
- Cross-linked as part of the concurrency workstream: this is the {#2790, #2789, #2791} cluster. Per Devin's note on **#2790**, #2789 is "the R-cluster member that carries a backfill" — same bug/fix as #2790's atomic-upsert direction, gated only by needing the de-dup migration to land before the UNIQUE index and `doUpdateSet` on `signature` can exist. Sequence: this migration → atomic upsert here → transaction (#2791).

### Regressions & testability (Sam Whitfield)
- **Wrap the whole de-dup in one `db.transaction()`** and make it idempotent / safe on an empty DB so it survives #2785's destructive-recovery replay (`resetDatabaseState` drops+re-runs all migrations); add `.ifNotExists()` to the new index create or the replay throws "index already exists".
- **Explicit orphan acceptance criterion:** after migration, zero `failure_notifications.group_id` reference a non-existent group; repoint occurrences via `UPDATE` (not delete+reinsert) so `failure_captures` / `failure_occurrence_screens` children survive.
- **Add a dialect test** asserting the upsert `.returning("id")` returns the pre-existing group id on the conflict path (hand-rolled RETURNING, untested for upsert).
- **Prove the concurrency test fails on `main` first** — `test/db/testDbHelper.ts`'s `createTestDatabase()` is a real in-memory `bun:sqlite` where the mutex-releases-across-await interleaving is real, so `Promise.all([...])` is a true red→green regression test in <100ms.

kaeawc:
## What

Makes `FailureAnalyticsRepository.recordFailure()` an **atomic upsert** and adds a **UNIQUE index** on `failure_groups.signature`, with a de-dup backfill migration for any duplicate-signature rows the old non-unique index let accumulate.

Closes #2789.

## Why

The daemon runs a single `bun:sqlite` connection behind an async mutex that **releases across every `await`** (`src/db/bunSqliteDialect.ts`). `recordFailure()` did a non-atomic get-or-create — `SELECT` the group, then `INSERT` a fresh-UUID group **or** `UPDATE total_count + 1` — keyed on a **non-unique** `signature` index. Two same-signature failures whose `await` windows overlap (crash-loop burst, ANR storm, simultaneous multi-device runs) either:

1. **Duplicate groups** — both read "no group", both `INSERT`; the non-unique index accepts both, fragmenting one signature across multiple dashboard cards.
2. **Lost increment** — both read `total_count = N`, both write `N + 1`; the second clobbers the first.

Occurrence rows were never lost (per-call UUID), but the aggregated group counts drift and duplicates accumulate — exactly when failure analytics matter most.

## How

**1. Migration `src/db/migrations/2026_07_01_000_failure_groups_signature_unique.ts`** — one `db.transaction()`:
- De-dupes duplicate-signature groups, keeping the deterministic keeper `ORDER BY first_occurrence ASC, id ASC` (min-id tiebreak survives a same-ms `first_occurrence` tie).
- **Repoints `failure_occurrences.group_id` and `failure_notifications.group_id` onto the keeper BEFORE deleting losers.** The migration connection runs `PRAGMA foreign_keys = ON` and these migrations are not wrapped in a DDL transaction, so `failure_occurrences`'s CASCADE FK would wipe the occurrences if losers were deleted first. `failure_notifications.group_id` is **not** an FK, so it is repointed explicitly to avoid silently stranded orphans.
- Back-fills the keeper: `total_count = SUM` (preserves historical increments past retention), `last_occurrence = MAX`, `unique_sessions = COUNT(DISTINCT session_id)` over the repointed occurrences.
- Drops the plain `idx_failure_groups_signature`, recreates it `UNIQUE IF NOT EXISTS`, and adds a composite `idx_failure_occurrences_group_session` so the per-event `COUNT(DISTINCT)` recompute does not scan the whole group.
- Idempotent / empty-DB safe so it survives #2785's destructive-recovery replay.

**2. `recordFailure()` atomic upsert** — `INSERT ... ON CONFLICT(signature) DO UPDATE SET total_count = total_count + 1 ... RETURNING id`:
- The count bump is now a single statement. `RETURNING id` yields the pre-existing id on the conflict path (so the id resolves atomically — no separate SELECT that could race fire-and-forget retention cleanup).
- `unique_sessions` is **derived** via a single-statement `COUNT(DISTINCT session_id)` after the occurrence insert (idempotent under interleave), not incremented on a stale read.
- Tool-call-info merge is preserved via a follow-up update on the conflict path.

**Test helper:** `createTestDatabase` gains an opt-in `foreignKeys` flag so the FK-cascade ordering hazard is exercised for real (production enables `foreign_keys = ON`; the test helper previously did not).

## Testing

- `test/db/failureAnalyticsRepository.test.ts` — new concurrency block: N concurrent `recordFailure` for one signature → exactly one group, `total_count === N`; M distinct sessions → `unique_sessions === M` (asserted **derived**); distinct-signature isolation; tool-call-info merge under concurrency. These **fail on `main`** (24 groups instead of 1) and pass here — a true red→green regression against the real in-memory `bun:sqlite`.
- `test/db/failureGroupsSignatureUniqueMigration.test.ts` — FK-safe repoint-before-delete (occurrences survive), notification orphan check (zero dangling `group_id`), deterministic tiebreak, count/`last_occurrence` backfill, uniqueness enforcement, empty-DB + idempotent re-run, `down()`.
- `test/db/bunSqliteDialectReturning.test.ts` — the hand-rolled dialect's `.onConflict(...).returning("id")` returns the **pre-existing** id on the conflict path.
- Full suite: **5734 pass, 0 fail** (`bun test`). `turbo run lint build` clean. Each unit test < 100ms, deterministic (no sleeps).

## Reviewed

`/auto-mobile-code-review` on the branch diff: one finding addressed — the keeper now inherits `MAX(last_occurrence)` so a collapsed group is not sorted stale on the dashboard. The tool-call-info merge retains a bounded read-modify-write on the conflict path (best-effort aggregation; the count-critical path is atomic) — noted as acceptable by design.

## Known limitations (tracked separately)

- **Multi-statement body is not yet wrapped in a single transaction (#2877).** `recordFailure` still issues its statements (upsert → occurrence insert → `unique_sessions` recompute → optional tool-info merge → screens/capture/notification) without an outer transaction — same as before this PR. The count bump now lands deterministically *before* the occurrence insert, so if the occurrence insert throws, the group `total_count` is incremented ahead of a missing occurrence row. Closing that window is tracked in #2877 (wrap the whole method); this PR intentionally does not, to keep the {#2790, #2789} DB-audit concurrency cluster sequencing clean. The tool-info merge here is already transactional as a targeted fix for its read-modify-write.

### DIFF
--- a/src/db/failureAnalyticsRepository.ts
+++ b/src/db/failureAnalyticsRepository.ts
@@ -1,13 +1,11 @@
-import type { Kysely } from "kysely";
+import { sql, type Kysely } from "kysely";
 import { getDatabase } from "./database";
 import type {
   Database,
-  NewFailureGroup,
   NewFailureOccurrence,
   NewFailureOccurrenceScreen,
   NewFailureCapture,
   NewFailureNotification,
-  FailureGroupUpdate,
 } from "./types";
 import { logger } from "../utils/logger";
 import type { Timer } from "../utils/SystemTimer";
@@ -132,55 +130,20 @@ export class FailureAnalyticsRepository {
     const occurrenceId = crypto.randomUUID();
 
     try {
-      // Check if group exists
-      const existingGroup = await db
-        .selectFrom("failure_groups")
-        .select(["id", "total_count", "unique_sessions", "first_occurrence", "tool_call_info_json"])
-        .where("signature", "=", input.signature)
-        .executeTakeFirst();
+      // Atomic get-or-create keyed on signature (#2789). The daemon's single
+      // connection releases its mutex across every await, so the former
+      // SELECT-then-INSERT/UPDATE interleaved: two same-signature failures could
+      // both INSERT (duplicate groups, now blocked by the UNIQUE index from
+      // migration 2026_07_01_000) or both read a stale total_count and clobber
+      // (lost increments). A single INSERT ... ON CONFLICT DO UPDATE makes the
+      // count bump atomic. The candidate id is used only on the first-seen path.
+      const candidateGroupId = crypto.randomUUID();
+      const nowIso = new Date().toISOString();
 
-      let groupId: string;
-
-      if (existingGroup) {
-        groupId = existingGroup.id;
-
-        // Count unique sessions including this one
-        const sessionExists = await db
-          .selectFrom("failure_occurrences")
-          .select("id")
-          .where("group_id", "=", groupId)
-          .where("session_id", "=", input.occurrence.sessionId)
-          .executeTakeFirst();
-
-        const newUniqueSessions = sessionExists
-          ? existingGroup.unique_sessions
-          : existingGroup.unique_sessions + 1;
-
-        // Merge tool call info if this is a tool failure
-        let mergedToolCallInfo = input.toolCallInfo;
-        if (input.type === "tool_failure" && existingGroup.tool_call_info_json) {
-          const existing = JSON.parse(existingGroup.tool_call_info_json) as AggregatedToolCallInfo;
-          mergedToolCallInfo = this.mergeToolCallInfo(existing, input.toolCallInfo, input.occurrence);
-        }
-
-        const update: FailureGroupUpdate = {
-          last_occurrence: now,
-          total_count: existingGroup.total_count + 1,
-          unique_sessions: newUniqueSessions,
-          tool_call_info_json: mergedToolCallInfo ? JSON.stringify(mergedToolCallInfo) : null,
-          updated_at: new Date().toISOString(),
-        };
-
-        await db
-          .updateTable("failure_groups")
-          .set(update)
-          .where("id", "=", groupId)
-          .execute();
-      } else {
-        groupId = crypto.randomUUID();
-
-        const group: NewFailureGroup = {
-          id: groupId,
+      const upserted = await db
+        .insertInto("failure_groups")
+        .values({
+          id: candidateGroupId,
           type: input.type,
           signature: input.signature,
           title: input.title,
@@ -192,11 +155,22 @@ export class FailureAnalyticsRepository {
           unique_sessions: 1,
           stack_trace_json: input.stackTrace ? JSON.stringify(input.stackTrace) : null,
           tool_call_info_json: input.toolCallInfo ? JSON.stringify(input.toolCallInfo) : null,
-          updated_at: new Date().toISOString(),
-        };
+          updated_at: nowIso,
+        })
+        .onConflict(oc =>
+          oc.column("signature").doUpdateSet(eb => ({
+            last_occurrence: now,
+            total_count: eb("failure_groups.total_count", "+", 1),
+            updated_at: nowIso,
+          }))
+        )
+        .returning("id")
+        .executeTakeFirstOrThrow();
 
-        await db.insertInto("failure_groups").values(group).execute();
-      }
+      const groupId = upserted.id;
+      // RETURNING gives the pre-existing id on the conflict path, so a returned
+      // id equal to our fresh candidate means this call created the group.
+      const isNewGroup = groupId === candidateGroupId;
 
       // Insert occurrence
       const occurrence: NewFailureOccurrence = {
@@ -217,6 +191,49 @@ export class FailureAnalyticsRepository {
       };
 
       await db.insertInto("failure_occurrences").values(occurrence).execute();
+
+      // Derive unique_sessions from the committed occurrence rows rather than
+      // incrementing a stale read (#2789). COUNT(DISTINCT session_id) is
+      // evaluated inside SQLite in one statement, so interleaved recordFailure
+      // calls both write the same correct value — idempotent, not racy. Backed
+      // by idx_failure_occurrences_group_session so it does not scan the group.
+      await db
+        .updateTable("failure_groups")
+        .set({
+          unique_sessions: sql<number>`(
+            SELECT COUNT(DISTINCT session_id)
+            FROM failure_occurrences
+            WHERE group_id = ${groupId}
+          )`,
+        })
+        .where("id", "=", groupId)
+        .execute();
+
+      // Preserve tool-call-info aggregation. On the first-seen path the upsert
+      // already stored this occurrence's info; on the conflict path we fold the
+      // new contribution into the existing aggregate via a follow-up update.
+      if (input.type === "tool_failure" && !isNewGroup) {
+        const current = await db
+          .selectFrom("failure_groups")
+          .select("tool_call_info_json")
+          .where("id", "=", groupId)
+          .executeTakeFirst();
+
+        let mergedToolCallInfo = input.toolCallInfo;
+        if (current?.tool_call_info_json) {
+          const existing = JSON.parse(current.tool_call_info_json) as AggregatedToolCallInfo;
+          mergedToolCallInfo = this.mergeToolCallInfo(existing, input.toolCallInfo, input.occurrence);
+        }
+
+        await db
+          .updateTable("failure_groups")
+          .set({
+            tool_call_info_json: mergedToolCallInfo ? JSON.stringify(mergedToolCallInfo) : null,
+            updated_at: new Date().toISOString(),
+          })
+          .where("id", "=", groupId)
+          .execute();
+      }
 
       // Insert screens visited
       if (input.occurrence.screensVisited && input.occurrence.screensVisited.length > 0) {
--- a/src/db/failureAnalyticsRepository.ts
+++ b/src/db/failureAnalyticsRepository.ts
@@ -221,8 +221,8 @@ export class FailureAnalyticsRepository {
       // read-modify-write (JSON merged in JS), so wrap it in a transaction: the
       // single-connection dialect reserves the connection for the transaction
       // owner (bunSqliteDialect.ts #reserveTransaction), serializing concurrent
-      // tool_failure merges so none clobber another's contribution. (The full
-      // recordFailure transaction is tracked separately in #2791.)
+      // tool_failure merges so none clobber another's contribution. (Wrapping
+      // the full recordFailure body in one transaction is tracked in #2877.)
       if (input.type === "tool_failure" && !isNewGroup) {
         await db.transaction().execute(async trx => {
           const current = await trx
--- a/src/db/failureAnalyticsRepository.ts
+++ b/src/db/failureAnalyticsRepository.ts
@@ -197,42 +197,55 @@ export class FailureAnalyticsRepository {
       // evaluated inside SQLite in one statement, so interleaved recordFailure
       // calls both write the same correct value — idempotent, not racy. Backed
       // by idx_failure_occurrences_group_session so it does not scan the group.
-      await db
-        .updateTable("failure_groups")
-        .set({
-          unique_sessions: sql<number>`(
-            SELECT COUNT(DISTINCT session_id)
-            FROM failure_occurrences
-            WHERE group_id = ${groupId}
-          )`,
-        })
-        .where("id", "=", groupId)
-        .execute();
-
-      // Preserve tool-call-info aggregation. On the first-seen path the upsert
-      // already stored this occurrence's info; on the conflict path we fold the
-      // new contribution into the existing aggregate via a follow-up update.
-      if (input.type === "tool_failure" && !isNewGroup) {
-        const current = await db
-          .selectFrom("failure_groups")
-          .select("tool_call_info_json")
-          .where("id", "=", groupId)
-          .executeTakeFirst();
-
-        let mergedToolCallInfo = input.toolCallInfo;
-        if (current?.tool_call_info_json) {
-          const existing = JSON.parse(current.tool_call_info_json) as AggregatedToolCallInfo;
-          mergedToolCallInfo = this.mergeToolCallInfo(existing, input.toolCallInfo, input.occurrence);
-        }
-
+      // Skip it on the first-seen path: the INSERT already set unique_sessions=1
+      // and this call's occurrence is the only one, so the recompute is a no-op
+      // there. A concurrent same-signature call conflicts (isNewGroup=false) and
+      // runs its own recompute, which observes every committed occurrence.
+      if (!isNewGroup) {
         await db
           .updateTable("failure_groups")
           .set({
-            tool_call_info_json: mergedToolCallInfo ? JSON.stringify(mergedToolCallInfo) : null,
-            updated_at: new Date().toISOString(),
+            unique_sessions: sql<number>`(
+              SELECT COUNT(DISTINCT session_id)
+              FROM failure_occurrences
+              WHERE group_id = ${groupId}
+            )`,
           })
           .where("id", "=", groupId)
           .execute();
+      }
+
+      // Preserve tool-call-info aggregation. On the first-seen path the upsert
+      // already stored this occurrence's info; on the conflict path we fold the
+      // new contribution into the existing aggregate. The merge is an inherent
+      // read-modify-write (JSON merged in JS), so wrap it in a transaction: the
+      // single-connection dialect reserves the connection for the transaction
+      // owner (bunSqliteDialect.ts #reserveTransaction), serializing concurrent
+      // tool_failure merges so none clobber another's contribution. (The full
+      // recordFailure transaction is tracked separately in #2791.)
+      if (input.type === "tool_failure" && !isNewGroup) {
+        await db.transaction().execute(async trx => {
+          const current = await trx
+            .selectFrom("failure_groups")
+            .select("tool_call_info_json")
+            .where("id", "=", groupId)
+            .executeTakeFirst();
+
+          let mergedToolCallInfo = input.toolCallInfo;
+          if (current?.tool_call_info_json) {
+            const existing = JSON.parse(current.tool_call_info_json) as AggregatedToolCallInfo;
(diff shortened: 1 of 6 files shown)


kaeawc:
## Review round: 3 independent perspectives + `/auto-mobile-code-review`

Ran the branch through `/auto-mobile-code-review` and three independent reviewers (SQLite-correctness, concurrency-regression, product-direction). All verdicts: **mergeable, no blocking bugs**. Findings and how they were addressed:

**SQLite/migration correctness (reproduced against real `bun:sqlite`):** de-dup correlated-subquery is well-defined and deterministic; FK repoint-before-delete ordering is load-bearing and correct (verified 3–5 dup collapse + adversarial rowid order); the owning `db.transaction()` is safe given the non-wrapping migrator (rollback atomicity verified); idempotent + #2785-replay safe; `ORDER BY first_occurrence ASC, id ASC` total order means `CREATE UNIQUE INDEX` can't throw on a residual dup. **No changes needed.**

**Concurrency/regression:** `isNewGroup` discriminator sound; `unique_sessions` COUNT(DISTINCT) recompute genuinely race-free; no new retention-cleanup race; red→green confirmed (origin/main produces 20 duplicate groups; HEAD 1). **One real finding addressed** — the tool-call-info merge was a lossy read-modify-write (measured 8/10 contributions dropped under concurrent `tool_failure` bursts) and its test was too lenient to catch it. Fixed by wrapping the merge in `db.transaction()` (dialect reserves the connection for the tx owner → serializes concurrent merges) and tightening the test to assert exact losslessness (10/10). Added a serial lossless test.

**Product/direction:** all of #2789's acceptance criteria verified met; direction matches in-repo upsert precedent. Micro-opt applied (skip the redundant `unique_sessions` recompute on the first-seen path). Documented in the migration why single-group lost-increment drift is intentionally not healed (blanket recompute would destroy retention-preserved counts). The no-outer-transaction window is noted in the PR body and deferred to #2791.

Follow-ups will be filed for the deferred items.

Issue #2802, opened 2026-07-01
perf(db): WAL hygiene — set journal_size_limit and checkpoint(TRUNCATE) on close


kaeawc:
## Summary
The database runs in WAL mode (`PRAGMA journal_mode = WAL`) but never issues an explicit checkpoint, and never caps the WAL file size via `journal_size_limit`. A repo-wide grep confirms **zero** `wal_checkpoint` and **zero** `journal_size_limit` anywhere in `src/`. On a clean shutdown (`closeDatabase` → `dialect.destroy()` → `BunSqliteConnectionState.close()`) the `-wal`/`-shm` sidecar files are left behind at whatever size the auto-checkpoint bound reached, rather than being truncated. This is a low-severity hygiene item — not a correctness bug. Found via a DB concurrency/migration audit (2026-07-01), finding P8wal, priority P2, verified against source.

## Background & history
- WAL mode + busy timeout + FK enforcement were introduced together in `configureSqliteDatabase`: commit `3dac40453` (2026-06-28), [PR #2565](https://github.com/kaeawc/auto-mobile/pull/2565) — "Fix SQLite busy timeout and support parallel daemon requests". That change enabled WAL for concurrent-read performance but did not add a `journal_size_limit` cap or any checkpoint policy.
- `closeDatabase` (which just calls `dbInstance.destroy()`) predates WAL mode: commit `a2d898d3f` (2025-12-28), "Add SQLite database infrastructure with Kysely". It was written before WAL existed, so it has never had a reason to checkpoint.
- The low-level `close()` in the dialect (`this.#db.close()`) currently reflects its blame to commit `1670cf174` (2026-06-29, PR #2632) due to a line move; behaviorally it has always been a bare `db.close()` with no checkpoint.

The code is shaped this way because WAL was added purely for read concurrency; nobody has yet needed to bound the sidecar files or clean them on shutdown. SQLite's default auto-checkpoint (1000 pages ≈ 4 MB) already caps steady-state WAL growth, so this went unnoticed.

## The problem
WAL mode is enabled but there is no size cap and no shutdown checkpoint:

```ts
// src/db/database.ts:39-48
export function configureSqliteDatabase(sqliteDb: SqlitePragmaDatabase): void {
  // Wait for transient writer locks instead of failing immediately with SQLITE_BUSY
  sqliteDb.exec(`PRAGMA busy_timeout = ${SQLITE_BUSY_TIMEOUT_MS};`);

[...]

PR #2880, opened 2026-07-03
fix(db): drain in-flight DB writes before closeDatabase() so shutdown cannot strand on Kysely ConnectionMutex (#2792)

### LINKED ISSUE #2792: fix(db): drain in-flight DB writes before closeDatabase() so shutdown cannot strand on Kysely ConnectionMutex
## Summary
On graceful shutdown, `Daemon` calls `closeDatabase()` with **no prior drain** of in-flight DB writes, while telemetry from device WebSockets is ingested as **fire-and-forget** (`void this.handleMessage(data)`). `closeDatabase()` → `dialect.destroy()` synchronously closes the `bun:sqlite` handle. Any query still queued behind Kysely's single-connection `ConnectionMutex` at that moment either errors ("Cannot use a closed database") or, in a specific interleaving, is left **permanently stranded** because Kysely's `RuntimeDriver.acquireConnection()` throws on the `#destroyPromise` guard after the mutex has been handed off, and the stranded promise never settles — hanging the shutdown path. Blast radius is shutdown-only inside the single daemon; it does not affect steady-state queries.

Found via a DB concurrency/migration audit (2026-07-01), finding C3, priority P2, verified against source.

## Background & history
- Telemetry ingest is fire-and-forget: `void this.handleMessage(data)` was introduced in `src/features/observe/DeviceServiceClient.ts:351` by commit `9d3784010` (2026-01-30). WS `message` events are dispatched without awaiting, so the `await recorder.record*Event(...)` DB writes inside `AndroidCtrlProxyClient.handleWebSocketMessage` run untracked by the shutdown sequence.
- `closeDatabase()` (`src/db/database.ts:199`) dates to commit `a2d898d3f` (2025-12-28, "Add SQLite database infrastructure with Kysely"). It does `await dbInstance.destroy()` with no drain.
- The daemon shutdown call site `await closeDatabase()` (`src/daemon/daemon.ts:1222`) was added in `365f05bfd` (2026-01-05, "Record test execution timing").
- The dialect's own hand-rolled query mutex (`#activeQueries` / `#waiters`) and the synchronous `close()` were added / reshaped in `916a75f1c` ("Fix SQLite parallel client safety", #2587, 2026-06-29, https://github.com/kaeawc/auto-mobile/pull/2587) and `1670cf174` (#2632, 2026-06-29, https://github.com/kaeawc/auto-mobile/pull/2632). Notably, `close()` closes the handle immediately without consulting `#activeQueries` or `#waiters`.
- Why it looks fine today: individual queries *are* serialized (both by Kysely's `ConnectionMutex` and the dialect's `#waiters`), so nobody noticed that shutdown skips the drain — the window only opens when `destroy()` races a query that hasn't yet acquired the connection.

## The problem
The concurrency model: a **single** `bun:sqlite` connection, guarded by a hand-rolled async mutex in `src/db/bunSqliteDialect.ts` that serializes individual queries but **releases across `await` boundaries**. On top of that, Kysely wraps the single-connection driver in its own `ConnectionMutex` (because `SqliteAdapter.supportsMultipleConnections === false`). Everything runs in ONE singleton daemon per uid — this is not a cross-process issue.

Telemetry writes are never awaited by anything the shutdown sequence can see:

```ts
// src/features/observe/DeviceServiceClient.ts:350
ws.on("message", (data: WebSocket.Data) => {
  void this.handleMessage(data);   // fire-and-forget: DB writes untracked
});
```

```ts
// src/features/observe/android/AndroidCtrlProxyClient.ts:2352 (inside handleWebSocketMessage)
const recorder = TelemetryRecorder.getInstance();
recorder.setContext(this.device.deviceId, null);
await recorder.recordNetworkEvent({ /* ... */ });   // reaches the shared bun:sqlite conn
```

Shutdown tears down sockets, then closes the DB with no barrier:

```ts
// src/daemon/daemon.ts:1193
await stopTelemetryPushSocketServer();
// ... more stop*() calls, HTTP server close, daemon-file cleanup ...
// src/daemon/daemon.ts:1222
await closeDatabase();   // <-- no drain of the fire-and-forget writes above
```

`closeDatabase()` closes the handle synchronously via the dialect:

```ts
// src/db/database.ts:199
export async function closeDatabase(): Promise<void> {
  if (dbInstance) {
    await dbInstance.destroy();   // -> BunSqliteDriver.destroy() -> state.close()
    dbInstance = null;
  }
}
```

```ts
// src/db/bunSqliteDialect.ts:88 (BunSqliteDriver.destroy)
async destroy(): Promise<void> {
  if (this.#connectionState) {
    this.#connectionState.close();      // closes handle immediately...
    this.#connectionState = undefined;  // ...and drops the state
  }
}

// src/db/bunSqliteDialect.ts:198 (BunSqliteConnectionState.close)
close(): void {
  if (this.#db) {
    this.#db.close();   // does NOT wait for #activeQueries or #waiters
    this.#db = null;
  }
}
```

The interleaving (reproduced): a query Q1 holds Kysely's `ConnectionMutex`; Q2 and Q3 are queued in `ConnectionMutex.obtainLock()`. `destroy()` fires, setting `RuntimeDriver.#destroyPromise` and closing the handle. When Q1 finishes it releases the lock; Q2 obtains it but then `RuntimeDriver.acquireConnection()` re-checks `if (this.#destroyPromise) throw` and throws *before* reaching `#driver.acquireConnection`. Depending on ordering, the throw path can bypass the `releaseLock()` that would unblock Q3, so **Q3's promise never settles**. With 2 queued waiters, one strands (the audit's "1 waiter settles, 2 waiters strand 1"). Best case, all in-flight writes just error out with "Cannot use a closed database" — silent telemetry loss on the way down.

Diagnosis correction vs. the original spec: the strand lives in **Kysely's `ConnectionMutex`** (`node_modules/kysely/dist/driver/connection-mutex.js` + `runtime-driver.js`), NOT in the dialect's own `#waiters`. Therefore a "drain the dialect's `#waiters` inside `close()`" fix is **ineffective** — by the time `close()` runs, the stranded waiter is parked one level up in Kysely.

## Impact
- Shutdown-only, single daemon. Two failure modes:
  1. **Silent data loss** (common): in-flight telemetry/failure writes that lose the race throw "Cannot use a closed database" and are dropped. Low severity — those rows were best-effort anyway.
  2. **Hung shutdown** (rare, timing-dependent): a stranded `ConnectionMutex` waiter leaves an unsettled promise. If anything on the shutdown path awaits that chain, the daemon fails to exit cleanly and must be force-killed / times out. Higher severity but narrow — needs a query queued in Kysely's mutex at the exact instant `destroy()` runs.
- Nobody sees this in steady state; it only bites on SIGTERM/SIGINT while a device is actively streaming telemetry. Hence P2.

## Proposed fix

> **The drain MUST be bounded — an unbounded drain can HANG shutdown forever, the exact outcome this issue is trying to prevent (per review).** Sam: if a write is itself wedged (a #2791 transaction that never commits because a mid-sequence await never settles, or a fire-and-forget telemetry write awaiting something upstream), a naive "await all in-flight writes before `closeDatabase()`" never completes and shutdown hangs — arguably worse than dropping a best-effort telemetry row, since the daemon won't die and only a process-manager SIGKILL escapes. Use `Promise.race([drain, timeout(N ms)])`, and on timeout log a warn and proceed to `close()` anyway (best-effort writes were best-effort).
>
> **"Await in-flight" needs a barrier/tracking primitive, not just a call (per review).** Telemetry is `void this.handleMessage(data)` — there is **no handle** to await, so draining requires FIRST introducing tracking: an app-level counter of outstanding `record*Event` promises (or quiescing at the Kysely `ConnectionMutex` level). Draining the dialect's `#activeQueries`/`#waiters` is ineffective (the strand is one level up in Kysely, as diagnosed above). The acceptance criteria must name WHICH quiesce point, or the fix will pass a happy-path test and still strand under the documented interleaving.
>
> **Enumerate every fire-and-forget DB writer live at shutdown (per review).** Sockets are stopped (`daemon.ts:1193`) before the DB close (`daemon.ts:1222`), so no NEW telemetry should arrive — good, that bounds the drain set. But verify there is no writer between those lines: session cleanup, `markStaleActiveSessionsExpired`, and especially `cleanupRetention()` in `failureAnalyticsRepository.ts:259` (`.catch(() => {})` fire-and-forget) can still be mid-flight at close. The drain must account for background cleanup tasks, not just WS telemetry.
>
> **Dependency:** #2802 (WAL checkpoint on close) depends on THIS landing first — a clean checkpoint requires the write set to be quiesced.

Quiesce in-flight writes **before** `closeDatabase()`, rather than trying to patch the mutex layer. Two complementary, minimal options:

1. **Shutdown barrier / draining flag.** Flip a `draining` flag early in shutdown so new telemetry writes reject fast (return without touching the DB), then `await` a short quiescence of outstanding writes before `closeDatabase()`. Track outstanding fire-and-forget writes with a counter and a settle promise, mirroring the deferred-work bookkeeping already used elsewhere.

```ts
// sketch — a small tracker owned by the daemon (or TelemetryRecorder)
private inFlight = 0;
private idle: Promise<void> = Promise.resolve();
private resolveIdle?: () => void;

track<T>(p: Promise<T>): Promise<T> {
  if (this.inFlight++ === 0) this.idle = new Promise(r => (this.resolveIdle = r));
  return p.finally(() => { if (--this.inFlight === 0) this.resolveIdle?.(); });
}

// shutdown, before closeDatabase():
this.draining = true;                 // new writes short-circuit
await Promise.race([this.idle, timeout(1000)]);  // bounded drain
await closeDatabase();
```

2. **Reject-fast on drain in `TelemetryRecorder`.** When draining, `record*Event` returns immediately (log at debug, per the swallow convention in CLAUDE.md) so no new query ever enters Kysely's mutex after `destroy()` is imminent.

Precedent to mirror for the "await outstanding background work before teardown" shape: `performanceAuditRepository.startPeriodicPruning` already owns a cancelable background timer with clean teardown; follow that ownership style. Keep the drain **bounded** (timeout) so a genuinely wedged write cannot itself hang shutdown.

Must NOT:
- Do not add a blanket `await`-every-WS-message change to `DeviceServiceClient` — that would serialize/slow steady-state ingest and change hot-path latency. Scope the barrier to shutdown.
- Do not attempt to "drain the dialect's `#waiters` in `close()`" — verified ineffective (strand is in Kysely's `ConnectionMutex`).
- Do not change query results or ordering for normal (non-shutdown) operation.

- [ ] With N in-flight telemetry writes outstanding when shutdown begins, `closeDatabase()` completes and the shutdown promise settles within the bounded drain window — **no unsettled/stranded promise** (assert the shutdown `Promise` resolves; a stranded waiter would hang the test).
- [ ] **(per review) A wedged in-flight write does not block shutdown beyond N ms:** with a deliberately-never-completing write outstanding, the bounded drain (`Promise.race([drain, timeout(N)])`) times out, logs a warn, and shutdown still proceeds to `close()` and resolves.
- [ ] **(per review) The barrier/tracking primitive is named and drives the drain:** the acceptance path specifies WHICH quiesce point is used (app-level in-flight `record*Event` counter OR Kysely `ConnectionMutex` quiescence) — NOT draining the dialect's `#waiters` (verified ineffective). A test asserts the drain awaits the tracked set.
- [ ] **(per review) All fire-and-forget DB writers live at shutdown are enumerated and covered by the barrier:** WS telemetry plus background cleanup tasks between `daemon.ts:1193` and `:1222` — session cleanup, `markStaleActiveSessionsExpired`, and `cleanupRetention()` (`failureAnalyticsRepository.ts:259`, `.catch(() => {})`).
- [ ] After the `draining` flag is set, subsequent `record*Event` calls return without issuing a query against the closed/closing connection (assert zero "Cannot use a closed database" errors are thrown to callers post-drain).
- [ ] A regression unit test reproduces the destroy-during-queued-query interleaving at the Kysely + `BunSqliteDialect` layer and asserts all queued queries settle (resolve or reject) — none hang. New file `test/db/bunSqliteDialect.test.ts` (none exists today); use interfaces + fakes + `FakeTimer` per CLAUDE.md, run in <100ms, non-flaky (use an in-memory `bun:sqlite` and a gated `beforeQuery` to deterministically hold the mutex — no real timers/sleeps for the race).
- [ ] **(per review) The test uses a controllable fake at the mutex boundary, NOT `Promise.all` + `sleep`.** The strand is timing-dependent, so a sleep-based repro is flaky and violates the <100ms / non-flaky rule. Inject a fake driver/connection whose `acquireConnection`/query resolution is gated manually: park a query holding the `ConnectionMutex`, queue two more, fire `destroy()`, then resolve waiters in a chosen order. Assert (a) all queued promises settle (none strand) and (b) shutdown resolves within the timeout even when a write is deliberately never completed.
- [ ] `bun test` for the new/affected files green; `turbo run lint build test` green.

## References
- `src/daemon/daemon.ts:1222` — `await closeDatabase()` with no prior drain
- `src/daemon/daemon.ts:1193` — `stopTelemetryPushSocketServer()` (shutdown sequence)
- `src/db/database.ts:199` — `closeDatabase()` → `dbInstance.destroy()`
- `src/db/bunSqliteDialect.ts:88` — `BunSqliteDriver.destroy()`
- `src/db/bunSqliteDialect.ts:198` — `BunSqliteConnectionState.close()` (synchronous, ignores `#activeQueries`/`#waiters`)
- `src/features/observe/DeviceServiceClient.ts:351` — `void this.handleMessage(data)` (fire-and-forget)
- `src/features/observe/android/AndroidCtrlProxyClient.ts:2352` — `await recorder.recordNetworkEvent(...)` inside untracked WS handler
- `node_modules/kysely/dist/driver/connection-mutex.js`, `node_modules/kysely/dist/driver/runtime-driver.js` — `ConnectionMutex.obtainLock` / `RuntimeDriver.acquireConnection` `#destroyPromise` guard (strand site)
- Finding C3, DB concurrency/migration audit 2026-07-01

**Regressions & testability (Sam Whitfield)**
- A drain with NO timeout can hang shutdown forever — the exact outcome this issue is meant to prevent. Bound it with `Promise.race([drain, timeout(N)])` + force-close fallback on timeout; add an acceptance criterion.
- Fire-and-forget telemetry writes are UNTRACKED by construction (`void this.handleMessage`), so "await in-flight" requires a barrier/tracking primitive first (app-level counter OR Kysely `ConnectionMutex` quiescence) — draining the dialect's `#waiters` is ineffective (strand is one level up). The criteria must name which quiesce point.
- Enumerate every fire-and-forget DB writer live at shutdown, not just WS telemetry: session cleanup, `markStaleActiveSessionsExpired`, and `cleanupRetention()` (`failureAnalyticsRepository.ts:259`, `.catch(() => {})`) can be mid-flight between `daemon.ts:1193` (sockets stopped) and `:1222` (DB close).
- The test must use a **controllable fake at the mutex boundary** (park/queue/resolve in a chosen order), not `Promise.all` + `sleep` — the strand is timing-dependent and a sleep repro will be flaky. Flagged in acceptance criteria.

**Cross-links:** #2791 (transactional graph write — a never-committing tx is a wedged-write source the bounded drain must survive); #2802 (WAL checkpoint on close) depends on this landing first.

kaeawc:
## What

On graceful shutdown the daemon closed the single `bun:sqlite` connection (`closeDatabase()` → `dialect.destroy()`) with **no prior drain** of fire-and-forget DB writes. A query still queued behind Kysely's single-connection `ConnectionMutex` at the instant `destroy()` ran could be **permanently stranded**, hanging shutdown; in the common case in-flight best-effort writes were silently dropped with "Cannot use a closed database".

This lands two complementary parts:

1. **Dialect hardening** (`src/db/bunSqliteDialect.ts`) — `destroy()` now marks the connection state closed but **keeps** it, so `acquireConnection()` keeps **resolving** a (closed) lease. Kysely releases its `ConnectionMutex` lock only when the driver's `acquireConnection` *resolves* — the old code **threw**, leaving the lock held and stranding the next waiter. Queued queries now reject cleanly with `Cannot use a closed database`, and the closed check also prevents the accidental reopen-after-close path.
2. **App-level drain barrier** (`src/db/dbWriteBarrier.ts`) — a shared `DbWriteBarrier` tracks in-flight best-effort writes. Writers route their DB work through `barrier.track()`; `daemon.stop()` calls a **bounded** `drain(1000ms)` before `closeDatabase()`. New writes short-circuit once draining.

Closes #2792.

## Why

Found via the 2026-07-01 DB concurrency/migration audit (finding C3, P2). The strand was **reproduced** against source: with Q1 parked in the mutex and Q2/Q3 queued, `destroy()` → q1 resolves, q2 rejects, **q3 never settles** (hang). Root cause confirmed in `kysely/dist/driver/runtime-driver.js`: `RuntimeDriver.acquireConnection` obtains the `ConnectionMutex` lock, then calls the driver's `acquireConnection`; a throw there never releases the lock, so the next waiter strands.

Two review points from the issue drove the design:
- **The drain MUST be bounded** — an unbounded "await all in-flight writes" would inherit the very hang it prevents if a write is itself wedged (e.g. a never-committing #2791 transaction). `drain()` uses `Promise.race([idle, timeout])`; on timeout it logs a warn and proceeds to close anyway (best-effort writes were best-effort).
- **"Await in-flight" needs a tracking primitive** — telemetry is `void this.handleMessage(...)` with no handle to await. `DbWriteBarrier` is that primitive. This is the named quiesce point (app-level in-flight counter), **not** draining the dialect's `#waiters` (verified ineffective — the strand is one level up in Kysely).

## How

- **`src/db/dbWriteBarrier.ts`** (new): `DbWriteBarrier` interface + `InMemoryDbWriteBarrier` (in-flight counter + settle promise + injected `Timer` for the bound) + shared `getDbWriteBarrier()`/`resetDbWriteBarrier()`. `track()` short-circuits (returns `undefined`, `logger.debug` per the swallow convention) once draining.
- **`src/db/bunSqliteDialect.ts`**: `#closed` flag; `close()` sets it, closes the handle, and wakes waiters; `executeQuery`/`beginTransaction` reject when closed (before reopening); `destroy()` keeps the state so acquire never throws.
- **`daemon.ts`**: `await getDbWriteBarrier().drain(DB_WRITE_DRAIN_TIMEOUT_MS)` before `closeDatabase()`, warn-on-timeout.
- Wired every fire-and-forget DB writer live at shutdown through the barrier:
  - `TelemetryRecorder.record{Network,Log,Os,Navigation,Storage,Layout}Event` (WS ingest).
  - `FailureAnalyticsRepository.cleanupRetention()` (`.catch(() => {})`).
  - `SessionManager` activity + lazy/cleanup expiry writes (the `void this.recordSessionActivity` / `void markReleased` paths).
  - `markStaleActiveSessionsExpired` was checked and **excluded** — it runs at startup (`initializeDatabase`), not shutdown.

## Testing

- **`test/db/dbWriteBarrier.test.ts`** (new): tracked-value passthrough, in-flight counting, drain resolves `true` when all settle, drain returns `false` after the bound elapses for a **wedged** write (FakeTimer, no sleep), short-circuit-while-draining, counter-safe on rejection.
- **`test/db/bunSqliteDialect.test.ts`** (new): destroy-during-queued interleaving via a **gated `beforeQuery`** (deferred promise, no timers/sleep) — asserts all three queued queries settle, none strand; plus a direct-driver test proving `acquireConnection` resolves after `destroy()` and the query rejects with a closed-db error.
- Draining coverage added to `TelemetryRecorder`, `failureAnalyticsRepository`, and `sessionManager` tests (skip DB write while draining; route through the barrier so drain awaits them).
- `bun test` for db + daemon + telemetry suites green; `bun run lint` and `bun run build` clean; `scripts/all_fast_validate_checks.sh` green. All new unit tests <100ms, non-flaky.

## Acceptance criteria (from #2792)

- [x] N in-flight writes → `closeDatabase()` completes and shutdown settles within the bounded window, no stranded promise.
- [x] A wedged in-flight write does not block shutdown beyond N ms (bounded `Promise.race` times out, warns, proceeds).
- [x] The barrier/tracking primitive is named and drives the drain (app-level in-flight `track()` counter — **not** the dialect `#waiters`).
- [x] All fire-and-forget DB writers live at shutdown enumerated and covered (WS telemetry, `cleanupRetention`, session activity/expiry); `markStaleActiveSessionsExpired` shown startup-only.
- [x] After `draining` is set, `record*Event` returns without querying the closing connection (no "Cannot use a closed database" thrown to callers).
- [x] New `test/db/bunSqliteDialect.test.ts` reproduces the interleaving with a controllable fake at the mutex boundary (no `Promise.all` + sleep), <100ms, non-flaky.
- [x] `lint build test` green.

## Follow-ups

- #2802 (WAL checkpoint on close) depends on this quiesce landing first — now unblocked.


kaeawc:
## Adversarial self-review (3 personas) + resolutions

Ran three independent review agents over the diff — a concurrency/async skeptic, a regression/blast-radius reviewer, and a staff architecture reviewer. Summary and how each finding was addressed:

**Concurrency (verified by reproduction):** the barrier `#idle`/`#resolveIdle` lifecycle is race-free, `drain()` never leaks the timer, and the Kysely mutex-strand fix was reproduced as fixed (destroy resolves; queued query + transaction reject cleanly, none strand). No Critical/High. Actioned nit: the `void barrier.track(...)` session writes now `.catch(logger.warn)` so a rejected best-effort write can't surface as an `unhandledRejection`.

**Regression / blast-radius:** no steady-state regressions. Confirmed the telemetry hot path, network `recordId` (`0 ?? null === 0`; ids are AUTOINCREMENT so never 0), dialect `destroy()` inertness, and drain placement all behave identically when not draining; `drain()` returns immediately when nothing is in flight (no added shutdown latency). Documented that the barrier's draining flag latches for the process lifetime (prod exits; tests inject a fresh barrier).

**Architecture / completeness:** direction sound, two parts both load-bearing and Part 1 is confirmed *not* the forbidden "drain `#waiters`" approach. Two enumeration gaps found:
- `markInstalledAppsStale` (WS-close, shutdown-time) — **now routed through the barrier**.
- `NavigationGraphManager.recordNavigationEvent` — **intentionally left untracked** and documented inline: it interleaves with `hierarchy_update` processing to preserve SDK screen names, and a drain await hop perturbs that hot-path ordering (verified: it breaks the `preserve SDK screen names` test by reordering, a real behavior change, not a test artifact). A shutdown race there is already made safe by Part 1's reject-on-closed handling, so the row drops cleanly rather than stranding.

Full suite green after changes: **5738 pass, 0 fail** (441 files); `lint` + `build` clean.

Issue #2786, opened 2026-07-01
fix(db): reset failed migrationsPromise and surface a cached startup error synchronously so a floating rejection can't kill the daemon


kaeawc:
## Summary
When startup migrations fail, `migrationsPromise` in `src/db/database.ts` is left as a truthy, *rejected* promise that is never reset. If the very first `getDatabase()` caller is a field initializer with no same-tick `await` on that promise (e.g. `MemoryThresholdManager`), Bun can report the rejection as an `unhandledRejection`, which the daemon's fatal handler turns into `process.exit(1)` — a hard daemon crash instead of a clean, catchable query error. This is narrow (only on a migration failure, and only when no default startup path has already attached an awaiter) but the failure mode is a full daemon kill.

Found via a DB concurrency/migration audit (2026-07-01), finding C2, priority P1, verified against source.

## Background & history
- Introduced: commit `916a75f1` (2026-06-29), [PR #2587](https://github.com/kaeawc/auto-mobile/pull/2587) — "Fix SQLite parallel client safety". This reworked migration gating to run migrations once via a shared `migrationsPromise` and added `waitForMigrationsBeforeQuery` as a `beforeQuery` hook, plus `ensureMigrationsStarted` / `createStartupMigrationError`. The `.catch(...)` that wraps the failure in a structured error was added here.
- The `migrationsPromise` field itself dates to `73b2dd3a` (2026-01-05, [PR #293](https://github.com/kaeawc/auto-mobile/pull/293)).
- The at-risk field-initializer opener `private db = getDatabase();` in `MemoryThresholdManager` predates this and comes from `699ce5db` (2025-12-31, [PR #117](https://github.com/kaeawc/auto-mobile/pull/117)).
- Why it's shaped this way: fail-fast is partly intentional. `ensureMigrationsStarted` deliberately does not auto-retry on a corrupt DB (re-running identical migrations just fails again — this pairs with finding C1). The gap is only the *delivery mechanism* of that failure: a floating rejected promise rather than a synchronous, catchable error.

## The problem
`ensureMigrationsStarted` sets `migrationsPromise` once, and on failure the promise stays truthy and rejected forever:

```ts
// src/db/database.ts:145
function ensureMigrationsStarted(dbPath: string): void {
  if (!migrationsRun && !migrationsPromise) {
    migrationsPromise = startMigrations(dbPath).catch(error => {
      throw createStartupMigrationError(error);   // re-throws → migrationsPromise stays REJECTED
[...]

PR #2889, opened 2026-07-03
fix(db): reset migration + path globals in closeDatabase() so a same-process reopen behaves like a cold start (#2796)

### LINKED ISSUE #2796: chore(db): closeDatabase() leaves migration flags set, so a same-process reopen would skip migration gating
## Summary
`closeDatabase()` in `src/db/database.ts` destroys the Kysely `dbInstance` but leaves the module-global `migrationsRun` / `migrationsPromise` flags set. If `getDatabase()` were called again in the same process after a `closeDatabase()`, the new connection would be handed the stale `migrationsRun === true` state and `waitForMigrationsBeforeQuery()` would return immediately — application queries would run without ever gating on migrations. This is latent today because the only caller (`src/daemon/daemon.ts:1222`) is terminal daemon shutdown with no reopen path, but the reset is the correct place to also harden test isolation.

Found via a DB concurrency/migration audit (2026-07-01), finding M5, priority P2, verified against source.

## Background & history
- Introduced: commit `a2d898d3` (2025-12-28), "Add SQLite database infrastructure with Kysely" — added `dbInstance`, `closeDatabase()`, and the singleton `getDatabase()`.
- `migrationsRun` was added in commit `9f95375e` (2025-12-31); `migrationsPromise` in `73b2dd3a` (2026-01-05, [#293](https://github.com/kaeawc/auto-mobile/pull/293)).
- The migration-gating handshake (`waitForMigrationsBeforeQuery`, `createStartupMigrationError`, "refusing to run queries until the daemon restarts") was hardened in `916a75f1` (2026-06-29, [#2587](https://github.com/kaeawc/auto-mobile/pull/2587), "Fix SQLite parallel client safety").
- Shaped this way because the daemon lifecycle is single-shot per process: the daemon opens the DB once at startup and only calls `closeDatabase()` during graceful shutdown right before exit, so the flags never needed to be reset. The `closeDatabase()` reset was simply never extended to cover the migration globals that arrived in later commits.

## The problem
`closeDatabase()` only nulls `dbInstance`; the migration state globals survive:

```ts
// src/db/database.ts:92-94
let dbInstance: Kysely<DatabaseSchema> | null = null;
let migrationsRun = false;
let migrationsPromise: Promise<void> | null = null;
```

```ts
// src/db/database.ts:199-204
export async function closeDatabase(): Promise<void> {
  if (dbInstance) {
    await dbInstance.destroy();
    dbInstance = null;
  }
}
```

The gating check short-circuits as soon as `migrationsRun` is truthy:

```ts
// src/db/database.ts:104-110
async function waitForMigrationsBeforeQuery(): Promise<void> {
  if (migrationsRun || !migrationsPromise) {
    return;
  }
  await migrationsPromise;
}
```

Trigger scenario (same process, sequential — this is not a cross-await interleave):
1. `getDatabase()` opens the DB, `startMigrations()` runs, sets `migrationsRun = true`.
2. `closeDatabase()` destroys `dbInstance` but leaves `migrationsRun === true`.
3. A subsequent `getDatabase()` sees `dbInstance === null`, rebuilds the Kysely instance via `createApplicationDatabase()` (which wires `waitForMigrationsBeforeQuery` as `beforeQuery`), but `ensureMigrationsStarted()` is a no-op because `migrationsRun` is still `true`.
4. The very first query on the reopened connection calls `waitForMigrationsBeforeQuery()`, which returns immediately — no migration run, no gate. Against a fresh/empty file (e.g. a different `AUTOMOBILE_DB_PATH`) the connection would issue queries against an unmigrated schema.

The concurrency model here is a single `bun:sqlite` connection guarded by a hand-rolled async mutex in `src/db/bunSqliteDialect.ts` inside one singleton daemon per uid. This particular bug is not an interleave across `await` boundaries — it is stale module-global state that outlives the connection it describes. `resolvedDbPath` (line 70) has the same lifetime issue: it is cached on first resolve and never cleared, so a reopen under a changed `AUTOMOBILE_DB_PATH`/`AUTOMOBILE_DB_DIR` would still bind the old path.

Production is safe today only because `daemon.ts:1222` calls `closeDatabase()` as the last DB action before process exit — there is no in-process reopen.

## Impact
- No production impact today: the sole caller is terminal daemon shutdown; there is no reopen path.
- Latent correctness hazard: any future feature that reopens the DB in-process (config reload, DB path switch, restart-without-exit) would silently skip migration gating and could query an unmigrated schema.
- Test isolation: tests that open the app DB share process-global `migrationsRun`/`migrationsPromise`/`resolvedDbPath`, so `closeDatabase()` does not fully reset state between cases — a source of order-dependent flakiness if such tests are added.

Severity is low/latent; this is hardening, not an active bug.

## Proposed fix
Reset the migration and path globals in `closeDatabase()` so a reopen behaves like a cold start:

```ts
// src/db/database.ts
export async function closeDatabase(): Promise<void> {
  if (dbInstance) {
    await dbInstance.destroy();
    dbInstance = null;
  }
  migrationsRun = false;
  migrationsPromise = null;
  migrationsError = null;   // (per review) fold in #2786's cached error — one state machine, reset as a set
  resolvedDbPath = null;
}
```

Notes / guardrails:
- Reset the flags unconditionally (outside the `if (dbInstance)` guard) so a partially-initialized state — e.g. `ensureMigrations()` started migrations but `dbInstance` was already nulled — is still cleaned up.
- **(per review)** If #2786 lands, it adds a module-global `migrationsError`; this reset MUST include `migrationsError = null` too. Otherwise a reopen after a prior failed boot starts with a **stale cached error** and throws on the first query of an otherwise-healthy new DB — a reopen that should behave like a cold start instead fails permanently. Whichever of {#2786, #2796} lands second folds it in. Cross-ref #2786.
- **(per review)** Clearing `resolvedDbPath` relaxes the daemon-lifetime invariant that `getDatabasePath()` always matches the path the DB was opened at (the documented rationale at `resolveDbPath()`). In the daemon this is inert (the only `closeDatabase()` is at shutdown-then-exit). The one hazard: if any code path calls `closeDatabase()` **mid-daemon-life** (not at exit) with the env changed in between, the reopen would bind a *different* file than the one the daemon reported earlier via `getDatabasePath()`. Confirm no such mid-life close exists; for test isolation the full reset is fine, but don't let a non-exit close reopen against a divergent env silently.
- Clearing `resolvedDbPath` is what lets a reopen re-read `AUTOMOBILE_DB_PATH`/`AUTOMOBILE_DB_DIR`; keep it consistent with the lazy-resolution rationale documented at `resolveDbPath()` (lines 72-90). If reviewers prefer to keep the path cache pinned for the daemon's lifetime, split it into a separate opt-in reset — but for test isolation the full reset is preferable.
- Must NOT change the happy-path daemon shutdown behavior: after the final `closeDatabase()` the process exits, so the extra assignments are inert there.
- This mirrors the existing "cold start" precedent in `getDatabase()`/`ensureMigrationsStarted()` which key entirely off `!migrationsRun && !migrationsPromise`.

- [ ] After `getDatabase()` (migrations complete) → `closeDatabase()` → `getDatabase()` in the same process, migrations are gated again: `waitForMigrationsBeforeQuery()` does not short-circuit until a fresh `migrationsPromise` resolves (assert `migrationsRun === false` immediately after `closeDatabase()`).
- [ ] After `closeDatabase()`, a reopen under a different `AUTOMOBILE_DB_PATH` resolves the new path (i.e. `getDatabasePath()` reflects the changed env), proving `resolvedDbPath` was cleared.
- [ ] **(per review)** If #2786 has landed (module-global `migrationsError`), `closeDatabase()` also resets it to `null`: a same-process reopen after a failed boot behaves like a cold start and does not throw a stale cached error on the first query of a healthy new DB. Cross-ref #2786.
- [ ] **(per review)** Confirm no mid-daemon-life (non-exit) `closeDatabase()` caller exists that would reopen against a changed env and diverge from the previously-reported `getDatabasePath()`; the "path always matches" invariant is preserved for the daemon.
- [ ] Unit test added following repo conventions (interfaces + fakes + FakeTimer where time is involved), runs in <100ms, non-flaky. Likely file: `test/db/databaseLazyPath.test.ts` (or a new `test/db/databaseReset.test.ts`).
- [ ] `bun test` for the affected file(s) green; `turbo run lint build test` green.
- [ ] No change to daemon shutdown semantics (`src/daemon/daemon.ts:1222` still closes the DB cleanly once, immediately before exit).

## References
- `src/db/database.ts:92-94` — module-global migration/instance flags.
- `src/db/database.ts:199-204` — `closeDatabase()` (missing flag reset).
- `src/db/database.ts:104-110` — `waitForMigrationsBeforeQuery()` short-circuit.
- `src/db/database.ts:145-151` — `ensureMigrationsStarted()` no-op when `migrationsRun` set.
- `src/db/database.ts:70,85-90` — `resolvedDbPath` cache and lazy resolution.
- `src/daemon/daemon.ts:1222` — sole (terminal) caller of `closeDatabase()`.
- Finding M5, DB concurrency/migration audit, 2026-07-01.

**Correctness (Mira)**
- Must reset `migrationsError` too if #2786 lands — the three migration globals are one state machine; leaving `migrationsError` set means a reopen after a failed boot throws a stale error against a healthy new DB. Folded into the sketch + acceptance. Cross-ref #2786.
- Resetting `resolvedDbPath` relaxes the "path always matches" daemon-lifetime invariant; inert at shutdown-then-exit, but a mid-life close with a changed env would bind a divergent file. Confirm no such caller exists. Added acceptance criterion.
- Direction matches the `getDatabase()`/`ensureMigrationsStarted()` cold-start keying (`!migrationsRun && !migrationsPromise`); unconditional reset (outside the `if (dbInstance)` guard) is correct.

**Direction (Devin)** — via #2784's review: #2796 is the same `database.ts` globals from the shutdown end; fold its reset into the shared DB-lifecycle workstream (#2784 anchor, #2786 the delivery seam) rather than as a separate drive-by. Cross-ref #2784, #2786.

**Regressions & testability (Sam)** — via #2786's review: `migrationsError` inherits this issue's exact reset bug; coordinate the clear so the coupling ships covered. No separate regression comment posted on this issue.

kaeawc:
## What

`closeDatabase()` in `src/db/database.ts` destroyed the Kysely `dbInstance` but left the module-global migration state machine — `migrationsRun` / `migrationsPromise` / `migrationsError` — and the cached `resolvedDbPath` set. A **same-process reopen** therefore inherited stale state: `ensureMigrationsStarted()` no-ops (because `migrationsRun` is still `true`), `waitForMigrationsBeforeQuery()` short-circuits, and the new connection issues queries against a possibly-unmigrated schema — or rethrows a stale startup error against an otherwise-healthy reopened DB.

This resets all four globals unconditionally so a reopen behaves like a true cold start.

Closes #2796.

## Why

No production impact today: the sole caller is `daemon.ts:1285`, terminal daemon shutdown-then-exit, with no reopen path. But the reset is the correct place to harden two things:

- **Test isolation** — tests that open the app DB share process-global `migrationsRun`/`migrationsPromise`/`resolvedDbPath`; without the reset, `closeDatabase()` doesn't fully restore cold-start state between cases (a source of order-dependent flakiness).
- **Latent correctness** — any future in-process reopen (config reload, DB path switch, restart-without-exit) would silently skip migration gating and could query an unmigrated schema.

`migrationsError` (#2786) has already landed, so — per Mira Kessler's review on the issue — it's folded into the reset: the three migration globals are one state machine, and leaving `migrationsError` set would make a reopen after a failed boot throw a stale cached error against a healthy new DB.

## How

`src/db/database.ts` — `closeDatabase()` now resets the four globals **outside** the `if (dbInstance)` guard, so a partially-initialized state (migrations started but `dbInstance` already nulled) is still cleaned up:

```ts
migrationsRun = false;
migrationsPromise = null;
migrationsError = null;
resolvedDbPath = null;
```

- The migration globals are reset as a set — mirrors the existing cold-start keying in `getDatabase()`/`ensureMigrationsStarted()` (`!migrationsRun && !migrationsPromise`).
- `resolvedDbPath` is cleared so a reopen re-reads `AUTOMOBILE_DB_PATH` / `AUTOMOBILE_DB_DIR`. Confirmed there is **no mid-daemon-life (non-exit) `closeDatabase()` caller** — the only production caller is `daemon.ts:1285` at shutdown-then-exit — so the "path always matches the DB the daemon opened" invariant from `resolveDbPath()` is preserved; the reset is inert there and only matters for test isolation.
- Daemon shutdown semantics are unchanged: after the final `closeDatabase()` the process exits, so the extra assignments are inert.

## Testing

- **`test/db/databaseReset.test.ts`** (new, TDD — written red first, then made green) covering the issue's acceptance criteria, all behavioral (the globals aren't exported):
  - `migrationsRun` reset — after `getDatabase()`(migrated) → `closeDatabase()` → reopen against a **fresh, empty** DB dir, the reopen binds the new file (`getDatabasePath()` proof) and migrations **re-run** (a migrated-table query returns `[]` instead of "no such table"). This assertion is sensitive to both the path and migration resets independently.
  - `resolvedDbPath` reset — after `closeDatabase()`, changing `AUTOMOBILE_DB_DIR` makes `getDatabasePath()` reflect the new path.
  - `migrationsError` reset — after a failed boot + `closeDatabase()`, `getMigrationsError()` is `null` and a healthy reopen queries cleanly without rethrowing the stale error.
- Follows repo conventions (fresh module import per case for global isolation, temp dirs, `defaultTimer` for retry). Full `test/db/` suite: **432 pass, 0 fail**. `turbo run lint build` clean.

## Reviewed

Ran `/auto-mobile-code-review` on the branch diff. Verdict: **correctly fixes** — resets the full state machine per Mira's riders (`migrationsError` folded in), confirmed no mid-life close caller so the path invariant holds, no change to daemon shutdown semantics. One non-blocking observation: a `closeDatabase()` that raced an in-flight migration would null `migrationsPromise` mid-run, but that path is guarded by `createFileMigrationLock` and is unreachable in the daemon (shutdown is sequential after drain).

### DIFF
--- a/src/db/database.ts
+++ b/src/db/database.ts
@@ -282,6 +282,26 @@ export async function closeDatabase(): Promise<void> {
     await dbInstance.destroy();
     dbInstance = null;
   }
+
+  // Reset the module-global state that outlives the connection so a same-process
+  // reopen behaves like a cold start (issue #2796). These are reset
+  // unconditionally — outside the `if (dbInstance)` guard — so a
+  // partially-initialized state (migrations started but `dbInstance` already
+  // nulled) is still cleaned up.
+  //
+  // - `migrationsRun` / `migrationsPromise` / `migrationsError` are one state
+  //   machine; leaving any set would let `ensureMigrationsStarted()` no-op and
+  //   `waitForMigrationsBeforeQuery()` short-circuit (skipping migration gating
+  //   on the new connection), or rethrow a stale startup error against an
+  //   otherwise-healthy reopened DB.
+  // - `resolvedDbPath` is cleared so a reopen re-reads AUTOMOBILE_DB_PATH /
+  //   AUTOMOBILE_DB_DIR. In the daemon the only caller is shutdown-then-exit
+  //   (daemon.ts), so this is inert there and the "path always matches" invariant
+  //   from resolveDbPath() holds; the reset is what gives tests full isolation.
+  migrationsRun = false;
+  migrationsPromise = null;
+  migrationsError = null;
+  resolvedDbPath = null;
 }
 
 /**
--- a/test/db/databaseReset.test.ts
+++ b/test/db/databaseReset.test.ts
@@ -1,0 +1,171 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import path from "path";
+import { mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { DAEMON_LAUNCH_CWD_ENV } from "../../src/utils/workingDirectory";
+import { defaultTimer } from "../../src/utils/SystemTimer";
+
+/**
+ * Regression tests for issue #2796.
+ *
+ * `closeDatabase()` destroys the Kysely instance but historically left the
+ * module-global migration state machine (`migrationsRun` / `migrationsPromise`
+ * / `migrationsError`) and the cached `resolvedDbPath` set. A same-process
+ * reopen therefore inherited stale state: `ensureMigrationsStarted()` no-ops
+ * because `migrationsRun` is still true, `waitForMigrationsBeforeQuery()`
+ * short-circuits, and the new connection issues queries against a possibly
+ * unmigrated schema — or throws a stale cached startup error against an
+ * otherwise-healthy new DB. The fix resets all four globals unconditionally so
+ * a reopen behaves like a cold start.
+ *
+ * These globals are module-scoped and not exported, so each assertion drives
+ * the reset through observable behavior (re-migration / re-resolution) rather
+ * than reading internals directly.
+ */
+describe("closeDatabase resets migration + path globals (issue #2796)", () => {
+  const originalLaunchCwd = process.env[DAEMON_LAUNCH_CWD_ENV];
+  const originalDbDir = process.env.AUTOMOBILE_DB_DIR;
+  const originalDbPath = process.env.AUTOMOBILE_DB_PATH;
+  const originalMigrationsDir = process.env.AUTOMOBILE_MIGRATIONS_DIR;
+  const originalLegacyMigrationsDir = process.env.AUTO_MOBILE_MIGRATIONS_DIR;
+
+  function restore(key: string, value: string | undefined): void {
+    if (value === undefined) {
+      delete process.env[key];
+    } else {
+      process.env[key] = value;
+    }
+  }
+
+  const tempDirs: string[] = [];
+
+  afterEach(async () => {
+    restore(DAEMON_LAUNCH_CWD_ENV, originalLaunchCwd);
+    restore("AUTOMOBILE_DB_DIR", originalDbDir);
+    restore("AUTOMOBILE_DB_PATH", originalDbPath);
+    restore("AUTOMOBILE_MIGRATIONS_DIR", originalMigrationsDir);
+    restore("AUTO_MOBILE_MIGRATIONS_DIR", originalLegacyMigrationsDir);
+    for (const dir of tempDirs.splice(0)) {
+      await removeTempDirWithRetry(dir);
+    }
+  });
+
+  async function importFreshDatabaseModule() {
+    // Fresh module instance so its lazy globals are not shared with other tests.
+    return import(`../../src/db/database.ts?reset-test=${Date.now()}-${Math.random()}`);
+  }
+
+  async function makeTempDbDir(prefix: string): Promise<string> {
+    const dir = await mkdtemp(path.join(tmpdir(), prefix));
+    tempDirs.push(dir);
+    return dir;
+  }
+
+  async function removeTempDirWithRetry(dir: string): Promise<void> {
+    for (let attempt = 0; attempt < 100; attempt += 1) {
+      try {
+        await rm(dir, { recursive: true, force: true });
+        return;
+      } catch (error) {
+        const code = (error as NodeJS.ErrnoException).code;
+        if (code !== "EBUSY" && code !== "EPERM" && code !== "ENOTEMPTY") {
+          throw error;
+        }
+        await defaultTimer.sleep(50);
+      }
+    }
+    await rm(dir, { recursive: true, force: true });
+  }
+
+  function queryToolCalls(db: any) {
+    return db
+      .selectFrom("tool_calls" as any)
+      .selectAll()
+      .execute();
+  }
+
+  test("reopen after close re-runs migrations against a fresh DB (migrationsRun reset)", async () => {
+    const firstDir = await makeTempDbDir("auto-mobile-reset-first-");
+    process.env.AUTOMOBILE_DB_DIR = firstDir;
+    delete process.env[DAEMON_LAUNCH_CWD_ENV];
+
+    const databaseModule = await importFreshDatabaseModule();
+
+    // Cold start on the first DB: migrations run, migrationsRun becomes true.
+    const firstDb = databaseModule.getDatabase();
+    expect(await queryToolCalls(firstDb)).toEqual([]);
+
+    await databaseModule.closeDatabase();
+
+    // Point at a brand-new, empty DB dir and reopen in the same process.
+    const secondDir = await makeTempDbDir("auto-mobile-reset-second-");
+    process.env.AUTOMOBILE_DB_DIR = secondDir;
+
+    // The reopen must bind the fresh file (resolvedDbPath reset), otherwise the
+    // stale path cache silently redirects back to the first (already-migrated)
+    // DB and masks the migrationsRun bug below.
+    expect(databaseModule.getDatabasePath()).toBe(path.join(secondDir, "auto-mobile.db"));
+
+    // And against that fresh, empty file migrations must re-run (migrationsRun
+    // reset). If they don't, ensureMigrationsStarted() no-ops and the query
+    // hits an unmigrated schema -> "no such table: tool_calls".
+    const secondDb = databaseModule.getDatabase();
+    expect(await queryToolCalls(secondDb)).toEqual([]);
+
+    await databaseModule.closeDatabase();
+  });
+
+  test("reopen after close re-resolves the DB path under a changed env (resolvedDbPath reset)", async () => {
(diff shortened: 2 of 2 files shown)


kaeawc:
## Multi-agent audit + review summary

Ran `/auto-mobile-code-review` plus a 3-persona adversarial audit (SQLite concurrency correctness / test-quality-regression / architecture-direction). Consensus: **ship as-is — no confirmed bugs**, all CI green including `windows-latest`.

**Verified strengths**
- Mutation-tested the new tests: reverting the fix (or resetting *only* `resolvedDbPath`) makes each assertion fail with the right symptom (`no such table: tool_calls` / stale path / non-null error), so all four globals are independently load-bearing. test-1 deliberately asserts `getDatabasePath()` **before** the re-migration query so a stale path cache can't mask the `migrationsRun` reset.
- Confirmed the only `closeDatabase()` caller is `daemon.ts:1285` (shutdown-then-exit) — no mid-daemon-life close — so the `resolveDbPath()` "path always matches" invariant is preserved and the reset is inert in the daemon.
- No unhandled rejection / migration-lock leak from nulling an in-flight `migrationsPromise` (rejection consumed at `.then(,onRejected)`; file lock released in `runMigrations`' own `finally`).

**Deferred (filed as follow-ups, out of scope for a `database.ts`-only P2 hardening fix):**
- #2896 — `dbWriteBarrier`'s drain latch is the one shutdown global that does *not* cold-start on reopen, so "reopen == cold start" isn't fully true (latent; no in-process reopen path exists today).
- #2898 — no generation fence: an in-flight migration completing after `closeDatabase()` + reopen could stomp the next generation's `migrationsError` (unreachable today).
- #2900 — encapsulate the four migration/path globals into a single resettable state object to structurally prevent the "added a 5th global, forgot to reset it" regression class (YAGNI-gated future refactor).

**Acknowledged, not blocking:** test-1 runs ~112ms, over the <100ms unit-test guideline, because behaviorally proving the `migrationsRun` reset *requires* two real migration runs (cold-start on DB-A, then re-migration on a fresh DB-B — a same-DB reopen can't distinguish reset from non-reset). This is a real-FS integration test consistent with its sibling `databaseLazyPath.test.ts`; non-flaky across repeated runs.

Issue #2761, opened 2026-07-01
Output reduction 6: observe diff for non-observe tools (--actions-diff-observe)


kaeawc:
Part of the MCP output-context reduction effort. See the dependency graph across issues in this series.

## Goal
`--actions-diff-observe` (default off): non-observe tools emit a **diff** against the last observation shown to the agent, instead of the full embedded observation.

## Behavior
- Baseline = "last observation output to the agent," stored in `session.cacheData.customData.lastRenderedObservation`, updated whenever an observation is output.
- `observe` **always** emits the full (sanitized) observation and resets the baseline. Non-observe actions emit a diff vs the baseline, then update it.
- **Fall back to full** (no diff) when: `activeWindow`/package changed (cross-screen diff is meaningless), baseline missing, or no `sessionUuid` (legacy single-agent path).

## Node identity
Nodes have no stable ID, so diffing uses a synthetic key: `resource-id` + `bounds` + `text` + sibling index.

## Scope
- `ObserveResultOutput.ts` — `diffObserveResult(baseline, next, cfg)` producing a compact diff (added / removed / changed nodes + changed scalar fields).
- Baseline read/write in `src/daemon/sessionManager.ts` `cacheData` (depends on the repaired cache write from #2758).
- Applied in `finalizeToolResponse` for non-observe tools when the flag is set.

## Acceptance
- Unit tests: diff, baseline reset on `observe`, cross-screen full-emit, legacy no-session full-emit.
- Diff output format signed off on first real output before finalizing.

## Dependencies
- Config plumbing (#2756).
- Finalize hook + `lastHierarchy` repair (#2758).

kaeawc:
Folded into the consolidated spec **#3026** and delivered by merged **PR #3034** (squash `09e49a71d`) — `--actions-diff-observe` consumer wired at `finalizeToolResponse` with the `resource-id + bounds + text + sibling index` synthetic key (ancestor-path scoped), `lastRenderedObservation` baseline, and cross-screen/no-session fallback. Follow-ups: #3051 (real-device format sign-off), #3052 (focused/awaited mirror fields), #3053 (stable node identity + internal-consumer guard). Safe to close as a duplicate of #3026.

kaeawc:
[...]

PR #2891, opened 2026-07-03
feat(server): finalizeToolResponse chokepoint + lastHierarchy cache repair (#2758)

### LINKED ISSUE #2758: Output reduction 3: finalizeToolResponse hook + lastHierarchy cache repair

kaeawc:
## What

Wire `sanitizeObserveResult` (#2757) in at the single serialization chokepoint and repair the dormant `lastHierarchy` session-cache write.

Handlers pre-serialize via `createStructuredToolResponse(result)` into an MCP envelope `{ content: [{ type: "text", text }], structuredContent }`. This adds a `finalizeToolResponse(response, { name, sessionUuid })` hook at the one post-handler return (`src/server/toolRegistry.ts`) that shrinks the single `ObserveResult` a payload may carry and rewrites **both** representations so the wire text and `structuredContent` never disagree.

It also fixes a latent bug: the `lastHierarchy` cache write read `response.viewHierarchy`, but `response` is the envelope, so the read was always `undefined` and the cache was **never populated**. The #2761 diff baseline depends on it.

Closes #2758.

## Why

An `ObserveResult` (~25k tokens, with `elements` duplicating the tree) rides on both the top-level `observe` result and every action result's `.observation`. Sanitizing once at the chokepoint — where the tool `name` and `sessionUuid` are in scope — is the minimal, single-owner place to shrink the wire output without touching any of the internal consumers that read the in-memory result.

## How

- **`src/server/finalizeToolResponse.ts`** (new):
  - Extract payload — prefer `structuredContent`, else guarded `JSON.parse(content[0].text)`.
  - Locate the `ObserveResult` — the payload itself for `observe`, else `payload.observation` (gated on a `viewHierarchy` marker so arbitrary success payloads aren't cloned for nothing).
  - Apply `sanitizeObserveResult` with `cfg.dropElements` from `serverConfig.isObserveResultDropElementsEnabled()` (`trimNodes` default on). Sanitize deep-clones, so it is **output-only** — the handler's in-memory result and anything already cached from it are untouched.
  - Rewrite **both** `structuredContent` and `content[0].text` (via `stringifyToolResponse`) from the same object.
  - Safe no-op for non-envelope, image, non-JSON-text, and non-observe payloads.
- **`toolRegistry.ts`** — call `finalizeToolResponse(response, { name, sessionUuid })` at the return chokepoint; repair the cache read to `response.structuredContent.viewHierarchy` (and the parallel screenshot read). The cache write runs **before** finalize, so it keeps the full, untrimmed hierarchy the diff baseline needs while the wire copy is sanitized.

## Not in scope (deferred, see follow-up)

- **`toolRegistry.ts:417` `response?.found`** — the swipeOn scroll-position update reads `response?.found`, but `createStructuredToolResponse` only hoists `success`/`error` to the envelope, so `found` is always `undefined` and the block is dead. Same envelope-vs-`structuredContent` class as the `lastHierarchy` bug fixed here, but outside this issue's scope. Filed as a follow-up.

## Testing

- `test/server/finalizeToolResponse.test.ts` — unit coverage (interface-free pure fn): observe + action `.observation` sanitized in **both** `structuredContent` and text; text/structuredContent stay byte-identical; elements-drop honors the config gate; caller's in-memory result never mutated; pass-through for non-observe / image / non-JSON-text / null / primitive / hierarchy-less payloads; structuredContent-absent text fallback.
- `test/server/toolRegistry.lastHierarchyCache.test.ts` — integration (real `SessionManager`/`DevicePool`/`DaemonState` autolock recipe, `FakeTimer`/`FakeDeviceUtils`, no device I/O): `lastHierarchy` is actually populated after an `observe` and holds the **full** hierarchy, while the returned wire response is sanitized at the chokepoint.
- Full server suite: **701 pass, 0 fail**. `turbo run lint build` clean.

## Reviewed

Ran `/auto-mobile-code-review` on the branch diff: verified sanitize is output-only (cache keeps the untrimmed tree), the text/structuredContent rewrite can't diverge, and every pass-through path is a no-op. Flagged the `response?.found` latent bug as a follow-up (above).

### DIFF
--- a/src/server/finalizeToolResponse.ts
+++ b/src/server/finalizeToolResponse.ts
@@ -1,0 +1,115 @@
+import type { ObserveResult } from "../models/ObserveResult";
+import { sanitizeObserveResult, type SanitizeObserveConfig } from "../features/observe/output/ObserveResultOutput";
+import { serverConfig } from "../utils/ServerConfig";
+import { stringifyToolResponse } from "../utils/toolUtils";
+
+/**
+ * Context needed to finalize a tool response at the serialization chokepoint.
+ * `name` selects where the `ObserveResult` lives (top-level for `observe`, else
+ * `.observation`); `sessionUuid` is accepted for symmetry with the call site and
+ * future per-session gating, though the current sanitize config is global.
+ */
+export interface FinalizeToolResponseContext {
+  name: string;
+  sessionUuid?: string;
+}
+
+/**
+ * Single post-handler serialization hook (issue #2758). Handlers pre-serialize
+ * via `createStructuredToolResponse` into an MCP envelope
+ * `{ content: [{ type: "text", text }], structuredContent }`. This runs once at
+ * `toolRegistry.ts` return and shrinks the one `ObserveResult` the payload may
+ * carry, rewriting BOTH representations so the wire text and `structuredContent`
+ * never disagree.
+ *
+ * Output-only: `sanitizeObserveResult` deep-clones, so the handler's in-memory
+ * result (and anything already cached from it, e.g. `lastHierarchy`) is
+ * untouched. Non-envelope, image, non-JSON-text, and non-observe payloads pass
+ * through unchanged — a safe no-op.
+ */
+export function finalizeToolResponse<T>(response: T, ctx: FinalizeToolResponseContext): T {
+  if (!response || typeof response !== "object") {
+    return response;
+  }
+
+  const envelope = response as {
+    content?: Array<{ type?: string; text?: string }>;
+    structuredContent?: unknown;
+  };
+
+  // Prefer structuredContent; fall back to the serialized text part when a tool
+  // returned text only. Anything else (image parts, non-JSON text) is left alone.
+  const hasStructured =
+    envelope.structuredContent !== undefined &&
+    envelope.structuredContent !== null &&
+    typeof envelope.structuredContent === "object";
+
+  const textPart =
+    Array.isArray(envelope.content) &&
+    envelope.content[0]?.type === "text" &&
+    typeof envelope.content[0].text === "string"
+      ? envelope.content[0]
+      : undefined;
+
+  let payload: Record<string, unknown> | undefined;
+  if (hasStructured) {
+    payload = envelope.structuredContent as Record<string, unknown>;
+  } else if (textPart) {
+    try {
+      const parsed = JSON.parse(textPart.text as string);
+      if (parsed && typeof parsed === "object") {
+        payload = parsed as Record<string, unknown>;
+      }
+    } catch {
+      // Not JSON — nothing to sanitize, leave the response as-is.
+      return response;
+    }
+  }
+
+  if (!payload) {
+    return response;
+  }
+
+  const cfg: SanitizeObserveConfig = {
+    dropElements: serverConfig.isObserveResultDropElementsEnabled(),
+  };
+
+  // Locate the ObserveResult: the payload itself for `observe`, else `.observation`.
+  let sanitizedPayload: Record<string, unknown> | undefined;
+  if (ctx.name === "observe" && isObserveResult(payload)) {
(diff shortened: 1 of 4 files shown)


kaeawc:
## Review summary (self + 3-perspective adversarial pass)

Ran `/auto-mobile-code-review` plus three independent reviewers (correctness/regression, architecture/direction, test-quality). Consolidated outcome:

**Correctness/regression:** No confirmed bugs. Every edge case reproduced and verified as a safe no-op — array/primitive `structuredContent`, image parts, non-JSON text, `null`/primitive responses, `viewHierarchy: null`, `observation: null`. The cache-before-finalize ordering is safe: the cache captures the reference before finalize reassigns `structuredContent` to a fresh deep-clone, so `lastHierarchy` keeps the full untrimmed tree while the wire copy is trimmed (proven end-to-end).

**Architecture/direction:** Sound — the single chokepoint (both `name` and `sessionUuid` in scope) is the correct seam; sanitizing inside `createStructuredToolResponse` would violate the output-only invariant, and per-handler would scatter the concern. No double-serialization on the hot path (non-observe tools hit an early guard before any clone/re-stringify).

**Test-quality:** Added in `bb80ee2` per the review's one must-add + high-value hardening:
- **End-to-end `tapOn` action test** through the real `toolRegistry` handler chokepoint (the issue names both `observe` AND `tapOn`) — `.observation` sanitized in text + structuredContent, byte-identical.
- `lastScreenshot` cache-repair assertion (symmetric `structuredContent` read).
- perf-audit raw-dump stripping + GFXINFO diagnostics truncation, `awaitedElement` extras preservation, elements-drop on a nested `.observation`, and an array-shaped root node — all through the wiring layer.

Full server suite: **706 pass, 0 fail**. `turbo run lint build` clean.

### Deferred work (follow-up issues to be filed)
1. **`toolRegistry.ts` `response?.found`** (swipeOn scroll-position update) reads the envelope, but `createStructuredToolResponse` only hoists `success`/`error` — so `found` is always `undefined` and the block is dead. Same envelope-vs-`structuredContent` class as the `lastHierarchy` bug fixed here.
2. **`toolResultsNoStructuredContent` flag** (`ServerConfig.isToolResultsNoStructuredContentEnabled()`) is fully plumbed from CLI/env but read by nothing. `finalizeToolResponse` is now its natural (and arguably only correct) home as an envelope-level transform — track which sibling issue consumes it.

PR #2903, opened 2026-07-03
fix(server): swipeOn scroll-position update reads found off structuredContent (#2897)

### LINKED ISSUE #2897: Latent bug: swipeOn scroll-position update reads response?.found off the MCP envelope (always undefined)
## Summary

`src/server/toolRegistry.ts:417` reads `response?.found` (and `response?.success`) off the **MCP envelope**, but handlers pre-serialize via `createStructuredToolResponse`, which only hoists `success`/`error` to the envelope top level — never `found`. So `response?.found` is always `undefined` and the swipeOn scroll-position update block is **dead code**:

```ts
// After swipeOn executes with lookFor, update the tool call with scroll position
if (name === "swipeOn" && args.lookFor && response?.success && response?.found) {
  const scrollPosition = UIStateExtractor.createScrollPosition(args);
  ...
  scrollNavManager.updateScrollPosition(scrollPosition);
}
```

`swipeOn` returns `createStructuredToolResponse({ message, observation, ...result })` (`src/server/interactionTools.ts:765`), where `result.found` lands inside `structuredContent`, not on the envelope. `createStructuredToolResponse` (`src/utils/toolUtils.ts:98-122`) hoists only `success`/`error`. Net effect: `NavigationGraphManager.updateScrollPosition` is never called after a `swipeOn` with `lookFor`, so scroll-position tracking in the navigation graph silently does nothing.

## Provenance

This is the **same envelope-vs-`structuredContent` bug class** fixed for `lastHierarchy` in #2758 / PR #2891. It was discovered while reviewing that PR but is out of its scope.

## Proposed fix

Read from `structuredContent`, matching the #2758 repair:

```ts
if (name === "swipeOn" && args.lookFor && response?.structuredContent?.success && response?.structuredContent?.found) {
```

(Verify `success` placement: `createStructuredToolResponse` sets both `response.success` and `response.structuredContent.success`, so the envelope `success` read happens to work today — but `found` only exists under `structuredContent`. Reading both from `structuredContent` is consistent and correct.)

## Acceptance / test

Add a toolRegistry integration test (mirror `test/server/toolRegistry.lastHierarchyCache.test.ts`): register a `swipeOn` handler returning `createStructuredToolResponse({ success: true, found: true, observation, scrollIterations: 1 })`, call the handler with `args.lookFor` set, and assert `NavigationGraphManager.updateScrollPosition` was invoked (inject/spy the manager). Assert it is NOT invoked when `found` is false.

## Context

- Fixed sibling bug: PR #2891 (finalizeToolResponse chokepoint + lastHierarchy cache repair), issue #2758.
- Envelope shape: `createStructuredToolResponse` at `src/utils/toolUtils.ts:98`.
- swipeOn return: `src/server/interactionTools.ts:750-775`.

kaeawc:
## What

Fix the dead swipeOn scroll-position update block in `toolRegistry`, plus an identical sibling dead-read in `DefaultUIStateSetup` surfaced by adversarial review.

Both read `found` off the **MCP envelope**, but handlers pre-serialize via `createStructuredToolResponse`, which hoists only `success`/`error` to the envelope top level — `found` lives under `structuredContent`. So `response?.found` was always `undefined`:

1. **`src/server/toolRegistry.ts`** — the swipeOn scroll-position block never ran, so `NavigationGraphManager.updateScrollPosition` was never called after a `swipeOn` with `lookFor`. Scroll-position tracking in the navigation graph silently did nothing.
2. **`src/features/navigation/DefaultUIStateSetup.ts`** — `setupScrollPosition` executes swipeOn via the registry handler and read `result?.found` off the envelope. The success branch was dead: setup always logged "could not find target element" and returned `null` even on a successful scroll, dropping the swipeOn setup action from replay.

This is the follow-up explicitly deferred from #2758 / PR #2891 (which fixed the sibling `lastHierarchy` bug of the same class).

Closes #2897.

## Why

`swipeOn`'s real result (`ScrollUntilVisible.ts:160-164` / `344-348`) carries `success: true, found: true`, which `swipeOnHandler` spreads into `createStructuredToolResponse({ message, observation, ...result })`. `createStructuredToolResponse` (`src/utils/toolUtils.ts:98-122`) copies the payload to `structuredContent` and hoists only `success`/`error` to the top level — so `found` exists **only** under `structuredContent`. Reading it off the envelope was a guaranteed `undefined`, making the branch unreachable. (In `lookFor` mode a not-found scroll throws `ActionableError`, so the guard's real purpose is gating the found path — which was never reachable.)

## How

- **`src/server/toolRegistry.ts`** — read both `success` and `found` from `response.structuredContent`, matching the #2758 `lastHierarchy` repair. Comment added explaining the envelope-vs-`structuredContent` hazard. The block runs before `finalizeToolResponse`, so `structuredContent` is intact when read.
- **`src/features/navigation/DefaultUIStateSetup.ts`** — same repair for `result?.structuredContent?.success && result?.structuredContent?.found`, with the same explanatory comment.

## Testing

- `test/server/toolRegistry.scrollPositionUpdate.test.ts` (new) — integration coverage mirroring `toolRegistry.lastHierarchyCache.test.ts` (real `SessionManager`/`DevicePool`/`DaemonState` autolock recipe, `FakeTimer`/`FakeDeviceUtils`, no device I/O): `found: true` + `lookFor` → `updateScrollPosition` **is** invoked with a matching `ScrollPosition`; `found: false` → **not** invoked; `lookFor` absent → **not** invoked.
- `test/features/navigation/DefaultUIStateSetup.test.ts` (extended) — `setupScrollPosition` returns the `swipeOn(lookFor: …)` marker action on `found: true`, `null` on `found: false`.
- Both TDD-verified red→green (each new positive test fails against the pre-fix envelope read, passes after).
- Full navigation + server suites: **988 pass, 0 fail**. `turbo run lint build` clean.

## Reviewed

Ran `/auto-mobile-code-review` plus three adversarial subagent passes (correctness skeptic, test-quality auditor, architecture/scope). Correctness and test-quality passes found no defects. The architecture pass surfaced the `DefaultUIStateSetup` sibling (now fixed here) and recommended a systemic follow-up: a typed `getStructuredField` accessor / typed envelope so this whole bug class is caught at the type checker rather than per-site — filed as a follow-up issue.

### DIFF
--- a/src/features/navigation/DefaultUIStateSetup.ts
+++ b/src/features/navigation/DefaultUIStateSetup.ts
@@ -159,7 +159,13 @@ export class DefaultUIStateSetup implements UIStateSetup {
       // Execute swipeOn with lookFor
       const result = await swipeOnTool.handler(swipeOnArgs);
 
-      if (result?.success && result?.found) {
+      // `swipeOn` pre-serializes into an MCP envelope via createStructuredToolResponse,
+      // which hoists only `success`/`error` to the top level — `found` lives under
+      // `structuredContent`. Reading `result?.found` off the envelope was always
+      // undefined, so this success branch was dead and setup always logged the
+      // "could not find" warning even on a successful scroll (issue #2897; same class
+      // as the toolRegistry scroll-position and #2758 lastHierarchy fixes).
+      if (result?.structuredContent?.success && result?.structuredContent?.found) {
         logger.info(`[UI_STATE_SETUP] Successfully scrolled to target element`);
         return `swipeOn(lookFor: ${JSON.stringify(scrollPosition.targetElement)})`;
       } else {
--- a/src/server/toolRegistry.ts
+++ b/src/server/toolRegistry.ts
@@ -413,8 +413,13 @@ class ToolRegistryClass {
             getMcpRecorder()?.record(name, args);
           }
 
-          // After swipeOn executes with lookFor, update the tool call with scroll position
-          if (name === "swipeOn" && args.lookFor && response?.success && response?.found) {
+          // After swipeOn executes with lookFor, update the tool call with scroll position.
+          // Handlers pre-serialize into an MCP envelope, so `found` lives under
+          // `structuredContent` — `createStructuredToolResponse` only hoists
+          // `success`/`error` to the envelope top level. The former `response?.found`
+          // read was always undefined, so this block was dead and scroll-position
+          // tracking never fired (issue #2897; same class as the #2758 lastHierarchy fix).
+          if (name === "swipeOn" && args.lookFor && response?.structuredContent?.success && response?.structuredContent?.found) {
             const scrollPosition = UIStateExtractor.createScrollPosition(args);
             if (scrollPosition) {
               const scrollNavManager = sessionUuid
--- a/test/features/navigation/DefaultUIStateSetup.test.ts
+++ b/test/features/navigation/DefaultUIStateSetup.test.ts
@@ -1,10 +1,13 @@
-import { describe, expect, test } from "bun:test";
+import { afterEach, describe, expect, test } from "bun:test";
 import { DefaultUIStateSetup, type ObserveScreenLike } from "../../../src/features/navigation/DefaultUIStateSetup";
 import { FakeAdbClient } from "../../fakes/FakeAdbClient";
 import type { AdbClient } from "../../../src/utils/android-cmdline-tools/AdbClient";
 import type { BootedDevice } from "../../../src/models";
 import type { ObserveResult } from "../../../src/models/ObserveResult";
 import type { NavigationEdge } from "../../../src/features/navigation/NavigationGraphManager";
+import { ToolRegistry } from "../../../src/server/toolRegistry";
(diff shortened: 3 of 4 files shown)


kaeawc:
### Adversarial review pass — summary

Ran `/auto-mobile-code-review` plus three adversarial subagents (correctness skeptic, test-quality auditor, architecture/scope). Results:

- **Correctness**: no defects. Verified the `structuredContent` read fires exactly when intended; confirmed that in `lookFor` mode a not-found scroll *throws* `ActionableError` (never returns `found:false`), so the guard's real job is gating the found path.
- **Test quality**: solid. Verified the spy resolves the same session-scoped `NavigationGraphManager` instance production uses, clean cross-test state teardown, and TDD red→green.
- **Architecture**: surfaced the identical sibling dead-read in `DefaultUIStateSetup.setupScrollPosition` — **now fixed in this PR** with regression coverage. Also recommended a systemic fix (typed `getStructuredField` / typed envelope so this class is caught at the type checker), filed as follow-up #2907.

PR #2905, opened 2026-07-03
fix(db): cold-start dbWriteBarrier drain latch on closeDatabase (#2896)

### LINKED ISSUE #2896: chore(db): dbWriteBarrier drain latch does not cold-start on a same-process DB reopen
## Summary
Follow-up from PR #2889 (issue #2796). #2796 made `closeDatabase()` reset the migration/path module globals (`migrationsRun` / `migrationsPromise` / `migrationsError` / `resolvedDbPath`) so a **same-process reopen behaves like a cold start**. A multi-agent audit of that PR found one shutdown-state global that is deliberately *not* reset by `closeDatabase()` and therefore does **not** cold-start on reopen: the `dbWriteBarrier` drain latch.

## Detail
`src/db/dbWriteBarrier.ts` documents (around lines 30-33) that its draining flag "latches for the process lifetime … a same-process restart (tests) starts from a fresh barrier via `resetDbWriteBarrier()`." At daemon shutdown (`src/daemon/daemon.ts:1278`) `getDbWriteBarrier().drain(...)` is called before `closeDatabase()` (`daemon.ts:1285`) to quiesce in-flight best-effort writes (issue #2792 ordering).

If a code path ever calls `drain()`/`beginDrain()` and then reopens the DB **in the same process** (config reload, DB path switch, restart-without-exit) without calling `resetDbWriteBarrier()`, the barrier stays permanently draining and every tracked best-effort write is silently skipped (`dbWriteBarrier.ts:68-72`) against the reopened DB.

## Why it's not fixed in #2796
- `closeDatabase()` (in `src/db/database.ts`) intentionally does not own the write-barrier lifecycle — that's a separate concern with its own `resetDbWriteBarrier()` entry point.
- **No production impact today:** the only `closeDatabase()` caller is `daemon.ts:1285`, which is shutdown-then-`process.exit` (`logger.close()` immediately follows). There is no in-process reopen path, so the latch is never observed post-drain in a reopened process.
- Latent only: it becomes real the moment any feature reopens the DB in-process after a drain.

## Proposed direction (for whoever picks this up)
- Decide the ownership boundary: either have the DB-lifecycle reset (the future single `reset()` — see the encapsulation follow-up) also clear the write barrier via `resetDbWriteBarrier()`, or document explicitly that any in-process reopen path MUST call `resetDbWriteBarrier()` alongside `closeDatabase()`.
- Add a regression test: `beginDrain()`/`drain()` → `closeDatabase()` → reopen → assert a tracked best-effort write is NOT skipped.

## Acceptance
- [ ] A same-process reopen after a real `drain()` accepts tracked best-effort writes again (no permanent skip).
- [ ] Test covering drain → close → reopen → write-not-skipped.
- [ ] The "reopen behaves like a cold start" contract from #2796 is either fully true or its documented exception (barrier) is called out where reopen paths are added.

## Provenance
Found by the PR #2889 multi-agent correctness audit (Mira Kessler persona, finding #3), 2026-07-03. No confirmed live bug — latent/hardening, priority P3.

kaeawc:
## What

Follow-up to #2889 (issue #2796). `closeDatabase()` resets the migration/path
module globals (`migrationsRun` / `migrationsPromise` / `migrationsError` /
`resolvedDbPath`) so a **same-process reopen behaves like a cold start** — but
it deliberately left one shutdown-state global untouched: the `dbWriteBarrier`
drain latch.

`dbWriteBarrier`'s draining flag latches for the process lifetime. Daemon
shutdown calls `getDbWriteBarrier().drain(...)` (`daemon.ts:1278`) to quiesce
in-flight best-effort writes **before** `closeDatabase()` (`daemon.ts:1285`,
issue #2792 ordering). If any future path reopens the DB **in the same process**
after that drain (config reload, DB path switch, restart-without-exit) without
also cold-starting the barrier, the shared barrier stays permanently draining
and every tracked best-effort write is silently skipped
(`dbWriteBarrier.ts` `track()`) against the reopened DB.

`closeDatabase()` now clears the shared barrier via `resetDbWriteBarrier()`,
extending the "reopen behaves like a cold start" contract from #2796 to the
write barrier for consumers that resolve `getDbWriteBarrier()` at use-time.
Construction-captured consumers keep their pinned barrier and remain a
documented reopen exception (tracked in #2912).

Closes #2896.

## Why

**No production impact today**, latent hardening only (issue is P3):

- The sole `closeDatabase()` caller is `daemon.ts:1285`, terminal daemon
  shutdown-then-`process.exit` (`logger.close()` immediately follows). There is
  no in-process reopen path, so the latch is never observed post-drain in a
  reopened process.
- It becomes a real, silent bug the moment any feature reopens the DB in-process
  after a drain.

**Ownership decision** (the issue asked to pick one): rather than documenting an
exception that every future reopen path must remember, the DB-lifecycle reset
now **owns** clearing the write barrier. This matches the existing #2796 pattern
where `closeDatabase()` already resets every other shutdown-state global as a
set. No import cycle is introduced — `dbWriteBarrier.ts` imports only
`SystemTimer`/`logger`, neither of which imports `database.ts`.

**Daemon shutdown is unaffected.** `stop()` tears down every writer — all socket
servers, the health/heartbeat/device timers, and HTTP sessions — *before*
`drain()`/`closeDatabase()`, so no live writer exists in the tiny window between
the reset and `process.exit`. The extra reset is inert there.

## How

`src/db/database.ts` — `closeDatabase()` calls `resetDbWriteBarrier()` after the
existing migration/path global resets, so the barrier cold-starts alongside them:

```ts
migrationsRun = false;
migrationsPromise = null;
migrationsError = null;
resolvedDbPath = null;
resetDbWriteBarrier();
```

`src/db/dbWriteBarrier.ts` — updated the `beginDrain()` and `resetDbWriteBarrier()`
doc comments to reflect that `closeDatabase()` now clears the shared barrier on
every DB close, so a same-process reopen always sees a fresh, non-draining
barrier.

## Testing

- **`test/db/dbWriteBarrierResetOnClose.test.ts`** (new, TDD — written red first,
  then made green) covering the issue's acceptance criteria:
  - `drain() → closeDatabase() → reopen → tracked best-effort write is NOT
    skipped` — models the daemon shutdown ordering, then asserts the reopened
    barrier is a **distinct, non-draining** instance and a tracked write actually
    runs (returns its value) instead of short-circuiting.
  - `closeDatabase() clears a drain latch even with no open connection` — the
    reset is unconditional, mirroring the migration-global reset, so a
    partially-initialized shutdown still cold-starts the barrier.
- Follows repo conventions (fresh `database.ts` import per case for global
  isolation, temp dirs, `defaultTimer` for retry, `resetDbWriteBarrier()` in
  `beforeEach`/`afterEach` for suite isolation).
- Full `test/db/` suite: **454 pass, 0 fail**. `test/daemon/`: **578 pass, 0
  fail**. `turbo run lint build` clean.

## Follow-up (filed: #2912)

The injected-barrier consumers — `TelemetryRecorder` (singleton), `SessionManager`
(`daemon.ts:141`), `FailureAnalyticsRepository` (module-level `const`) — capture
`getDbWriteBarrier()` at construction. A future in-process reopen path must
reconstruct them (or they'd hold the stale, drained barrier). Moot today (no
reopen path exists); flagged in the `beginDrain()` doc comment and tracked in
#2912 (which also captures a self-limiting, already-swallowed closed-DB write
window in the daemon shutdown-exit path surfaced by the review).

### DIFF
--- a/src/db/database.ts
+++ b/src/db/database.ts
@@ -14,6 +14,7 @@ import {
   extractMissingPackageName,
   isMissingMigrationDependencyError,
 } from "./migrationDependencyIntegrity";
+import { resetDbWriteBarrier } from "./dbWriteBarrier";
 
 type BunDatabaseConstructor = typeof import("bun:sqlite").Database;
 type BunDatabase = import("bun:sqlite").Database;
@@ -302,6 +303,17 @@ export async function closeDatabase(): Promise<void> {
   migrationsPromise = null;
   migrationsError = null;
   resolvedDbPath = null;
+
+  // Cold-start the write barrier too (issue #2896, follow-up to #2796). Its
+  // draining flag latches for the process lifetime, so shutdown's
+  // `getDbWriteBarrier().drain(...)` before this close (issue #2792 ordering)
+  // would otherwise leave the shared barrier permanently draining. Any future
+  // in-process reopen (config reload / DB path switch / restart-without-exit)
+  // would then silently skip every tracked best-effort write against the
+  // reopened DB. Resetting here makes the "reopen behaves like a cold start"
+  // contract fully true instead of carrying a barrier-shaped exception. Inert
+  // in the daemon, where the only caller is shutdown-then-`process.exit`.
+  resetDbWriteBarrier();
 }
 
 /**
--- a/src/db/database.ts
+++ b/src/db/database.ts
@@ -310,9 +310,13 @@ export async function closeDatabase(): Promise<void> {
   // would otherwise leave the shared barrier permanently draining. Any future
   // in-process reopen (config reload / DB path switch / restart-without-exit)
   // would then silently skip every tracked best-effort write against the
-  // reopened DB. Resetting here makes the "reopen behaves like a cold start"
-  // contract fully true instead of carrying a barrier-shaped exception. Inert
-  // in the daemon, where the only caller is shutdown-then-`process.exit`.
+  // reopened DB. Resetting here extends the "reopen behaves like a cold start"
+  // contract to the write barrier for consumers that resolve
+  // `getDbWriteBarrier()` at use-time; construction-captured consumers keep
+  // their pinned barrier and remain the documented reopen exception (#2912).
+  // Inert in the daemon, where the only caller is
+  // shutdown-then-`process.exit`. If the migration/path resets above are ever
+  // consolidated into a single reset(), this must move with them.
   resetDbWriteBarrier();
 }
 
--- a/src/db/dbWriteBarrier.ts
+++ b/src/db/dbWriteBarrier.ts
@@ -27,9 +27,12 @@ export interface DbWriteBarrier {
 
   /**
    * Flip the draining flag so subsequent {@link track} calls short-circuit.
-   * The flag latches for the process lifetime: production exits after shutdown,
-   * and a same-process restart (tests) starts from a fresh barrier via
-   * {@link resetDbWriteBarrier} or an injected instance.
+   * The flag latches for the instance's lifetime; a fresh cold start comes from
+   * a new barrier. `closeDatabase()` clears the shared barrier via
+   * {@link resetDbWriteBarrier} on every DB close, so a same-process reopen —
+   * whether a real shutdown drain-then-reopen or a test restart — always sees a
+   * fresh, non-draining barrier (issue #2896). Injected instances are reset by
+   * their owner.
    */
   beginDrain(): void;
 
@@ -123,7 +126,13 @@ export function getDbWriteBarrier(): DbWriteBarrier {
   return sharedBarrier;
 }
 
-/** Reset the shared barrier (testing only). */
+/**
+ * Discard the shared barrier so the next {@link getDbWriteBarrier} lazily
+ * creates a fresh, non-draining one. Called by `closeDatabase()` so a
+ * same-process DB reopen cold-starts the barrier (issue #2896) — its draining
+ * flag would otherwise latch for the process lifetime after a shutdown drain —
+ * and by tests for isolation.
+ */
 export function resetDbWriteBarrier(): void {
   sharedBarrier = null;
 }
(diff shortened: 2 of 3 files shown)


kaeawc:
## Multi-agent adversarial review (3 personas)

Ran three independent adversarial reviewers on this PR before merge — a concurrency skeptic, a completeness auditor, and a design/maintainability critic.

**Consensus:** ship-worthy, latent hardening with no live consumer. No blocker, no confirmed live bug.

**Addressed in this PR:**
- **Doc over-claim (all three, major):** the comments claimed the cold-start contract is "fully true" / a reopen "always sees a fresh, non-draining barrier." That is false for the three construction-captured consumers (`TelemetryRecorder`, `FailureAnalyticsRepository`, `SessionManager`), which pin the barrier at construction and don't observe the identity-swap. Corrected the `beginDrain()` / `resetDbWriteBarrier()` JSDoc and the `closeDatabase()` comment to scope the contract to use-time consumers and name the captured-consumer exception (commit baed7b55a).

**Filed as follow-up #2912 (deferred, P3):**
- Captured-reference staleness — decide per-call resolution vs. in-place `reset()` vs. consumer reconstruction when an in-process reopen path is actually added.
- A self-limiting, already-`logger.warn`-swallowed closed-DB write window in the daemon shutdown-exit path (`AndroidCtrlProxyClient.markInstalledAppsStale` resolves the barrier fresh), microseconds before `process.exit`.

**Verified NOT problems:** no concurrency/reentrancy hazard (`resetDbWriteBarrier()` is a synchronous field assignment, cannot interleave with an in-flight `drain()`/`track()`); the tests genuinely exercise the shared singleton (the fresh-`database.ts`-import still resolves the same `dbWriteBarrier` module — probed empirically); the design decision to keep the null-swap over in-place reset is correct for the only production caller (in-place reset would widen the daemon exit window).

PR #2906, opened 2026-07-03
fix(db): store evaluated timestamp for datetime('now') column defaults (#2895)

### LINKED ISSUE #2895: fix(db): migration columns store literal "datetime('now')" instead of a timestamp (48 columns, 24 migrations)
## Summary

Every migration column declared with `col.defaultTo("datetime('now')")` stores the **literal string** `datetime('now')` as its default value, not an evaluated timestamp. Kysely treats a plain string argument to `.defaultTo(...)` as a **value/parameter**, not raw SQL — so the column DDL becomes `DEFAULT 'datetime(''now'')'` instead of `DEFAULT (datetime('now'))`. Any row inserted without an explicit value for these columns gets the useless text `datetime('now')` in what is meant to be a timestamp.

Discovered during review of #2790 / PR #2892 (a reviewer flagged it as an out-of-scope aside; PR #2892 does **not** touch it).

## Reproduction (verified)

```ts
// against a real migrated in-memory bun:sqlite (test/db/testDbHelper.ts)
const db = await createTestDatabase();
await db.insertInto("navigation_apps").values({ app_id: "x", updated_at: "u" }).execute();
const row = await db.selectFrom("navigation_apps").selectAll()
  .where("app_id", "=", "x").executeTakeFirstOrThrow();
console.log(JSON.stringify(row.created_at));
// => "datetime('now')"     ← literal text, not a timestamp
```

## Scope

`grep -rn "defaultTo(\"datetime('now')\")" src/db/migrations/` → **48 columns across 24 migration files** (every `created_at`, several `updated_at`/timestamp columns). There are **zero** correct `sql\`...\``-tagged defaults in the migrations for contrast, so the pattern is uniformly wrong wherever it appears. Affected files:

```
2025_12_28_000_initial_schema.ts            2026_01_14_000_appearance_config.ts
2025_12_30_000_performance_thresholds.ts    2026_01_25_001_test_run_details.ts
2025_12_30_001_navigation_graph.ts          2026_01_27_000_crash_anr_monitoring.ts
2025_12_31_000_accessibility_baselines.ts   2026_01_27_000_failures.ts
2025_12_31_001_memory_audit.ts              2026_01_29_000_named_nodes.ts
2026_01_02_000_prediction_history.ts        2026_01_30_000_performance_live_metrics.ts
2026_01_03_000_feature_flags.ts             2026_03_15_000_telemetry_events.ts
2026_01_03_000_test_executions.ts           2026_03_18_000_navigation_events.ts
2026_01_05_000_tool_calls.ts                2026_03_19_000_storage_events.ts
2026_01_08_000_test_coverage.ts             2026_03_19_001_layout_events.ts
2026_01_09_000_video_recordings.ts          2026_04_01_000_drop_custom_events.ts
2026_01_10_000_device_snapshots.ts          2026_04_02_000_device_sessions.ts
```

## Impact

- Any `created_at`/timestamp that relies on the column default (rather than an explicit app-supplied value) is garbage text `datetime('now')`. In practice most write paths pass an explicit ISO/`Date.now()` value, which masks the bug — but any code that reads such a defaulted column (ordering by `created_at`, age/retention math, diagnostics) silently mis-behaves.
- Low blast radius today precisely because most inserts supply the value; this is a **latent correctness landmine** for any new column/insert path that leans on the default.
- Note the interaction with PR #2892: `getOrCreateApp` now returns the DB row's `created_at` (via `returningAll()`), so it returns `"datetime('now')"` on the create path — but its only caller (`NavigationGraphManager.ts:264`) discards the return value, so there is no functional regression from #2892. This issue is strictly pre-existing.

## Proposed fix

Replace the string literal with a raw-SQL default so SQLite evaluates the function at insert time:

```ts
import { sql } from "kysely";
// ...
.addColumn("created_at", "text", col => col.notNull().defaultTo(sql`(datetime('now'))`))
```

- Sweep all 48 columns across the 24 files.
- **Migration strategy:** SQLite cannot alter a column default in place cheaply, and existing rows already store the literal text. Decide whether to (a) only fix forward (new tables/columns emit the correct default) plus a one-off data-repair migration that rewrites rows whose timestamp column literally equals `datetime('now')`, or (b) accept that historical defaulted rows stay wrong and just fix the DDL going forward. Given most inserts pass explicit values, a targeted repair migration is likely low-risk.
- **Add a guard test:** insert a row omitting the defaulted column against the real migrated schema and assert the stored value parses as a real timestamp (not the literal string) — this both proves the fix and prevents regression. Consider a lint/CI grep that fails on `defaultTo("datetime('now')")` to stop the pattern reappearing.

## Acceptance criteria
- [ ] No migration uses `defaultTo("datetime('now')")` (string literal); all use `defaultTo(sql\`(datetime('now'))\`)` or equivalent.
- [ ] A test inserts a row without the defaulted column and asserts a real timestamp is stored (real in-memory `bun:sqlite`, per `test/db/testDbHelper.ts`).
- [ ] Decision recorded on repairing existing literal-`datetime('now')` rows (forward-fix vs. data migration).
- [ ] Guard (lint grep or test) prevents the pattern from returning.

## References
- Root-caused during PR #2892 review (issue #2790). SQLite/Kysely reviewer noted: "`defaultTo("datetime('now')")` is a string-literal default, not a `sql\`\`` expression, so it stores the literal text — a pre-existing quirk, out of scope for #2790."
- Kysely `defaultTo` docs: a non-`sql`-tagged string is bound as a value, not SQL.
- Milestone: Database (#2803 audit tracking).

kaeawc:
## What

Every migration column declared with `col.defaultTo("datetime('now')")` stored the **literal string** `datetime('now')` as its default, not an evaluated timestamp. Kysely binds a plain string argument to `.defaultTo(...)` as a **value/parameter**, so the DDL became `DEFAULT 'datetime(''now'')'` instead of `DEFAULT (datetime('now'))`. Any row inserted without an explicit value for these columns got the useless text `datetime('now')` in what is meant to be a timestamp.

This sweeps all **48 `datetime('now')` columns across 24 files** plus **1 `CURRENT_TIMESTAMP` column** (`recomposition_metrics`, the identical bug with a different literal, caught in review) to `defaultTo(sql`(<expr>)`)` so SQLite evaluates the function at insert time, adds a **data-repair migration** that heals rows already poisoned by either literal, and adds a shell/bats regression guard plus a runtime behavior test.

Closes #2895.

## Why

Low blast radius today only because most write paths pass an explicit ISO/`Date.now()` value, which masks the bug. But any code that reads a defaulted column (ordering by `created_at`, age/retention math, diagnostics) silently mis-behaves, and it is a latent correctness landmine for any new column/insert path that leans on the default. Root-caused during the #2790 / PR #2892 review as an out-of-scope aside.

Verified before/after DDL against a real migrated in-memory `bun:sqlite`:
- Before: `created_at` text default `'datetime(''now'')'` → stored value `"datetime('now')"`.
- After: `created_at` text default `(datetime('now'))` → stored value `2026-07-03 12:00:00` (parses as a real timestamp).

## How

**Forward DDL fix (all 48 columns).** Replaced `defaultTo("datetime('now')")` with `defaultTo(sql`(datetime('now'))`)` and added the `sql` value import where the file only imported `type Kysely` (three files already imported `sql`, left untouched).

**Editing historical migrations is safe.** Kysely's migrator identifies executed migrations by **name only** (`migrator.js:454`) — there is no content checksum, and its "corrupted migrations" guard fires solely on missing/out-of-order *names*. So editing an already-run migration body never re-runs it and never trips recovery: existing DBs keep their old column default (their poisoned rows are healed by the repair migration below), and freshly-created/replayed DBs get the corrected DDL.

**Repair decision: forward-fix + a data-repair migration.** `2026_07_03_000_repair_datetime_now_defaults.ts` introspects the live schema (`sqlite_master` + `pragma_table_info`) and, for every TEXT-affinity column, runs `UPDATE <t> SET <c> = datetime('now') WHERE <c> = 'datetime(''now'')'`. Generic introspection guarantees completeness (a defaulted column can never be silently missed the way a hand-maintained list could); the exact-match predicate makes it safe — no legitimate write path stores that exact string. Wrapped in one `db.transaction()` for atomicity (matching the `2026_07_01_000_failure_groups_signature_unique` convention). Idempotent and a no-op on fresh/replayed schemas, so it survives the destructive-recovery replay.

## Testing

- `test/db/migrationDatetimeDefaults.test.ts`
  - **Grep guard** (regression backstop): scans all migration sources and fails if `defaultTo("datetime('now')")` reappears. Confirmed **red** (50 offenders) before the fix.
  - **Behavior**: insert into `navigation_apps` omitting the defaulted `created_at` against the real migrated in-memory `bun:sqlite`, assert the stored value is not the literal and `Date.parse` succeeds. Confirmed **red** before the fix.
- `test/db/repairDatetimeNowDefaultsMigration.test.ts` — seeds a poisoned literal row, runs the repair, asserts it becomes a real timestamp; asserts legitimate ISO timestamps are left untouched; asserts idempotency.
- `bun test test/db/` green: **462 pass, 0 fail** (incl. the full migration-chain test with the new migration). `eslint` and `bun build.ts` clean.

## Acceptance criteria (from #2895)

- [x] No migration passes a SQL time-expression as a string literal to `defaultTo(...)` (`datetime('now')` **and** `CURRENT_TIMESTAMP`); all use `defaultTo(sql`(<expr>)`)`.
- [x] A test inserts a row without the defaulted column and asserts a real timestamp is stored (real in-memory `bun:sqlite`).
- [x] Decision recorded on repairing existing literal-`datetime('now')` rows: **forward-fix DDL + generic data-repair migration**.
- [x] Guard prevents the pattern from returning: `scripts/validate-no-datetime-now-literal.sh` + `test/bats/validate-no-datetime-now-literal.bats`, wired into `all_fast_validate_checks.sh` (matches the `validate-no-debug-log-tags.sh` convention); rejects the whole class of SQL time-expressions-as-strings, not just the two seen.

### DIFF
--- a/scripts/all_fast_validate_checks.sh
+++ b/scripts/all_fast_validate_checks.sh
@@ -40,6 +40,7 @@ add_check "codex-skills" "\"$PROJECT_ROOT/scripts/validate_codex_skills.sh\"" "c
 add_check "lychee" "\"$PROJECT_ROOT/scripts/lychee/validate_lychee.sh\"" "docs,links" "Validate documentation links"
 add_check "dependabot" "\"$PROJECT_ROOT/scripts/validate_dependabot.sh\"" "config,yaml" "Validate Dependabot config"
 add_check "debug-tags" "\"$PROJECT_ROOT/scripts/validate-no-debug-log-tags.sh\"" "lint" "Reject stray [*-DEBUG] log tags in src/"
+add_check "datetime-now-literal" "\"$PROJECT_ROOT/scripts/validate-no-datetime-now-literal.sh\"" "lint" "Reject string-literal SQL time-expression defaults in migrations"
 
 print_usage() {
   cat <<'EOF'
--- a/scripts/validate-no-datetime-now-literal.sh
+++ b/scripts/validate-no-datetime-now-literal.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/env bash
+#
+# Guard against a SQL time-expression column default being passed as a plain
+# string in migrations.
+#
+# `col.defaultTo("datetime('now')")` (or `"CURRENT_TIMESTAMP"`) passes a plain
+# string to Kysely, which binds it as a VALUE — the DDL becomes
+# DEFAULT 'datetime(''now'')' / DEFAULT 'CURRENT_TIMESTAMP', so any defaulted row
+# stores the literal text instead of an evaluated timestamp. The correct form is
+# defaultTo(sql`(datetime('now'))`). This check fails CI if the broken pattern
+# returns for any SQL-time expression (not just the two that triggered #2895).
+#
+# See issue #2895 (48 datetime('now') columns + 1 CURRENT_TIMESTAMP column stored
+# the literal string across 25 migrations).
+
+set -euo pipefail
+
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+MIGRATIONS_DIR="${ROOT_DIR}/src/db/migrations"
+
+# Match `defaultTo("<sql-time-expr>...")` where the string literal opens with a
+# SQL date/time function or CURRENT_* keyword — the shape that binds as a value
+# instead of raw SQL. Legitimate value defaults ("{}", "", "success") do not
+# match. The correct sql`...`-tagged form is not a string literal, so it is
+# excluded. Case-insensitive so CURRENT_TIMESTAMP / current_timestamp both trip.
+PATTERN="defaultTo\\(\\s*[\"'](datetime\\(|strftime\\(|date\\(|time\\(|julianday\\(|unixepoch\\(|CURRENT_TIMESTAMP|CURRENT_TIME|CURRENT_DATE)"
+
+matches="$(grep -rEni --include='*.ts' "$PATTERN" "$MIGRATIONS_DIR" || true)"
+
+if [[ -n "$matches" ]]; then
+  echo "error: SQL time-expression passed as a string literal to defaultTo(...) in migrations — use defaultTo(sql\`(<expr>)\`) so SQLite evaluates it instead of storing the literal text:" >&2
+  echo "$matches" >&2
+  exit 1
+fi
+
+echo "No string-literal SQL time-expression defaults found in migrations."
--- a/src/db/migrations/2025_12_28_000_initial_schema.ts
+++ b/src/db/migrations/2025_12_28_000_initial_schema.ts
@@ -1,4 +1,4 @@
-import type { Kysely } from "kysely";
+import { type Kysely, sql } from "kysely";
 
 export async function up(db: Kysely<unknown>): Promise<void> {
   // Create device_configs table
@@ -11,7 +11,7 @@ export async function up(db: Kysely<unknown>): Promise<void> {
     .addColumn("active_mode", "text")
     .addColumn("config_json", "text", col => col.notNull().defaultTo("{}"))
     .addColumn("created_at", "text", col =>
-      col.notNull().defaultTo("datetime('now')")
+      col.notNull().defaultTo(sql`(datetime('now'))`)
     )
     .addColumn("updated_at", "text", col => col.notNull())
     .execute();
--- a/src/db/migrations/2025_12_30_000_performance_thresholds.ts
+++ b/src/db/migrations/2025_12_30_000_performance_thresholds.ts
@@ -1,4 +1,4 @@
-import type { Kysely } from "kysely";
+import { type Kysely, sql } from "kysely";
 
 export async function up(db: Kysely<unknown>): Promise<void> {
   // Create performance_thresholds table
@@ -19,7 +19,7 @@ export async function up(db: Kysely<unknown>): Promise<void> {
     .addColumn("touch_latency_threshold_ms", "real", col => col.notNull())
     .addColumn("weight", "real", col => col.notNull().defaultTo(1.0)) // For weighted average
     .addColumn("created_at", "text", col =>
-      col.notNull().defaultTo("datetime('now')")
+      col.notNull().defaultTo(sql`(datetime('now'))`)
     )
     .addColumn("ttl_hours", "integer", col => col.notNull().defaultTo(24)) // TTL in hours
     .execute();
@@ -70,7 +70,7 @@ export async function up(db: Kysely<unknown>): Promise<void> {
     .addColumn("touch_latency_ms", "real")
     .addColumn("diagnostics_json", "text") // JSON blob for detailed diagnostics
     .addColumn("created_at", "text", col =>
-      col.notNull().defaultTo("datetime('now')")
+      col.notNull().defaultTo(sql`(datetime('now'))`)
     )
     .execute();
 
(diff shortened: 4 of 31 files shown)


kaeawc:
## Multi-perspective review — findings addressed

Ran an AutoMobile code review plus three independent review agents (DB-engineer, regression/QA, architecture). Summary of what landed in `8c5b3f61d`:

**🐞 Confirmed bug (DB-engineer) — incomplete sweep.** `recomposition_metrics.created_at` used `defaultTo("CURRENT_TIMESTAMP")` — the identical "string binds as a value" bug with a different literal, which the original sweep, repair predicate, and guard all missed. Fixed the DDL to `defaultTo(sql`(CURRENT_TIMESTAMP)`)`, broadened the repair to heal rows equal to **either** `datetime('now')` or `CURRENT_TIMESTAMP`, and broadened the guard to reject the whole class of SQL time-expressions passed as strings (`datetime(`, `strftime(`, `date(`, `time(`, `CURRENT_TIMESTAMP/TIME/DATE`, …). A full audit (`grep 'defaultTo("…")'` over migrations) confirms the only remaining string-literal defaults are legitimate values (`""`, `"{}"`, `"success"`).

**🧭 Convention (architecture) — guard home.** Moved the source-grep guard from a `bun test` into `scripts/validate-no-datetime-now-literal.sh` + `test/bats/validate-no-datetime-now-literal.bats`, wired into `all_fast_validate_checks.sh`, matching the established `validate-no-debug-log-tags.sh` pattern (CLAUDE.md: shell guards live under `scripts/`). The runtime behavior test stays in `bun test`.

**🕰 Format decision (architecture + QA).** Kept `datetime('now')` (SQLite `YYYY-MM-DD HH:MM:SS`) rather than switching to ISO — it matches the codebase's other server-evaluated-now writes (`sql`datetime('now')`` in `ThresholdManager`/`MemoryThresholdManager`); switching to ISO would create a *new* divergence from those siblings. The DB-engineer agent audited every `orderBy`/comparison site and found **no** column that is both defaulted and written explicit-ISO at a compared site, so there's no ordering hazard today. Decision recorded in the repair migration docblock.

**✅ Test gaps (QA) closed.** Added: `CURRENT_TIMESTAMP` heal test, `down()` no-op coverage, a chain-integration test (asserts the repair migration is in the executed `kysely_migration` history), and a `recomposition_metrics` defaulted-insert behavior test.

**Verified clean:** `bun test test/db/` 460 pass · shellcheck + bats (4/4) · eslint 0 · `bun build.ts` ok. Kysely identifies migrations by name only (no content checksum), so editing historical migration bodies is safe — existing DBs keep their old default and are healed by the repair migration; fresh/replayed DBs get the corrected DDL.

Non-blocking note (QA/DB-engineer): several `updated_at` columns are `.notNull()` with no default (asymmetric with their now-defaulting `created_at` siblings). Pre-existing, not a bug (all writers supply `updated_at` explicitly in ISO) — filing a follow-up rather than expanding this PR's scope.

kaeawc:
Both Codex inline findings on this PR are addressed in follow-up #2915 (they landed here before the inline comments were seen — inline comments don't block auto-merge, and the repair migration was still unreleased so it's corrected in place):

- **P1 (re-poisoning on upgraded DBs):** the repair now **rebuilds** each column whose stored default is the broken literal (SQLite table-rebuild), so `getOrCreateApp` and other default-path inserts stop re-storing `datetime('now')` — not just a one-shot row fix.
- **P2 (arbitrary-data corruption):** the row repair is now scoped via `pragma_table_info.dflt_value` to columns that actually carry the broken default, so `storage_events.value` and other arbitrary-content columns are never touched.

Thanks Codex — both were real.

Issue #2893, opened 2026-07-03
perf(db): drop now-redundant single-column device indexes on telemetry event tables (follow-up to #2788)


kaeawc:
## Summary

Follow-up to #2788 / PR #2890, which added composite `(device_id, timestamp)` indexes (`idx_<table>_device_timestamp`) to the six telemetry event tables (`network_events`, `log_events`, `os_events`, `navigation_events`, `storage_events`, `layout_events`). PR #2890 was deliberately kept **additive-only** per #2788's explicit scope ("do not modify the existing single-column indexes"). This issue tracks the deferred optimization it left on the table: the standalone single-column `idx_<table>_device` indexes are now **functionally redundant** and should be dropped to cut per-insert write amplification.

## Background

A composite index `(device_id, timestamp)` serves any query a standalone `(device_id)` index can, via the left-prefix rule — equality on `device_id`, `device_id` as a covering column, and `COUNT(*) WHERE device_id=?`. Verified during PR #2890's review with `EXPLAIN QUERY PLAN` against `network_events` (2000 rows): with the composite present and the standalone `idx_network_events_device` **absent**, every `device_id=?` access path is served by the composite (bare `WHERE device_id=?` → `SEARCH ... USING INDEX idx_network_events_device_timestamp (device_id=?)`; `COUNT(*) WHERE device_id=?` → covering scan of the composite). There is no `ANALYZE`/`sqlite_stat1` in `src/db/`, so the planner chooses structurally and deterministically.

Consequently each of the six tables currently carries **4 indexes where 3 suffice** (`timestamp`, the category column, the redundant `device_id`, and the new composite). At the 10k-row retention cap with a high insert rate, the redundant `device_id` B-tree is maintained on every insert for no read benefit.

## Important distinction (do NOT drop the timestamp index)

The standalone **`timestamp`** index must stay: the retention cutoff query (`ORDER BY timestamp DESC LIMIT 1 OFFSET RETENTION_MAX_ROWS`, no device filter — e.g. `src/db/networkEventRepository.ts:184-190`) uses it as a covering scan, and the composite **cannot** substitute because `device_id` leads the composite. Only the standalone `device_id` index is redundant.

The category-column indexes (`idx_network_events_host`, `idx_log_events_tag`, `idx_os_events_category`) are also out of scope — different query shapes.

## The redundant indexes to drop

- `idx_network_events_device` (`src/db/migrations/2026_03_15_000_telemetry_events.ts:54-58`)
[...]

PR #2904, opened 2026-07-03
perf(db): drop now-redundant single-column device indexes on telemetry event tables (#2893)

kaeawc:
## What

Follow-up to #2788 / PR #2890, which added composite `(device_id, timestamp)` indexes (`idx_<table>_device_timestamp`) to the six telemetry event tables but was deliberately kept **additive-only** per #2788's scope ("do not modify the existing single-column indexes"). This PR completes the optimization it deferred: it drops the six now-redundant standalone `idx_<table>_device` indexes to cut per-insert write amplification.

New forward-only migration `2026_07_03_000_drop_redundant_device_indexes.ts` (sorts after `2026_07_02_000_event_composite_indexes.ts`) drops the six device indexes in `up()` and recreates them in `down()`.

Closes #2893.

## Why

A composite `(device_id, timestamp)` index serves every access path a standalone `(device_id)` index can, via SQLite's left-prefix rule — so each of the six tables was carrying **4 indexes where 3 suffice** (`timestamp`, the category column, the redundant `device_id`, and the composite). At the retention cap with a high insert rate, the redundant `device_id` B-tree was maintained on every insert for no read benefit.

There is no `ANALYZE`/`sqlite_stat1` in `src/db/`, so the planner chooses structurally and deterministically — the composite is picked for `device_id=?` access whether or not the standalone index exists.

## How

Dropped indexes (one per table), confirmed against each source migration:

- `idx_network_events_device`, `idx_log_events_device`, `idx_os_events_device` (`2026_03_15_000_telemetry_events.ts`)
- `idx_navigation_events_device` (`2026_03_18_000_navigation_events.ts`)
- `idx_storage_events_device` (`2026_03_19_000_storage_events.ts`)
- `idx_layout_events_device` (`2026_03_19_001_layout_events.ts`)

`up()` uses `.dropIndex(...).ifExists()`; `down()` recreates each with `.createIndex(...).ifNotExists().column("device_id")` — byte-identical to the base migrations — so the migration is idempotent, reversible, and replay-safe under #2785's destructive-recovery path.

**Deliberately left intact** (different query shapes, explicitly out of scope):
- The standalone **`timestamp`** indexes — the retention cutoff (`ORDER BY timestamp DESC LIMIT 1 OFFSET <cap>`, no device filter) uses them as a covering scan, and the composite **cannot** substitute because `device_id` leads it.
- The category/content indexes (`idx_network_events_host`, `idx_log_events_tag`, `idx_os_events_category`).

## Testing

[...]

Issue #2788, opened 2026-07-01
perf(db): add composite (device_id, timestamp) indexes to telemetry event tables to eliminate ORDER BY filesorts


kaeawc:
## Summary

Every telemetry event getter queries `WHERE device_id=? [AND timestamp>=?] ORDER BY timestamp DESC LIMIT n`, but the six event tables (`network_events`, `log_events`, `os_events`, `navigation_events`, `storage_events`, `layout_events`) only have **single-column** indexes on `device_id` and `timestamp` separately. SQLite can use at most one of them per query, so the common `device_id` filter forces a temp B-tree filesort for the `ORDER BY timestamp DESC`. Adding a composite `(device_id, timestamp)` index per table lets SQLite satisfy both the filter and the ordering from one index. Found via a DB concurrency/migration audit (2026-07-01), finding P3idx, priority P1, verified against source.

## Background & history

- Introduced: commit `d20e1f0e` (2026-03-16), [PR #1535](https://github.com/kaeawc/auto-mobile/pull/1535) — created `network_events`/`log_events`/`os_events` (plus the now-dropped `custom_events`) with one index each on `timestamp`, one on `device_id`, and one on a category column.
- The sibling tables followed the same single-column pattern: `navigation_events` in `2026_03_18_000_navigation_events.ts`, `storage_events` in `2026_03_19_000_storage_events.ts`, and `layout_events` in `2026_03_19_001_layout_events.ts` — all added by commit `fdb94336` (2026-03-23), [PR #1571](https://github.com/kaeawc/auto-mobile/pull/1571).
- The getter query shape (`getNetworkEvents`) was finalized in commit `73130e51` (2026-03-28), [PR #1624](https://github.com/kaeawc/auto-mobile/pull/1624).
- Why it's shaped this way: each table was indexed by copy-pasting the same "one index per hot column" template. Nobody added the composite because the retention cap (10k rows) keeps tables small, so the filesort was never a visible bottleneck. Notably, the repo already establishes the correct pattern: `2025_12_30_000_performance_thresholds.ts:78-83` creates `idx_performance_audit_results_device_timestamp` via `.columns(["device_id", "timestamp"])`.

## The problem

All six repositories share an identical getter shape. Example:

```ts
// src/db/networkEventRepository.ts:141-167
let q = getDb(db).selectFrom("network_events").selectAll();

if (query.deviceId) {
  q = q.where("device_id", "=", query.deviceId);
}
[...]

PR #2918, opened 2026-07-03
test(db): add CI drift guard for telemetry event tables' index inventory (#2908)

### LINKED ISSUE #2908: perf(db): add a CI drift guard for the telemetry event tables' index inventory (follow-up to #2893)
## Summary

Follow-up surfaced during the multi-persona review of PR #2904 (closes #2893, which dropped the six now-redundant single-column `idx_<table>_device` indexes on the telemetry event tables).

Two independent reviewers (test-quality + DBA personas) noted that **no CI drift guard exists** to keep the telemetry event tables' index inventory consistent over time. The composite-index migration (#2890 / `2026_07_02_000_event_composite_indexes.ts`) and the drop migration (#2904 / `2026_07_03_000_drop_redundant_device_indexes.ts`) both rely on the reviewer/author remembering the intended per-table index set. Nothing stops a future migration from silently re-adding an `idx_<table>_device` index (re-introducing the exact write-amplification #2893 removed) or dropping a `timestamp` index the retention cutoff depends on.

This is deliberately **out of scope** for #2893/#2904 (which were narrowly scoped to the additive→drop pair per #2788), and is filed on its own so a drift guard is designed and reviewed independently.

## Desired outcome

A test (or lightweight CI check) that asserts the canonical index inventory for the six telemetry event tables (`network_events`, `log_events`, `os_events`, `navigation_events`, `storage_events`, `layout_events`) after a full `runMigrations` replay, so any future migration that perturbs it must consciously update the guard.

Expected per-table inventory today (post-#2904):
- `idx_<table>_timestamp` (retention cutoff covering scan — **must stay**)
- `idx_<table>_device_timestamp` (composite; serves every `device_id=?` access path)
- Category/content index where applicable: `idx_network_events_host`, `idx_log_events_tag`, `idx_os_events_category`
- **No** standalone `idx_<table>_device` (dropped in #2904)

## Evaluation criteria

- [ ] A test builds the full schema via `runMigrations` and asserts `PRAGMA index_list` for each of the six tables equals the canonical set above (ignoring the auto-created PK/rowid index).
- [ ] The test fails loudly if a standalone `idx_<table>_device` reappears or a `timestamp`/composite index is removed.
- [ ] Runs <100ms against `:memory:` (per CLAUDE.md), non-flaky, no real device/network.
- [ ] Documented rationale linking back to #2788 → #2890 (#2788) → #2893 (#2904) so the intended shape is discoverable.

## Context / provenance

- #2788 scoped the composite-index work as additive-only.
- PR #2890 added the composite indexes.
- #2893 / PR #2904 dropped the redundant device indexes and added `test/db/dropRedundantDeviceIndexes.test.ts` (per-table `EXPLAIN QUERY PLAN` + `PRAGMA index_list` assertions for the drop specifically — but only for that migration, not a standing whole-inventory guard).
- Persona-review threads on PR #2904 (DBA + test-quality) both flagged the absence of a standing drift guard as the one non-blocking gap; the drop itself was verified regression-free (reproduced with stock `sqlite3` + `EXPLAIN QUERY PLAN`).

kaeawc:
## What

Adds a standing **drift guard** for the telemetry event tables' index inventory — the CI check requested in #2908 (follow-up surfaced during the multi-persona review of #2904 / issue #2893).

New test: `test/db/telemetryEventIndexInventory.test.ts`. It replays the **full** on-disk migration chain against a fresh `:memory:` DB and pins the canonical per-table explicit-index inventory — **name and column composition** — for the six event tables (`network_events`, `log_events`, `os_events`, `navigation_events`, `storage_events`, `layout_events`).

Closes #2908.

## Why

#2890 added the `(device_id, timestamp)` composite indexes (additive-only per #2788); #2904 dropped the now-redundant standalone `idx_<table>_device` indexes (a composite serves every `device_id=?` path via the left-prefix rule, so the single-column index was pure per-insert write amplification).

`dropRedundantDeviceIndexes.test.ts` proves that **one migration** is correct, but nothing stops a **future** migration from silently:
- re-adding an `idx_<table>_device` index (re-introducing the exact write amplification #2893 removed),
- dropping a `timestamp` index the retention-cutoff scan depends on, or
- recreating a same-named index over the **wrong columns/order** (e.g. reordering the composite to `(timestamp, device_id)`, which breaks the `device_id=?` left-prefix path that is the whole rationale).

Both reviewers on #2904 (DBA + test-quality personas) flagged the absence of this standing whole-inventory guard as the one non-blocking gap. This PR is that guard, filed and reviewed independently as #2893/#2904 intended.

## How

- `beforeAll` builds the schema once via `runMigrations(db)` with **no provider override**, so it asserts the *shipped* chain — a new migration is exercised the moment it lands. A single shared DB is safe because every test is strictly read-only.
- Canonical inventory is a single source-of-truth `Record<table, Record<indexName, columns[]>>`. Each table is asserted with **deep equality** against it, catching all drift directions at once (added, removed, or repointed/reordered index). Its keys are also the authoritative table list.
- **Derives** the live `%_events` table set from `sqlite_master` and asserts it equals the canonical keys — so a future 7th event table can't silently escape the guard (`pragma_index_list` on a missing table returns `[]`, which would otherwise make per-table assertions pass vacuously).
- Inventory is read from `PRAGMA index_list` filtered to `origin = 'c'` (explicit `CREATE INDEX`), ignoring the auto-created PK/rowid index exactly as the issue asks; columns come from `pragma_index_info`.
- Rationale linking #2788 → #2890 → #2893/#2904 documented in the file header.

## Testing

- `bun test test/db/telemetryEventIndexInventory.test.ts` → **5 pass, 0 fail** (~180ms, one shared full-replay, `:memory:`, no device/network, non-flaky).
- **Red proof** (each reverted after): re-adding `idx_network_events_device` → fails; reordering the composite to `(timestamp, device_id)` → fails the column-composition + left-prefix tests; adding a 7th `%_events` table → fails the live-table-set test.
- `bun test test/db/` → **483 pass, 0 fail** (no regressions). `eslint` clean.

## Evaluation criteria (#2908)
- [x] Builds full schema via `runMigrations` and asserts `PRAGMA index_list` per table equals the canonical set, ignoring the auto-created PK/rowid index.
- [x] Fails loudly if a standalone `idx_<table>_device` reappears or a `timestamp`/composite index is removed (proven with probe migrations).
- [x] Runs <100ms/test against `:memory:`, non-flaky, no real device/network.
- [x] Documented rationale linking #2788 → #2890 → #2893/#2904.

### DIFF
--- a/test/db/telemetryEventIndexInventory.test.ts
+++ b/test/db/telemetryEventIndexInventory.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { afterAll, beforeAll, describe, expect, test } from "bun:test";
 import { Database as BunDatabase } from "bun:sqlite";
 import { Kysely, sql } from "kysely";
 import { BunSqliteDialect } from "../../src/db/bunSqliteDialect";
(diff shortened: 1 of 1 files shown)


kaeawc:
### Multi-persona review — addressed in `08264aa3a`

Ran three independent persona reviews (DBA, test-quality, pragmatic-maintainer). Consensus findings and resolution:

- **Assert index columns, not just names** (DBA, maintainer) — a same-named index recreated over the wrong columns/order (e.g. composite reordered to `(timestamp, device_id)`) previously stayed green. ✅ Now pins `name → ordered columns` via `pragma_index_info`; red-proofed.
- **A 7th `%_events` table could silently escape a hardcoded table list** (maintainer, DBA) — the `custom_events` precedent shows the set has changed before. ✅ Now derives the live `%_events` set from `sqlite_master` and asserts it equals the canonical keys; red-proofed. (Confirmed `custom_events` is dropped by `2026_04_01_000_drop_custom_events.ts` before the composite migration, so it's correctly out of the live set.)
- **`pragma_index_list` on a missing table returns `[]`, making `not.toContain` assertions vacuously pass** (test-quality) — ✅ closed by the independent live-table-set assertion above.
- **`beforeEach` ran a full replay per read-only test** (test-quality, maintainer) — ✅ switched to `beforeAll` + one shared DB (the sibling drop test needs `beforeEach` only because it mutates).

Not actioned (with rationale): pinning the retention-cutoff `EXPLAIN QUERY PLAN` here is redundant now that column composition is asserted (DBA agreed it's belt-and-suspenders); the sibling drop test already pins query plans. The three named-invariant tests are kept for legible failure messages and are near-free under `beforeAll`.

PR #2921, opened 2026-07-03
refactor(tools): typed accessor + typed envelope for structuredContent reads (#2907)

### LINKED ISSUE #2907: Typed accessor / typed envelope to eliminate the envelope-vs-structuredContent dead-read bug class
## Summary

Introduce a typed accessor / typed envelope so the recurring **envelope-vs-`structuredContent` dead-read** bug class is caught at the type checker instead of by per-site code review.

## Background

Tool handlers pre-serialize results via `createStructuredToolResponse` (`src/utils/toolUtils.ts:98-122`) into an MCP envelope:

```ts
{ content: [{ type: "text", text }], structuredContent: <payload>, success?, error? }
```

`createStructuredToolResponse` hoists **only** `success` and `error` to the envelope top level. Every other field of the payload — `found`, `viewHierarchy`, `screenshot`, etc. — exists **only** under `structuredContent`. A consumer that reads a non-hoisted field off the top level (`response.found`) silently gets `undefined`: no type error, no runtime error, just a dead branch.

## Recurrence (this is a bug *class*, not a one-off)

Confirmed instances of exactly this mistake:

1. **`lastHierarchy` cache** — read `response.viewHierarchy` off the envelope; cache was never populated. Fixed in #2758 / PR #2891.
2. **`toolRegistry.ts` swipeOn scroll-position update** — read `response?.found`; `NavigationGraphManager.updateScrollPosition` never fired. Fixed in #2897 / PR #2903.
3. **`DefaultUIStateSetup.setupScrollPosition`** — read `result?.found`; success branch dead, always returned `null`/"could not find" even on a successful scroll. Fixed in #2897 / PR #2903.

Three instances across two PRs is enough signal that per-site patching is a band-aid.

## Proposed direction (either or both)

- **Typed accessor:** add `getStructuredField<T>(response, key): T | undefined` in `toolUtils.ts` that reads `response?.structuredContent?.[key]`, and route the existing `response?.structuredContent?.X` sites through it (`toolRegistry.ts` scroll-position + lastHierarchy + screenshot reads, `DefaultUIStateSetup`). Documents intent and removes the top-level-vs-`structuredContent` foot-gun.
- **Stronger — typed envelope:** make `createStructuredToolResponse<T>` return a type where `structuredContent: T` is the payload and the top level exposes only `success`/`error`. Then `response.found` becomes a **compile error**, catching the whole class at the type checker rather than by per-site archaeology. This is the more durable fix.

## Acceptance

- A single typed seam for reading payload fields off a tool-call envelope.
- Reading a non-hoisted field off the envelope top level is a compile error (typed-envelope path) or is impossible-by-convention because all reads go through the accessor.
- Existing call sites migrated; no behavior change (pure refactor + type tightening).
- Grep audit confirms no remaining raw `response.<payloadField>` / `result.<payloadField>` reads on an MCP envelope outside `success`/`error`.

## Provenance

Recommended by adversarial architecture review of PR #2903 (issue #2897). Deferred from that PR as out-of-scope (that PR is a targeted fix + its sibling). Sibling fixes for grounding: PR #2891 (#2758), PR #2903 (#2897).

kaeawc:
## What

Introduce a single **typed seam** for reading payload fields off an MCP tool-call envelope, so the recurring **envelope-vs-`structuredContent` dead-read** bug class (#2907) is caught by convention/type instead of by per-site code review.

Closes #2907.

## Why — a bug *class*, not a one-off

Handlers pre-serialize results via `createStructuredToolResponse` (`src/utils/toolUtils.ts`) into an MCP envelope:

```ts
{ content: [{ type: "text", text }], structuredContent: <payload>, success?, error? }
```

Only `success`/`error` are hoisted to the top level. Every other field (`found`, `viewHierarchy`, `screenshot`, …) lives **only** under `structuredContent`. A raw `response.found` read silently yields `undefined` — no type error, no runtime error, just a dead branch. This has bitten three confirmed sites across two PRs (#2758/PR #2891 lastHierarchy; #2897/PR #2903 swipeOn scroll-position and `DefaultUIStateSetup`). Per-site patching is a band-aid.

## How

`src/utils/toolUtils.ts`:
- **`StructuredToolResponse<T>`** — typed envelope interface. `structuredContent: T` is the payload; the top level exposes only the serialized `content` plus hoisted `success`/`error`.
- **`createStructuredToolResponse<T>(content: T): StructuredToolResponse<T>`** — now generic and typed (was `any`). Same hoisting behavior.
- **`getStructuredField<T>(response, key): T | undefined`** — the one accessor every payload-field read routes through. Reads `response?.structuredContent?.[key]`, safely returns `undefined` for null/undefined responses, a missing/non-object `structuredContent`, or an absent field.

Migrated call sites (behavior-preserving):
- `src/server/toolRegistry.ts` — swipeOn scroll-position (`success`+`found`), observe `lastHierarchy`, observe `screenshot`.
- `src/features/navigation/DefaultUIStateSetup.ts` — `success`+`found`.

Pure refactor + type tightening; no behavior change.

## Testing

- New `test/utils/toolUtilsStructuredField.test.ts` (12 tests): accessor returns nested field / by-reference / `undefined` on every miss path; the foot-gun (raw top-level read is `undefined` while the accessor gets the value); typed envelope hoists only `success`/`error`, serializes payload, tolerates non-object payloads.
- Regressions green: `toolRegistry.scrollPositionUpdate`, `toolRegistry.lastHierarchyCache`, `DefaultUIStateSetup`, `finalizeToolResponse`.
- Full suite: **5900 pass, 0 fail** (`bun test`). `eslint` 0, `bun build.ts` ok.

## Acceptance criteria (#2907)
- [x] A single typed seam for reading payload fields off a tool-call envelope (`getStructuredField`).
- [x] Reading a non-hoisted field is impossible-by-convention — all payload-field reads route through the accessor; the typed envelope documents that only `success`/`error` are hoisted.
- [x] Existing call sites migrated; no behavior change (pure refactor + type tightening).
- [x] Grep audit confirms no remaining raw `response.<payloadField>` / `result.<payloadField>` reads on an MCP envelope outside `success`/`error` (remaining `result.viewHierarchy`/`.found`/`.observation` reads are on internal `ObserveResult`/action-result objects, not envelopes).

### DIFF
--- a/test/utils/toolUtilsStructuredField.test.ts
+++ b/test/utils/toolUtilsStructuredField.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, test } from "bun:test";
 import {
   createStructuredToolResponse,
   getStructuredField,
+  getStructuredPayload,
 } from "../../src/utils/toolUtils";
 
 /**
@@ -55,6 +56,47 @@ describe("getStructuredField", () => {
     expect((response as Record<string, unknown>).found).toBeUndefined();
     // Accessor gets the real value.
     expect(getStructuredField<boolean>(response, "found")).toBe(true);
+  });
+
+  test("does not resolve inherited (non-own) keys like prototype methods", () => {
+    const response = createStructuredToolResponse({ success: true });
+    // `toString`/`constructor` exist on Object.prototype but are not payload
+    // fields — an own-key guard must keep them from leaking through.
+    expect(getStructuredField(response, "toString")).toBeUndefined();
+    expect(getStructuredField(response, "constructor")).toBeUndefined();
+    expect(getStructuredField(response, "__proto__")).toBeUndefined();
+  });
+
+  test("resolves an own key whose value is falsy (found: false)", () => {
+    const response = createStructuredToolResponse({ success: true, found: false });
+    expect(getStructuredField<boolean>(response, "found")).toBe(false);
+  });
+});
+
+describe("getStructuredPayload", () => {
+  test("returns the whole payload object under structuredContent", () => {
+    const payload = { success: true, found: true, awaitTimeout: false };
+    const response = createStructuredToolResponse(payload);
+    expect(getStructuredPayload(response)).toEqual(payload);
+    expect(getStructuredPayload(response)).toBe(response.structuredContent);
+  });
+
+  test("returns undefined for null / undefined responses", () => {
+    expect(getStructuredPayload(null)).toBeUndefined();
+    expect(getStructuredPayload(undefined)).toBeUndefined();
+  });
+
+  test("returns undefined when structuredContent is missing or not an object", () => {
+    expect(getStructuredPayload({ content: [] })).toBeUndefined();
+    expect(getStructuredPayload({ structuredContent: "text" })).toBeUndefined();
+    expect(getStructuredPayload({ structuredContent: null })).toBeUndefined();
+  });
+
+  test("does NOT fall back to parsing the serialized text part", () => {
+    // A text-only envelope carries the payload as JSON in content[0].text, not
+    // under structuredContent — getStructuredPayload deliberately ignores it.
+    const textOnly = { content: [{ type: "text", text: JSON.stringify({ success: true }) }] };
+    expect(getStructuredPayload(textOnly)).toBeUndefined();
   });
 });
 
--- a/test/utils/toolUtilsStructuredField.test.ts
+++ b/test/utils/toolUtilsStructuredField.test.ts
@@ -29,14 +29,11 @@ describe("getStructuredField", () => {
     expect(getStructuredField(response, "found")).toBeUndefined();
   });
 
-  test("returns undefined when the field only exists at the envelope top level", () => {
-    // This is the whole point: `success`/`error` are hoisted, but reading them
-    // as payload fields off `structuredContent` still works because the payload
-    // also carries them. A field that the caller mistakenly expects at the top
-    // level (never hoisted) must not leak through.
-    const response = createStructuredToolResponse({ success: true, error: "boom" });
-    // A hoisted-only field like a hypothetical top-level `found` is not present.
-    expect(getStructuredField(response, "found")).toBeUndefined();
+  test("ignores a field that exists only at the envelope top level, not in structuredContent", () => {
+    // Hand-built envelope where `found` sits at the top level but NOT in the
+    // payload — the accessor must read only `structuredContent` and not leak it.
+    const envelope = { content: [], structuredContent: { success: true }, found: true };
+    expect(getStructuredField<boolean>(envelope, "found")).toBeUndefined();
   });
 
   test("returns undefined for null / undefined responses", () => {
--- a/test/utils/toolUtilsStructuredField.test.ts
+++ b/test/utils/toolUtilsStructuredField.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, test } from "bun:test";
+import {
+  createStructuredToolResponse,
+  getStructuredField,
+} from "../../src/utils/toolUtils";
+
+/**
+ * Typed seam for the envelope-vs-`structuredContent` dead-read bug class (issue
+ * #2907). `createStructuredToolResponse` hoists ONLY `success`/`error` to the
+ * envelope top level; every other payload field lives under `structuredContent`.
+ * A raw `response.found` read silently yields `undefined`. `getStructuredField`
+ * is the single accessor every payload-field read must route through.
+ */
+describe("getStructuredField", () => {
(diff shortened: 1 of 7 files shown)


test/utils/toolUtilsStructuredField.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    // This is the whole point: `success`/`error` are hoisted, but reading them
+    // as payload fields off `structuredContent` still works because the payload
+    // also carries them. A field that the caller mistakenly expects at the top
+    // level (never hoisted) must not leak through.
+    const response = createStructuredToolResponse({ success: true, error: "boom" });
+    // A hoisted-only field like a hypothetical top-level `found` is not present.
+    expect(getStructuredField(response, "found")).toBeUndefined();
+  });
+
+  test("returns undefined for null / undefined responses", () => {
```
kaeawc, line 39: Minor (non-blocking): this case is titled "field only exists at the envelope top level" but the payload `{ success, error }` puts both fields under `structuredContent` too, so it's really re-asserting the missing-`found` path. Not wrong — just weaker than the label implies. The real top-level-vs-structuredContent foot-gun is already proven by the `response.found === undefined` assertion below. Fine to leave.

kaeawc (commented):
## AutoMobile code review — PR #2921 (typed envelope seam, #2907)

**Verdict: Correctly fixes (with one scope caveat worth noting).** Verified locally: `bun test` full suite 5900 pass / 0 fail, `eslint` 0, `bun build.ts` ok. The three `toolRegistry.ts` reads and the `DefaultUIStateSetup.ts` read are migrated to `getStructuredField` with byte-identical semantics — `updateSessionCache(cacheValue: any)` accepts the `unknown` return, and `getStructuredField` preserves the original `?.structuredContent?.x` truthiness short-circuit. Grep audit clean: no remaining raw `structuredContent?.<field>` envelope reads; the surviving `result.viewHierarchy`/`.found`/`.observation` reads are on internal `ObserveResult`/action-result objects, not MCP envelopes.

**Scope caveat (not blocking):** the issue offered a stronger "typed envelope so `response.found` is a *compile error*" path. This PR delivers the typed envelope (`StructuredToolResponse<T>`) + the accessor-by-convention path, which fully satisfies the "impossible-by-convention" acceptance bullet. But `response` in `toolRegistry.ts:311` is declared `any` (handler return type is `any`), so a raw `response.found` there is still *not* a compile error — the type-checker enforcement is only partially realized. The accessor convention is the actual guardrail here. Reasonable to defer tightening handler return types as follow-up.

**Most important thing to verify first:** that no external consumer relied on `createStructuredToolResponse`'s return being `structuredContent: any` to index arbitrary fields — the full-project `bun build.ts` compiling clean confirms this, so I'm confident.

kaeawc:
## Addressed adversarial review (3 perspectives: type-system skeptic, completeness auditor, contrarian architect)

Pushed `7089b9dfb` addressing the findings:

**F4 — machine-enforce the convention (contrarian architect, highest leverage).** Added a `no-restricted-syntax` ESLint rule banning raw `x.structuredContent.<field>` reads (mirrors the existing `SystemTimer` `setTimeout` ban at `eslint.config.mjs:164`). This catches the *actual* failure mode — authors who don't know the field isn't hoisted — at CI, not just those who remember the accessor. Green tree-wide.

**F3 / GAP 1–2 — genuine consolidation (auditor + architect).** Added `getStructuredPayload<T>(response)` companion for whole-payload reads and routed `PlanExecutor.parseStructuredToolPayload`, `PlanExecutor.mergeToolDiagnosticsIntoStepDetails`, and `finalizeToolResponse` through it. There is now one seam per read shape (single-field vs whole-payload) instead of three hand-rolled `structuredContent && typeof === "object"` narrowings. Behavior-preserving (each site keeps its own text-fallback/write-back).

**F2 + JSDoc nit (all three).** Corrected the JSDoc — the accessor is a *structural* guard (reads from `structuredContent`, own-key only); it does not validate `key`/`T`. Also hardened `getStructuredField` with `Object.hasOwn` so inherited keys (`toString`, `constructor`, `__proto__`) can't leak.

**F1 — the compile-error typed envelope (deferred, tracked).** All three reviewers correctly noted that `StructuredToolResponse<T>`'s payoff (making `response.found` a *compile error*) requires re-typing the `Promise<any>` `ToolHandler` boundary, which is out of scope here (the issue itself deferred the stronger path). Filed as follow-up — see below. The typed envelope + accessor + lint rule together deliver the "impossible-by-convention" arm now and lay the type foundation for the durable arm later.

New tests: `getStructuredPayload` (payload/null/non-object/no-text-fallback) + `getStructuredField` own-key/falsy-value cases. Full suite **5906 pass / 0 fail**, eslint 0, build ok.

PR #2923, opened 2026-07-03
test(db): harden Windows file-DB close/reopen tests against EBUSY livelock (#2916)

### LINKED ISSUE #2916: Flaky CI: windows-latest file-backed SQLite close/reopen tests (EBUSY/livelock) intermittently fail
## Summary

The `Node TypeScript Build and Test (windows-latest)` CI job intermittently fails on **real file-backed SQLite close/reopen tests** — a *different* test each run, always a `mkdtemp` + `auto-mobile.db` open→close→reopen scenario, with a multi-second stall (5–11s) characteristic of a Windows file-handle EBUSY / cleanup livelock. `ubuntu-latest` and `macos-latest` pass the same tests every time.

Observed across two consecutive runs of PR #2915 (a migration-only change that does not touch these tests):

- Run 1: `(fail) closeDatabase resets migration + path globals (issue #2796) > reopen after close re-runs migrations against a fresh DB (migrationsRun reset) [10688.00ms]` — `test/db/databaseReset.test.ts` (uses `mkdtemp` + real `auto-mobile.db`).
- Run 2: `(fail) closeDatabase cold-starts the dbWriteBarrier drain latch (issue #2896) > a tracked best-effort write after drain -> close -> reopen is NOT skipped [5282.00ms]`.

Both are non-deterministic: re-running the failed Windows job turned the build green with no code change, confirming a flake rather than a real regression. This forced a manual "re-run failed jobs" to land PR #2915.

## Impact

- **Blocks auto-merge intermittently.** Any PR can be held up by an unrelated Windows flake and needs a manual job re-run.
- Erodes trust in the Windows signal — a real Windows-only regression could hide behind "just re-run it."

## Root cause (known pattern)

Real file-backed `bun:sqlite` databases don't release the file handle cleanly/promptly on Windows, so a subsequent `rm`/reopen races the OS (`EBUSY`), and the test's close→reopen loop can livelock. This is the same class noted when the `failure_groups` file-DB tests were made Windows-safe: prefer `:memory:` + a temp **lock file** over a real file-backed DB, and never assert on real-file cleanup timing.

## Proposed fix

Harden the file-backed close/reopen tests for Windows without weakening what they assert:
- Where the test only needs "close resets globals / reopen re-runs migrations," back it with `:memory:` and a separate temp **lock file** for the path-global assertions (the pattern already used elsewhere in `test/db`), instead of a real `auto-mobile.db`.
- Where a real file is genuinely required, add a bounded retry-with-backoff around `rm`/reopen on Windows (`EBUSY`) and raise the per-test timeout, or gate the real-file variant off `process.platform === "win32"` with an explicit skip + comment.
- Candidate files: `test/db/databaseReset.test.ts` (#2796), the dbWriteBarrier drain-latch test (#2896), and any other `mkdtemp(... auto-mobile.db ...)` close/reopen test.

## Acceptance criteria
- [ ] `Node TypeScript Build and Test (windows-latest)` is green across N consecutive runs with no code change.
- [ ] No file-backed SQLite close/reopen test relies on prompt Windows file-handle release / real-file cleanup timing.
- [ ] The original behavior each test proves (global reset on close, migrations re-run on reopen, drain-latch cold start) is still asserted.

## References
- Surfaced while landing PR #2915 (follow-up to #2895 / #2906); the flaky tests are unrelated to that migration change.
- Related but distinct flake already tracked: #2914 (DefaultProcessExecutor stdout timeout on macos-latest).
- Prior art: the `failure_groups` de-dup migration tests moved to `:memory:` + temp lock file for exactly this Windows portability reason.

kaeawc:
## What

Harden the file-backed SQLite **close/reopen** tests against the `windows-latest`
`EBUSY`/livelock flake tracked in #2916. Test-only change — no production code
touched.

Closes #2916.

## Why

`bun:sqlite` can hold the `.db`/`-wal`/`-shm` file handles past
`Kysely.destroy()` on Windows, so removing the enclosing `mkdtemp` dir throws
`EBUSY`/`EPERM`/`ENOTEMPTY`. Several `test/db` files each duplicated a cleanup
helper that retried **100 × 50 ms = 5 s per dir**. A close/reopen test that
creates two temp dirs therefore stalled ~10 s in `afterEach`, blowing the
per-test timeout — matching the observed `10688 ms` / `5282 ms` failures. Each
run flaked a *different* file-DB test, and re-running the job turned it green
with no code change, forcing manual "re-run failed jobs" to land unrelated PRs.
`ubuntu`/`macos` release handles immediately, so the retry path never triggers
there and local `bun test` can't catch it.

## How

1. **Shared bounded, best-effort cleanup helper** — `test/db/tempDbDir.ts`
   (`removeTempDbDir`). Retries a small **bounded** number of times (default
   5 × 50 ms) on the transient Windows lock codes, then **gives up**, leaving the
   throwaway temp dir for the OS temp sweeper instead of blocking. Unexpected
   error codes are rethrown. `rm`/`timer` are injectable so the retry/give-up
   behavior is unit-tested with a fake filesystem + `FakeTimer` — the real handle
   livelock has no off-Windows reproduction. Each test uses a fresh unique
   `mkdtemp` dir, so a leaked dir never contends with another test and
   correctness never depends on removal succeeding.
2. **Adopted the helper** across the flaky files, deleting the duplicated 5 s
   retry loops: `databaseReset`, `dbWriteBarrierResetOnClose`, `databaseLazyPath`,
   `databaseIncompleteExtraction`, and the previously un-retried
   `databaseMigrationFailure`.
3. **`databaseReset` path-reset test opens no real DB.** The "resolvedDbPath
   reset" assertion is on the pure string path cache (`resolveDatabasePathFrom­Environment`
   never touches the filesystem), so it now drives `getDatabasePath()` alone and
   opens no `auto-mobile.db` — removing its file-handle dependency entirely.
   `closeDatabase()` resets the path cache unconditionally (even with no open
   connection), so the reset is still exercised; the full close-with-open-connection
   path stays covered by the `migrationsRun` test.
4. **Generous timeouts** (30 s) on the genuinely file-backed close/reopen tests
   (they need a shared on-disk file across the migration and app connections —
   `:memory:` gives each connection its own empty DB) as belt-and-suspenders
   against slow Windows disk I/O.

Every behavioral assertion is preserved: migrations re-run on reopen, path
re-resolved under a changed env, `migrationsError` cleared on close, and the
`dbWriteBarrier` drain-latch cold start.

## Testing

New `test/db/tempDbDir.test.ts` (fake `rm` + `FakeTimer`, no real fs/clock,
31 ms) proves the two properties that keep CI from stalling: a persistent lock
**gives up without throwing within a strictly bounded sleep budget**, and
transient locks retry a bounded number of times before succeeding/giving up;
plus first-try success without sleeping and rethrow of unexpected codes.

- `bun test test/db/` — **484 pass, 0 fail** (7.0 s).
- The six affected files run in **369 ms**.
- `eslint` on changed files: 0 problems. `bun build.ts`: ok.

The `windows-latest` acceptance criterion (green across N runs) can only be
confirmed in CI, since the handle livelock has no off-Windows reproduction.

## Acceptance criteria (#2916)
- [x] No file-backed SQLite close/reopen test relies on prompt Windows
      file-handle release / real-file cleanup timing (bounded best-effort
      cleanup; path-reset test opens no file).
- [x] The original behavior each test proves is still asserted (global reset on
      close, migrations re-run on reopen, drain-latch cold start).
- [ ] `windows-latest` green across N consecutive runs with no code change —
      verify in CI post-merge.

### DIFF
--- a/test/db/databaseIncompleteExtraction.test.ts
+++ b/test/db/databaseIncompleteExtraction.test.ts
@@ -1,9 +1,10 @@
 import { afterEach, describe, expect, test } from "bun:test";
 import path from "path";
-import { mkdir, mkdtemp, rm, writeFile } from "node:fs/promises";
+import { mkdir, mkdtemp, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { defaultTimer } from "../../src/utils/SystemTimer";
 import { isIncompleteExtractionError } from "../../src/db/migrationDependencyIntegrity";
+import { removeTempDbDir } from "./tempDbDir";
 
 /**
  * Issue #2833: the incomplete-extraction mapping is scoped to KNOWN migration
@@ -34,31 +35,13 @@ describe("startup migration failure does not mislabel a genuine bad import", () 
     restore("AUTOMOBILE_DB_PATH", originalDbPath);
     restore("AUTOMOBILE_MIGRATIONS_DIR", originalMigrationsDir);
     if (tempDir) {
-      await removeTempDirWithRetry(tempDir);
+      await removeTempDbDir(tempDir);
       tempDir = undefined;
     }
   });
 
   async function importFreshDatabaseModule() {
     return import(`../../src/db/database.ts?incomplete-extraction-test=${Date.now()}-${Math.random()}`);
-  }
-
-  // Windows holds the sqlite file until the handle is fully released; retry the
-  // removal past a transient EBUSY/EPERM rather than failing the test on cleanup.
-  async function removeTempDirWithRetry(dir: string): Promise<void> {
-    for (let attempt = 0; attempt < 100; attempt += 1) {
-      try {
-        await rm(dir, { recursive: true, force: true });
-        return;
-      } catch (error) {
-        const code = (error as NodeJS.ErrnoException).code;
-        if (code !== "EBUSY" && code !== "EPERM" && code !== "ENOTEMPTY") {
-          throw error;
-        }
-        await defaultTimer.sleep(50);
-      }
-    }
-    await rm(dir, { recursive: true, force: true });
   }
 
   test("an unknown missing package surfaces the generic error, not incomplete-extraction", async () => {
--- a/test/db/databaseLazyPath.test.ts
+++ b/test/db/databaseLazyPath.test.ts
@@ -1,9 +1,9 @@
 import { afterEach, describe, expect, test } from "bun:test";
 import path from "path";
-import { mkdtemp, rm } from "node:fs/promises";
+import { mkdtemp } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { DAEMON_LAUNCH_CWD_ENV } from "../../src/utils/workingDirectory";
-import { defaultTimer } from "../../src/utils/SystemTimer";
+import { removeTempDbDir } from "./tempDbDir";
 
 /**
  * Regression test for the module-load-time-vs-runtime path hazard.
@@ -39,23 +39,6 @@ describe("database path lazy resolution", () => {
   async function importFreshDatabaseModule() {
     // Fresh module instance so its lazy cache is not shared with other tests.
     return import(`../../src/db/database.ts?lazy-db-path-test=${Date.now()}-${Math.random()}`);
-  }
-
-  async function removeTempDirWithRetry(dir: string): Promise<void> {
-    for (let attempt = 0; attempt < 100; attempt += 1) {
-      try {
-        await rm(dir, { recursive: true, force: true });
-        return;
-      } catch (error) {
-        const code = (error as NodeJS.ErrnoException).code;
-        if (code !== "EBUSY" && code !== "EPERM" && code !== "ENOTEMPTY") {
-          throw error;
-        }
-        await defaultTimer.sleep(50);
-      }
-    }
-
-    await rm(dir, { recursive: true, force: true });
   }
 
   test("resolves relative AUTOMOBILE_DB_DIR against the launch cwd set AFTER import", async () => {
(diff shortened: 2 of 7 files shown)


kaeawc:
### Multi-perspective review pass (reliability / coverage / architecture)

Ran three independent review perspectives against the branch. Summary + what I changed in `29458f16a`:

**Reliability — root cause confirmed.** The 5–11s stall is in the `afterEach` cleanup loop, not `closeDatabase()`/`destroy()` or the reopen path (`bunSqliteDialect.close()`/`destroy()` are single synchronous calls, no backoff). The old `100×50ms=5s`-per-dir retry × 2 temp dirs ≈ the observed `10688ms`; one locked dir ≈ the `5282ms` drain-latch failure. The load-bearing fix is the bounded ~200ms/dir give-up.
- Fixed a **misleading comment**: a test's third-arg timeout does **not** govern `afterEach` hooks in bun (they use the default hook timeout independently). Reworded so the 30s timeout is documented as covering the slow test *body* only; cleanup is bounded separately by `removeTempDbDir`.

**Coverage — no weakened assertions.** The `resolvedDbPath` reset test now uses `getDatabasePath()` (no file opened) but still catches the regression it names, and the "close *with* an open connection resets the path" coverage it dropped is still exercised by the `migrationsRun` test. Added unit coverage for the previously-untested default `rm`/`onGiveUp` paths.

**Architecture — direction acceptable.** `:memory:` genuinely can't back the migration-rerun tests (migration and app connections are separate DBs). `file::memory:?cache=shared` *could* eliminate the file, but bun:sqlite URI-filename support + the real file-based migration lock (`createFileMigrationLock(dbPath)`) make it higher-risk than a real temp file with bounded cleanup — now documented in `tempDbDir.ts`. Swapped `console.warn` → shared `logger.warn`.

**Deliberately not done:** migrating `backupDatabaseFile.test.ts`'s `rmSync(..., maxRetries:10, retryDelay:100)` — it's already bounded (~1s, single dir) and wrapped best-effort, so not a flake risk; a sync sibling helper would be scope creep. Noted for a possible future consistency pass.

Issue #2265, opened 2026-05-21
feat(ios): invoke custom accessibility actions via tapOn


kaeawc:
## Summary

When an iOS element exposes `UIAccessibilityCustomAction` items (e.g., "Share", "Archive", "Open Link"), AutoMobile can see them in the hierarchy via `extras["sdk.accessibilityCustomActions"]` but cannot invoke them. This means elements like attributed text views — where the entire block is a single accessibility element with custom actions for embedded links or interactive regions — are not actionable.

## Current State

- **SDK walker** captures `view.accessibilityCustomActions` names and sends them to CtrlProxy
- **HierarchyMerger** serializes them into `extras["sdk.accessibilityCustomActions"]` as a comma-separated string
- **ElementFinder** does not search `extras` — only `text`, `content-desc`, and `ios-accessibility-label`
- **GesturePerformer** only handles generic actions (`tap`, `long_click`, `scroll_forward`, etc.)
- **XCUITest has no API to invoke custom accessibility actions** — coordinate taps don't help because the parent element consumes the tap

## Proposed Behavior

```json
{
  "selector": { "text": "Open Link" },
  "action": "tap",
  "platform": "ios"
}
```

If the selector text matches a custom accessibility action name (and no standard element text match is found), `tapOn` should invoke that action directly rather than coordinate-tapping.

## Research: macOS AXUIElement Bridge (No SDK Required)

We prototyped and confirmed that the macOS Accessibility API can invoke custom accessibility actions on iOS Simulator elements **without the SDK embedded in the target app**.

### How It Works

The iOS Simulator bridges its iOS accessibility tree to macOS via `AccessibilityPlatformTranslation.framework`. From a macOS process:

1. Get the Simulator's PID via `NSRunningApplication`
2. Create an `AXUIElementCreateApplication(pid)` reference
3. Navigate the window tree to find `[AXGroup/iOSContentGroup]` — this contains the full iOS element tree
4. iOS elements expose `AXCustomActions` attribute and custom actions appear inline in the element's action list as `"Name:<action_name>\nTarget:0x0\nSelector:(null)"`
5. Call `AXUIElementPerformAction(element, actionName)` to invoke the custom action

### Proof of Concept Results

| Step | Result |
|------|--------|
| Find Simulator via PID | ✅ Works |
| Navigate to `iOSContentGroup` subrole in window | ✅ Works |
| Read iOS element tree (labels, IDs, values) | ✅ Works |
[...]

PR #2950, opened 2026-07-03
test(ios): add control-proxy wire-parity drift tripwire + fix request_voiceover_action mismatch

### LINKED ISSUE #2857: test(ios): add control-proxy wire-parity drift tripwire + fix request_voiceover_action mismatch
## Summary

The iOS `control-proxy` TS client ↔ Swift runner have **no cross-language contract test**, so command-set drift is only caught at runtime on-device as an "Unknown command type" error. This issue asks for a drift tripwire (the iOS analog of the Android/TS guard added in #2835 / #2771) and to fix the one **live** mismatch it immediately surfaces: `request_voiceover_action`.

Follow-up from the typed-command-model migration #2846 / #2854 (merged `aa25414a`) and its 3-perspective review.

## Background — how this was found

During the #2854 review, two independent reviewers flagged that the wire-parity guarantee is one-sided:

- The Swift runner **does** lock its own inbound set: `TypedRequestDecodeTests.testEveryRequestTypeDecodesToMatchingCase` iterates `RequestType.allCases` (`ios/control-proxy/Sources/CtrlProxy/Models.swift`), so a dropped/renamed **enum case** fails a test.
- But **nothing** asserts the set of command `type` strings the **TS client emits** is a subset of the Swift `RequestType` rawValues. A new/renamed TS command silently compiles and only fails at runtime on device. This is exactly the "three parallel implementations drift" #2846 warns about (cf. the `unsubscribe_storage` TS/Android mismatch found during #2752, fixed in #2832).

Response correlation is by `requestId` only (`src/utils/RequestManager.ts`), and unknown types now decode-fail with a `requestId`-correlated error carrying "Unknown command type: <type>" (see #2854 + the TS `rewriteUnknownCommandError` diagnostic). So the drift degrades silently rather than crashing.

## The live mismatch: `request_voiceover_action`

`src/features/observe/ios/CtrlProxyVoiceOver.ts:98` (`requestVoiceOverActivate`) sends `messageType: "request_voiceover_action"`, and its response `voiceover_action_result` is already handled (`src/features/observe/ios/decodeCtrlProxyMessage.ts:131`). **But `request_voiceover_action` has never existed in the Swift `RequestType`** (not on main, not before the #2854 refactor).

This is **not dormant** — it's reached:
- `src/features/action/TapOnElement.ts:1518` calls `client.requestVoiceOverActivate(label, voiceOverAction)` inside `executeIOSTapWithVoiceOver` (the iOS `tapOn`/`longPress` VoiceOver-activation path).
- The runner rejects it with "Unknown command type: request_voiceover_action", so `result.success` is false, and `TapOnElement.ts:~1525` logs a warning and **falls back to a coordinate tap**.

Net effect: the iOS VoiceOver-based activation path is effectively dead-on-arrival — it always fails and falls back, emitting an error log each time. (Likely related to the accessibility-action work in #2265.)

## Scope / checklist

- [ ] Add a TS-side contract test (e.g. under `test/features/observe/ios/`) asserting every `messageType`/`type:` string the iOS client can emit (across `src/features/observe/ios/*.ts` + the shared `src/features/observe/shared/SharedTextDelegate.ts` / `DeviceServiceUtils.ts`) is a member of a checked-in copy of the iOS `RequestType` rawValues — mirroring Android's `KNOWN_REQUEST_TYPES` approach (#2835). Adding a TS command without listing it should be a compile/test failure.
- [ ] Resolve `request_voiceover_action`: either (a) implement it in the runner — a `RequestType` case + payload struct + handler in `CommandHandler.swift` emitting a `voiceover_action_result` (decoder already exists), or (b) remove/repair the TS sender + its `TapOnElement` call if the capability isn't wanted. Decide in light of #2265.
- [ ] (Optional, stronger) Add a field-name wire-parity backstop: a fixture of JSON snapshots transcribed from the TS builders, decoded by the Swift payloads, so renaming a **field** (not just a command) is caught. Today a rename like `disableAllFiltering` → `disableFiltering` silently passes every test except the one that reads it.

## Risk

Low blast radius (test + a well-scoped runner addition or TS removal). The tripwire is the durable win — it prevents the next silent iOS drift.

kaeawc:
## What

Adds a **cross-language wire-parity tripwire** for the iOS `control-proxy` TS client ↔ Swift runner — the iOS analog of the Android/TS guard from #2835 — and fixes the one **live** mismatch it immediately surfaces: `request_voiceover_action`.

Closes #2857.

## Why

The Swift runner locks its own inbound set (`TypedRequestDecodeTests` iterates `RequestType.allCases`), but **nothing** asserted that the command `type` strings the **TS client emits** stay a subset of the Swift `RequestType` rawValues. A new/renamed TS command compiled cleanly and only failed at runtime on-device as an `Unknown command type: <type>` error. This is the "three parallel implementations drift" #2846 warns about.

The tripwire immediately caught a **live** mismatch: `src/features/observe/ios/CtrlProxyVoiceOver.ts` sent `messageType: "request_voiceover_action"`, but **no such case has ever existed in the Swift `RequestType`**. It is reached (not dormant): `TapOnElement.ts:1518` (`executeIOSTapWithVoiceOver`, the iOS VoiceOver-activation tap/longPress path) → the runner rejected it → `result.success` was false → it always logged a warning and fell back to a coordinate tap. The VoiceOver activation path was **dead-on-arrival**.

## How

**1. Tripwire (the durable win).** New `src/features/observe/ios/ctrlProxyRequestTypes.ts` exports `IOS_KNOWN_REQUEST_TYPES` — a checked-in copy of the Swift `RequestType` rawValues (the iOS analog of Android's `KNOWN_REQUEST_TYPES`). New `test/features/observe/ios/ctrlProxyWireParity.test.ts` asserts two independent guarantees:
- `IOS_KNOWN_REQUEST_TYPES` equals the `RequestType` rawValues **read live out of `Models.swift`** (plus a hand-transcribed list that must also match, so neither can rot independently) — a case renamed/added/removed on the runner fails here.
- Every command `type`/`messageType` the iOS client can emit — scanned from `observe/ios/*.ts` + the shared `SharedTextDelegate`/`SharedGestureDelegate` (all three emit shapes: `messageType:`, `type:` object/ternary literals, and `CtrlProxyDatabase.request(...)` first-args) — is a member of `IOS_KNOWN_REQUEST_TYPES`. Adding a TS command the runner can't decode now fails a test.

**2. Resolve `request_voiceover_action` (issue option b — consolidate).** Rather than add a second near-identical runner command, VoiceOver activation now rides the **existing** `request_action` command, which already finds an element by label and dispatches `activate`/`tap` / `long_press` in the Swift `performAction`. This *reduces* the wire surface (serving the anti-drift goal and the repo's "one canonical primitive per concern" rule) instead of adding a parallel command that can itself drift. #2265 (custom accessibility actions via a macOS AXUIElement bridge) is a genuinely different future mechanism that would not reuse this command, so keeping the phantom "for #2265" is not justified.
- `CtrlProxyVoiceOver.requestVoiceOverActivate` now sends `request_action` (was the phantom `request_voiceover_action`); the runner replies `action_result`, which decodes into the same shape.
- Swift `GesturePerformer.performAction` gains an `"activate"` case (→ tap for an element located by label), grouped with `click`/`tap`.
- Removed the now-dead `voiceover_action_result` decoder branch (no runner ever emitted it; `action_result` decoding is already covered).

## Testing

- **New** `ctrlProxyWireParity.test.ts` (6 tests): live Swift-enum read, no-dupes, `IOS_RUNNER_FEATURE_COMMANDS` subset, and the emit-subset guard. Confirmed it goes **RED** on the pre-fix tree with exactly `["request_voiceover_action"]` as the sole unknown, then GREEN after the fix.
- **New** `CtrlProxyVoiceOver.test.ts` cases: `requestVoiceOverActivate` sends `request_action` (not the phantom), carries `label`+`action`, and resolves from `action_result`, for both `activate` and `long_press`.
- **New** Swift `CommandHandlerTests`: `request_action` with `action:"activate"`/`"long_press"` + a label decodes and reaches `performAction` with the label preserved, replying `action_result`.
- Existing `TapOnElement.voiceover.test.ts` (action-mapping) and `decodeCtrlProxyMessage.test.ts` pass; removed the obsolete `voiceover_action_result` decode test.
- Verified locally: `eslint --fix` 0 issues, `bun build.ts` ok, full `bun test` **5895 pass / 0 fail / 2 skip**, `./scripts/ios/swift-test.sh` all packages pass (control-proxy incl. the new tests).

## Acceptance criteria (#2857)
- [x] TS-side contract test: every command string the iOS client can emit is a member of a checked-in copy of the Swift `RequestType` rawValues; adding an unlisted TS command fails a test.
- [x] The checked-in list is verified against the **live** `Models.swift` enum, not just a transcription.
- [x] `request_voiceover_action` resolved: routed through the existing `request_action` command; the VoiceOver activation path is now functional instead of always falling back.
- [ ] *(Deferred, issue's optional item 3)* field-name wire-parity backstop (JSON snapshot fixtures) — filed as a follow-up.

## Delivery note
The `activate` case is compiled into the iOS XCUITest runner, so the on-device VoiceOver fix only reaches end users after the next iOS runner release is cut (the pinned-release checksum registry `src/constants/release.ts` is CI-generated, not edited here). The TS-side wire fix + tripwire land immediately on merge.

## Post-open hardening (commit af73b2d)
Follow-up commit hardens the tripwire scanner per the 3-perspective review: widened the command-token class to allow digits (`[a-z][a-z0-9_]*`) on both the emit-scan and the Swift-enum read; tolerate a `type:` key/value line-break; and a new guard test forbids template-literal command discriminators (the one dynamic shape the literal scan cannot follow). Residual const-hoist / import-graph coverage tracked as follow-ups.

### DIFF
--- a/ios/control-proxy/Sources/CtrlProxy/GesturePerformer.swift
+++ b/ios/control-proxy/Sources/CtrlProxy/GesturePerformer.swift
@@ -779,7 +779,9 @@ public class GesturePerformer: GesturePerforming {
 
             try runOnMainThread {
                 switch action.lowercased() {
-                case "click", "tap":
+                case "click", "tap", "activate":
+                    // "activate" is the VoiceOver activation gesture (issue #2857); for an
+                    // element located by label it resolves to a tap, matching "click"/"tap".
                     found.tap()
                 case "long_click", "long_press":
                     found.press(forDuration: 1.0)
--- a/ios/control-proxy/Tests/CtrlProxyTests/CommandHandlerTests.swift
+++ b/ios/control-proxy/Tests/CtrlProxyTests/CommandHandlerTests.swift
@@ -905,6 +905,45 @@ final class CommandHandlerTests: XCTestCase {
         )
     }
 
+    // MARK: - Node Action Tests
+
+    /// VoiceOver activation (issue #2857) rides `request_action` with action "activate"
+    /// and an accessibility label. The command must decode and reach `performAction`
+    /// with the label preserved, replying `action_result`.
+    func testActionActivateByLabelReachesPerformAction() {
+        let request = WebSocketRequest.action(RequestAction(
+            requestId: "act-activate-1",
+            action: "activate",
+            resourceId: nil,
+            label: "Submit"
+        ))
+
+        guard let actionResponse = handleRequest(request, as: WebSocketResponse.self) else { return }
+        XCTAssertEqual(actionResponse.success, true)
+        XCTAssertEqual(actionResponse.type, "action_result")
+
+        let history = fakeGesturePerformer.getActionHistory()
+        XCTAssertEqual(history.count, 1)
+        XCTAssertEqual(history[0].action, "activate")
+        XCTAssertEqual(history[0].label, "Submit")
+        XCTAssertNil(history[0].resourceId)
+    }
+
+    /// `request_action` decoded straight from the wire the TS client now emits for
+    /// VoiceOver long-press activation.
+    func testActionLongPressDecodesFromWire() throws {
+        let request = try decodeWebSocketRequest(
+            #"{"type":"request_action","requestId":"act-lp-1","action":"long_press","label":"Row"}"#
+        )
+
+        guard let actionResponse = handleRequest(request, as: WebSocketResponse.self) else { return }
+        XCTAssertEqual(actionResponse.success, true)
+
+        let history = fakeGesturePerformer.getActionHistory()
+        XCTAssertEqual(history[0].action, "long_press")
+        XCTAssertEqual(history[0].label, "Row")
+    }
+
     // MARK: - Device Control Tests
 
     func testPressHomeSuccess() {
--- a/src/features/observe/ios/CtrlProxyVoiceOver.ts
+++ b/src/features/observe/ios/CtrlProxyVoiceOver.ts
@@ -78,13 +78,20 @@ export class CtrlProxyVoiceOver {
   }
 
   /**
-   * Request VoiceOver accessibility action on an element by label.
+   * Activate an element by its accessibility label via the runner's node-action path.
+   *
+   * Rides the existing `request_action` command (element lookup by label, then
+   * `activate`→tap / `long_press`→press in the Swift `performAction`) rather than a
+   * dedicated `request_voiceover_action` command, which never existed in the Swift
+   * `RequestType` and so failed to decode on-device — always falling back to a
+   * coordinate tap (issue #2857). The runner replies `action_result`, which decodes
+   * into the same shape as `CtrlProxyVoiceOverActionResult`.
    *
    * @param label - The accessibility label of the target element
    * @param action - The action to perform: "activate" or "long_press"
    * @param timeoutMs - Request timeout in milliseconds (default: 5000)
    * @param perf - Optional performance tracker
-   * @returns VoiceOver action result
+   * @returns Action result
    */
   async requestVoiceOverActivate(
     label: string,
@@ -94,14 +101,14 @@ export class CtrlProxyVoiceOver {
   ): Promise<CtrlProxyVoiceOverActionResult> {
     return sendCommand<CtrlProxyVoiceOverActionResult>(this.context, {
       idPrefix: "voiceover_action",
-      responseType: "voiceover_action",
-      messageType: "request_voiceover_action",
+      responseType: "action",
+      messageType: "request_action",
       params: { label, action },
       timeoutMs,
       perf,
       cancelScreenshotBackoff: false,
       notConnectedError: () => ({ success: false, error: "Not connected to CtrlProxy" }),
-      timeoutError: () => ({ success: false, error: "Timeout waiting for voiceover_action_result" }),
+      timeoutError: () => ({ success: false, error: "Timeout waiting for action_result" }),
     });
   }
 }
--- a/src/features/observe/ios/ctrlProxyRequestTypes.ts
+++ b/src/features/observe/ios/ctrlProxyRequestTypes.ts
@@ -1,0 +1,83 @@
+/**
+ * Checked-in copy of the iOS CtrlProxy runner's inbound command contract.
+ *
+ * The Swift runner locks its own inbound set with `RequestType` (a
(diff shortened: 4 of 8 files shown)


kaeawc:
## 3-perspective review summary

Ran three independent reviewers (protocol/soundness, iOS-Swift runtime, maintainer/direction). **No must-fix blockers**; the protocol fix is verified sound. Actions taken + follow-ups below.

### Verified sound
- **Protocol fix** — `request_voiceover_action` was a phantom (never in Swift `RequestType`); routing onto the existing `request_action` + new `activate` case is correct, and correlation-by-`requestId` is preserved (changing `responseType` is cosmetic — `RequestManager` matches by id only).
- **Swift runtime** — `activate`→`tap()` on a label-resolved element is the correct XCUITest primitive; the new case is additive-only (no existing `request_action` caller sends `activate`); empty-label is guarded upstream in `TapOnElement`. The real tap mapping is XCUITest-only, so the CommandHandler+Fake test is the correct off-device ceiling.

### Addressed in-PR (commit af73b2d)
Protocol reviewer proved the scanner had adjacency-based false negatives:
- Widened the command token to `[a-z][a-z0-9_]*` on **both** the emit scan and the Swift-enum read (a digit-bearing rawValue could previously slip past both sides).
- Tolerate a `type:` key/value line-break.
- Added a guard test forbidding template-literal command discriminators — the one dynamic shape a literal scan can't follow — so the assumption fails loudly, not silently.
- Documented the shared-delegate cross-platform invariant inline.

### Filed as follow-ups (see linked issues)
- Structural (AST/lint) scan to also catch const-hoisted `type` values, + derive the shared-delegate list from the iOS import graph.
- Merge the now-redundant `CtrlProxyVoiceOverActionResult` into `CtrlProxyActionResult`.
- Field-name wire-parity backstop via JSON snapshot fixtures (issue #2857's optional item 3).

kaeawc:
Follow-ups filed for deferred/out-of-scope work surfaced in review: #2954 (field-name wire-parity backstop — #2857 item 3), #2955 (structural AST + import-graph scan), #2956 (merge redundant CtrlProxyVoiceOverActionResult).

Issue #2899, opened 2026-07-03
Wire toolResultsNoStructuredContent flag into finalizeToolResponse (flag plumbed but dormant)


kaeawc:
## Summary

The `toolResultsNoStructuredContent` output-reduction flag is **fully plumbed but consumed by nothing in `src/`**. It flows CLI/env → `outputReductionFlags` → `ServerConfig`, but no production code reads `serverConfig.isToolResultsNoStructuredContentEnabled()`:

- `src/utils/outputReductionFlags.ts:13,49,88` — field + parsing
- `src/utils/ServerConfig.ts:40,232,235-236` — setter/getter
- `src/daemon/manager.ts:1036`, `src/daemon/daemon.ts:206`, `src/daemon/types.ts:191` — wiring
- **Readers in `src/`: none** (only tests reference the getter)

So enabling `--tool-results-no-structured-content` (or its env/feature-flag) currently has **no effect** on tool output.

## Why `finalizeToolResponse` is the home

PR #2891 (#2758) added `src/server/finalizeToolResponse.ts` — the single post-handler serialization chokepoint that owns the MCP envelope `{ content, structuredContent }` and runs exactly once per tool call. Stripping `structuredContent` from the envelope is an **envelope-level transform** (not an `ObserveResult`-shape transform), so it cannot live in `ObserveResultOutput.ts`/`sanitizeObserveResult`. `finalizeToolResponse` is the natural — and arguably only correct — owner.

This was flagged during the #2891 architecture review: the PR built the flag's natural home but left it dormant, and there is now no other seam that owns it.

## Proposed implementation

In `finalizeToolResponse` (after the observe/action sanitize step), when `serverConfig.isToolResultsNoStructuredContentEnabled()` is true, delete `structuredContent` from the returned envelope (keeping `content[0].text` as the single source of truth). Considerations:

- MCP tools declared with an `outputSchema` are **required** to return `structuredContent` — verify the host tolerates its omission, or gate the strip to tools without an output schema. This interaction needs to be resolved before implementing (may require the tool `name`/schema in `FinalizeToolResponseContext`, which already carries `name`).
- Decide precedence vs `observeResultCompact` (a separate, still-unwired flag — an `ObserveResult`-shape concern that belongs inside `sanitizeObserveResult`, tracked separately).

## Acceptance / test

[...]

PR #2958, opened 2026-07-03
feat(mcp): gate structuredContent duplication (drop for no-schema tools) (#2759)

### LINKED ISSUE #2759: Output reduction 4: gate structuredContent duplication (drop for no-schema tools)

## Goal
Stop billing the same tree twice. `src/utils/toolUtils.ts` returns both `content[0].text` (pretty JSON) and `structuredContent` (the same object); clients that ingest `structuredContent` pay for the tree twice.

## Key fact
Per MCP (2025-06-18), `structuredContent` is only meaningful when a tool declares an `outputSchema`. **`observe` — the largest payload — declares NO `outputSchema`, yet still returns `structuredContent`.** Only 5 tools declare one: `tapOn`, `setUIState`, `executePlan`, `accessibility`, `accessibilityFocus`.

The text block is the universal, model-facing representation and must always stay. `structuredContent` is the one to drop/gate — never the text.

## Scope
1. **No-schema tools (immediate, no spec risk):** stop emitting `structuredContent` when no `outputSchema` is registered. Switch `observe` (`src/server/observeTools.ts:271`) to text-only, or make `createStructuredToolResponse` omit `structuredContent` absent a schema. Kills the duplicate on the biggest payload.
2. **Schema tools (flag-gated):** `--tool-results-no-structured-content` drops `outputSchema` + `structuredContent` together for the 5 schema tools (can't drop one without the other — spec requires conforming `structuredContent` when `outputSchema` is advertised). Conditional registration at `src/server/toolRegistry.ts:716`.

## Why this precedes #2760
Compressed text output (TOON/compact) yields **zero** context savings for clients that bill `structuredContent` unless this lands first — it's the same lever. `structuredContent` must remain valid JSON, so it can't hold TOON.

## Acceptance
- `observe` response contains no `structuredContent`; MCP spec compliance preserved (no `outputSchema` advertised without `structuredContent`).
- Schema tools unchanged unless the flag is set.

## Dependencies
- Config plumbing (#2756) for the flag.
- Must land before the compact encoding (#2760).

kaeawc:
## Summary

Closes #2759. Stops billing the same tree twice. Tool handlers pre-serialize into an MCP envelope carrying **both** `content[0].text` (pretty JSON, the universal model-facing surface) **and** a byte-for-byte `structuredContent` duplicate. Per MCP 2025-06-18, `structuredContent` is only meaningful when a tool declares an `outputSchema` — yet `observe` (the largest payload) declares none and still ships the duplicate.

The text block always stays. `structuredContent` is gated:

1. **No-schema tools (unconditional):** stop emitting `structuredContent` when the tool registers no `outputSchema`. Kills the duplicate on `observe`, the biggest payload.
2. **Schema tools (flag-gated):** `--tool-results-no-structured-content` drops `outputSchema` **and** `structuredContent` together (the spec forbids advertising one without the other).

## Changes

- **`src/server/stripToolResultStructuredContent.ts`** (new) — pure helper. Drops `structuredContent` when the tool has no `outputSchema` OR the flag is set; keeps it only for schema tools with the flag off. Preserves `content`/`success`/`error`; never mutates input; no-ops on non-envelope/image/text-only responses.
- **`src/server/index.ts`** — wire the strip at the **CallTool wire boundary**, not the shared handler. Internal `tool.handler(...)` callers (e.g. `DefaultUIStateSetup`, #2897) keep the full envelope.
- **`src/server/toolRegistry.ts`** — `getToolDefinitions` omits the `outputSchema` advertisement when the flag is on.

The flag was already plumbed through `OutputReductionFlags`/`ServerConfig`/daemon (#2756, #2933) but never consumed — this PR is the missing consumer.

## Why this precedes #2760

Compressed text output (TOON/compact) yields **zero** savings for clients that bill `structuredContent` unless this lands first — `structuredContent` must remain valid JSON, so it can't hold TOON.

## Transport coverage

Both paths gated: direct MCP calls hit the strip in `index.ts`; daemon mode runs `createMcpServer`, and `proxyServer.ts` forwards the daemon's already-stripped result + already-omitted `tools/list`.

## Test plan (TDD, EC1–EC8)

- `test/server/stripToolResultStructuredContent.test.ts` — the full keep/strip decision matrix, immutability, field preservation, non-envelope no-ops.
- `test/server/tools/structuredContentGating.test.ts` — end-to-end via in-memory MCP client:
  - EC1: no-schema tool → `content[]` only, no `structuredContent` key.
  - EC2: schema tool, flag off → conforming `structuredContent` kept; `tools/list` advertises `outputSchema`.
  - EC3: schema tool, flag on → `structuredContent` dropped; `tools/list` omits `outputSchema`.
  - EC4: large no-schema payload → serialized wire size drops ~40% (the compact duplicate).
  - EC5: text block preserved (model-facing surface intact).
  - EC6: internal `tool.handler(...)` callers still receive `structuredContent`.
  - EC8: never `structuredContent` without an advertised `outputSchema`.

`bunx turbo run lint build test` → 5938 pass, 0 fail.

## Acceptance (from issue)

- [x] `observe` response contains no `structuredContent`; MCP spec compliance preserved.
- [x] Schema tools unchanged unless the flag is set.

### DIFF
--- a/src/daemon/daemonMcpProxy.ts
+++ b/src/daemon/daemonMcpProxy.ts
@@ -3,6 +3,7 @@ import { DaemonManager, type DaemonManagerLike } from "./manager";
 import { logger } from "../utils/logger";
 import { SOCKET_PATH, DAEMON_STARTUP_TIMEOUT_MS, CONNECTION_TIMEOUT_MS, DAEMON_VERSION, DAEMON_VERSION_RESTART_COOLDOWN_MS } from "./constants";
 import type { DaemonOptions } from "./types";
+import { OUTPUT_REDUCTION_FLAG_SPECS } from "../utils/outputReductionFlags";
 import { compareVersions } from "../server/deviceMatcher";
 import { releaseVersion } from "../utils/mcpVersion";
 import { defaultTimer, type Timer } from "../utils/SystemTimer";
@@ -184,6 +185,43 @@ export interface ProxiedResourceTemplate {
   name: string;
   description?: string;
   mimeType?: string;
+}
+
+/**
+ * The daemon startup options that change its observable MCP behavior and so must
+ * match before a running daemon can be reused: `embeddedSdk` plus every
+ * output-reduction flag. A same-build MCP client that requests one of these
+ * against an already-running daemon started without it (or vice versa) would
+ * otherwise silently get the wrong tool-output contract until a manual restart
+ * (issue #2759 — the `toolResultsNoStructuredContent` case). The output-reduction
+ * fields are derived from `OUTPUT_REDUCTION_FLAG_SPECS` (whose `field` names map
+ * 1:1 to `DaemonOptions`) so a new flag is covered automatically.
+ */
+const REUSE_CRITICAL_OPTION_KEYS: (keyof DaemonOptions)[] = [
+  "embeddedSdk",
+  ...OUTPUT_REDUCTION_FLAG_SPECS.map(spec => spec.field),
+];
+
+/**
+ * Boolean-valued startup options that differ between the requested MCP config
+ * and the running daemon. Each option is compared as a strict boolean
+ * (`=== true`), so `undefined` and `false` are treated identically. Returns a
+ * human-readable list (empty when everything matches) for logging and error
+ * messages.
+ */
+function startupOptionMismatches(
+  requested: DaemonOptions | undefined,
+  running: DaemonOptions | undefined
+): string[] {
+  const mismatches: string[] = [];
+  for (const key of REUSE_CRITICAL_OPTION_KEYS) {
+    const want = requested?.[key] === true;
+    const have = running?.[key] === true;
+    if (want !== have) {
+      mismatches.push(`${key} (requested=${want}, running=${have})`);
+    }
+  }
+  return mismatches;
 }
 
 /**
@@ -477,20 +515,20 @@ export class DaemonMcpProxy {
       return;
     }
 
-    const requestedEmbeddedSdk = this.config.daemonOptions?.embeddedSdk === true;
-    const runningEmbeddedSdk = status.options?.embeddedSdk === true;
-    if (requestedEmbeddedSdk === runningEmbeddedSdk) {
+    const requested = this.config.daemonOptions;
+    const mismatches = startupOptionMismatches(requested, status.options);
+    if (mismatches.length === 0) {
       return;
     }
 
     if (!this.config.autoStartDaemon) {
       throw new DaemonUnavailableError(
-        "Daemon startup options differ from MCP server options and auto-start is disabled"
+        `Daemon startup options differ from MCP server options (${mismatches.join(", ")}) and auto-start is disabled`
       );
     }
 
     logger.info(
-      `[DaemonMcpProxy] Daemon embeddedSdk=${runningEmbeddedSdk} differs from MCP server embeddedSdk=${requestedEmbeddedSdk}, restarting daemon`
+      `[DaemonMcpProxy] Daemon startup options differ (${mismatches.join(", ")}), restarting daemon`
     );
     await this.daemonManager.restart(this.config.daemonOptions ?? {});
     const ready = await this.daemonManager.waitForReady(DAEMON_STARTUP_TIMEOUT_MS);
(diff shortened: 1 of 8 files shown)


kaeawc:
## Multi-agent review + code review — addressed

Ran the AutoMobile code review plus three independent persona reviewers (spec-compliance skeptic, regression hunter, staff-engineer/direction). **All three came back clean — no blockers, no majors, no regressions.** Highlights verified:

- **Spec invariant holds** across all 4 flag×schema states — never advertise `outputSchema` without returning conforming `structuredContent`, and vice versa.
- **No regressions:** every internal `structuredContent` consumer (`DefaultUIStateSetup`, `PlanExecutor`, `lastHierarchy`/scroll-position session cache, `finalizeToolResponse`) reads *before* the wire boundary; daemon/proxy relay the already-stripped result; `--cli` and `schemas/tool-definitions.json` generation unaffected (flag defaults off).
- **Error paths never reach the strip** (they `throw` above it); the strip only acts on success envelopes.

**Folded into this PR** from review feedback: unified the two "has `outputSchema`" predicates (written as an explicit null-check vs truthiness at the two gate sites) into a single `toolHasOutputSchema` helper, so the wire result and the advertised schema can never desync. Added a test asserting both sites share the predicate.

**Deferred as follow-ups** (cross-cutting / optional):
- #2962 — debug-log intentional `structuredContent` omission (HarperZ9's intentional-vs-accidental signal).
- #2963 — emit `tools/list_changed` when output-reduction flags toggle at runtime (applies to all sibling flags, benign today since flags are set at startup).

kaeawc:
## Addressed Codex P2 — honor the flag when reusing a daemon

Fixed in 59fd178. `DaemonMcpProxy.ensureStartupOptionsMatch` compared only `embeddedSdk`, so a same-build MCP client requesting `--tool-results-no-structured-content` against an already-running daemon started without it would silently reuse that daemon — schema tools kept returning `structuredContent` and advertising `outputSchema` until a manual restart (and the reverse mismatch stripped unexpectedly).

The reuse gate now compares `embeddedSdk` **plus every output-reduction flag**, derived from `OUTPUT_REDUCTION_FLAG_SPECS` (whose `field` names map 1:1 to `DaemonOptions`), so the whole latent class is fixed and future flags are covered automatically. Any mismatch restarts the daemon; error/log messages now name the differing options. Added two `daemonMcpProxy.test.ts` cases (restart-on-`toolResultsNoStructuredContent`-mismatch; no-restart-when-matching).

## Failing CI (`XCTestRunner Simulator Tests` / `iOS`)

Not caused by this PR — it's the known-flaky iOS Reminders e2e job failing at "step 2 (observe)", a device/CtrlProxy observe flake. Verified:
- The Swift plan executor reads only `result["content"][].text` (`AutoMobilePlanExecutor.swift:415-426`), **never** `structuredContent`; this PR always preserves `content[0].text`, so the wire-strip cannot break it.
- The executor + `ios/control-proxy` are unchanged on `main` since this branch's base (no fix I'm missing), and `main` itself went red on this same job across several unrelated PRs on 07-03 before recovering — an intermittent environmental flake. Re-running the job.

kaeawc:
## Rebased onto latest main + reconciled with #2899

Rebasing surfaced that main already shipped **#2899/#2933** (`d1aaea40e`), which independently added a same-named `stripToolResultStructuredContent.ts`. That landed the **flag-gated** strip (drops `structuredContent` only when `--tool-results-no-structured-content` is set) + `outputSchema` suppression. But #2759's core ask — the **unconditional no-schema strip**, so `observe` (the largest payload) drops the duplicate *by default* with no flag — was **not** on main.

Reconciled rather than duplicated:
- Extended main's `stripToolResultStructuredContent` to take `hasOutputSchema` and drop when `!hasOutputSchema || flag` (no-schema unconditional + the existing flag path). Still at the CallTool wire boundary, so internal `tool.handler(...)` callers keep the full envelope.
- Added a shared `toolHasOutputSchema` predicate — the single source both the wire strip and the `getToolDefinitions` `outputSchema` advertisement key off, so they can't disagree.
- Updated main's #2899 unit + wire-boundary tests to the new signature (the wire probe now declares an `outputSchema`, matching its exportPlan-like intent) and added `structuredContentGating.test.ts` for the no-schema unconditional strip, `tools/list` gating, size win, and the internal-caller carve-out.
- Kept the daemon-reuse flag-gate fix (Codex P2), re-applied on main's `ensureStartupOptionsMatch`.

Local: lint + build clean; all new/updated tests pass. The only local test failures are pre-existing `jimp`/wasm image failures (`ImageTransformer.ts:194`, `PerceptualHasher`) that reproduce on a clean `main` checkout — unrelated to this diff.

PR #2959, opened 2026-07-03
fix(db): make NavigationGraphManager.recordNavigationEvent graph write transactional (#2791)

### LINKED ISSUE #2791: fix(db): make NavigationGraphManager.recordNavigationEvent graph write transactional

kaeawc:
## What

`NavigationGraphManager.recordNavigationEvent` wrote node + node-visit + node-modals + edge + edge-traversal + UI-elements + edge-modals + scroll-position + `touchApp` as a sequence of **separately-awaited** statements with **no enclosing transaction**. A mid-sequence throw (e.g. a UNIQUE conflict) or a daemon interrupted mid-`await` left the on-disk navigation graph **partially applied** — a node/edge persisted while its coverage/metadata rows were missing, so `getNavigationGraph`/coverage queries under- or mis-count traversals for edges that visibly exist.

This wraps **only the DB writes** in a single `db.transaction()`, so the graph write is all-or-nothing.

Closes #2791.

## Why

Per finding R4 of the 2026-07-01 DB concurrency/migration audit (P1). All navigation repositories resolve to the **same singleton `getDatabase()`** connection, so one transaction can cover both `NavigationRepository` and `TestCoverageRepository`, but today none of these writes are transactional.

The daemon runs a **single** `bun:sqlite` connection guarded by a hand-rolled async mutex that serializes *individual* queries but **releases across every `await`**. So between (say) `createEdge` and `recordEdgeTraversal` the event loop yields; if the next statement throws, the node + edge are already committed with no rollback.

**Headline risk this design defends against (per reviews — Mira Kessler & Sam Whitfield):** getting the transaction threading wrong doesn't degrade performance — it **deadlocks the entire daemon**, which is strictly worse than the partial-write bug. A `db.transaction()` reserves the one connection under an owner symbol (`bunSqliteDialect.ts` `#reserveTransaction`). Any query NOT dispatched through that same `trx` gets a different owner and `#enterQuery` loops forever — there is **no `busy_timeout` escape** (in-process JS mutex, not SQLite's lock). Every repo method resolves its handle via `this.getDb()` → singleton, so a single stray call inside the transaction wedges the daemon's only connection.

## How

**Structural safety over per-call-site threading.** Rather than thread a `trx` param through ~35 `getDb()` call sites (a single miss = silent deadlock), the whole repository instance is rebound to the transaction:

- **`NavigationRepository.withExecutor(trx)` / `TestCoverageRepository.withExecutor(trx)`** return an instance whose `getDb()` returns `trx`, so **every** read and write on it uses the transaction automatically — it is *impossible* for a bound instance to touch the singleton. This eliminates the stray-`getDb()` deadlock class the reviewers flagged, at the instance level.
- **`NavigationRepository.runInTransaction(fn)`** → `getDb().transaction().execute(fn)`, mirroring `installedAppsRepository`'s in-repo transaction so the manager never handles a raw db.
- **`getOrCreateUIElement` is now transaction-aware** via Kysely's `db.isTransaction`. It opens its own transaction on the singleton path (it's a dynamic-match SELECT-then-INSERT that can't be a single upsert), but when already bound to a caller's `trx` it runs the body **directly on the executor** — a nested `BEGIN` on the same connection is rejected by `#reserveTransaction` ("Nested transactions are not supported").
- **`recordNavigationEvent`** now wraps only the DB writes in `runInTransaction`, using `withExecutor(trx)`-bound repos for every call inside — including the reads (`getNode`, `getNodeModals`) and the final `touchApp`. `storeUIElements`/`storeScrollPosition` take the bound repo so their writes join the same transaction.
- **Side effects moved OUTSIDE the transaction** (only after commit): `TelemetryRecorder.recordNavigationEvent(...)`, `notifyGraphUpdated()`, and the in-memory `this.currentScreen` / `this.activeNavigation` assignments. `previousScreen` is captured before the transaction so the edge is computed correctly and, on rollback, `currentScreen` stays the previous screen.

## Testing

- `bun test test/features/navigation/NavigationGraphManager.test.ts` → **49 pass, 0 fail** (~0.27s), including 4 new transaction tests.
- `bun test test/db/navigationRepository.test.ts test/db/testCoverageRepository.test.ts` → **57 pass, 0 fail**.
- Full suite: `bun test` → **5925 pass, 2 skip, 0 fail**.
- `turbo run lint build` → clean; `schemas/tool-definitions.json` regenerated, no drift.

New tests (real in-memory `createTestDatabase()` + a fake `TestCoverageRepository` that throws on `recordEdgeTraversal` and overrides `withExecutor` so the throw survives rebinding, <100ms, no device):
- **Rollback** — injected throw in `recordEdgeTraversal` → zero new `navigation_nodes`/`navigation_edges`/`test_edge_coverage` rows (counts before/after), and `currentScreen` unchanged.
- **Happy path** — node + visit + edge + traversal rows all present.
- **Side-effect ordering** — telemetry + `notifyGraphUpdated` observe `touchApp` (final in-txn write) already called → they run after commit, never inside; and neither fires on rollback.
- **Concurrent writer** — `recordNavigationEvent` run alongside an independent writer on the same connection completes (guarded by a timeout that fails on deadlock).
- The existing scroll-position tests already route `getOrCreateUIElement` through the new transaction, so the `isTransaction` nested-avoidance path is exercised end-to-end.

## Acceptance criteria (#2791)
- [x] Mid-sequence throw → `recordNavigationEvent` rolls back; zero new node/edge/coverage rows persisted (counts before/after).
- [x] On success, node + visit + edge + traversal rows all present (happy path preserved).
- [x] `TelemetryRecorder.recordNavigationEvent` / `notifyGraphUpdated()` NOT invoked on rollback; not held open across them on success.
- [x] No statement in the sequence uses `this.getDb()` — all go through `trx` (enforced structurally by `withExecutor` binding the whole instance); `touchApp` is inside the txn.
- [x] Concurrent-writer regression test proves completion (no daemon stall).
- [x] In-memory field rollback invariant: `currentScreen`/`activeNavigation` unchanged on rollback.
- [x] Txn body issues only DB statements — side effects asserted to fire only after the `trx` callback returns.
- [x] Regression/unit tests in `NavigationGraphManager.test.ts` using real in-memory DB + fakes, <100ms, non-flaky, no real device.

## Notes
- **Sequenced after #2790** (atomic upserts remove the most common mid-sequence throw); this transaction is independently warranted for the crash/shutdown-interruption case.

### DIFF
--- a/src/db/navigationRepository.ts
+++ b/src/db/navigationRepository.ts
@@ -38,6 +38,35 @@ export class NavigationRepository {
     }
     return getDatabase();
   }
+
+  /**
+   * Return a repository instance whose every read and write runs on the supplied
+   * executor (a Kysely transaction handle) instead of the shared singleton
+   * connection.
+   *
+   * This is the safe way to enlist navigation writes in a caller-owned
+   * transaction: binding the whole instance means no method can accidentally fall
+   * back to `getDatabase()` mid-transaction. A stray singleton query while a
+   * transaction holds the single bun:sqlite connection deadlocks the daemon —
+   * `#enterQuery` (bunSqliteDialect) spins forever for a non-owner and there is no
+   * busy_timeout escape (the in-process JS mutex, not SQLite's lock).
+   */
+  withExecutor(executor: Kysely<Database>): NavigationRepository {
+    return new NavigationRepository(executor);
+  }
+
+  /**
+   * Run `fn` inside a single transaction on this repository's connection, passing
+   * the transaction executor to thread into `withExecutor`. Mirrors
+   * `installedAppsRepository`'s in-repo transaction so callers never touch a raw
+   * database handle. Keep the callback body to DB statements only — the
+   * transaction takes exclusive ownership of the single connection, so any awaited
+   * non-DB IO inside it stalls every other daemon writer for its duration.
+   */
+  async runInTransaction<T>(fn: (trx: Kysely<Database>) => Promise<T>): Promise<T> {
+    return this.getDb().transaction().execute(fn);
+  }
+
   /**
    * Get or create a navigation app record.
    */
@@ -343,74 +372,100 @@ export class NavigationRepository {
     // callers for the same element serialize rather than both inserting duplicate rows
     // (R2356). Keep the body to DB reads/writes only — no IO — since it blocks the daemon
     // for its duration.
-    return db.transaction().execute(async trx => {
-      // Try to find existing element with same properties
-      let query = trx
-        .selectFrom("ui_elements")
-        .selectAll()
-        .where("app_id", "=", appId);
+    //
+    // When this repo is already bound to a caller's transaction (via withExecutor),
+    // opening another transaction would be a nested BEGIN on the same connection —
+    // `#reserveTransaction` throws "Nested transactions are not supported". Run the
+    // body directly on the existing executor instead; it is already serialized and
+    // atomic under the enclosing transaction.
+    if (db.isTransaction) {
+      return this.getOrCreateUIElementWithin(db, appId, element, timestamp);
+    }
+    return db.transaction().execute(trx =>
+      this.getOrCreateUIElementWithin(trx, appId, element, timestamp)
+    );
+  }
 
-      if (element.text !== undefined) {
-        query = query.where("text", "=", element.text);
-      }
-      if (element.resourceId !== undefined) {
-        query = query.where("resource_id", "=", element.resourceId);
-      }
-      if (element.contentDescription !== undefined) {
-        query = query.where("content_description", "=", element.contentDescription);
-      }
-      if (element.className !== undefined) {
-        query = query.where("class_name", "=", element.className);
-      }
-      if (element.bounds) {
-        query = query
-          .where("bounds_left", "=", element.bounds.left)
-          .where("bounds_top", "=", element.bounds.top)
-          .where("bounds_right", "=", element.bounds.right)
-          .where("bounds_bottom", "=", element.bounds.bottom);
-      }
+  private async getOrCreateUIElementWithin(
+    trx: Kysely<Database>,
+    appId: string,
+    element: {
+      text?: string;
+      resourceId?: string;
+      contentDescription?: string;
+      className?: string;
+      bounds?: { left: number; top: number; right: number; bottom: number };
+      clickable?: boolean;
+      scrollable?: boolean;
+    },
+    timestamp: number
+  ): Promise<UIElement> {
+    // Try to find existing element with same properties
+    let query = trx
+      .selectFrom("ui_elements")
+      .selectAll()
+      .where("app_id", "=", appId);
 
-      const existing = await query.executeTakeFirst();
+    if (element.text !== undefined) {
+      query = query.where("text", "=", element.text);
+    }
+    if (element.resourceId !== undefined) {
+      query = query.where("resource_id", "=", element.resourceId);
+    }
+    if (element.contentDescription !== undefined) {
+      query = query.where("content_description", "=", element.contentDescription);
+    }
+    if (element.className !== undefined) {
+      query = query.where("class_name", "=", element.className);
+    }
+    if (element.bounds) {
+      query = query
+        .where("bounds_left", "=", element.bounds.left)
+        .where("bounds_top", "=", element.bounds.top)
+        .where("bounds_right", "=", element.bounds.right)
+        .where("bounds_bottom", "=", element.bounds.bottom);
+    }
 
-      if (existing) {
-        // Update last_seen_at
-        await trx
-          .updateTable("ui_elements")
-          .set({ last_seen_at: timestamp })
-          .where("id", "=", existing.id)
-          .execute();
+    const existing = await query.executeTakeFirst();
 
-        return {
-          ...existing,
-          last_seen_at: timestamp,
-        };
-      }
+    if (existing) {
+      // Update last_seen_at
+      await trx
+        .updateTable("ui_elements")
+        .set({ last_seen_at: timestamp })
+        .where("id", "=", existing.id)
+        .execute();
 
-      // Create new UI element
-      const newElement: NewUIElement = {
-        app_id: appId,
-        text: element.text || null,
-        resource_id: element.resourceId || null,
-        content_description: element.contentDescription || null,
-        class_name: element.className || null,
-        bounds_left: element.bounds?.left || null,
-        bounds_top: element.bounds?.top || null,
-        bounds_right: element.bounds?.right || null,
-        bounds_bottom: element.bounds?.bottom || null,
-        clickable: element.clickable !== undefined ? (element.clickable ? 1 : 0) : null,
-        scrollable: element.scrollable !== undefined ? (element.scrollable ? 1 : 0) : null,
-        first_seen_at: timestamp,
+      return {
+        ...existing,
         last_seen_at: timestamp,
       };
+    }
 
-      const result = await trx
-        .insertInto("ui_elements")
-        .values(newElement)
-        .returningAll()
-        .executeTakeFirstOrThrow();
+    // Create new UI element
+    const newElement: NewUIElement = {
+      app_id: appId,
+      text: element.text || null,
+      resource_id: element.resourceId || null,
+      content_description: element.contentDescription || null,
+      class_name: element.className || null,
+      bounds_left: element.bounds?.left || null,
+      bounds_top: element.bounds?.top || null,
+      bounds_right: element.bounds?.right || null,
+      bounds_bottom: element.bounds?.bottom || null,
+      clickable: element.clickable !== undefined ? (element.clickable ? 1 : 0) : null,
+      scrollable: element.scrollable !== undefined ? (element.scrollable ? 1 : 0) : null,
+      first_seen_at: timestamp,
+      last_seen_at: timestamp,
+    };
 
-      return result;
-    });
+    const result = await trx
+      .insertInto("ui_elements")
+      .values(newElement)
+      .returningAll()
+      .executeTakeFirstOrThrow();
+
+    return result;
   }
 
   /**
--- a/src/db/testCoverageRepository.ts
+++ b/src/db/testCoverageRepository.ts
@@ -37,6 +37,20 @@ export class TestCoverageRepository {
   }
 
   /**
+   * Return a repository instance whose every read and write runs on the supplied
+   * executor (a Kysely transaction handle) instead of the shared singleton
+   * connection, so coverage writes can enlist in a caller-owned transaction
+   * alongside navigation writes on the same connection.
+   *
+   * Binding the whole instance means no method can fall back to `getDatabase()`
+   * mid-transaction — a stray singleton query while a transaction holds the single
+   * bun:sqlite connection deadlocks the daemon (see `NavigationRepository.withExecutor`).
+   */
+  withExecutor(executor: Kysely<Database>): TestCoverageRepository {
+    return new TestCoverageRepository(this.timer, executor);
+  }
+
+  /**
    * Start a new test coverage session for an app.
    */
   async startSession(sessionUuid: string, appId: string): Promise<TestCoverageSession> {
--- a/src/features/navigation/NavigationGraphManager.ts
+++ b/src/features/navigation/NavigationGraphManager.ts
@@ -202,6 +202,15 @@ export class NavigationGraphManager implements NavigationGraphService {
 
   /**
    * Create a new instance for testing with injected dependencies.
+   *
+   * Precondition for recordNavigationEvent's atomicity: `repository` and
+   * `testCoverageRepository` MUST resolve to the same underlying connection
+   * (the default is both unbound → shared `getDatabase()` singleton). The
+   * transactional write opens one transaction on the navigation connection and
+   * enlists the coverage repo onto it via withExecutor; injecting a coverage
+   * repo bound to a DIFFERENT Kysely handle would split writes across
+   * connections and silently defeat the rollback guarantee. Tests that inject a
+   * db must pass the SAME instance to both repositories.
    */
   public static createForTesting(
     repository?: NavigationRepository,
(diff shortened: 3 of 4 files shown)


kaeawc:
## Multi-perspective review (3 personas) + feedback addressed

Ran three independent reviews (concurrency-correctness, regression/testability, architecture/direction). **No blocking findings** — the transactional conversion is correct; the two hard parts (side-effect placement outside the txn, nested-txn avoidance) are handled. Feedback addressed in the follow-up commit:

**Mira Kessler (SQLite correctness)** — Verified deadlock-safe: every DB call inside the txn body runs on a `withExecutor(trx)`-bound repo, so `#enterQuery` always sees the same owner and never spins; connection identity holds because both repos rebind to the same `trx`. `getOrCreateUIElement`'s `db.isTransaction` branch correctly avoids a nested `BEGIN`. Rollback safe (session total counters roll back with the txn). *Nit → fixed:* the rollback test now also asserts `test_coverage_sessions.total_*` counters revert.

**Sam Whitfield (regressions/testability)** — Mutation-verified the rollback and deadlock tests genuinely catch their failure modes. No behavior regression: `activeNavigation`/`notifyGraphUpdated` moving post-commit is strictly more correct, and both production callers (`AndroidCtrlProxyClient`, `IosSdkEventIngestor`) already `try/catch` the awaited call. *Weak test → fixed:* the "side effects fire only after commit" test now records touchApp's call count at *every* notify/telemetry fire and asserts exactly one fire after the final in-txn write, so a stray *early* side effect is caught (mutation-verified) instead of masked by a later post-commit fire.

**Devin Osei (architecture/direction)** — `withExecutor` (whole-instance rebind) is the right abstraction vs per-method `trx` threading (which would reintroduce the stray-singleton deadlock risk); keeping orchestration in the manager (vs pushing into the repo) is correct since it spans two repos + correlation logic. *In-PR tweak → done:* documented the load-bearing precondition that both repos must share the same connection, at the transaction site and on `createForTesting`.

### Follow-ups filed (out of scope for this PR)
- #2968 — transactionalize `promoteSuggestion` (structural writes) + `recordHierarchyNavigation` Case 1; same `withExecutor`/`runInTransaction` vehicle. Also folds in Devin's note to promote the `isTransaction` dual-path into a documented convention if more transactional callers land.
- #2969 — isolate this test file's singleton-backed describe blocks (pre-existing cold-first-run flakiness both Mira and Sam hit; unrelated to this PR's logic).

PR #2976, opened 2026-07-04
feat(observe): wire observeResultCompact flag into sanitizeObserveResult (#2951)

### LINKED ISSUE #2951: Wire observeResultCompact flag into sanitizeObserveResult (flag plumbed but dormant)
## Summary

The `observeResultCompact` output-reduction flag is **fully plumbed but consumed by nothing in `src/`** — the identical dormant state that #2899 (PR #2933) just fixed for `toolResultsNoStructuredContent`.

It flows CLI/env → `outputReductionFlags` → `ServerConfig`, but no production code reads `serverConfig.isObserveResultCompactEnabled()`:

- `src/utils/outputReductionFlags.ts:12,42-47` — field + spec (`--observe-result-compact` / `AUTOMOBILE_OBSERVE_RESULT_COMPACT` / feature-flag key `observe-result-compact`)
- `src/utils/ServerConfig.ts:39,223-229` — setter/getter
- `src/features/featureFlags/FeatureFlagApplier.ts` + `src/daemon/*` — wiring
- **Readers in `src/`: none** (only the getter definition and tests reference it)

So enabling `--observe-result-compact` (or its env/feature-flag) currently has **no effect** on tool output.

## Where it belongs (unlike #2899)

Per the #2899 issue text and the review of PR #2891: `observeResultCompact` is an **`ObserveResult`-shape** concern, so it belongs **inside `sanitizeObserveResult`** (`src/features/observe/output/ObserveResultOutput.ts:69`), gated by a new field on `SanitizeObserveConfig` (`ObserveResultOutput.ts:49`) — *not* at the envelope/wire boundary where `toolResultsNoStructuredContent` lives.

`finalizeToolResponse` already threads `SanitizeObserveConfig` and reads `serverConfig.isObserveResultDropElementsEnabled()` (`src/server/finalizeToolResponse.ts`); the natural implementation adds `compact: serverConfig.isObserveResultCompactEnabled()` to that config object and implements the compaction in `sanitizeObserveResult`, exactly mirroring how `dropElements` is handled.

## Open design questions to resolve first

- **Define "compact".** The doc (`docs/design-docs/mcp/feature-flags.md`) only says "Emit `observe` results in a compact form." Decide the concrete transform: e.g. array-of-arrays / positional node encoding, single-letter or abbreviated attribute keys, dropping default-valued attributes beyond today's allow-list, or flattening `bounds` (currently a 4-key object) to `[l,t,r,b]`. Whatever is chosen must round-trip losslessly enough for downstream consumers (`lastHierarchy` diff baseline #2761, `elements` extraction).
- **Precedence / composition with `observeResultDropElements`** (drops `elements`) and the wire-level `toolResultsNoStructuredContent` (#2899). These compose: compact reshapes the payload, drop-elements removes a sub-tree, strip removes the duplicate envelope copy. Confirm they stack cleanly and add a test with all three enabled.
- **Consumer tolerance.** Anything reading the observe payload shape (IDE, plan diffing, `UIStateExtractor`) must accept the compact form, or the compaction must be output-only (like `sanitizeObserveResult`, which deep-clones and never mutates the cached `lastHierarchy`).

## Acceptance / test

- Unit test in `test/features/observe/output/ObserveResultOutput*.test.ts`: with the compact gate enabled, `sanitizeObserveResult` emits the compact form; disabled, it emits today's shape. Cover interplay with `dropElements`.
- Unit test confirming `finalizeToolResponse` passes the compact flag through (mirrors the existing `dropElements` gate test in `test/server/finalizeToolResponse.test.ts`).
- Confirm the in-memory `ObserveResult` / `lastHierarchy` cache is never mutated (output-only), as with the existing sanitize.

## Context

- Sibling flag wired by: PR #2933 (issue #2899) — `toolResultsNoStructuredContent`, wire-boundary strip.
- Part of the Output Size Reduction milestone (#2755–#2762).
- Flagged during the PR #2933 adversarial review (completeness pass): "`observeResultCompact` is fully plumbed but consumed by nothing — the identical dormant state this PR is fixing."

### LINKED ISSUE #2978: observeResultCompact: resolve mixed bounds representation (node tuples vs. object elements/root/window bounds)
## Summary

Follow-up from #2951 / PR #2976, which wired the `--observe-result-compact` flag by flattening **view-hierarchy node** `bounds` from the object `{left,top,right,bottom}` to the positional tuple `[left,top,right,bottom]`.

That PR deliberately scoped compaction to `ViewHierarchyNode.bounds` only. Several **sibling `bounds` fields on the same observe payload keep the object shape**, so a single compacted response now carries `bounds` in *two* representations simultaneously:

- `ViewHierarchyNode.bounds` → **tuple** (compacted)
- `Element.bounds` in `elements[]` (`src/models/Element.ts:7`) → object
- `Hierarchy.bounds` (iOS root, `src/models/ViewHierarchyResult.ts:75`) → object
- `ViewHierarchyWindowInfo.bounds`, `ContentHiddenRegion.bounds` → object

Any wire consumer reading `bounds` generically must now branch on `Array.isArray(bounds)`. Two independent reviewers (architecture + correctness) flagged this as the main coherence smell, while agreeing the node-only scoping is a defensible v1 (the node hierarchy holds the vast majority of bounds — 130 vs a handful in the baseline fixture — and `--observe-result-drop-elements` removes `elements` entirely, which incidentally eliminates the mix for that flag combination).

## Decision needed

Pick the permanent end state:
- **(a) Compact everywhere** — extend compaction to `elements[].bounds`, `Hierarchy.bounds`, and `ViewHierarchyWindowInfo.bounds` so *all* `bounds` on a compacted response are tuples. If so, extract a shared `compactBounds()` helper rather than duplicating the `node.bounds as unknown` cast dance (`src/features/observe/output/ObserveResultOutput.ts:204-210`).
- **(b) Document node-hierarchy-scoped forever** — state explicitly in `docs/design-docs/mcp/feature-flags.md` (and ideally a wire-schema note) that compaction only applies to the node hierarchy, and that `elements`/root/window bounds remain objects.

## Consumer note

The in-repo Kotlin consumer `RealLayoutDataSource.parseBounds` was taught the tuple form in PR #2976 (`android/desktop-core/.../RealLayoutDataSource.kt`). If option (a) lands, no further consumer change is needed there. External MCP clients are the only other decoders and see the compact form only when they opt into the (off-by-default) flag.

## Also from the PR review (smaller follow-ups)
- Add a test confirming `--observe-result-compact` composes with `--actions-diff-observe` (the diffed post-action observation path, which runs outside `finalizeToolResponse`).
- Consider documenting the tuple order `[left,top,right,bottom]` in the wire/tool schema, not just the flag doc, since external clients are the decoders.

## Context
- Issue: #2951
- PR: #2976
- Sibling dormant-flag wiring: #2899 / PR #2933

kaeawc:
## What

Wires the `--observe-result-compact` output-reduction flag (env: `AUTOMOBILE_OBSERVE_RESULT_COMPACT`) so it actually reshapes `observe` output, and defines "compact" as flattening **every** `bounds` object to a positional tuple.

Closes #2951. Closes #2978.

## Why

The flag was **fully plumbed but consumed by nothing** — the identical dormant state #2899 (PR #2933) fixed for `toolResultsNoStructuredContent`. It flowed CLI/env → `outputReductionFlags` → `ServerConfig` and was even applied to `serverConfig` (`daemon.ts:203`, `FeatureFlagApplier.ts:61`), but **no production code read** `serverConfig.isObserveResultCompactEnabled()`. Enabling it had **no effect**.

## How

**Define "compact":** every `bounds` object in the served observe result is flattened from the 4-key object `{left, top, right, bottom}` to the positional tuple `[left, top, right, bottom]`. `bounds` is the most-repeated object shape after `trimNodes` (~130 occurrences / 7.7% / ~975 tokens on the baseline `android-home.json`), and each carries the four repeated key strings, so this is the largest remaining lossless reduction. The order is fixed and documented, so it round-trips losslessly.

**1. `sanitizeObserveResult` (`src/features/observe/output/ObserveResultOutput.ts`).** New `compact?: boolean` on `SanitizeObserveConfig`; when set, `compactObserveBounds` does a single deep walk of the deep-cloned output and flattens **every** `bounds` site — view-hierarchy nodes, the `elements` arrays, the `Hierarchy` root, window/content-hidden-region bounds, and the focused/accessibility-focused/awaited element fields — via a shared `compactBounds()` helper. So a compacted response never carries a **mix** of object- and tuple-shaped bounds (the coherence concern raised in review). It runs after `trimNodes`, before `dropElements`, so it is **output-only** (the in-memory `ObserveResult` / `lastHierarchy` diff baseline and internal consumers are untouched) and idempotent (`!Array.isArray` guard → safe under proxy+daemon double relay).

Only fields literally named `bounds` are flattened, so insets (`systemInsets`, a `{top,bottom,left,right}` object) and other shapes are never mis-compacted. `rawViewHierarchy` is skipped — `raw: true` still returns the unshaped hierarchy.

**2. `finalizeToolResponse` (`src/server/finalizeToolResponse.ts`).** Adds `compact: serverConfig.isObserveResultCompactEnabled()` to the sanitize config, mirroring the existing `dropElements` gate. Both `structuredContent` and `content[0].text` are rewritten from the same object so they never diverge, on both the top-level `observe` and nested `.observation` (tapOn/swipeOn) paths.

**3. In-repo wire consumer (`android/desktop-core/.../RealLayoutDataSource.kt`).** `parseBounds` previously handled `JsonObject`/`JsonPrimitive` but let a `JsonArray` fall to `else -> ElementBounds(0,0,0,0)` — with the flag on, the desktop Layout Inspector would silently render every element at a zero rect. Added a `JsonArray` branch decoding the `[l,t,r,b]` tuple. Compiled locally with JDK 21 (`:desktop-core:compileKotlin` BUILD SUCCESSFUL), ktfmt-formatted.

**4. Docs.** `docs/design-docs/mcp/feature-flags.md` states the concrete transform, its coverage, the insets/raw exclusions, and composition with the sibling flags.

Compaction is opt-in and off by default, so external wire consumers only see the tuple form after explicitly enabling it. Verified all in-memory `.bounds` consumers read during the observe pipeline, before `finalizeToolResponse` clones — none see the compact form.

## Testing

TDD — tests written first (red), then implementation (green).

- `test/features/observe/output/ObserveResultOutput.test.ts` — `compact bounds flatten (gated)`: node flatten / keeps object when off / **round-trips losslessly** / output-only / measurably shrinks **bytes and tokens** on the real fixture / no-op on bounds-less nodes / array-shaped root + nested children / composes with `trimNodes` + `dropElements`. Plus `compact flattens EVERY bounds site (#2978)`: all four `elements` buckets, Hierarchy root, windows, content-hidden-regions, focused/accessibility-focused/awaited fields, **never compacts `systemInsets`**, leaves `rawViewHierarchy` unshaped, output-only across every site.
- `test/server/finalizeToolResponse.test.ts` — EC-C: gate on → tuple bounds in **both** `structuredContent` and `content[0].text` / gate off → object bounds / output-only / **nested `.observation` (tapOn) path** / composes with `drop-elements` + the wire-level `tool-results-no-structured-content` strip (all three on).

Mutation-tested (reverting the compaction to a no-op fails the compact tests). Verified: `bun test` **5988 pass** (the 6 failures are pre-existing jimp/wasm image + memory-leak tests, reproduced on a clean tree without this change); `bun run lint` clean; `bun run build` ok; pre-commit `schemas/tool-definitions.json` regeneration confirms **no drift** (default-off flag).

## Acceptance criteria (#2951, #2978)
- [x] Compact gate enabled → `sanitizeObserveResult` emits the compact form; disabled → today's shape. Interplay with `dropElements` covered.
- [x] `finalizeToolResponse` passes the compact flag through (mirrors the `dropElements` gate test), including the nested `.observation` path.
- [x] In-memory `ObserveResult` / `lastHierarchy` cache never mutated (output-only), asserted at the unit and finalize layers and across every bounds site.
- [x] Single wire representation: **all** bounds are tuples in compact mode (no object/tuple mix); `systemInsets` and `rawViewHierarchy` correctly excluded.
- [x] In-repo Kotlin consumer teaches the tuple shape (no silent zero-rects).


kaeawc:
## Adversarial review (3 perspectives) + fixes

Ran three independent reviewers — correctness/skeptic, architecture/direction, test-rigor — against the diff. Summary and what changed since the first commit:

### Verdicts
- **Correctness:** transform is pure, output-only (deep-clone boundary at `ObserveResultOutput.ts:89` before any mutation), idempotent (`!Array.isArray` guard, safe under proxy+daemon double relay), and type-safe (`bun run build`/`tsc` clean). All ~40 in-memory `.bounds` consumers read the original object, unaffected.
- **Architecture:** good v1 direction. Quantified savings on `android-home.json`: ~3,900 bytes / 7.7% / ~975 tokens — bounds is the most-repeated object shape after `trimNodes`. The riskier levers (key abbreviation, positional node arrays) correctly deferred as they'd require a decoder on every consumer.
- **Test rigor:** mutation-tested clean (reverting `compactHierarchyBounds` to a no-op fails 7 tests), <100ms, pure in-memory, strong round-trip + output-only assertions at both layers.

### Addressed in `4aa6a28`
1. **Silent-zero-rect break in an in-repo wire consumer (correctness, substantive).** `RealLayoutDataSource.parseBounds` (`android/desktop-core/.../RealLayoutDataSource.kt`) handled `JsonObject`/`JsonPrimitive` but a `JsonArray` fell to `else -> ElementBounds(0,0,0,0)` — with `--observe-result-compact` on, the desktop Layout Inspector would render every element at a zero rect (no exception, just wrong). Added a `JsonArray` branch that decodes the `[l,t,r,b]` tuple. Compiled locally with JDK 21 (`:desktop-core:compileKotlin` BUILD SUCCESSFUL), ktfmt-formatted.
2. **Missing coverage on the action `.observation` path (test rigor, insist).** Every EC-C test used `{name:"observe"}`; the nested `.observation` branch (tapOn/swipeOn — the common case) is distinct code. Added a `finalizeToolResponse` test asserting compact tuples on `.observation` in both `structuredContent` and mirrored `content[0].text`.

### Deferred to follow-ups (both reviewers converged here)
- **Mixed `bounds` representation.** Compact touches only `ViewHierarchyNode.bounds`; `elements[].bounds`, the iOS `Hierarchy.bounds` root, and `ViewHierarchyWindowInfo.bounds` stay objects, so a response can carry both shapes. Node-only scoping is defensible for v1 (node hierarchy holds the vast majority of bounds; `dropElements` removes `elements` entirely), but the end state should be decided. Filing a follow-up.
- `actionsDiffObserve` composition test; wire-schema doc note for external decoders. Follow-up.

kaeawc:
### Update: implemented "compact everywhere" (option a from #2978)

Rather than defer the mixed-representation question, generalized compaction to **every** `bounds` site in one deep walk (`compactObserveBounds` + shared `compactBounds()` helper): view-hierarchy nodes, all four `elements[]` buckets, the `Hierarchy` root, window/content-hidden-region bounds, and the focused/accessibility-focused/awaited element fields. A compacted response now carries `bounds` in **one shape everywhere** — no `Array.isArray` branching for consumers.

Guards: only fields literally named `bounds` are flattened (so `systemInsets` `{top,bottom,left,right}` is never mis-compacted), `rawViewHierarchy` is skipped (raw stays raw), still output-only + idempotent. New tests cover every site plus the systemInsets/raw exclusions. `bun test` 5988 pass. This PR now **Closes #2978** as well.

PR #2977, opened 2026-07-04
fix(db): assert nav + coverage repos share one connection before txn (#2968)

### LINKED ISSUE #2968: fix(db): transactionalize promoteSuggestion + recordHierarchyNavigation graph writes (siblings of #2791)
## Summary

## The problem

```ts
```

## Proposed fix

## Acceptance criteria

## References

kaeawc:
## What

Adds a lightweight runtime assertion that `NavigationGraphManager`'s navigation
repository (`this.repository`) and test-coverage repository
(`this.testCoverageRepository`) resolve to the **same** underlying database
connection **before** `recordNavigationEvent` opens its shared transaction. A
mistakenly foreign-bound coverage repo now fails loudly with an `ActionableError`
instead of silently splitting writes across connections.

Closes #2968.

## Why

Deferred hardening from the PR #2959 architecture review (Devin Osei). #2959 made
`recordNavigationEvent` transactional by opening one `db.transaction()` on the
navigation connection and enlisting the coverage repo onto that same `trx` via
`withExecutor`. Its atomicity guarantee rests on a precondition — **both
repositories share one connection** — that until now was enforced only by
convention (both default to the `getDatabase()` singleton) plus doc comments at
the transaction site and `createForTesting`.

If a coverage repo were bound to a *different* connection, `withExecutor(trx)`
silently rebinds it to the navigation `trx`, so its writes land on the navigation
connection rather than where the injected repo pointed — quietly ignoring caller
intent and, if the `withExecutor` rebinding were ever removed in a future
extension, splitting the write and defeating the rollback. The guard makes this
misconfiguration fail loudly at the point it matters.

## How

- **`NavigationRepository.resolveConnection()` / `TestCoverageRepository.resolveConnection()`**
  — narrow public accessors returning the bound executor (constructor/`withExecutor`)
  or the shared `getDatabase()` singleton when unbound. Two unbound repositories
  resolve to the same memoized singleton, so identity (`===`) holds by construction.
- **`NavigationGraphManager.assertSharedConnection()`** — throws `ActionableError`
  (error-handling convention strategy 1) when the two repos resolve to different
  connections. Called at the top of `recordNavigationEvent`'s write section,
  **before** `runInTransaction`, so the single bun:sqlite connection is never
  reserved on a guard failure.
- Kept to the issue's primary "lightweight runtime assertion" option rather than a
  speculative generic `runInTransactionWith` helper — there is exactly one
  transaction site today (YAGNI; grow the interface when a second arrives).
- Updated the `createForTesting` doc comment to note the precondition is now
  runtime-enforced.

## Testing

- `bun test test/features/navigation/NavigationGraphManager.test.ts test/db/navigationRepository.test.ts test/db/testCoverageRepository.test.ts` → **116 pass, 0 fail** (~1.2s), including the new guard + `resolveConnection` tests.
- `turbo run lint build` → clean.
- Full `bun test` → 5979 pass, 2 skip; the 6 failures are pre-existing and
  unrelated (jimp/wasm image + memory-leak tests fail identically on `main` with
  my changes stashed).

New tests (real in-memory `createTestDatabase()`, injected repos, no device/network, <100ms):
- **`resolveConnection`** (both repo test files) — returns the bound executor for a
  bound repo; two unbound repos resolve to the same singleton; `withExecutor` rebinds.
- **Guard** (`NavigationGraphManager.test.ts`, new `#2968` describe) — coverage repo
  on a *different* connection → `recordNavigationEvent` rejects with `ActionableError`
  whose message names the connection invariant; **zero** rows written on either
  connection, `currentScreen` stays null, telemetry never fires (guard precedes the
  transaction); and the happy path still persists the node when both share a
  connection.

- [x] Runtime assertion of connection identity between `this.repository` and
      `this.testCoverageRepository` before the transaction opens.
- [x] A foreign-bound coverage repo fails loudly (`ActionableError`) instead of
      silently splitting writes.
- [x] Production path unaffected — both repos default to the `getDatabase()`
      singleton, invariant holds by construction; existing transaction tests green.
- [x] Guard fires before the connection is reserved (no partial writes on failure).
- [x] Unit tests follow Interface+Fake conventions, <100ms, non-flaky, no real device.

### DIFF
--- a/src/db/navigationRepository.ts
+++ b/src/db/navigationRepository.ts
@@ -40,6 +40,17 @@ export class NavigationRepository {
   }
 
   /**
+   * The Kysely connection this repository resolves to: the executor bound via
+   * `withExecutor`/the constructor, or the shared `getDatabase()` singleton when
+   * unbound. Exposed so a caller enlisting a second repository in one transaction
+   * can assert both resolve to the SAME connection before opening it — two unbound
+   * repositories share the singleton, so identity (`===`) holds by construction.
+   */
+  resolveConnection(): Kysely<Database> {
+    return this.getDb();
+  }
+
+  /**
    * Return a repository instance whose every read and write runs on the supplied
    * executor (a Kysely transaction handle) instead of the shared singleton
    * connection.
--- a/src/db/testCoverageRepository.ts
+++ b/src/db/testCoverageRepository.ts
@@ -37,6 +37,18 @@ export class TestCoverageRepository {
   }
 
   /**
+   * The Kysely connection this repository resolves to: the executor bound via
+   * `withExecutor`/the constructor, or the shared `getDatabase()` singleton when
+   * unbound. Exposed so a transaction owner (NavigationGraphManager) can assert
+   * this coverage repo shares the navigation repo's connection before enlisting
+   * both in one transaction — a foreign-bound coverage repo would otherwise split
+   * writes and silently defeat the rollback guarantee.
+   */
+  resolveConnection(): Kysely<Database> {
+    return this.getDb();
+  }
+
+  /**
    * Return a repository instance whose every read and write runs on the supplied
    * executor (a Kysely transaction handle) instead of the shared singleton
    * connection, so coverage writes can enlist in a caller-owned transaction
--- a/src/features/navigation/NavigationGraphManager.ts
+++ b/src/features/navigation/NavigationGraphManager.ts
@@ -4,6 +4,7 @@ import { BackStackInfo } from "../../models";
 import { NavigationRepository } from "../../db/navigationRepository";
 import { TelemetryRecorder } from "../telemetry/TelemetryRecorder";
 import { TestCoverageRepository } from "../../db/testCoverageRepository";
+import { ActionableError } from "../../models/ActionableError";
 import type { NavigationEdge as DBNavigationEdge, NavigationNode as DBNavigationNode, TestCoverageSession } from "../../db/types";
 import {
   NavigationGraph,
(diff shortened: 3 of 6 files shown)


kaeawc:
## Multi-perspective review + fixes

Ran three independent review passes (DB-concurrency, API-design/YAGNI, and test-rigor). No correctness blockers in the guard itself — the `getDatabase()` reset TOCTOU was verified clean (two synchronous `resolveConnection()` calls, no interleave), and the inline-assertion-over-shared-helper choice is the right YAGNI call for a single transaction site.

**Addressed in c9150ac:**
- **Load-bearing test fix (test-rigor, major):** the original "no rows on either connection" test never started a session, so the coverage enlistment path was unreachable and the `coverageDb` assertion passed vacuously. Replaced with an **active-session** test that proves the guard converts an otherwise-opaque cross-connection FK violation into a clear `ActionableError`, plus a `test_node_coverage` assertion. Verified all four guard tests now fail if the guard is removed.
- **Stronger message assertion:** `/connection/i` → `/different database connections/i`.
- **Comment cleanup:** collapsed the duplicated atomicity-precondition block; guard now precedes the transaction-body comment.

**Filed as follow-up (out of scope for #2968's letter):**
- #2980 — `recordHierarchyNavigation` enlists the coverage repo **unguarded and non-atomically** (same invariant, one method over). All three reviewers flagged it.

Note: one reviewer initially read #2968 as asking to transactionalize `promoteSuggestion`/`recordHierarchyNavigation` — that's a misread; #2968's text asks verbatim for "a lightweight runtime assertion of connection identity … before opening the transaction," which is what this PR implements.

PR #2982, opened 2026-07-04
fix(nav): assert shared connection before recordNavigationEvent transaction (#2968)

### LINKED ISSUE #2968: fix(db): transactionalize promoteSuggestion + recordHierarchyNavigation graph writes (siblings of #2791)

kaeawc:
## What

Adds a lightweight runtime assertion that `NavigationGraphManager`'s navigation and coverage repositories resolve to the **same** underlying database connection **before** `recordNavigationEvent` opens its write transaction. If they don't, it throws an `ActionableError` immediately — before any row is written — instead of silently splitting writes across two connections and defeating the transaction's rollback guarantee.

Closes #2968.

## Why

Deferred hardening from the PR #2959 architecture review (Devin Osei). `recordNavigationEvent` opens a transaction on the navigation repo's connection and enlists the coverage repo onto that same `trx` via `withExecutor`. The "both repositories share the same connection" precondition was enforced only by convention (both default to the `getDatabase()` singleton) plus doc comments at the transaction site and `createForTesting`.

A coverage repo mistakenly constructed against a *different* Kysely handle would split its writes onto a separate connection, so a mid-sequence rollback would no longer be atomic across node + coverage rows. The invariant holds by construction in production (both default to the singleton), but nothing made a misconfiguration fail loudly. This closes that gap with a cheap reference-identity check.

## How

- **`NavigationRepository.getConnection()` / `TestCoverageRepository.getConnection()`** — a narrow accessor that returns the repository's resolved executor (the injected handle, or the shared `getDatabase()` singleton when unbound). Reference identity is the intended comparison.
- **`NavigationGraphManager.assertSharedConnection()`** — compares the two repos' resolved connections and throws an `ActionableError` on mismatch, naming the split-connection cause and the fix (construct both with the same handle).
- **`recordNavigationEvent`** calls `assertSharedConnection()` after computing the modal stack and **before** `runInTransaction`, so a misconfiguration fails before the transaction opens (no partial state to roll back).

This is the "runtime assertion" option from the issue rather than a shared multi-repo `runInTransaction` helper — the base `withExecutor` already rebinds the coverage repo onto `trx`, so the assertion is defense-in-depth that catches wiring mistakes and future subclass `withExecutor` overrides early, exactly as #2968 requested ("fails loudly instead of silently splitting writes").

## Testing

- `bun test test/features/navigation/NavigationGraphManager.test.ts` → **51 pass, 0 fail** (~0.3s), including 2 new tests.
- `bun test test/db/navigationRepository.test.ts test/db/testCoverageRepository.test.ts` → **57 pass, 0 fail**.
- `bunx turbo run lint build` → clean; `schemas/tool-definitions.json` regenerated, no drift.
- Full suite: 6 pre-existing failures only, all in image-processing modules (`PerceptualHasher`, `ScreenshotUtils`) + one memory-leak test — confirmed failing on a clean `git stash` of this branch, unrelated to this change.

New tests (real in-memory `createTestDatabase()`, <100ms, no device):
- **Shared connection happy path** — both repos bound to the same in-memory db → `recordNavigationEvent` records normally.
- **Foreign connection** — coverage repo bound to a *different* connection → `recordNavigationEvent` rejects with a `/connection/i` error, **zero** rows written on either connection, `currentScreen` still null, and telemetry not invoked (assertion fires before the transaction opens).

- [x] Runtime assertion of connection identity between `this.repository` and `this.testCoverageRepository` before opening the transaction.
- [x] A foreign-bound coverage repo fails loudly (`ActionableError`) instead of silently splitting writes.
- [x] No behavior change when both repos share a connection (production default + existing tests unaffected).
- [x] Assertion fires before any write — no partial state on failure.
- [x] lint + build + navigation/repository tests green.

## Notes
- Follow-up from #2791 / PR #2959. The sibling `recordHierarchyNavigation` path is intentionally out of scope: its writes are not wrapped in a transaction, so the cross-connection atomicity invariant doesn't apply there.

### DIFF
--- a/src/db/navigationRepository.ts
+++ b/src/db/navigationRepository.ts
@@ -40,6 +40,19 @@ export class NavigationRepository {
   }
 
   /**
+   * Resolve the underlying Kysely connection this repository will execute
+   * against (the injected handle, or the shared `getDatabase()` singleton when
+   * unbound). Exposed so a caller enlisting multiple repositories in one
+   * transaction can assert they share a connection before opening it — a
+   * coverage repo bound to a different connection would split writes and defeat
+   * the enclosing transaction's rollback guarantee. Reference identity is the
+   * intended comparison.
+   */
+  getConnection(): Kysely<Database> {
+    return this.getDb();
+  }
+
+  /**
    * Return a repository instance whose every read and write runs on the supplied
    * executor (a Kysely transaction handle) instead of the shared singleton
    * connection.
--- a/src/db/testCoverageRepository.ts
+++ b/src/db/testCoverageRepository.ts
@@ -37,6 +37,19 @@ export class TestCoverageRepository {
   }
 
   /**
+   * Resolve the underlying Kysely connection this repository will execute
+   * against (the injected handle, or the shared `getDatabase()` singleton when
+   * unbound). Exposed so a caller enlisting this repo alongside a navigation repo
+   * in one transaction can assert they share a connection before opening it — a
+   * coverage repo bound to a different connection would split writes and defeat
+   * the enclosing transaction's rollback guarantee. Reference identity is the
+   * intended comparison.
+   */
+  getConnection(): Kysely<Database> {
+    return this.getDb();
+  }
+
+  /**
    * Return a repository instance whose every read and write runs on the supplied
    * executor (a Kysely transaction handle) instead of the shared singleton
    * connection, so coverage writes can enlist in a caller-owned transaction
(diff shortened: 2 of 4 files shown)


kaeawc:
Closing as redundant: the connection-identity assertion this PR implements is already merged to `main` via PR #2977 (commit 0a1a77ee5), which added `resolveConnection()` + `assertSharedConnection()` — functionally identical. This PR was cut from a stale base and conflicts. The current #2968 scope (transactionalize `promoteSuggestion` + `recordHierarchyNavigation`) is being addressed in a fresh PR.

PR #2987, opened 2026-07-04
fix(db): transactionalize promoteSuggestion + recordHierarchyNavigation graph writes (#2968)

### LINKED ISSUE #2968: fix(db): transactionalize promoteSuggestion + recordHierarchyNavigation graph writes (siblings of #2791)

kaeawc:
## What

`NavigationGraphManager` had two remaining multi-write sequences with the same non-transactional partial-write bug class that #2791 fixed for `recordNavigationEvent`. This wraps both in a single `runInTransaction` / `withExecutor(trx)` transaction so each graph write is all-or-nothing.

Closes #2968.

### 1. `promoteSuggestion` (structural writes)
`getOrCreateNode` (creates the named node) + `repository.promoteSuggestion` (which does `getOrCreateFingerprint` + an `UPDATE` to link the suggestion) ran as separately-awaited statements. A throw during the link `UPDATE` — or a daemon interrupted mid-`await` — left an **orphaned node/fingerprint** with the suggestion still unpromoted.

### 2. `recordHierarchyNavigation` Case 1 (counters)
The existing-node branch did `updateNodeVisit` + coverage `recordNodeVisit` + `touchApp` across **two repos**, non-transactional. A mid-sequence throw left the node-visit increment applied while coverage counters were not (or vice-versa).

## Why

Per the architecture review (Devin Osei) on PR #2959, siblings of #2791. The daemon runs a **single** `bun:sqlite` connection guarded by an async mutex that serializes individual queries but **releases across every `await`** — so between two awaited statements the event loop yields and a throw leaves the graph partially applied with no rollback.

## How

Same structural-safety vehicle as `recordNavigationEvent` (#2791 / PR #2959):

- **`promoteSuggestion`** wraps `getOrCreateNode` + `promoteSuggestion` in `this.repository.runInTransaction(trx => …)`, binding the repo via `withExecutor(trx)` so every statement — including the nested `getOrCreateFingerprint` upsert — runs on the transaction. Only the navigation repo participates, so **no coverage enlistment / shared-connection assertion** is needed. `notifyGraphUpdated()` stays **outside** (post-commit).
- **`recordHierarchyNavigation` Case 1** wraps `updateNodeVisit` + coverage `recordNodeVisit` + `touchApp` in the same transaction, binding **both** repos via `withExecutor(trx)`. Because two repos are enlisted onto one connection, it calls the existing `assertSharedConnection()` (added by #2977) **before** opening the transaction. The in-memory `this.currentScreen` assignment and `notifyGraphUpdated()` move **outside** the transaction (post-commit), so on rollback `currentScreen` is unchanged and no notify fires.
- Branch-selector reads (`getNodeByFingerprint`) stay outside the write transaction, matching the issue's specified write boundary and the `recordNavigationEvent` precedent.

The other `recordHierarchyNavigation` cases (2/3/4) each issue a single write on one repo, so there is no multi-write sequence to make atomic.

## Testing

- `bun test test/features/navigation/NavigationGraphManager.test.ts` → **58 pass, 0 fail** (~0.46s), including 4 new tests.
- `bun test test/db/navigationRepository.test.ts test/db/testCoverageRepository.test.ts` → **63 pass, 0 fail**.
- `bunx turbo run lint build` → clean; `schemas/tool-definitions.json` regenerated, no drift.
- Full suite: 6 pre-existing failures only, all in image-processing modules (`PerceptualHasher`, `ScreenshotUtils`) + one memory-leak test — confirmed failing on a clean checkout, unrelated to this change.

New tests (real in-memory `createTestDatabase({ foreignKeys: true })`, <100ms, no device; TDD — the rollback tests fail on `main`'s non-transactional code and pass after the fix):
- **promoteSuggestion happy path** — node + fingerprint + suggestion-link all persisted; suggestion leaves the unpromoted list.
- **promoteSuggestion rollback** — an invalid `suggestionId` makes `repository.promoteSuggestion` throw *after* `getOrCreateNode` created the node; asserts **zero** new nodes/fingerprints persist (on `main` the orphan node survives → 1).
- **recordHierarchyNavigation Case 1 happy path** — visit_count +1, one coverage row, session total +1, `currentScreen` updated to the matched screen.
- **recordHierarchyNavigation Case 1 rollback** — a coverage repo that throws on `recordNodeVisit` (with an active session) → asserts `visit_count`, coverage rows, and session total all **unchanged**, `currentScreen` stays the pre-event screen, and no `notifyGraphUpdated` fires (on `main` the visit increment leaks → 2).

- [x] `promoteSuggestion` DB writes are atomic (node + fingerprint + suggestion-link all-or-nothing); rollback test with an injected mid-sequence throw asserts zero new rows.
- [x] `recordHierarchyNavigation` Case 1 writes are atomic; rollback leaves node visit/coverage counters unchanged.
- [x] No statement inside either transaction uses the singleton `getDb()` — all go through `withExecutor(trx)`.
- [x] Side effects (`notifyGraphUpdated`, in-memory field updates) stay outside the transaction.
- [x] Tests in `NavigationGraphManager.test.ts` using real in-memory `createTestDatabase()` + fakes, <100ms, non-flaky.

## Notes
- Sibling of #2791 (PR #2959); same `withExecutor` / `runInTransaction` vehicle. The shared-connection precondition is enforced by `assertSharedConnection()` (#2977) for the two-repo Case-1 transaction.
- Convention note (Devin §4): no new repo method here opens its own transaction while also being enlisted in an outer one, so the `getOrCreateUIElement` `isTransaction` dual-path guard is not needed for these call sites.

### DIFF
--- a/src/features/navigation/NavigationGraphManager.ts
+++ b/src/features/navigation/NavigationGraphManager.ts
@@ -523,7 +523,7 @@ export class NavigationGraphManager implements NavigationGraphService {
 
       // Persist the counter writes atomically: updateNodeVisit + coverage recordNodeVisit +
       // touchApp span two repos on the shared connection. Enlist both onto one `trx` via
-      // withExecutor so a mid-sequence throw (e.g. the coverage upsert) rolls back the node-visit
+      // withExecutor so a mid-sequence throw (e.g. the coverage recordNodeVisit) rolls back the node-visit
       // increment too, instead of leaving the graph partially applied. Assert the shared-connection
       // precondition first (the coverage repo is enlisted onto the navigation `trx`). In-memory
       // field updates and notifyGraphUpdated stay OUTSIDE so they never run on rollback and the
--- a/src/features/navigation/NavigationGraphManager.ts
+++ b/src/features/navigation/NavigationGraphManager.ts
@@ -519,18 +519,37 @@ export class NavigationGraphManager implements NavigationGraphService {
     // Case 1: Check if fingerprint is already correlated to a named node (scoped to this app)
     const existingNode = await this.repository.getNodeByFingerprint(this.currentAppId, fingerprintHash);
     if (existingNode) {
-      // Update the existing node's visit count and last_seen_at
-      await this.repository.updateNodeVisit(existingNode.id, timestamp);
+      const appId = this.currentAppId;
 
-      // Record node visit for test coverage if session is active
-      if (this.activeTestSession) {
-        await this.testCoverageRepository.recordNodeVisit(
-          this.activeTestSession.id,
-          existingNode.id,
-          timestamp
-        );
-      }
+      // Persist the counter writes atomically: updateNodeVisit + coverage recordNodeVisit +
+      // touchApp span two repos on the shared connection. Enlist both onto one `trx` via
+      // withExecutor so a mid-sequence throw (e.g. the coverage upsert) rolls back the node-visit
+      // increment too, instead of leaving the graph partially applied. Assert the shared-connection
+      // precondition first (the coverage repo is enlisted onto the navigation `trx`). In-memory
+      // field updates and notifyGraphUpdated stay OUTSIDE so they never run on rollback and the
+      // exclusive connection is not held across non-DB work.
+      this.assertSharedConnection();
 
+      await this.repository.runInTransaction(async trx => {
+        const repository = this.repository.withExecutor(trx);
+        const testCoverageRepository = this.testCoverageRepository.withExecutor(trx);
+
+        // Update the existing node's visit count and last_seen_at
+        await repository.updateNodeVisit(existingNode.id, timestamp);
+
+        // Record node visit for test coverage if session is active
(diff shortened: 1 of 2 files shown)


kaeawc:
Merged manually (squash). The `Node TypeScript Build and Test (windows-latest)` job failed on **pre-existing, unrelated** Windows-only file-DB lifecycle flakes — two runs failed on *different* tests, both green on ubuntu-latest + macos-latest, both in files this PR does not touch:
- run 1: `test/db/databaseLazyPath.test.ts` (migration-gating wait, 9.2s)
- run 2: `test/db/dbWriteBarrierResetOnClose.test.ts` (#2896 cold-start, 10.3s)

These match the documented Windows EBUSY/livelock class already targeted by #2916/#2923 and are tracked in #2992. This PR only changes `NavigationGraphManager` transaction logic (navigation tests passed on all platforms). Follow-up tech-debt note: #2994.

PR #2999, opened 2026-07-04
perf(db): add storage_events key-lookup seek index + thread previousValue (#2798)

### LINKED ISSUE #2798: perf(db): add covering index for recordStorageEvent previous-value lookup
## Summary
Every `recordStorageEvent` insert first runs a "find the previous value for this key" `SELECT ... ORDER BY timestamp DESC LIMIT 1`. The `storage_events` table only has single-column indexes on `timestamp` and `device_id`, so this query seeks on `device_id` but then does a **temp B-tree sort** over every matching row to satisfy the `ORDER BY`. Adding a composite covering index `(device_id, file_name, key, timestamp)` collapses it to a single index seek + read of the last row. This is a hot-path read on every telemetry storage event; the fix is a one-line migration plus a caller-side option to skip the read entirely.

Found via a DB concurrency/migration audit (2026-07-01), finding P4storidx, priority P2, verified against source.

## Background & history
- Introduced: commit `fdb94336` (2026-03-23), PR https://github.com/kaeawc/auto-mobile/pull/1571 — "feat: comprehensive telemetry dashboard enhancements". This commit added both the `storage_events` table/migration and the `recordStorageEvent` repository with its previous-value lookup.
- Schema provenance: the table and its two indexes were created in `src/db/migrations/2026_03_19_000_storage_events.ts` (`idx_storage_events_timestamp` on `timestamp`, `idx_storage_events_device` on `device_id`). Neither index covers the `(device_id, file_name, key)` filter plus the `timestamp` sort. The `previous_value` column itself was added later in `src/db/migrations/2026_03_19_002_storage_events_previous_value.ts`.
- Why it's shaped this way: the lookup was a convenience so the repository could auto-populate `previous_value` when the caller didn't already know it. At the time the table was tiny and the missing index wasn't noticeable; the table now retains up to `RETENTION_MAX_ROWS = 10_000` rows, so the sort scans a meaningful slice per insert.

## The problem
On every insert, when `previousValue` is not supplied, the repository issues a lookup ordered by `timestamp DESC`:

```ts
// src/db/storageEventRepository.ts:31-50
// Look up the previous value for this key if not already provided
let previousValue: string | null = input.previousValue ?? null;
if (previousValue === null && input.key !== null && input.deviceId !== null) {
  try {
    const q = d
      .selectFrom("storage_events")
      .select("value")
      .where("device_id", "=", input.deviceId)
      .where("file_name", "=", input.fileName)
      .where("key", "=", input.key)
      .orderBy("timestamp", "desc")
      .limit(1);
    const prev = await q.executeTakeFirst();
    if (prev) {
      previousValue = prev.value;
    }
  } catch {
    // best-effort lookup
  }
}
```

The only usable index for this predicate is `idx_storage_events_device` (single column `device_id`). SQLite seeks that index for `device_id=?`, then must apply the `file_name`/`key` filters and sort the surviving rows to find the `MAX(timestamp)` row. The audit verifier confirmed the query plan as:

```
SEARCH storage_events USING INDEX idx_storage_events_device (device_id=?)
USE TEMP B-TREE FOR ORDER BY
```

The `USE TEMP B-TREE FOR ORDER BY` is the cost: SQLite materializes and sorts the matching rows on every single insert instead of doing a single indexed seek to the newest row.

This is not a correctness/concurrency defect — it is purely a per-insert read amplification. (For completeness: the concurrency model here is a single `bun:sqlite` connection guarded by a hand-rolled async mutex in `src/db/bunSqliteDialect.ts` that serializes individual queries but releases across `await`; the lookup + insert are two separate awaited statements, so a concurrent event for the same key could read a stale previous value — but this finding is scoped to the missing index, not that race.)

Worse, no current caller supplies `previousValue`: `TelemetryRecorder.recordStorageEvent` (`src/features/telemetry/TelemetryRecorder.ts:227`) doesn't even expose the field on its input shape, so the lookup fires on **every** storage telemetry event from both the Android and iOS CtrlProxy clients.

## Impact
- Every storage-changed telemetry event pays for a temp-btree sort read before its insert. With the retention window at 10k rows and many events sharing a `device_id`, the sorted slice grows with device activity.
- No user sees a hard failure (the lookup is wrapped in a best-effort `try/catch`), so this is a latent throughput/latency cost on the telemetry ingest path, not a crash. Severity is genuinely narrow — it is a hot but background write path — hence P2.

## Proposed fix
Two independent, additive changes:

> **Terminology correction (per review):** the proposed `(device_id, file_name, key, timestamp)` index is **NOT actually a covering index** — the query does `.select("value")` and `value` is not in the index, so the plan is `SEARCH … USING INDEX idx_storage_events_key_lookup (device_id=? AND file_name=? AND key=?)` + a rowid lookup for `value` (one B-tree descent, no sort, but **not** index-only). Reframe this as a **seek-reduction index** (e.g. `idx_storage_events_key_recent`) rather than "covering," OR make it genuinely covering by appending `value` as the last column. Mira recommends **not** appending `value` (it can be an arbitrarily large TEXT blob — widening the index for a true index-only read is a bad trade); accept the rowid lookup and just fix the name/framing. The seek + no-sort win (killing the temp B-tree) is preserved either way.

> **Coordinate with #2788 (per review):** #2788 adds `idx_storage_events_device_timestamp (device_id, timestamp)` to `storage_events`; this issue adds the `(device_id, file_name, key, timestamp)` seek index. Both are legitimate (different query shapes) but must land in **one** `storage_events` migration file, sanity-checked that they aren't near-duplicates the planner will only ever pick one of. See #2788.

1. **Add the composite seek-reduction index** in a new forward-only migration mirroring `src/db/migrations/2026_03_19_000_storage_events.ts`:

```ts
// src/db/migrations/2026_07_01_000_storage_events_key_lookup.ts (new)
export async function up(db: Kysely<unknown>): Promise<void> {
  await db.schema
    .createIndex("idx_storage_events_key_lookup")
    .ifNotExists()
    .on("storage_events")
    .columns(["device_id", "file_name", "key", "timestamp"])
    .execute();
}

export async function down(db: Kysely<unknown>): Promise<void> {
  await db.schema.dropIndex("idx_storage_events_key_lookup").ifExists().execute();
}
```

With this index the predicate `device_id=? AND file_name=? AND key=?` is a prefix seek and the trailing `timestamp` column makes the `ORDER BY timestamp DESC LIMIT 1` a single indexed read — the temp B-tree disappears.

2. **Let callers skip the read on the hot path** by threading an optional `previousValue` through `TelemetryRecorder.recordStorageEvent` (add the field to its input object in `src/features/telemetry/TelemetryRecorder.ts:227` and pass it into `RecordStorageEventInput`). When the source already knows the prior value, the repository's `if (previousValue === null ...)` guard already short-circuits the lookup entirely.

3. **Fix the bare-swallow `catch` (per review — CLAUDE.md violation).** The best-effort lookup's `catch {}` at `storageEventRepository.ts:~32-34` swallows the error with no log, violating the error-handling convention (log-and-continue must log at `debug`/`warn` with a one-line reason). Add logging so a failing lookup leaves a trace. Also confirm `file_name` is always non-null at this call site: the `if` guards `input.key !== null && input.deviceId !== null` but has **no null-guard for `file_name`**; a NULL `file_name` bind makes `WHERE file_name = ?` match nothing (SQL NULL semantics, `= NULL` is never true), so the lookup returns nothing regardless of the index.

Must NOT: change the auto-populate behavior when `previousValue` is absent (existing callers rely on it), and must NOT alter query results — the seek-reduction index changes only the plan, not the returned row.

Precedent to mirror: forward-only migrations under `src/db/migrations/` with `.ifNotExists()` (see `2026_03_19_000_storage_events.ts`); optional-field threading like other `TelemetryRecorder` inputs.

- [ ] After migration, `EXPLAIN QUERY PLAN` for the lookup query shows a seek on `idx_storage_events_key_lookup` and **no** `USE TEMP B-TREE FOR ORDER BY`.
- [ ] Behavioral: for a given `(device_id, file_name, key)` with N prior events, `recordStorageEvent` with no `previousValue` sets `previous_value` to the `value` of the most-recent (max-timestamp) prior row — identical result to before the index.
- [ ] Behavioral: when a caller supplies `previousValue`, no lookup `SELECT` is issued and the supplied value is stored verbatim.
- [ ] Unit test in `test/db/storageEventRepository.test.ts` using interfaces + fakes + FakeTimer per CLAUDE.md, runs in <100ms, non-flaky, covering both the auto-lookup and the supplied-`previousValue` skip paths.
- [ ] `bun test` for the touched files green; `turbo run lint build test` green.
- [ ] (per review) Index is named/framed as a **seek-reduction** index (not "covering"), OR `value` is appended to make it genuinely covering — the `EXPLAIN QUERY PLAN` assertion runs against the **real predicate** (`device_id=? AND file_name=? AND key=?`) and asserts the new index is used and no `USE TEMP B-TREE FOR ORDER BY` appears.
- [ ] (per review) The best-effort lookup `catch` logs (no bare swallow — CLAUDE.md convention) and `file_name` is confirmed non-null at the call site (else the lookup silently returns nothing).
- [ ] (per review) The `storage_events` index migration is coordinated with #2788's `idx_storage_events_device_timestamp` in a single migration file, verified not to be a planner-redundant near-duplicate.
- [ ] Perf note: measure per-insert lookup time before/after on a `storage_events` table seeded near `RETENTION_MAX_ROWS` (10k rows) for one device; confirm the plan change and that stored `previous_value` values are unchanged.

- **Index topology is correct** — `(device_id, file_name, key, timestamp)` gives a three-column equality prefix seek plus trailing `timestamp` for `ORDER BY timestamp DESC LIMIT 1`; SQLite walks it backward to the newest row without the temp B-tree, killing the `USE TEMP B-TREE FOR ORDER BY` the audit measured.
- **NOT a covering index (substantive fix):** the query does `.select("value")` and `value` is not in the index, so the plan does a rowid table lookup for `value` — a single B-tree descent, no sort, but **not** index-only. Either rename/reframe to a *seek-reduction* index or append `value` (Mira advises against appending — `value` is an arbitrarily large TEXT blob; accept the rowid lookup and fix the name). Reflected in the Proposed fix and acceptance above.
- **`file_name` has no null-guard at the call site.** The `if` guards `key`/`deviceId` non-null but not `file_name`; a NULL `file_name` bind matches nothing (`= NULL` is never true), so the lookup returns nothing regardless of the index. Confirm `file_name` is always non-null here.
- **The best-effort `catch` is a bare swallow — CLAUDE.md violation.** Add logging so a failing lookup leaves a trace.
- **`EXPLAIN QUERY PLAN`-verify against the real predicate** (including the NULL semantics), asserting the after-plan shows the new index and no temp B-tree, so a future column-order edit can't silently regress it.
- **Caller-side `previousValue` skip is the bigger win and is race-relevant.** The lookup + insert are two separate awaited statements on the mutex-released connection, so even with the index a concurrent same-key event can read a stale previous value. Threading `previousValue` from a source that already knows the prior value (Android/iOS CtrlProxy) short-circuits the lookup and *also* closes that race — prefer wiring the field over relying on the index for correctness-adjacent callers.
- **Coordinate the migration with #2788** (single `storage_events` migration file; not a planner-redundant near-duplicate of `idx_storage_events_device_timestamp`).

## References
- `src/db/storageEventRepository.ts:31-50` (previous-value lookup)
- `src/db/storageEventRepository.ts:18` (`RETENTION_MAX_ROWS = 10_000`)
- `src/db/migrations/2026_03_19_000_storage_events.ts:20-32` (existing single-column indexes)
- `src/db/migrations/2026_03_19_002_storage_events_previous_value.ts` (`previous_value` column)
- `src/features/telemetry/TelemetryRecorder.ts:227-240` (caller; no `previousValue` field)
- `src/features/observe/android/AndroidCtrlProxyClient.ts:2483`, `src/features/observe/ios/IOSCtrlProxyClient.ts:975` (event sources)
- `test/db/storageEventRepository.test.ts` (existing test file)
- Finding P4storidx, audit date 2026-07-01

kaeawc:
## What

Adds one forward-only migration creating a composite **seek-reduction** index `idx_storage_events_key_lookup (device_id, file_name, key, timestamp)` on `storage_events`, threads an optional `previousValue` through `TelemetryRecorder.recordStorageEvent`, and fixes both bare-swallow `catch` blocks in the repository.

Closes #2798.

## Why

On every storage telemetry insert that doesn't already carry `previousValue`, `recordStorageEvent` runs a lookup:

```sql
SELECT value FROM storage_events
WHERE device_id=? AND file_name=? AND key=?
ORDER BY timestamp DESC LIMIT 1
```

**The issue's original premise is now stale — reframed honestly.** #2798 was filed against the pre-#2788 schema, where the only usable index was the standalone `idx_storage_events_device` and the plan showed `USE TEMP B-TREE FOR ORDER BY`. That temp B-tree is **already gone**: PR #2890 (#2788) added `idx_storage_events_device_timestamp (device_id, timestamp)` and PR #2904 (#2893) dropped the standalone device index, so the composite's `device_id` prefix now supplies the DESC order for free. Verified empirically — the current "before" plan is `SEARCH storage_events USING INDEX idx_storage_events_device_timestamp (device_id=?)` with **no** temp B-tree.

What remains is the **seek width**. With only `(device_id, timestamp)` the planner seeks on `device_id=?` and then scans that device's rows in timestamp order applying `file_name=?`/`key=?` row-by-row until the first match. For a device with many distinct keys, that scan dominates.

**Benchmark (10 000 rows, one device, 500 keys × 20 timestamps, 20 000 lookups, `:memory:`):**

| | plan | per lookup |
|---|---|---|
| before (`device_timestamp` composite only) | `SEARCH … idx_storage_events_device_timestamp (device_id=?)` | **~670 µs** |
| after (`key_lookup` index) | `SEARCH … idx_storage_events_key_lookup (device_id=? AND file_name=? AND key=?)` | **~0.96 µs** |

≈ **700× faster** per insert's lookup — the "before" is effectively an O(device-rows) scan, the "after" a direct 3-column prefix seek. One extra index on a 10k-capped table is clearly worth it.

## How

- **New migration** `src/db/migrations/2026_07_04_000_storage_events_key_lookup.ts` — auto-discovered by `FileMigrationProvider`; sorts after `2026_07_03_*` so `storage_events` exists on a full replay. `.ifNotExists()` up / `.ifExists()` down for #2785 destructive-recovery replay safety.
- **Seek-reduction, NOT covering** (per Mira Kessler's review on #2798): `value` is deliberately excluded — it's an arbitrarily large TEXT blob, so widening the index for an index-only read is a bad trade. Plan = prefix seek + one rowid lookup for `value`. Named `_key_lookup`, not `_covering`.
- **Distinct from #2788's composite, not planner-redundant**: `idx_storage_events_device_timestamp` serves the device-scoped getter; this index serves the three-equality lookup. The planner picks each for its own shape. (#2788 already merged, so the issue's "land them in one file" note is moot.)
- **Thread optional `previousValue`** through `TelemetryRecorder.recordStorageEvent` → `RecordStorageEventInput`. The guard is `input.previousValue !== undefined` (not `?? null`), so an **explicit `previousValue: null`** ("there is no prior value") is honored verbatim and does **not** trigger the lookup — only an omitted field does. This closes the mutex-released read-then-insert stale-read window for callers that supply the value. **Forward-looking hook**: no runner emits a prior value on the wire today (`AndroidCtrlProxyClient.ts`, `IosSdkEventIngestor.ts` both omit it), so the practical win in this PR is the index; wiring a source to emit `oldValue` is a filed follow-up.
- **Fix both bare-swallow `catch` blocks** (the lookup *and* the retention `cleanupIfNeeded` twin one function away) to `logger.warn` — CLAUDE.md error-handling convention, leaves a trace at the default level. Confirmed `file_name` is non-null at the call site.

## Not in scope (follow-ups filed)

- Having the Android/iOS CtrlProxy runners emit the prior value on the wire so the `previousValue` skip path activates (release-gated runner protocol change).
- Pre-existing: the iOS `recordStorageEvent` call passes an undeclared `operation` and omits `valueType`/`changeType` — a latent type error masked only by Bun skipping type-checking.

## Testing

- **New** `test/db/storageEventsKeyLookup.test.ts` — index columns via `pragma_index_info`; **`EXPLAIN QUERY PLAN`** against the real predicate (no index-reference + no temp B-tree before; 3-column prefix seek + no temp B-tree after); same most-recent value before/after; idempotent `up`, reversible `down`, partial-up→`down` no-op; #2788 composite + retention `timestamp` index untouched; full `runMigrations` replay leaves the index present.
- **`test/db/storageEventRepository.test.ts`** — supplied `previousValue` skips the lookup even with a competing prior row; **explicit `previousValue: null` honored verbatim** (not overwritten by a stale row); max-timestamp row chosen regardless of insertion order; lookup scoped by `file_name`/`key`.
- **`test/features/telemetry/TelemetryRecorder.test.ts`** — explicit `previousValue` threads through; absent field stays `undefined` (auto-lookup path).
- **`test/db/telemetryEventIndexInventory.test.ts`** (#2908 drift guard) updated for the new index.
- `bun test test/db/` + `test/features/telemetry/` green; changed files lint clean.

## Reviewed

Ran `/auto-mobile-code-review` plus three independent adversarial reviewers (SQLite-correctness, regression/completeness, architecture-direction). All three reproduced the EXPLAIN plan change and confirmed the index is correct, replay-safe, and distinct from #2788's composite. Their substantive findings are all addressed here: the explicit-`null` ambiguity (now guarded on `!== undefined`), the missing perf measurement (benchmark above), and the `cleanupIfNeeded` bare-catch twin (now logged). Remaining items are filed as follow-ups.

### DIFF
--- a/src/db/migrations/2026_07_04_000_storage_events_key_lookup.ts
+++ b/src/db/migrations/2026_07_04_000_storage_events_key_lookup.ts
@@ -1,0 +1,60 @@
+import type { Kysely } from "kysely";
+
+/**
+ * Seek-reduction index for `recordStorageEvent`'s per-insert previous-value
+ * lookup (#2798).
+ *
+ * On every storage telemetry insert that doesn't already carry `previousValue`,
+ * `storageEventRepository.recordStorageEvent` runs:
+ *
+ *   SELECT value FROM storage_events
+ *   WHERE device_id=? AND file_name=? AND key=?
+ *   ORDER BY timestamp DESC LIMIT 1
+ *
+ * #2798 was filed against the pre-#2788 schema, where the only usable index was
+ * the standalone `idx_storage_events_device` and the plan showed
+ * `USE TEMP B-TREE FOR ORDER BY`. That temp B-tree is ALREADY gone: PR #2890
+ * (#2788) added `idx_storage_events_device_timestamp (device_id, timestamp)` and
+ * PR #2904 (#2893) dropped the standalone device index, so the composite's
+ * `device_id` prefix now supplies the DESC order for free.
+ *
+ * What remains — and what this index fixes — is the SEEK width. With only the
+ * `(device_id, timestamp)` composite the planner seeks on `device_id=?` and then
+ * scans every row for that device in timestamp order, applying the `file_name=?`
+ * and `key=?` filters row-by-row until the first match. A device with many
+ * distinct keys makes that scan arbitrarily long. `(device_id, file_name, key,
+ * timestamp)` turns the three equality predicates into a single prefix seek that
+ * descends straight to the target key's rows, and the trailing `timestamp` column
+ * still yields the newest row without a sort (SQLite walks the ascending index
+ * backward for `ORDER BY timestamp DESC`).
+ *
+ * NOT a covering index (per Mira Kessler's review on #2798): the query selects
+ * `value`, which is deliberately NOT in the index — `value` is an arbitrarily
+ * large TEXT blob, and widening the index to make the read index-only is a bad
+ * trade. The plan is a prefix seek + one rowid lookup for `value`; the name is
+ * `_key_lookup` (a seek-reduction index), NOT `_covering`.
+ *
+ * Distinct from `idx_storage_events_device_timestamp` (#2788), not a
+ * planner-redundant near-duplicate: that composite serves the device-scoped
+ * getter (`WHERE device_id=? ORDER BY timestamp DESC`) and cannot serve this
+ * three-equality predicate as a prefix seek; this index cannot serve the
+ * device-only getter's ordering as efficiently. The planner picks each for its
+ * own shape.
+ *
+ * Additive and forward-only, `.ifNotExists()` / `.ifExists()` so the migration
+ * survives #2785's destructive-recovery replay (drop-all then replay every
+ * migration on an empty DB) in unusual orders. Sorts after
+ * `2026_07_03_*`, so `storage_events` exists when it runs during a full replay.
+ */
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await db.schema
+    .createIndex("idx_storage_events_key_lookup")
+    .ifNotExists()
+    .on("storage_events")
+    .columns(["device_id", "file_name", "key", "timestamp"])
+    .execute();
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await db.schema.dropIndex("idx_storage_events_key_lookup").ifExists().execute();
+}
--- a/src/db/storageEventRepository.ts
+++ b/src/db/storageEventRepository.ts
@@ -1,6 +1,7 @@
 import type { Kysely } from "kysely";
 import type { Database } from "./types";
 import { getDatabase } from "./database";
+import { logger } from "../utils/logger";
 
 export interface RecordStorageEventInput {
   deviceId: string | null;
@@ -28,9 +29,21 @@ export async function recordStorageEvent(
 ): Promise<void> {
   const d = getDb(db);
 
-  // Look up the previous value for this key if not already provided
+  // Look up the previous value for this key only when the caller did not supply
+  // one. `!== undefined` (not `?? null`) is deliberate: a caller that passes an
+  // explicit `previousValue: null` is asserting "there is no prior value" and
+  // must NOT trigger the lookup — otherwise a stale older row would silently
+  // override that intent. Omitting the field (undefined) is the only auto-lookup
+  // trigger.
+  //
+  // The lookup predicate (device_id=? AND file_name=? AND key=? ORDER BY
+  // timestamp DESC LIMIT 1) is served by idx_storage_events_key_lookup as a
+  // prefix seek (#2798). `file_name` is a non-null column on both the table and
+  // RecordStorageEventInput, so it never binds NULL here (a NULL bind would make
+  // `file_name = ?` match nothing regardless of the index).
   let previousValue: string | null = input.previousValue ?? null;
-  if (previousValue === null && input.key !== null && input.deviceId !== null) {
+  const previousValueSupplied = input.previousValue !== undefined;
+  if (!previousValueSupplied && input.key !== null && input.deviceId !== null) {
     try {
       const q = d
         .selectFrom("storage_events")
@@ -44,8 +57,10 @@ export async function recordStorageEvent(
       if (prev) {
         previousValue = prev.value;
       }
-    } catch {
-      // best-effort lookup
+    } catch (error) {
+      // Best-effort lookup: a failure here must not block the insert, but log so
+      // it leaves a trace (CLAUDE.md error-handling convention — no bare swallow).
+      logger.warn(`storage_events previous-value lookup failed: ${error}`, error);
     }
   }
 
@@ -124,8 +139,11 @@ async function cleanupIfNeeded(db?: Kysely<Database>): Promise<void> {
           .execute();
       }
     }
-  } catch {
-    // best-effort cleanup
+  } catch (error) {
+    // Best-effort retention cleanup: a failure must not surface to the telemetry
+    // write path, but log so it leaves a trace (CLAUDE.md error-handling
+    // convention — no bare swallow).
+    logger.warn(`storage_events retention cleanup failed: ${error}`, error);
   } finally {
     cleanupInProgress = false;
   }
--- a/src/features/telemetry/TelemetryRecorder.ts
+++ b/src/features/telemetry/TelemetryRecorder.ts
@@ -241,6 +241,9 @@ export class TelemetryRecorder {
     value: string | null;
     valueType: string | null;
     changeType: string;
+    // When the source already knows the prior value, thread it through to skip
+    // the repository's per-insert previous-value lookup (#2798).
+    previousValue?: string | null;
   }): Promise<void> {
     const { deviceId, sessionId } = this.snapshotContext();
     const input: RecordStorageEventInput = { deviceId, sessionId, ...event };
--- a/test/db/storageEventRepository.test.ts
+++ b/test/db/storageEventRepository.test.ts
@@ -46,6 +46,87 @@ describe("StorageEventRepository", () => {
     expect(events[0].previousValue).toBe("v1");
   });
 
+  test("recordStorageEvent skips the lookup when previousValue is supplied, even if a prior row exists", async () => {
+    // A prior event for the same (device, file, key) would be found by the
+    // auto-lookup. Supplying previousValue must short-circuit that lookup and
+    // store the supplied value verbatim rather than the prior row's value.
+    await recordStorageEvent({
+      deviceId: "d1", timestamp: 1000, applicationId: null, sessionId: null,
+      fileName: "prefs.xml", key: "theme", value: "light", valueType: null, changeType: "add",
+    }, db);
+    await recordStorageEvent({
+      deviceId: "d1", timestamp: 2000, applicationId: null, sessionId: null,
+      fileName: "prefs.xml", key: "theme", value: "dark", valueType: null, changeType: "modify",
+      previousValue: "caller-knows",
+    }, db);
+    const events = await getStorageEvents({ deviceId: "d1" }, db);
+    // Most-recent first; supplied value wins over the "light" the lookup would find.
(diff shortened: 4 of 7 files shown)


kaeawc:
Follow-ups filed for the two scoped-out items: #3000 (have the CtrlProxy runners emit the prior storage value so the `previousValue` skip path activates) and #3001 (pre-existing iOS `recordStorageEvent` type mismatch masked by Bun's no-typecheck).

PR #2989, opened 2026-07-04
test(ios): add executable guard for iOS pinch geometry (#2979)

kaeawc:
## What

Adds an **executable** guard for the iOS pinch endpoint geometry so it can no longer silently
diverge from the Android runner. Extracts the pure pinch trig out of `synthesizePinch` into a
platform-agnostic C function `ObjCExceptionCatcher_computePinchPoints` (no XCTest/UIKit
dependency), and adds `PinchGeometryTests.swift` — the direct analog of Android's
`PinchGeometryTest` — plus a **shared golden-vector parity table** mirrored byte-identically on
both platforms.

Closes #2979.

## Why

Follow-up to #2911 / PR #2931. `rotationDegrees` (default `0`) is how far the two-finger axis
rotates *during* the pinch: fingers start on the **horizontal** axis and finish on an axis
**rotated by `rotationDegrees`** (a combined pinch+rotate). This convention is deliberately shared
by both runners, but after #2931 only Android had an executable guard (`PinchGeometryTest` over the
extracted `computePinchPoints`). The iOS side was pinned by a **human-maintained code comment** in
`ObjCExceptionCatcher.m` — nothing failed if someone edited the iOS `dxStart/dyStart/dxEnd/dyEnd`
math (or Android's) without updating the other, which is exactly the silent-drift class #2911
worried about. This PR closes that asymmetry. Low severity, no default-path impact
(`rotationDegrees` defaults to 0; no production caller passes a non-zero value) — this is
hardening, not a bug fix.

## How

- **Extract (Option 1)** — the pinch endpoint trig (`endRadians`, `dxStart/dyStart`,
  `dxEnd/dyEnd`, and the `center ∓/± offset` point construction) moves from inside
  `ObjCExceptionCatcher_synthesizePinch` into a new pure `ObjCExceptionCatcher_computePinchPoints`
  returning an `ObjCPinchPoints` struct. Declared in the header **without** the `TARGET_OS_IOS`
  guard so it is callable from the macOS test host. `synthesizePinch` now **calls** this function,
  so the test guards the real synthesis math rather than a mirror of it. The iOS-specific
  minimum-distance floor (`distance > 1 ? distance : 1`) stays in the synthesis wrapper, so
  behavior is byte-identical.
- **Per-platform invariants** — `PinchGeometryTests.swift` asserts the same convention as
  Android: radii = distance/2, start axis always horizontal, end axis rotated by `rotationDegrees`,
  rotation sign respected, zero-rotation no-op.
[...]

Issue #2979, opened 2026-07-04
Add executable guard for iOS pinch geometry (parity with Android computePinchPoints)


kaeawc:
## Summary

Follow-up to #2911 / PR #2931. The shared pinch `rotationDegrees` convention (finger axis
starts horizontal, ends rotated by `rotationDegrees` — a combined pinch+rotate) is now pinned by
an **executable** regression guard on Android (`PinchGeometryTest` exercising the extracted pure
`computePinchPoints()`), but the **iOS** runner side is guarded only by a **human-maintained code
comment** in `ios/control-proxy/Sources/ObjCExceptionCatcher/ObjCExceptionCatcher.m`. Nothing fails
if the two implementations silently diverge.

## Context / history

- #2911 surfaced that `rotationDegrees` is applied only to the *end* finger positions (start axis
  horizontal, end axis rotated), so a non-zero rotation is a combined pinch+rotate rather than a
  pinch along a fixed rotated axis. The convention is shared by both runners, so there is no
  current cross-platform mismatch.
- PR #2931 chose **Option 1 (document)**: it documented the convention across the tool reference,
  MCP schema, TS models, and both runners, and — to prevent silent regression — extracted the
  Android endpoint math into a pure `internal fun computePinchPoints()` and added
  `PinchGeometryTest` (5 cases: radii, start-horizontal invariant, end-axis rotation, negative
  rotation sign, zero-rotation no-op).
- During PR #2931 review, the test-quality reviewer noted the iOS math lives in
  `ObjCExceptionCatcher.m` behind XCUITest private-API event synthesis (`XCPointerEventPath` /
  `XCSynthesizedEventRecord`) and is not unit-testable as-is; a code comment cross-referencing
  Android's `computePinchPoints` was deemed sufficient for that PR. This issue tracks closing that
  asymmetry properly.

## Why it matters

The only thing keeping iOS and Android pinch geometry in agreement is that a human keeps the two in
sync. If someone edits the iOS `dxStart/dyStart/dxEnd/dyEnd` math (or the Android side) without
updating the other, cross-platform pinch results silently diverge and no test catches it — exactly
the class of drift #2911 was worried about.

## Options

1. **Extract + unit-test the iOS geometry.** Pull the pure trig
   (`ObjCExceptionCatcher.m:~135-165`: `endRadians`, `dxStart=startRadius; dyStart=0`,
[...]

PR #3015, opened 2026-07-04
test(pinch): single-source golden vectors with cross-platform drift guard (#2997)

### LINKED ISSUE #2997: Harden pinch-geometry parity: single-source golden vectors instead of hand-duplicated tables (follow-up to #2979)
## Summary

Follow-up to #2979 / PR #2989. The Android and iOS pinch-geometry parity guard is currently a
**hand-duplicated golden-vector table**: identical `(centerX, centerY, distanceStart, distanceEnd,
rotationDegrees) -> expected 4-point set` literals live in **two** files and are kept in sync only
by a cross-reference comment. This issue tracks replacing that with a mechanism that cannot silently
drift.

## Context / history

PR #2989 (closing #2979) added an executable guard for the iOS pinch endpoint geometry to match
Android's `computePinchPoints`/`PinchGeometryTest`:

- iOS trig extracted into the pure `ObjCExceptionCatcher_computePinchPoints`
  (`ios/control-proxy/Sources/ObjCExceptionCatcher/ObjCExceptionCatcher.m`), consumed by
  `synthesizePinch`, and unit-tested in
  `ios/control-proxy/Tests/CtrlProxyTests/PinchGeometryTests.swift`.
- A shared golden-vector table asserts cross-platform parity:
  `testGoldenVectorsMatchAndroidParity` (Swift) and `golden vectors match iOS parity`
  (`android/control-proxy/src/test/kotlin/dev/jasonpearson/automobile/ctrlproxy/PinchGeometryTest.kt`).

Both tables carry a prominent `SHARED GOLDEN TABLE — keep byte-identical with ...` header.

## The gap

The duplicated table catches the **common** regression (someone edits one platform's endpoint math
without updating that platform's golden literals — its assertion fails loudly). It does **not**
catch a **coordinated one-sided edit**: if someone intentionally changes the convention on platform
A *and* updates only A's golden literals, platform B keeps its old math + old (now-divergent)
literals and both suites still pass green — the runners diverge in production undetected. A single
shared source of truth would close this.

During PR #2989 review this was flagged by two of three independent reviewers as a real (but
well-flagged, reasonably-chosen) maintainability hazard. A single shared fixture was deferred
because it is not cleanly loadable across the iOS test target's two build systems (SPM + the
xcodegen'd `.xcodeproj`) *and* Gradle test resources, and because a naively-shared fixture read by
both sides would still not catch the coordinated one-sided edit.

## Options

1. **Committed golden JSON fixture + per-platform loaders.** One `pinch-golden-vectors.json` at a
   shared path; SPM test target declares it via `resources: [.copy(...)]` (and the xcodegen project
   mirrors it), Gradle test reads it from test resources. Removes literal duplication; still needs
   both platforms to read the same file (doesn't alone catch coordinated edits — pair with option 3).
2. **Generator.** A small script emits both tables (or the JSON) from one source; CI fails if the
   committed artifacts are stale. Heavier for a 4-6 row table.
3. **Drift guard.** A lightweight CI/test check that the Swift and Kotlin golden literals are
   byte-identical (parse both, compare). Cheap; directly targets the "tables silently diverge"
   failure mode and composes with option 1.

## Acceptance criteria

- [ ] The Android and iOS pinch golden vectors derive from, or are verified against, a **single
      source** so the two cannot silently diverge.
- [ ] A coordinated one-sided convention edit (change math + golden on one platform only) fails a
      test or CI check.
- [ ] Solution stays fast/offline and fits both build systems (SPM + xcodegen `.xcodeproj`, Gradle).

Low severity, hardening only — `rotationDegrees` defaults to 0 and no production caller passes a
non-zero value (verified in #2931/#2979).

kaeawc:
## What

Replaces the **hand-duplicated** pinch golden-vector tables (identical literals in
`ios/control-proxy/Tests/CtrlProxyTests/PinchGeometryTests.swift` and
`android/control-proxy/.../PinchGeometryTest.kt`, kept in sync only by a cross-reference comment)
with a **single source of truth** — `test/fixtures/pinch-golden-vectors.json` — plus a
build-system-agnostic drift guard (`test/parity/pinchGoldenVectorParity.test.ts`) that verifies
**both** platforms' committed literals against the JSON.

Closes #2997.

## Why

Follow-up to #2979 / PR #2989. The duplicated table caught the **common** regression (edit one
platform's math without updating its own golden literals → that platform's assertion fails loudly).
It did **not** catch a **coordinated one-sided edit**: change the convention on platform A *and*
update only A's golden literals, and platform B keeps its old math + old (now-divergent) literals
while **both suites stay green** — the runners diverge in production undetected. Two of three
reviewers on #2989 flagged this as a real (if well-flagged) maintainability hazard. Low severity,
hardening only (`rotationDegrees` defaults to 0; no production caller passes a non-zero value —
verified in #2931/#2979).

## How

The issue's recommended **Option 1 (canonical JSON) + Option 3 (drift guard)**, implemented as a TS
test so it sidesteps the "can't cleanly load a shared fixture across SPM + xcodegen `.xcodeproj` +
Gradle" problem the issue called out:

- **`test/fixtures/pinch-golden-vectors.json`** — the single source: 5 rows of inputs
  (`centerX, centerY, distanceStart, distanceEnd, rotationDegrees`) + expected unordered 4-point
  set. Values are exactly the current Swift/Kotlin literals.
- **`test/parity/pinchGoldenVectorParity.test.ts` + `pinchGoldenVectors.ts`** — parse each
  platform's inline golden table out of source (balanced-delimiter extraction, comment-stripped
  number sequence, chunked into rows) and `diffGoldenTables` each against the canonical JSON.
  Inputs compared in order; the four endpoints compared as an **order-independent set** (mirrors the
  runtime `sortedPoints`/`sortPoints`), so the runners' opposite finger-labeling isn't false drift.
  A coordinated one-sided edit now leaves the untouched platform's literals disagreeing with the
  JSON → guard fails **regardless of which side changed**. Negative tests model both a point-set
  edit and an input-tuple edit.
- **Platform headers** — the `SHARED GOLDEN TABLE` KDoc/doc comments in both files now point at the
  JSON + guard and say "edit all three together"; the inline tables stay (fast, no runtime resource
  IO) but are no longer the source of truth.

Verify-against, not generate-from — satisfies AC1 ("derive from, **or are verified against**, a
single source"). The guard reads the test files as plain text, so it needs no SPM/xcodegen/Gradle
wiring and runs offline under `bun test` in <30ms (AC3). The per-platform runtime golden tests still
enforce *math↔literals*; this guard adds *literals↔JSON across both platforms* — together they close
the silent-divergence gap (AC2).

## Testing

- **New** `test/parity/pinchGoldenVectorParity.test.ts` — 8 cases, `bun test`, ~20ms.
- End-to-end teeth check: temporarily mutating a real Swift golden literal (`60`→`61`) on disk fails
  `AC1: iOS…` + the transitive-agreement test; reverted cleanly.
- Full suite green: **6032 pass, 0 fail, 2 skip** (`bun test`, 468 files).
- `eslint` clean on new files; `ktfmt --google-style` clean on the Kotlin file; `swiftformat --lint`
  clean on the Swift file (comment-only platform edits); `bun build.ts` OK.

### Known limitation

The guard verifies literals agree with the JSON; it does not execute platform math or assert the
per-platform runtime golden test functions still exist — those tests are the precondition for the
*math↔literals* half of the chain. Guarding their existence too is possible but low value; noted as
a potential follow-up.

### DIFF
--- a/android/control-proxy/src/test/kotlin/dev/jasonpearson/automobile/ctrlproxy/PinchGeometryTest.kt
+++ b/android/control-proxy/src/test/kotlin/dev/jasonpearson/automobile/ctrlproxy/PinchGeometryTest.kt
@@ -115,13 +115,16 @@ class PinchGeometryTest {
   }
 
   /**
-   * SHARED GOLDEN TABLE — keep byte-identical with the iOS runner's `PinchGeometryTests.swift`
-   * (`testGoldenVectorsMatchAndroidParity`). Each row is an input tuple and its expected
-   * *unordered* set of four endpoints. The comparison is order-independent because the two runners
-   * label which finger is "first" oppositely (Android builds center±offset, iOS center∓offset
-   * first) while producing the same two touch points. If either platform's endpoint math changes,
-   * its golden assertion fails loudly here or in the Swift mirror — closing the silent-divergence
-   * gap from issues #2911 / #2979.
+   * SHARED GOLDEN TABLE — the single source of truth is `test/fixtures/pinch-golden-vectors.json`;
+   * `test/parity/pinchGoldenVectorParity.test.ts` verifies this table and the iOS mirror
+   * (`PinchGeometryTests.swift`'s `testGoldenVectorsMatchAndroidParity`) against it. Each row is an
+   * input tuple and its expected *unordered* set of four endpoints. The comparison is
+   * order-independent because the two runners label which finger is "first" oppositely (Android
+   * builds center±offset, iOS center∓offset first) while producing the same two touch points. If
+   * either platform's endpoint math changes, its golden assertion fails loudly here or in the Swift
+   * mirror. If the two tables silently diverge — including a coordinated one-sided convention edit
+   * (change math + golden on one platform only) — the parity guard fails. Edit the JSON and both
+   * platform tables together. See issues #2911 / #2979 / #2997.
    */
   @Test
   fun `golden vectors match iOS parity`() {
--- a/ios/control-proxy/Tests/CtrlProxyTests/PinchGeometryTests.swift
+++ b/ios/control-proxy/Tests/CtrlProxyTests/PinchGeometryTests.swift
@@ -93,13 +93,16 @@ final class PinchGeometryTests: XCTestCase {
 
     // MARK: - Cross-platform golden parity (issue #2979, Option 2)
 
-    /// SHARED GOLDEN TABLE — keep byte-identical with `PinchGeometryTest.kt`
-    /// (`goldenVectorsMatchIosParity`). Each row is an input tuple and its expected *unordered*
-    /// set of four endpoints (start1, start2, end1, end2). We compare as an order-independent set
-    /// because the two runners label which finger is "first" oppositely (iOS builds center∓offset
-    /// first, Android center±offset first) while producing the same two touch points. If either
-    /// platform's endpoint math changes, its golden assertion fails loudly here or in the Kotlin
-    /// mirror — closing the silent-divergence gap from #2911.
+    /// SHARED GOLDEN TABLE — the single source of truth is `test/fixtures/pinch-golden-vectors.json`;
+    /// `test/parity/pinchGoldenVectorParity.test.ts` verifies this table and the Android mirror
+    /// (`PinchGeometryTest.kt`'s `golden vectors match iOS parity`) against it. Each row is an input
+    /// tuple and its expected *unordered* set of four endpoints (start1, start2, end1, end2). We
+    /// compare as an order-independent set because the two runners label which finger is "first"
+    /// oppositely (iOS builds center∓offset first, Android center±offset first) while producing the
+    /// same two touch points. If either platform's endpoint math changes, its golden assertion fails
+    /// loudly here or in the Kotlin mirror. If the two tables silently diverge — including a
+    /// coordinated one-sided convention edit (change math + golden on one platform only) — the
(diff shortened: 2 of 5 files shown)


kaeawc:
### Multi-perspective review + hardening (self-review round)

Ran three independent review passes (adversarial-correctness, build/CI-pragmatist, architecture/direction). Consensus: **no blockers**, guard's core promise holds — no reviewer could construct a false-negative or fail-open; every mis-parse fails closed via the `NUMBERS_PER_ROW` alignment check.

Addressed feedback in `36f751a`:
- **Strip comments before the balanced-delimiter scan** (flagged by two reviewers) — a stray `[`/`]`/`(`/`)` inside a comment can no longer mis-terminate `extractBalanced`.
- **Assert the per-platform runtime golden loops still exist** (architecture reviewer's substantive point) — this guard proves *literals↔JSON*; the *math↔literals* half lives in the platform test files' runtime loops. New assertion trips if `computePinchPoints`/`ObjCExceptionCatcher_computePinchPoints` + `assertEquals`/`XCTAssertEqual` are deleted, closing the symmetric hole. Verified end-to-end (removing the Swift `XCTAssertEqual` loop fails the guard).
- **Honest docs** on exotic numeric forms (exponent/underscore/hex fail closed, not silently mis-parsed).

Left as noted non-blocking follow-up: nothing is *generated from* the JSON (verify-against, which AC1 explicitly permits); string-literals inside the table region aren't stripped (no such literals exist today — messages live outside the balanced region).

Guard: 9 tests, ~24ms. Full local suite: 6032 pass / 0 fail.

Issue #2965, opened 2026-07-03
iOS runner: opaque decode-boundary error for out-of-range numeric literals (1e309) — surface an actionable message


kaeawc:
## Summary

When a client sends a malformed/out-of-range numeric literal (e.g. `1e309`, `1e400`) in **any** iOS CtrlProxy command, the runner surfaces an **opaque** error over the wire — `"The data couldn't be read because it isn't in the correct format."` — rather than an actionable message. This is the decode-boundary counterpart to #2928 (which hardened the *handler* boundary), surfaced during PR #2957's review.

## Detail

- Apple's `JSONDecoder` rejects an overflow numeric literal during **top-level pre-parse** with `DecodingError.dataCorrupted`, **empty `codingPath`**, underlying *"Number 1e309 is not representable in Swift"* — *before* any `Decodable`/`init(from:)` runs. (Verified on Swift 6.2 / macOS 15; see PR #2957 `testOverflowCoordinateLiteralIsRejectedAtDecode`.)
- `WebSocketServer` catches the decode failure and calls `sendErrorResponse` → `buildErrorResponseData`, which emits `WebSocketResponse.error(type: "error", error: error.localizedDescription, …)` (`ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift:~275-307`). For `dataCorrupted`, `localizedDescription` is the generic *"The data couldn't be read…"* — no field, no hint that a number was out of range.
- Because the failure is at pre-parse with an **empty codingPath**, a custom `init(from:)` **cannot** attribute it to a field. So a per-field message (like #2928's `"Invalid value 'inf' for parameter 'x'"`) is **infeasible** at the Decodable layer for this class of input — the only lever is the generic decode-error surfacing in `buildErrorResponseData`.

## Why #2928 / PR #2957 did not cover this

#2928's premise assumed `1e309` decodes to `Double.infinity` into the widened coordinate field, where the new `isFinite` handler guard would catch it. Empirically it does not (pre-parse rejection), so the handler guard never runs for `1e309` on Apple Foundation — the value is already rejected, just with an opaque message. #2928 correctly hardened the handler boundary (defense-in-depth for the non-wire/computed path); this issue is the separate, broader concern of decode-boundary error *legibility*, which spans all commands, not just gestures.

## Proposed fix (options)

[...]

PR #3019, opened 2026-07-04
fix(android): emit structured error envelope on CtrlProxy decode/handler failures (#2985)

### LINKED ISSUE #2985: Android CtrlProxy runner swallows decode/handler failures — no error envelope sent to client
## Summary

The Android CtrlProxy runner **swallows inbound decode failures and handler exceptions without sending any error response**, so a client that sends a malformed command (or triggers a handler throw) gets **no reply at all** — the TS daemon's awaiter hangs until it times out, instead of receiving an actionable error envelope.

This is the Android counterpart to #2965 (iOS opaque-but-present decode error). It's arguably worse: iOS at least returns a structured `type:"error"` envelope (now actionable after #2965 / PR #2983); Android returns nothing.

## Detail

`android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/WebSocketServer.kt`:

```kotlin
// handleClientMessage(...)
val request =
  try {
    protocolJson.decodeFromString<ProtocolRequest>(message)
  } catch (e: Exception) {
    Log.w(TAG, "Failed to parse client message: $message", e)
    return                         // <-- no error envelope sent to the client
  }

...
try {
  val response = handler.handleMessage(request)
  if (response != null) broadcast(response)
} catch (e: Exception) {
  Log.e(TAG, "Error handling message via handler", e)   // <-- also no error envelope
}
```
(WebSocketServer.kt:301-322 as of PR #2983's base)

Both the decode-failure `catch` (`:304-306`) and the handler-exception `catch` (`:320-321`) log and return without emitting a response. A client awaiting a reply for that `requestId` never gets one.

## Impact

- A malformed/unknown command → daemon awaiter hangs to timeout rather than surfacing a fast, actionable failure.
- Violates the spirit of the repo's error-handling convention (a system/protocol boundary failure should surface a structured error to the client, not be silently swallowed).

## Proposed fix

Mirror the iOS `WebSocketServer.buildErrorResponseData` / `sendErrorResponse` path: on decode failure and on handler exception, build and `broadcast` a structured error response for the (best-effort extracted) `requestId`, e.g. `{ "type":"error", "success":false, "requestId":<id|null>, "error":<actionable message>, "timestamp":<ms> }`. Extract `requestId` from the raw JSON best-effort (as iOS's `extractRequestId` does) so the client can correlate. Consider the same out-of-range / malformed-JSON legibility mapping #2965 introduced on iOS, for symmetry.

## Acceptance criteria

- [ ] A malformed inbound command produces a structured `type:"error"` response (not a silent `return`), correlated by `requestId` when present.
- [ ] A handler exception produces a structured error response instead of only a log line.
- [ ] Unit test (Robolectric/JUnit, no device) asserting a malformed payload yields an error frame with `success:false` and a non-empty `error`.

## Context

Found during the multi-agent review of PR #2983 (issue #2965) — the "wire-protocol / TS-consumer" reviewer confirmed the swallow at `WebSocketServer.kt:304-306` and `:320-321`. See PR #2983 for the iOS decode-boundary legibility fix this parallels.

kaeawc:
## What

Make the **Android** CtrlProxy runner emit a structured `type:"error"` envelope when an inbound command fails to decode or its handler throws — instead of logging and silently `return`ing, which left the daemon's awaiter hanging until timeout.

Closes #2985. This is the Android counterpart to the iOS decode/handler-boundary work (#2965 / PR #2983). Android was arguably worse: iOS at least returned *some* structured error; Android returned **nothing**.

Affected files:
- `android/protocol/.../WebSocketResponse.kt` — new `ErrorResponse` sealed subtype (`type:"error"`).
- `android/control-proxy/.../WebSocketServer.kt` — emit on decode failure + handler exception; `extractRequestId` / `describeDecodeFailure` helpers.
- `src/features/observe/android/AndroidCtrlProxyClient.ts` — consume the `type:"error"` frame and fail the correlated request fast.

## Why

`WebSocketServer.handleClientMessage` had two swallow points (as flagged in #2985):

```kotlin
val request = try { protocolJson.decodeFromString<ProtocolRequest>(message) }
              catch (e: Exception) { Log.w(...); return }   // no reply
...
try { ... } catch (e: Exception) { Log.e(...) }             // no reply
```

A client awaiting a reply for that `requestId` never got one, so a malformed/unknown command turned into a slow timeout rather than a fast, actionable failure — and it violated the repo's error-handling convention (a protocol-boundary failure should surface a structured error to the client).

## How

**Runner (Kotlin):**
- Added `ErrorResponse(timestamp, requestId?, success=false, error)` (`@SerialName("error")`) to the `WebSocketResponse` sealed hierarchy — serializes to `{"type":"error","success":false,"requestId":…,"error":…,"timestamp":…}`, mirroring the iOS envelope.
- **Decode failure:** best-effort `extractRequestId(raw)` (lenient parse of the raw payload, mirroring iOS `extractRequestId`) + `describeDecodeFailure(raw, e)` → broadcast an `ErrorResponse`. `describeDecodeFailure` maps an unknown/unregistered command type to `"Unknown command type: <type>"` (symmetric to the iOS `CommandError.unknownCommand` contract) and everything else to `"Malformed request: <cause>"` — parallel to the #2965 legibility mapping for out-of-range / malformed-JSON input.
- **Handler exception:** broadcast an `ErrorResponse` correlated by the decoded `request.requestId`. `CancellationException` is re-thrown first so cooperative cancellation still unwinds the read loop cleanly (it is never converted into an error frame).

**Consumer (TS):** the runner frame is only useful if the daemon acts on it — `AndroidCtrlProxyClient.handleWebSocketMessage` now has a `type:"error"` branch that calls `RequestManager.resolveError(requestId, error)`, so the awaiting request fails fast instead of hanging. A null/uncorrelated `requestId` is a safe no-op (mirrors the iOS `decodeCtrlProxyMessage` default-case reject).

## Testing

All no-device (Robolectric/JUnit + bun), following repo conventions.

- `WebSocketResponseTest.kt` — `ErrorResponse` serializes with `type:"error"` / `success:false` / requestId (and `null` when absent) and round-trips through the sealed serializer.
- `WebSocketServerTest.kt` — pure unit tests for `extractRequestId` (present / absent / non-string / unparseable → null) and `describeDecodeFailure` (unknown-type names the type; malformed is non-empty).
- `WebSocketServerIntegrationTest.kt` — end-to-end over a real WebSocket:
  - malformed/unknown command → `type:"error"`, `requestId` echoed, `success:false`, and the wire error **names the offending type** (proves the legibility mapping fires on the *real* kotlinx `SerializationException`, not just a synthetic one).
  - unparseable payload → error frame with **null** requestId (best-effort correlation).
  - handler throw → structured error frame correlated by requestId.
- `CtrlProxyClient.test.ts` — a `type:"error"` frame correlated by requestId fails the awaiting request fast (was: hang to timeout); a non-matching/uncorrelated error frame does not disturb other in-flight requests.

Local runs: `:protocol:test`, `:control-proxy:testDebugUnitTest` (JDK 21), and the bun CtrlProxy/RequestManager suites all green. (One pre-existing, unrelated `preserve SDK screen names` nav-graph test fails on unmodified main too — not touched here.)

> **Delivery note:** `android/control-proxy` ships as a pinned release APK the daemon downloads, so this reaches default installs only after the runner release is re-cut — same delivery gate as other runner-source changes.

### DIFF
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/WebSocketServer.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/WebSocketServer.kt
@@ -2,6 +2,7 @@ package dev.jasonpearson.automobile.ctrlproxy
 
 import android.util.Log
 import dev.jasonpearson.automobile.ctrlproxy.perf.PerfProvider
+import dev.jasonpearson.automobile.protocol.ErrorResponse
 import dev.jasonpearson.automobile.protocol.SdkEvent
 import dev.jasonpearson.automobile.protocol.WebSocketMessageHandler
 import dev.jasonpearson.automobile.protocol.WebSocketRequest as ProtocolRequest
@@ -23,6 +24,8 @@ import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.asSharedFlow
 import kotlinx.serialization.json.Json
 import kotlinx.serialization.json.JsonElement
+import kotlinx.serialization.json.JsonObject
+import kotlinx.serialization.json.JsonPrimitive
 
 /**
  * WebSocket server that streams view hierarchy updates to connected clients and dispatches inbound
@@ -41,6 +44,56 @@ class WebSocketServer(
 ) {
   companion object {
     private const val TAG = "WebSocketServer"
+
+    /** Lenient parser for best-effort field extraction from a raw (possibly malformed) payload. */
+    private val lenientJson = Json {
+      ignoreUnknownKeys = true
+      isLenient = true
+    }
+
+    /**
+     * Best-effort extraction of a top-level string [field] from a raw JSON payload. Returns null
+     * when the payload is unparseable, the field is absent, or the field is not a JSON string.
+     */
+    private fun extractStringField(raw: String, field: String): String? =
+      try {
+        (lenientJson.parseToJsonElement(raw) as? JsonObject)?.get(field)?.let { element ->
+          (element as? JsonPrimitive)?.takeIf { it.isString }?.content
+        }
+      } catch (e: Exception) {
+        // Expected for genuinely malformed payloads; correlation is best-effort by design.
+        null
+      }
+
+    /**
+     * Best-effort extraction of `requestId` from a raw JSON payload for error correlation,
+     * mirroring the iOS runner's `extractRequestId`. Returns null when it can't be determined.
+     * See #2985.
+     */
+    internal fun extractRequestId(raw: String): String? = extractStringField(raw, "requestId")
+
+    /**
+     * Maps an inbound-decode failure into an actionable, legible wire message. An unknown/
+     * unregistered command type surfaces "Unknown command type: <type>" (symmetric to the iOS
+     * `CommandError.unknownCommand` contract); everything else surfaces "Malformed request:
+     * <cause>" so an out-of-range numeric literal or a JSON syntax error is actionable rather than
+     * opaque. See #2985 (parallels the iOS #2965 legibility mapping).
+     */
+    internal fun describeDecodeFailure(raw: String, throwable: Throwable): String {
+      val cause =
+        throwable.message?.takeIf { it.isNotBlank() }
+          ?: throwable::class.simpleName
+          ?: "unknown error"
+      val looksLikeUnknownType =
+        cause.contains("polymorphic", ignoreCase = true) ||
+          cause.contains("class discriminator", ignoreCase = true)
+      if (looksLikeUnknownType) {
+        extractStringField(raw, "type")?.let { type ->
+          return "Unknown command type: $type"
+        }
+      }
+      return "Malformed request: $cause"
+    }
   }
 
   private var server: EmbeddedServer<*, *>? = null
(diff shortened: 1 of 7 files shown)


kaeawc:
### Multi-agent review summary

Ran a self code-review plus three independent review passes (Kotlin/coroutines skeptic, TS wire-protocol/consumer, architecture/direction). **All three: no confirmed bugs / regressions — ship-ready.** Key verified points:

- Decode-catch not rethrowing `CancellationException` is correct (synchronous non-suspending decode can't observe cooperative cancellation); the handler catch rethrows it, which is necessary.
- `broadcast`-to-all matches the existing Android response delivery convention; the TS consumer correlates by `requestId` and `resolveError` no-ops on non-owners — no info leak, no cross-request disturbance (covered by a new test).
- Adding `ErrorResponse` to the sealed hierarchy breaks no exhaustive `when` (only `WebSocketRequest` has one, in `CtrlProxyMessageHandler`); adding `"error"` to the TS union breaks no exhaustive switch.
- The malformed-command integration test feeds the **real** kotlinx `SerializationException` and asserts the wire error names the offending type — proving the legibility mapping fires on the real decode path, not just a synthetic exception.

Two non-blocking gaps were deferred to follow-ups (both explicitly classified as follow-up, not blocker, by the reviewers):
- #3022 — iOS parity: apply an Android-appropriate `rewriteUnknownCommandError`, add an out-of-range-literal legibility case, consider targeted (vs broadcast) send.
- #3023 — async action failures inside launched coroutines still hang the awaiter (this PR covers only the two synchronous decode/handler catches).

Issue #3026, opened 2026-07-04
Dormant output-reduction flags: --actions-diff-observe and --actions-no-observe are plumbed but have no consumer


kaeawc:
## Summary

Two output-reduction feature flags are **fully plumbed but consumed by nothing** — the same dormant state #2899/PR #2933 fixed for `toolResultsNoStructuredContent` and #2951/PR #2976 fixed for `observeResultCompact`:

- **`--actions-diff-observe`** (env `AUTOMOBILE_ACTIONS_DIFF_OBSERVE`) — "Return only the diff of the post-action observation instead of the full observation."
- **`--actions-no-observe`** (env `AUTOMOBILE_ACTIONS_NO_OBSERVE`) — "Skip returning the post-action observation entirely to reduce output size."

## Evidence (dormant)

Both flags flow CLI/env → `src/utils/outputReductionFlags.ts` → `ServerConfig` → daemon startup (`src/daemon/daemon.ts:209-213`, `src/features/featureFlags/FeatureFlagApplier.ts:66-70`), but **no production code reads the getters**:

```
$ git grep -n "isActionsDiffObserveEnabled" src/
src/utils/ServerConfig.ts:243:  isActionsDiffObserveEnabled(): boolean {   # definition only

$ git grep -n "isActionsNoObserveEnabled" src/
src/utils/ServerConfig.ts:251:  isActionsNoObserveEnabled(): boolean {     # definition only
```

Enabling either flag today has **no effect**.

## Tasks

1. **`--actions-no-observe`** (simpler): after an action's post-action observation is produced, drop the `observation` field from the served payload when the flag is on. Likely a small addition at the same serialization chokepoint (`src/server/finalizeToolResponse.ts`) or in the action tool wrappers. Composes trivially with `--observe-result-compact` (nothing left to flatten).

[...]

Issue #2990, opened 2026-07-04
observeResultCompact follow-ups: actionsDiffObserve composition test + wire-schema tuple doc


kaeawc:
## Summary

Two smaller items were named in #2978 but **not** implemented — PR #2976 (merged) landed only the "compact everywhere" bounds work and closed #2978, so these riders would otherwise be lost. Filing here to preserve them.

Background: PR #2976 wired `--observe-result-compact` (env `AUTOMOBILE_OBSERVE_RESULT_COMPACT`) to flatten every `bounds` object in an `observe` result from `{left,top,right,bottom}` to the positional tuple `[left,top,right,bottom]`, via `compactObserveBounds` in `src/features/observe/output/ObserveResultOutput.ts`, gated in `src/server/finalizeToolResponse.ts`. The flag is off by default. See #2951 (original wiring), #2978 (mixed-representation resolution), and PR #2976.

## 1. Test that `--observe-result-compact` composes with `--actions-diff-observe`

`observeResultDropElements` and `observeResultCompact` are applied in `sanitizeObserveResult` / `finalizeToolResponse` (the serialization chokepoint). `actionsDiffObserve` ("return only the diff of the post-action observation") runs on a **different** path — it reshapes the post-action observation before/around that chokepoint. PR #2976 added composition tests for compact × `dropElements` × `toolResultsNoStructuredContent`, but **not** compact × `actionsDiffObserve`.

- **Task:** add a unit test proving a diffed post-action observation still has its `bounds` compacted to tuples when both flags are on (and that the diff logic doesn't choke on tuple-shaped bounds — e.g. if the differ compares bounds objects field-by-field, a tuple vs object comparison could misbehave).
- **Risk to check:** does the diff computation read `bounds.left`-style fields? If so, and if compaction runs *before* the diff, the diff could break. Confirm the ordering (compaction is output-only at finalize, so likely runs after the diff — verify).

## 2. Document the tuple order in the wire / tool schema, not just the flag doc

The tuple order `[left, top, right, bottom]` is currently documented only in `docs/design-docs/mcp/feature-flags.md`. External MCP clients are the only consumers that see the compact form (opt-in), and they are the ones that must decode the tuple. The generated `schemas/tool-definitions.json` `observe` output schema still describes `bounds` as an object.

[...]

Issue #2917, opened 2026-07-03
SessionCacheData typed lastHierarchy/lastScreenshot slots are dormant decoys; real data lives in customData with a different type


kaeawc:
## Summary

`SessionCacheData` declares typed slots that **nothing writes**, while the actual observe-cache data is written into the untyped `customData` map under keys of the same name but a **different type**. The typed fields are dormant decoys, and a consumer that trusts the interface (e.g. the #2761 diff baseline) will read `undefined` — or, worse, assume the wrong type.

### The mismatch

`src/daemon/sessionManager.ts:14-19`:
```ts
export interface SessionCacheData {
  lastHierarchy?: string;      // Last observed view hierarchy   <-- typed string, never written
  lastScreenshot?: string;     // Base64 encoded last screenshot <-- typed string, never written
  lastObserveTime?: number;    // Timestamp of last hierarchy observation
  customData?: Record<string, any>; // Custom data set by tools  <-- where the data ACTUALLY lands
}
```

The write path (`src/server/toolRegistry.ts:443-448` → `ToolExecutionContext.updateSessionCache`, `src/server/ToolExecutionContext.ts:251`) does:
```ts
session.cacheData.customData[cacheKey] = cacheValue;
```
so:
- `customData.lastHierarchy` = a **`ViewHierarchyResult` object** (NOT the typed top-level `lastHierarchy: string`)
- `customData.lastScreenshot` = a base64 **string** (NOT the typed top-level `lastScreenshot: string`)

Grep confirms **no writer** targets `cacheData.lastHierarchy` / `cacheData.lastScreenshot` / `cacheData.lastObserveTime` directly; all observe caching flows through `customData`.

## Why it matters

- **Type lie:** `lastHierarchy` is declared `string` but the real value is a `ViewHierarchyResult` object. Any reader coded against the interface gets a compile-time-valid but runtime-wrong shape.
- **Wrong slot:** a reader doing `session.cacheData.lastHierarchy` gets `undefined`; the data is at `session.cacheData.customData.lastHierarchy`.
- **Blast radius:** the #2761 hierarchy-diff baseline depends on reading this cache (see #2758 / PR #2891, which *repaired* the write so the cache is finally populated). If #2761 reads the typed slot, it silently sees nothing.

Discovered while answering whether the `lastHierarchy` cache is session-independent (it is — keyed by `sessionUuid` → per-`Session` `cacheData.customData`). The keying is correct; only the type/slot modelling is wrong.

[...]

PR #3034, opened 2026-07-04
feat(observe): un-dormant --actions-diff-observe + --actions-no-observe (#3026)

kaeawc:
## What

Un-dormants the two remaining output-reduction action flags — both plumbed but consumed by nothing — at the `finalizeToolResponse` serialization chokepoint:

- **`--actions-diff-observe`** (`AUTOMOBILE_ACTIONS_DIFF_OBSERVE`): non-observe tools emit a compact **diff** of their post-action observation against the last one shown to the agent, instead of the full embedded observation.
- **`--actions-no-observe`** (`AUTOMOBILE_ACTIONS_NO_OBSERVE`): strip the embedded `observation` from non-observe tool results entirely.

Both remain **off by default**.

**Tracking issue: #3026** (the consolidated spec). This PR delivers **both** of #3026's tasks. It supersedes and folds in **#2761** (`--actions-diff-observe`) and **#2762** (`--actions-no-observe`) — please close those as duplicates of #3026 when this merges (not closing them here per process).

**Scope decision: option (b) — deliver both flags in one PR.** #2762 is already closed and #3026 wants both consumed (or removed); landing them together proves the shared **precedence contract** with tests rather than leaving `--actions-no-observe` orphaned-dormant.

## Why

Both flags flowed CLI/env → `outputReductionFlags` → `ServerConfig` → daemon startup, but **no production code read the getters** (`isActionsDiffObserveEnabled()` / `isActionsNoObserveEnabled()` were definitions only) — the identical dormant state #2899/#2933 fixed for `toolResultsNoStructuredContent` and #2951/#2976 for `observeResultCompact`. Enabling either did nothing. Every action currently re-embeds the full post-action observation (~25k tokens); between two same-screen actions almost nothing changes, so re-sending the whole tree is the dominant output cost of an interaction loop.

## How

**Placement — the `finalizeToolResponse` chokepoint (#2758/PR #2891), where the embedded `observation` is attached to non-observe results** (`BaseVisualChange` → `.observation`). The diff is **not** folded into `sanitizeObserveResult` (which is observe-tool shape transforms); it is a separate pure helper applied only at finalize.

**1. `diffObserveResult(baseline, next, cfg)` + `isSameObservationScreen` (`ObserveResultOutput.ts`).** Nodes carry no stable id, so identity is a synthetic **positional path key**: each node's local key `resource-id + bounds + text + sibling index` (NUL-joined) prefixed by its **ancestor chain** (depth-joined) — so identically-shaped repeated list cells in different subtrees never collide. `boundsKey` normalizes both the object shape and the `--observe-result-compact` tuple in both the key and the attribute compare. The diff is `{ isDiff: true, added, removed, changed, fields }` (`changed` carries per-attribute `{from,to}`; `fields` are changed top-level scalars — `updatedAt` excluded as churn). Pure/output-only.

**2. Baseline storage (`sessionManager.ts`).** Typed top-level slot `SessionCacheData.lastRenderedObservation` + `setLastRenderedObservation` (canonical typed-slot pattern per #2917, deviating from the issue's `customData` text). Holds the wire-shaped sanitized observation so diffs compare like-for-like; distinct from `lastHierarchy` (full untrimmed, internal reuse).

**3. `finalizeToolResponse`.**
- `observe` always emits the full sanitized observation and **resets** the baseline (never stripped by no-observe).
- Non-observe action:
  - **`--actions-no-observe` (higher precedence):** delete `payload.observation` from the served payload (both representations). The diff flag is moot.
  - else **`--actions-diff-observe`:** emit `diffObserveResult(baseline, sanitized)` on a same-screen baseline, then update the baseline. **Falls back to the full observation** on cross-screen change, missing baseline, or no `sessionUuid` (legacy single-agent path).
- Gated on the flags plus an injected `ObservationBaselineStore` (interface + fake) so finalize keeps no `sessionManager`/`DaemonState` dependency.

**4. `toolRegistry`.** Backs the store with the session cache only when a session + initialized daemon exist. Inert by default.

**5. Composition contract (#3026).** Precedence `no-observe > diff` is explicit and tested. The #2990 fence tests (EC-D1/EC-D2) still pass — they pass a `sessionUuid` but no store, so no diff is produced and the full compacted observation is emitted; compaction stays gated solely by `isObserveResultCompactEnabled()`. The stale fence comment (which speculated a handler-side diff) is updated to the shipped finalize placement.

**6. Docs.** `feature-flags.md` rows for both flags expanded to concrete behavior + precedence.

## Testing

TDD — tests first (red), then green.

- `diffObserveResult.test.ts` (21): empty diff; added/removed child; `checked`-toggle → `changed`; bounds change → remove+add; scalar-field diff; `updatedAt` never reported; **cross-subtree collision** (fails on the old local-only key); **object-vs-tuple bounds equivalence**; **real-fixture size proof** — a one-node change on `android-home.json` diffs to <10% of the full observation (the flag's whole point, measured with the #2755 `measureValue` harness); empty/absent `viewHierarchy`; array-shaped root; removed-subtree; no-mutation; <100ms. Plus `isSameObservationScreen`.
- `finalizeToolResponse.test.ts` (+15): diff-emit (observe resets baseline; action emits diff in both representations; baseline updates; fallback on missing baseline / cross-screen / no-sessionUuid; observe resets after a diff; output-only; compact-on diff carries tuple bounds) and **no-observe + precedence** (strips from a non-observe action in both representations; never strips observe; flag-off no-op; **both flags on → stripped, no diff, baseline untouched**; no-observation passthrough).

Verified: `bun run lint` clean, `bun run build` ok, `bun test test/server test/daemon test/features/observe/output` → **1405 pass / 0 fail**; no schema drift (both flags default-off). Unit tests use interface+fake, no device/network, <100ms.

## Acceptance (#3026)
- [x] `--actions-diff-observe` consumer wired at finalize; diff shape + synthetic key; baseline reset on observe; cross-screen / no-session fallback.
- [x] `--actions-no-observe` consumer wired at finalize; observation stripped from non-observe results; internal success detection unaffected (output-only).
- [x] Precedence `no-observe > diff` defined + tested.
- [x] No plumbing duplicated (getters already existed).
- [x] Byte-reduction measured on the #2755 fixture.
- [ ] Diff format signed off on first **real device** output — human step this PR can't satisfy; safe to land the default-off plumbing ahead of it.

## Bookkeeping
- Tracking: **#3026**. Delivers both its tasks.
- Fold in / close as duplicates on merge: **#2761**, **#2762**.


kaeawc:
## Addressed multi-perspective review (correctness / architecture / test-quality)

Three independent reviewers ran against the initial commit. Findings and resolutions (commit 5b5da7e0e):

**Correctness — ship-blocker fixed.** The node identity key used only the *local* sibling index, so identically-shaped repeated list cells in different subtrees collided and mis-paired → a phantom `checked false→true` change plus removing the wrong cell. Fixed: identity is now a globally-positional `pathKey` (ancestor chain of local keys, depth-joined with U+0001); the emitted `key` stays the readable leaf. New test `identical cells in different subtrees do not collide` fails on the old local-only key.

**Correctness — hardening (were latent, now closed).** `diffAttributes` compares `bounds` via `boundsKey`, so an object-shaped baseline vs a compacted-tuple next is not a spurious change. `valuesEqual` uses a key-order-insensitive `stableStringify` so a differently-ordered nested object can't produce a phantom `changed`.

**Test-quality — gaps closed.** Added: object-vs-tuple bounds equivalence (kills the `boundsKey→""` mutation), a **real-fixture size proof** (a one-node change on `android-home.json` diffs to <10% of the full observation — the whole point of the flag), empty/absent `viewHierarchy`, array-shaped root, removed-subtree, and an end-to-end compact-on diff carrying tuple bounds through `finalizeToolResponse`.

**Architecture — confirmed direction; doc fix applied.** Finalize placement (not handler-side), the typed `lastRenderedObservation` slot (per #2917, deviating from the issue's `customData` text), the `{isDiff, added, removed, changed, fields}` keyed-multiset shape over JSON-Patch, and the 2-method `ObservationBaselineStore` DI interface were all endorsed. Updated the stale #2990 fence comment (the diff ships *in* finalize now).

**Acceptance note:** #2761 lists "diff output format signed off on first real device output before finalizing" — a human sign-off this PR can't satisfy. The flag is **off by default**, so the plumbing is safe to land ahead of that sign-off; I'll leave that acceptance box for a real-device pass.

Verified after fixes: `bun run build` ok, `bun run lint` clean, `bun test test/server test/daemon` → 1344 pass / 0 fail; focused suites 122 pass / 0 fail; no schema drift (default-off flag).

kaeawc:
## Reconciled against the consolidated spec #3026

Backlog dedup flagged that **#2761** and **#2762** are both superseded by **#3026** (the authoritative, consolidated spec for the two dormant action flags). Re-aligned so nothing is built against a stale contract:

- **Placement confirmed at the `finalizeToolResponse` chokepoint** (where the embedded `.observation` is attached, #2758/PR #2891) — not folded into `sanitizeObserveResult`. The diff is a separate pure helper applied only at finalize.
- **No plumbing re-added** — both `ServerConfig` getters already existed (dormant); this wires the consumers only.
- **Scope: option (b)** — expanded to deliver **both** flags so the shared **precedence contract** (`--actions-no-observe` > `--actions-diff-observe`) lands here with tests, instead of leaving `--actions-no-observe` orphaned-dormant (#2762 is already closed).
- **Composition honored** — precedence is explicit (`diffActive = diffEnabled && !noObserveEnabled`) and tested (both-flags-on → observation stripped, no diff, baseline untouched). The #2990 EC-D1/EC-D2 fence tests still pass; compaction stays gated solely by `isObserveResultCompactEnabled()`.
- **Bookkeeping:** PR now tracks **#3026** and delivers both its tasks. **#2761** and **#2762** should be closed as duplicates of #3026 when this merges — flagging for the maintainer rather than closing them here.

Byte-reduction is measured on the #2755 `android-home.json` fixture (a one-node change diffs to <10% of the full observation). The only open acceptance item is the human "sign off the diff format on real device output" step, which the default-off plumbing can safely land ahead of.

Issue #2964, opened 2026-07-03
Android runner: reject non-finite gesture coordinates (isFinite guard) — mirror iOS #2928, load-bearing on the kotlinx wire


kaeawc:
## Summary

Mirror the iOS non-finite gesture-coordinate guard (#2928 / PR #2957) on the **Android** CtrlProxy runner. Android issue #2927 / PR #2946 widens the Android gesture request coordinate fields `Int → Double` (symmetric to iOS #2909 / #2919), but — unlike iOS — adds **no `isFinite` guard** at the handler boundary. Once #2946 merges, Android re-opens exactly the non-finite exposure that #2928 just closed for iOS.

## Why this matters (and why Android is *more* exposed than iOS)

- On **Apple/Darwin Foundation**, `JSONDecoder` rejects an overflow literal like `1e309` at pre-parse (`dataCorrupted`, empty `codingPath`) — it never reaches the widened `Double` field. So on iOS the guard is defense-in-depth for the non-wire construction path.
- On **Android/kotlinx.serialization**, a JSON overflow literal like `1e309` decodes to `Double.POSITIVE_INFINITY` **directly into the widened field** (kotlinx does not reject it). So after #2946, a non-finite coordinate can arrive **over the wire** and flow unchecked into the gesture dispatch (`android/control-proxy/.../CtrlProxy.kt` `perform*`), where on iOS the guard now catches it. The guard is therefore *load-bearing on the wire* for Android, not merely defense-in-depth.

(Reference: PR #2957's `testOverflowCoordinateLiteralIsRejectedAtDecode` documents the Darwin pre-parse behavior; the Swift-runtime review lens on #2957 flagged that swift-corelibs/kotlinx-style decoders coerce to Infinity instead.)

## Proposed fix

After #2946's widening, add an `isFinite` guard at the Android gesture handler boundary — the direct analog of iOS `CommandHandler.requireFinite(_:field:)` (`ios/control-proxy/Sources/CtrlProxy/CommandHandler.swift`) — rejecting any non-finite coordinate (`NaN`/`±Infinity`) for `request_tap_coordinates` / `request_swipe` / `request_(two|multi)_finger_swipe` / `request_drag` / `request_pinch` with a clean, actionable structured error response (per-command `*_result`, `success:false`) instead of dispatching into the accessibility gesture engine. A small shared `requireFinite(value, field)` helper keeps it DRY across the handlers.

## Acceptance criteria

[...]

Issue #2928, opened 2026-07-03
iOS runner: reject non-finite gesture coordinates (isFinite guard) at the decode/handler boundary


kaeawc:
## Summary

Hardening follow-up to #2909 / PR #2919. Widening the iOS gesture request coordinate fields `Int → Double` (correct, and required to accept fractional coordinates) has a narrow side effect: JSON overflow literals like `1e309` now decode to `Double.infinity` at the decode boundary, where the old `Int` field would have **rejected** them with a clean decode error. Surfaced by the Swift-runtime review lens on PR #2919.

## Detail

- JSON has no `NaN` literal, so `NaN` cannot arrive over the wire — but `1e309` (valid JSON) decodes to `+Infinity` into the now-`Double` coordinate fields (`ios/control-proxy/Sources/CtrlProxy/Models.swift` gesture request structs).
- There is no `isFinite` / range guard anywhere on the coordinate path (`GesturePerformer.tap/swipe/drag/pinch`, `CommandHandler.handle*`). A non-finite value flows into `CGVector`/`CGFloat` → `XCUICoordinate.withOffset` / the ObjC synthesizers.
- Partial existing mitigation: the ObjC pinch/multi-finger synthesis is wrapped in `@try/@catch` (`ObjCExceptionCatcher.m`) and converts a thrown exception to a structured `false`/error response; `pinch` also clamps radius (`distanceStart > 1 ? … : 1`). But `tap`/`swipe`/`drag` are **not** exception-wrapped, so a non-finite coordinate there is undefined behavior (likely a silent no-op rather than a crash).

## Why this is a follow-up, not part of PR #2919

- The exposure to negative/huge magnitudes is **pre-existing** — the old `Int` fields already admitted negative and arbitrarily-large integers. The PR only newly admits `Infinity` (from overflow), and only if a client deliberately sends `{"x": 1e309}`. The legitimate TS caller never does (JS coordinates are finite and, on the shipped path, still rounded).
- Both independent reviewers judged this out of scope for the fractional-coordinate fix and recommended a separate hardening issue rather than expanding that PR.

## Proposed fix

[...]

Issue #2927, opened 2026-07-03
Android runner: widen gesture request coordinate fields to Double (symmetric to iOS #2909)


kaeawc:
## Summary

Follow-up to #2909 / PR #2919 (iOS). The Android CtrlProxy runner has the **same latent `Int`-decode gap** on gesture request coordinate fields. Surfaced during the 3-perspective review of PR #2919 (cross-platform wire-protocol lens).

## Detail

`android/protocol/src/main/kotlin/dev/jasonpearson/automobile/protocol/WebSocketRequest.kt` declares coordinate fields as `Int` (kotlinx.serialization `@Serializable`):
- `RequestTapCoordinates.x/y` (`:52-53`)
- `RequestSwipe.x1/y1/x2/y2` (`:61-64`)
- `RequestTwoFingerSwipe.x1/y1/x2/y2` (`:72-75`)
- `RequestDrag.x1/y1/x2/y2` (`:84-87`)
- `RequestPinch.centerX/centerY/distanceStart/distanceEnd` (`:108-111`)

A fractional JSON number into a kotlinx `Int` field throws a decode exception — the same opaque failure #2909 describes for Swift.

## Why the severity is LOWER than iOS (why this is a separate, non-urgent follow-up)

Android's TS wire path uses `roundCoordinates: true` (`src/features/observe/android/CtrlProxyGestures.ts:21`), and the Android-only two-finger path additionally hard-rounds with `Math.round(...)` before serializing (`android/CtrlProxyGestures.ts:66-72`). So rounding is **centralized at the delegate layer** (effectively the issue's "option 2") and no fractional value can reach the wire from any current caller. iOS was higher-severity because its whole path runs `roundCoordinates:false` and depended on per-caller rounding in `PinchOn.ts`.

This is a defensible asymmetry, not a live bug — but #2909's "repo-wide robustness gap" framing is only half-closed until Android's protocol is symmetric (robust to any client regardless of the shared `coord()` rounding, which must not regress).

## Proposed fix

Widen the five gesture request `data class` coordinate fields in `WebSocketRequest.kt` from `Int` to `Double` (mirroring PR #2919 on iOS). `Double` deserializes both integer and fractional JSON, so the existing rounded path is unaffected. Verify the Android gesture engine / accessibility-service dispatch consumes `Double` (or convert at the handler boundary where a pixel grid is required, as the iOS `GesturePerformer` already does).

## Acceptance criteria

[...]

PR #3035, opened 2026-07-04
fix(android): route two-finger swipe through RequestManager (#2988)

### LINKED ISSUE #2988: Android two-finger swipe result correlation is orphaned — resolves only via 5s timeout (handleTwoFingerSwipeResult has no caller)
## Summary

The TypeScript Android two-finger-swipe result correlation is **orphaned**: `CtrlProxyGestures.handleTwoFingerSwipeResult` and its `pendingSwipeRequestId`/`pendingSwipeResolve` promise mechanism have **zero callers**, so a `two_finger_swipe` request's `swipe_result` is never routed to resolve its promise. The two-finger swipe promise resolves **only via its 5s timeout**, regardless of whether the runner reported success or failure.

Surfaced during the multi-agent review of PR #2984 (issue #2964, the Android non-finite gesture-coordinate guard). **Pre-existing and orthogonal to that PR** — #2984 is neutral to this bug — filed as a follow-up.

## Evidence (verified by reading `src/`)

- `src/features/observe/android/CtrlProxyGestures.ts`
  - `twoFingerSwipe(...)` (~L40) generates `requestId = two_finger_swipe_<ts>_<rand>` (L50), stores `this.pendingSwipeRequestId = requestId` / `this.pendingSwipeResolve = resolve`, and sends the request via a **raw** `this.context.getWebSocket()?.send(message)` (L84). It never calls `requestManager.register(...)`.
  - `handleTwoFingerSwipeResult(requestId, result)` (L93) is the intended resolver — but `grep -rn "handleTwoFingerSwipeResult" src/ test/` returns **only the definition**, no caller.
- `src/features/observe/android/AndroidCtrlProxyClient.ts`
  - On `swipe_result` (L1976-1984) it calls `this.requestManager.resolve<A11ySwipeResult>(message.requestId, {...})`. That resolves only ids present in the `RequestManager.pending` map. The two-finger id was never registered there, so `resolve(...)` no-ops for two-finger results.
- Net effect: `swipe_result` frames for a `two_finger_swipe_*` requestId are dropped; the promise from `twoFingerSwipe(...)` is settled only by its `timeoutMs` (default 5000ms) timer (`CtrlProxyGestures.ts` L56-66), returning `{ success: false, error: "Two-finger swipe timeout ..." }` even when the runner succeeded.

## Impact

- Every two-finger swipe (used for TalkBack-mode scrolling) blocks ~5s and reports a **timeout error** to the caller even on success — a latent correctness + latency bug. (If callers ignore the boolean and just await, they still eat the 5s stall.)

## Proposed fix (pick one, match the codebase convention)

1. **Register via `RequestManager` like the other gestures** (preferred, consistent): have `twoFingerSwipe(...)` go through the same `sendCommand`/`requestManager.register(requestId, ...)` path the single-finger swipe/tap/drag/pinch use, so the existing `AndroidCtrlProxyClient` `swipe_result` handler resolves it. Then delete the dead `pendingSwipeRequestId`/`pendingSwipeResolve`/`handleTwoFingerSwipeResult` mechanism.
2. **Wire the existing resolver**: in `AndroidCtrlProxyClient`'s `swipe_result` branch, if `message.requestId?.startsWith("two_finger_swipe_")`, route to `gestures.handleTwoFingerSwipeResult(message.requestId, {...})` before/instead of `requestManager.resolve`. Lower-churn but keeps two parallel mechanisms.

Option 1 removes the divergence entirely and is the DRY choice.

## Acceptance criteria

- [ ] A `two_finger_swipe` that the runner completes resolves its promise from the actual `swipe_result` (`success:true`) **without** waiting for the timeout.
- [ ] A runner-reported failure (`success:false`, incl. the new #2964 non-finite guard error) resolves the promise promptly with that error.
- [ ] The dead correlation mechanism is removed (option 1) or actually invoked (option 2) — no orphaned resolver remains.
- [ ] Test: a fake WebSocket delivering a `swipe_result` for a `two_finger_swipe_*` requestId resolves the promise before the fake timer's timeout fires (Interface + Fake + FakeTimer, per CLAUDE.md).

## Context / provenance

- Found by the "protocol-compat / TS-consumer" review lens on PR #2984.
- Related: #2964 (Android non-finite guard), PR #2946 (#2927, Double widening), iOS #2928 / PR #2957.

kaeawc:
## What

Route the Android **two-finger swipe** (`CtrlProxyGestures.requestTwoFingerSwipe`) through the shared `sendCommand` → `RequestManager.register` path that every sibling gesture already uses, and delete the dead bespoke promise mechanism.

Closes #2988.

Affected files:
- `src/features/observe/android/CtrlProxyGestures.ts` — rewrite `requestTwoFingerSwipe`; remove `pendingSwipeRequestId` / `pendingSwipeResolve` / `handleTwoFingerSwipeResult`.
- `test/features/observe/android/CtrlProxyGestures.test.ts` — new TDD coverage.

## Why

The two-finger swipe result correlation was **orphaned**. `requestTwoFingerSwipe` sent its request via a **raw** `getWebSocket()?.send(message)` and stored `pendingSwipeRequestId`/`pendingSwipeResolve` for a resolver — `handleTwoFingerSwipeResult` — that had **zero callers** (`grep -rn handleTwoFingerSwipeResult src/ test/` returned only the definition). The runner replies with a `swipe_result` frame, which `AndroidCtrlProxyClient` resolves through `RequestManager` **by requestId**; but the two-finger id was never registered there, so the frame was dropped and the promise settled **only via its 5s timeout** — returning `{ success:false, error:"Two-finger swipe timeout…" }` even when the runner succeeded.

Impact: every two-finger swipe (TalkBack-mode scrolling, `TalkBackSwipeExecutor`) blocked ~5s and reported a timeout error, even on success.

## How

Option 1 from the issue (the DRY choice):

- `requestTwoFingerSwipe` now calls `sendCommand<A11ySwipeResult>(...)` with `idPrefix: "two_finger_swipe"`, `messageType: "request_two_finger_swipe"`, and the `{x1,y1,x2,y2,duration,offset}` params. `sendCommand` runs the same `cancelScreenshotBackoff → ensureConnected → register → send → await` flow as swipe/tap/drag/pinch.
- The Android runner **shares the `swipe_result` response type** for two-finger swipes (`android/.../CtrlProxyMessageHandler.kt:124-125`; `SwipeResult` is `@SerialName("swipe_result")` with a `requestId`). So the existing `AndroidCtrlProxyClient` `swipe_result` handler resolves the two-finger promise by requestId as soon as the runner replies — success **or** the #2964 non-finite guard failure — with no timeout wait.
- Deleted the dead `pendingSwipeRequestId` / `pendingSwipeResolve` / `handleTwoFingerSwipeResult` members. No orphaned resolver remains.

Preserved behavior:
- **Wire format** — `createMessage` emits `{"type":"request_two_finger_swipe","requestId":…,"x1":…,"y1":…,"x2":…,"y2":…,"duration":…,"offset":…}`, byte-identical field order to the old serializer.
- **Integer coordinate rounding** — explicit `Math.round(...)` retained (intentional per the existing comments).
- **requestId prefix** — still `two_finger_swipe_…`.
- **Not-connected** — still returns `{success:false, error:"Not connected"}` without sending.
- `perf` is now threaded through `sendCommand` instead of ignored (minor improvement).

## Testing

New `test/features/observe/android/CtrlProxyGestures.test.ts` (Interface + Fake WebSocket + `FakeTimer` + real `RequestManager`, per CLAUDE.md; <100ms):

- Resolves from a `swipe_result` frame **before** the timeout fires (`success:true`), timer never advanced, no dangling pending request.
- Propagates a runner-reported failure (`success:false` + error, incl. the #2964 non-finite guard error) promptly, no timeout wait.
- Sends the correct wire message with rounded coordinates + `offset`.
- Still times out when the runner never replies.
- Returns `Not connected` without sending when the connection can't be established.

```
bun test test/features/observe/android/CtrlProxyGestures.test.ts   # 5 pass
bun test test/features/observe/ test/features/action/swipeon/       # green (see note)
npx eslint <changed files>                                          # clean
bun build.ts                                                        # ok
```

Note: `test/features/observe/CtrlProxyClient.test.ts > getLatestHierarchy > should preserve SDK screen names…` fails **identically on `origin/main` without this change** (verified via `git stash`) — pre-existing and unrelated to this PR.

## Related

- #2964 (Android non-finite gesture guard), PR #2946 / #2927 (Double widening), iOS #2928 / PR #2957.

### DIFF
--- a/src/features/observe/android/CtrlProxyGestures.ts
+++ b/src/features/observe/android/CtrlProxyGestures.ts
@@ -8,15 +8,10 @@
 import type { PerformanceTracker } from "../../../utils/PerformanceTracker";
 import { NoOpPerformanceTracker } from "../../../utils/PerformanceTracker";
 import { SharedGestureDelegate } from "../shared/SharedGestureDelegate";
+import { sendCommand } from "../DeviceServiceUtils";
 import type { DelegateContext, A11ySwipeResult } from "./types";
-import { generateSecureId } from "./types";
-import { ctrlProxyRequests, serializeCtrlProxyRequest } from "./ctrlProxyProtocol";
 
 export class CtrlProxyGestures extends SharedGestureDelegate {
-  // Legacy pending request state for two-finger swipe (uses manual promise pattern)
-  private pendingSwipeRequestId: string | null = null;
-  private pendingSwipeResolve: ((result: A11ySwipeResult) => void) | null = null;
-
   constructor(context: DelegateContext) {
     // `roundCoordinates: true` centralizes coordinate rounding for Android at the delegate layer
     // (SharedGestureDelegate.coord()), so integer coordinates reach the runner wire. This is
@@ -28,8 +23,12 @@ export class CtrlProxyGestures extends SharedGestureDelegate {
   }
 
   /**
-   * Request a two-finger swipe gesture for TalkBack mode.
-   * This is Android-only and uses a legacy manual promise pattern.
+   * Request a two-finger swipe gesture for TalkBack mode. Android-only.
+   *
+   * Routed through `sendCommand`/`RequestManager` like every other gesture (#2988): the runner's
+   * `swipe_result` frame is correlated by requestId and resolves this promise as soon as it
+   * arrives, instead of the promise only ever settling via its timeout. The requestId keeps the
+   * `two_finger_swipe_` prefix so callers/loggers that key on it are unaffected.
    */
   async requestTwoFingerSwipe(
     x1: number,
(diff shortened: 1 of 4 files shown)


kaeawc:
## Adversarial review round (3 independent perspectives)

Ran three independent reviewers — regression hunter, completeness/spec auditor, architecture/direction — against this branch. Consensus: **ship-worthy, no regression, all four acceptance criteria met.**

- **Regression hunter:** No regression. Wire message byte-identical (same field set/order), no single↔two-finger `swipe_result` mis-route (correlation is by unique `requestId`; distinct `idPrefix`), `isCommandSupported` gate inert on Android (not set on the delegate context), `two_finger_swipe_` prefix preserved and no consumer pattern-matches it. Caller tests (`TalkBackSwipeExecutor`) green.
- **Completeness auditor:** All 4 criteria met. The "resolves without waiting for the timeout" claim is genuinely proven — the `timer.getCurrentTime() === 0` assertions would fail under the old timeout-only behavior. No orphaned resolver / dead members remain anywhere in `src/` or `test/`.
- **Architecture:** Right solution — converges Android onto the same `sendCommand` path iOS already uses; correctly left as an Android override (iOS's two-finger analogue is a distinct `request_multi_finger_swipe` command). YAGNI on pushing into `SharedGestureDelegate`.

### Addressed
All three flagged the now-orphaned `ctrlProxyRequests.requestTwoFingerSwipe` **builder function** as the top (non-blocking) cleanup — it was the sole gesture builder left with zero production consumers and escapes CI dead-code gates. Removed in 27dbcf5a8 (kept the type + union + registry entry the wire drift-guard needs).

### Consciously declined (YAGNI / out of scope)
- Making `SharedGestureDelegate.coord()` `protected` to reuse instead of the explicit `Math.round` — the two-finger path bypasses `coord()` entirely (calls `sendCommand` directly), so there is no double-round; the explicit rounding is the sole, necessary rounding with a strong intentional comment. Single consumer → not worth widening the shared base.
- Test `flush()` helper's fixed `await Promise.resolve()` count is mildly brittle but correct today; test-only.

kaeawc:
Merged. Deferred / non-blocking review items filed as follow-ups: #3048 (two-finger swipe concurrency + `type:"error"` resolution test coverage) and #3049 (optional `protected coord()` reuse + `flush()` test-helper robustness).

Issue #3052, opened 2026-07-04
actions-diff-observe: diff omits top-level focusedElement / accessibilityFocusedElement / awaitedElement changes


kaeawc:
## Context
Follow-up to merged **PR #3034** (`--actions-diff-observe`, tracking #3026). `diffObserveResult` (`src/features/observe/output/ObserveResultOutput.ts`) diffs the view-hierarchy nodes and a curated set of top-level **scalar** fields (`DIFF_SCALAR_FIELDS`: rotation, wakefulness, userId, intentChooserDetected, notificationPermissionDetected, awaitTimeout, error).

It does **not** diff the top-level convenience **Element** mirror fields on `ObserveResult`:
- `focusedElement`
- `accessibilityFocusedElement`
- `awaitedElement`

## Impact
A focus change *is* reflected in the hierarchy (a node's `focused`/`accessibility-focused` attribute flips → shows as a `changed` node), so no information is strictly lost. But a consumer that reads the top-level `focusedElement`/`awaitedElement` mirror on an action result will not see it change in the emitted diff (those fields simply carry over from... nowhere — the diff replaces the whole `.observation`). For `observe waitFor` results surfaced through a non-observe action, `awaitedElement`/`awaitDuration`/`awaitTimeout` (only `awaitTimeout` is in the scalar set) may be under-represented.

## Options
1. Add `focusedElement` / `accessibilityFocusedElement` / `awaitedElement` (+ `awaitDuration`) to a diffed set, emitting `{from,to}` when they change (compare by a stable element key, or shallow-diff).
2. Explicitly document that focus/await state must be read from the hierarchy nodes in diff mode, and leave the mirrors out (they are redundant with the hierarchy).

Recommend (1) for `awaitedElement`/`awaitDuration` at least, since those have no hierarchy equivalent for the wait outcome.

## Pointers
- `diffObserveResult` + `DIFF_SCALAR_FIELDS` in `src/features/observe/output/ObserveResultOutput.ts`.
- `ObserveResult` field definitions in `src/models/ObserveResult.ts` (focusedElement:102, accessibilityFocusedElement:108, awaitedElement:152).
- Tests: `test/features/observe/output/diffObserveResult.test.ts`.

kaeawc:
[...]

PR #3059, opened 2026-07-04
feat(observe): diff top-level focusedElement/accessibilityFocusedElement/awaitedElement (#3052)

kaeawc:
## What

Extends `diffObserveResult` (`src/features/observe/output/ObserveResultOutput.ts`) to diff the top-level **Element mirror fields** on `ObserveResult`, which the `--actions-diff-observe` diff previously ignored:

- `focusedElement`
- `accessibilityFocusedElement`
- `awaitedElement`

plus the scalar `awaitDuration`. Each is emitted as `{from, to}` in the existing `ObserveDiff.fields` map when it changes. Closes **#3052** (follow-up to PR #3034 / tracking #3026).

## Why

`diffObserveResult` diffed the view-hierarchy nodes and a curated scalar set (`DIFF_SCALAR_FIELDS`), but not the convenience Element mirrors. A focus change *is* reflected in the hierarchy (a node's `focused` / `accessibility-focused` flips → a `changed` node), so no information was strictly lost — but a consumer that reads the top-level `focusedElement` / `awaitedElement` mirror off a non-observe action's diff would never see it change (the diff replaces the whole `.observation`, and these fields were absent from it). For an `observe waitFor` surfaced through a non-observe action, the **wait outcome** (`awaitedElement` / `awaitDuration`) has *no* hierarchy equivalent, so it was under-represented — the issue's primary motivation.

This implements the issue's **recommended Option 1**.

## How

**1. `DIFF_ELEMENT_FIELDS`** — the three Element mirror fields. Diffed into the same `fields` map as scalars (keeps the `ObserveDiff` shape stable; consumers already read `fields`). A focus gain reads as `{from: undefined, to: element}`, a loss as `{from: element, to: undefined}`.

**2. Bounds-tolerant compare.** Element mirrors carry a `bounds` that `--observe-result-compact` flattens from `{left,…}` to `[l,t,r,b]`. `stableStringify` gained a `boundsTolerant` flag that canonicalizes every `bounds` (object or tuple) to its `boundsKey` string at any depth, so a compaction toggle between captures is **not** a spurious change — mirroring `diffAttributes`' existing bounds handling for hierarchy nodes. `JSON.stringify(undefined)` is `undefined`, so an absent mirror only equals another absent mirror (clean focus gain/loss semantics).

**3. `awaitDuration` → `DIFF_SCALAR_FIELDS`.** A scalar with no hierarchy equivalent (the wait outcome). Not churn like `updatedAt`: present only on wait results, so an ordinary action carries `undefined` on both sides and reports nothing.

**4. `DiffObserveConfig.elementFields`** override, for symmetry with `scalarFields` (testability).

No consumer change: the finalize hook (`finalizeToolResponse.ts`) already emits `diff.fields`; the additions flow through unchanged.

## Testing

TDD — tests written first (red: module-missing export), then green.

`diffObserveResult.test.ts` (+9, 21 → 30 pass):
- changed `focusedElement` captured in `fields`
- focus gained (`undefined → element`) and lost (`element → undefined`) both surface
- unchanged `focusedElement` not reported
- `focusedElement` differing only in bounds shape (object vs tuple) → **not** a change
- `accessibilityFocusedElement` + `awaitedElement` changes captured
- `awaitDuration` captured as a scalar field; `DIFF_SCALAR_FIELDS` contains it
- `DIFF_ELEMENT_FIELDS` covers exactly the three mirrors
- element-field diffing does not mutate either input

Verified locally: `bun run lint` clean, `bun run build` ok, `bun test test/server test/features/observe/output` → **829 pass / 0 fail**, `scripts/all_fast_validate_checks.sh` → **11/11**. Unit tests are pure (no device/network), <100ms.

## Acceptance (#3052)
- [x] `focusedElement` / `accessibilityFocusedElement` / `awaitedElement` diffed, emitting `{from,to}` on change (Option 1).
- [x] `awaitDuration` diffed (the wait outcome with no hierarchy equivalent).
- [x] Bounds-tolerant so a `--observe-result-compact` toggle is not a spurious change.
- [x] Pure / output-only; inputs never mutated; existing node + scalar diff behavior unchanged.


kaeawc:
## Adversarial review pass — findings + responses

Ran three independent reviewers (correctness/regression, architecture/direction, completeness/tests). Summary and how each was addressed:

**Correctness (verdict: no bugs, ship-ready).** All 7 scrutiny points verified non-issues: the new `boundsTolerant` flag is opt-in so existing scalar/node compares are byte-identical; no churn (`undefined`-vs-`undefined` short-circuits); no mutation (diff is serialize-only, inputs deep-cloned each action); focus gain/loss `undefined` semantics correct.

**Architecture (verdict: mergeable, but flagged the encoding).** Emitting the *full* `Element` as `{from,to}` can re-embed a `node` child subtree, undercutting the token-reduction purpose. Verified real: `parseNodeBounds` shallow-copies the source node, and CtrlProxy nodes have no `.$` wrapper, so a focused *container* retains its children.
→ **Addressed** in 574d34b57: mirror elements are now **`node`-subtree-stripped** for both compare and emit (`leanElementForDiff`). The subtree is redundant with the node diff and is the only unbounded part of an `Element`, so stripping it bounds the size while keeping the useful "which element is focused/awaited" answer. A child-only change no longer double-reports the mirror. Kept all three mirrors (bounded cost now) for consumer ergonomics — the mirror exists precisely so a reader doesn't have to correlate node `focused` flips.

**Completeness (verdict: one blocking doc gap + test follow-ups).**
- **Blocking:** `docs/design-docs/mcp/feature-flags.md` said the diff `fields` are "scalars only" — stale. → **Fixed**: now documents the element mirror fields + `awaitDuration`.
- Untested `elementFields` config override + combined scalar/element/node scenario. → **Added** both tests (plus subtree-strip and child-only-no-op cases).

Test count on the diff suite: 21 → 34, all green; `bun test test/features/observe/output test/server` → 833 pass / 0 fail; lint + build clean; `scripts/all_fast_validate_checks.sh` 11/11.

Issue #3023, opened 2026-07-04
Android CtrlProxy: async action failures inside launched coroutines still hang the awaiter


kaeawc:
## Summary

Follow-up from #2985 / PR #3019. That PR closed the two **synchronous** swallow points in `WebSocketServer.handleClientMessage` — a failed decode and a synchronous throw from `handler.handleMessage(...)` now broadcast a structured `type:"error"` frame. But a deeper class of silent hang remains: **actions that respond asynchronously from their own launched coroutine and throw after `handleMessage` has already returned.**

## Detail

`CtrlProxyMessageHandler.handleMessage` (`android/control-proxy/.../CtrlProxyMessageHandler.kt`) dispatches most commands as **fire-and-forget** — the KDoc says "Every current command is fire-and-forget (the action broadcasts its own response asynchronously), so `handleMessage` always returns `null`." The action implementations in `CtrlProxy.kt` (e.g. `requestScreenshot`, `requestSwipe`, gesture/hierarchy handlers) do their real work inside `serviceScope.launch { … }` / callbacks and broadcast their own result later.

PR #3019's handler `catch` only wraps the **synchronous** portion of `handler.handleMessage(request)` in `WebSocketServer.handleClientMessage`:

```kotlin
try {
  val response = handler.handleMessage(request)   // returns null immediately for async actions
  if (response != null) broadcast(response)
} catch (e: CancellationException) { throw e }
  catch (e: Exception) { /* broadcast ErrorResponse */ }
```

If an action throws **inside its own launched coroutine** (after `handleMessage` returned `null`), that exception never reaches this `catch`. No error frame is emitted and the awaiting daemon request hangs to timeout — the exact symptom #2985 set out to eliminate, just on the async path.

## Impact

- An action-internal failure (e.g. an exception in the screenshot capture coroutine, a gesture dispatch failure inside `serviceScope.launch`) → daemon awaiter hangs to `RequestManager` timeout instead of a fast, correlated `type:"error"` failure.
- Scope-limited: only affects actions whose failure path currently logs-and-returns inside the launched coroutine without broadcasting any `*_result` / error. Actions that already broadcast a `success:false` result on their own error path are unaffected.

## Proposed direction

Per-action error plumbing so every launched action coroutine broadcasts a correlated failure on throw. Options:
[...]

Issue #3022, opened 2026-07-04
Android CtrlProxy error-envelope: iOS parity (rewriteUnknownCommandError, out-of-range legibility, targeted send)


kaeawc:
## Summary

Follow-up from #2985 / PR #3019 (Android CtrlProxy structured error envelope). That PR made the Android runner emit a `type:"error"` frame on decode/handler failures and wired the TS daemon consumer to fail the awaiter fast. Three points remain where **Android** is not yet at parity with the **iOS** decode-boundary work (#2965 / PR #2983). None were regressions; all were explicitly deferred as follow-ups by the multi-agent review of PR #3019.

### 1. TS consumer does not rewrite "Unknown command type" into an upgrade hint
- iOS runs the runner error text through `rewriteUnknownCommandError` before resolving the request — `src/features/observe/ios/decodeCtrlProxyMessage.ts:30-36`, called at `decodeCtrlProxyMessage.ts:275` — turning terse `"Unknown command type: X"` into actionable *"the runner is likely older than this daemon; rebuild/redeploy…"* guidance.
- The Android branch added in PR #3019 (`src/features/observe/android/AndroidCtrlProxyClient.ts`, the `message.type === "error"` case) passes `message.error` **verbatim** to `requestManager.resolveError`.
- The Kotlin runner *already emits the symmetric wire text*: `WebSocketServer.describeDecodeFailure` returns `"Unknown command type: $type"` (`android/control-proxy/.../WebSocketServer.kt`), specifically so a rewrite can match — but the daemon never applies one.
- **Note:** the iOS `rewriteUnknownCommandError` message is iOS-specific ("iOS CtrlProxy runner", `--manage-ios-runner`). Android needs its **own** guidance (rebuild/redeploy the `android/control-proxy` APK, or the Android equivalent override such as `AUTOMOBILE_CTRL_PROXY_APK_PATH` / `AUTOMOBILE_SKIP_ACCESSIBILITY_DOWNLOAD_IF_INSTALLED`) — do **not** reuse the iOS string verbatim. Consider a small shared `rewriteUnknownCommandError(error, platform)` helper rather than duplicating the regex.

### 2. Out-of-range numeric literal message not tested / not specifically mapped on Android
[...]

PR #3060, opened 2026-07-04
fix(android): route runner type:"error" frames into hierarchy wait path (#3032)

### LINKED ISSUE #3032: Android CtrlProxy error-envelope: hierarchy handler failures still hang the awaiter (type:"error" not routed into waitForFreshData)
## Summary

Follow-up from #2985 / PR #3019. That PR added a structured `type:"error"` frame from the Android CtrlProxy runner and wired `AndroidCtrlProxyClient.handleWebSocketMessage` to call `RequestManager.resolveError(requestId, …)` on it. This closes the hang for commands that await through `RequestManager`. But **`request_hierarchy` does not await through `RequestManager`**, so a hierarchy handler failure still hangs to timeout.

This is the un-addressed **P2** review comment from PR #3019 (chatgpt-codex-connector, on `src/features/observe/android/AndroidCtrlProxyClient.ts:1975`).

## Detail

- `CtrlProxyHierarchy.sendHierarchyRequest` (`src/features/observe/android/CtrlProxyHierarchy.ts:606`) generates and sends its **own** requestId and does **not** register a pending request with `RequestManager`.
- Hierarchy callers wait via `CtrlProxyHierarchy.waitForFreshData` (`CtrlProxyHierarchy.ts:527`) — they block until a fresh `hierarchy_update` push arrives, not on a `RequestManager` promise.
- The new `type:"error"` handling in `AndroidCtrlProxyClient.handleWebSocketMessage` (`AndroidCtrlProxyClient.ts:1969-1974`) only calls `this.requestManager.resolveError(message.requestId, …)`. Because the hierarchy request ID was never registered with `RequestManager`, `resolveError` **no-ops** and the branch returns.

Result: when the runner emits a `type:"error"` for a `request_hierarchy` handler failure, the hierarchy caller sits in `waitForFreshData` until its timeout — the exact slow-hang symptom #2985 set out to eliminate, on the hierarchy path.

## Impact

- A `request_hierarchy` decode/handler failure → `observe` / hierarchy sync hangs to `waitForFreshData` timeout instead of failing fast with the correlated error text.
- Scope-limited to the hierarchy request path (and any other command that uses a bespoke wait path instead of `RequestManager`).

## Proposed direction

Route runner `type:"error"` frames into the hierarchy wait path when the `requestId` correlates to an in-flight hierarchy request. Options:

1. Register hierarchy request IDs with `RequestManager` (or a parallel pending-hierarchy map) so `resolveError` can reject the correct waiter, and have `waitForFreshData` race that rejection; **or**
2. Have `CtrlProxyHierarchy` track its own outstanding requestId and expose a `rejectPendingHierarchy(requestId, error)` that `AndroidCtrlProxyClient` calls from the `type:"error"` branch (in addition to `RequestManager.resolveError`) so `waitForFreshData` unblocks with the error.

Either way the `type:"error"` branch must fan the error out to **all** wait mechanisms, not just `RequestManager`.

## Acceptance criteria

- [ ] A runner `type:"error"` frame whose `requestId` matches an in-flight hierarchy request rejects that hierarchy wait fast (no `waitForFreshData` timeout), surfacing the runner's error text.
- [ ] Non-correlatable error frames (null/unknown requestId) remain a safe no-op (current behavior preserved).
- [ ] Unit test: a `type:"error"` frame for an outstanding hierarchy requestId causes `requestHierarchySync` to reject/fail fast rather than time out.

## Context

- Parent: #2985, PR #3019 (merged, commit `00c4aea42`).
- Sibling async-path follow-up: #3023 (Kotlin-side async coroutine throws that never reach the synchronous handler `catch`). This issue is the **TS-consumer / hierarchy-correlation** half; #3023 is the **Kotlin async-broadcast** half. Together they cover the remaining hang classes after #3019.
- iOS-parity follow-up: #3022. Field-attributed decode legibility: #3017.
- Original review thread: PR #3019 inline comment on `AndroidCtrlProxyClient.ts:1975` ("Propagate hierarchy error frames").

kaeawc:
## What

Route the Android CtrlProxy runner's structured `type:"error"` frame into the **hierarchy** wait path so a `request_hierarchy` handler/decode failure fails fast instead of hanging to the `waitForFreshData` timeout.

Closes #3032. Follow-up to #2985 / PR #3019, which added the `type:"error"` envelope and wired `AndroidCtrlProxyClient.handleWebSocketMessage` to call `RequestManager.resolveError`. That closed the hang for commands that await through `RequestManager` — but **`request_hierarchy` does not**, so a hierarchy failure still hung. This is the un-addressed **P2** review comment from PR #3019 (on `AndroidCtrlProxyClient.ts:1975`).

Affected files:
- `src/features/observe/android/CtrlProxyHierarchy.ts` — track outstanding hierarchy request IDs + `rejectPendingHierarchy`; `waitForFreshData` races a runner-error rejection.
- `src/features/observe/android/AndroidCtrlProxyClient.ts` — `type:"error"` branch fans out to the hierarchy delegate in addition to `RequestManager`.
- `test/features/observe/CtrlProxyClient.test.ts` — new #3032 correlation tests.

## Why

`CtrlProxyHierarchy.sendHierarchyRequest` generates and sends its **own** requestId and does **not** register with `RequestManager`. Hierarchy callers block in `CtrlProxyHierarchy.waitForFreshData` until a fresh `hierarchy_update` push arrives — not on a `RequestManager` promise. So the `type:"error"` branch's `resolveError(requestId, …)` **no-op'd** for a hierarchy requestId, and the caller sat in `waitForFreshData` until timeout — the exact slow-hang #2985 set out to eliminate, on the hierarchy path.

## How

Implemented the issue's **option 2** (self-tracked pending map), keeping it local to the delegate that owns the bespoke wait:

- **`CtrlProxyHierarchy`** gains a `pendingHierarchyRejectors: Map<requestId, (error) => void>` and a public `rejectPendingHierarchy(requestId, error): boolean` — a safe no-op (`false`) on unknown/uncorrelated ids.
- **`sendHierarchyRequest`** now returns its `requestId` (was `boolean`) so `requestHierarchySync` can correlate the wait. `null` still signals the ADB-broadcast fallback.
- **`waitForFreshData`** takes an optional `requestId`; when present it registers a rejector that fails the wait fast when a runner error arrives. All settle paths (fresh / timeout / abort / screen-off / error) route through single-shot `settleResolve`/`settleReject` helpers that `cleanup()` the interval **and** the rejector — no leak, no double-settle.
- **`AndroidCtrlProxyClient`** `type:"error"` branch calls both `requestManager.resolveError(...)` and `this._hierarchy?.rejectPendingHierarchy(...)`. They key on disjoint maps (a hierarchy id is never in `RequestManager.pending`; a gesture id is never in `pendingHierarchyRejectors`), so there is no double-resolution and each is a no-op on the other's ids.

**Preserved behavior:** `getLatestHierarchy`'s passive wait calls `waitForFreshData` **without** a requestId, so it never registers a rejector and is byte-for-byte unchanged. The ADB-broadcast fallback (no WS requestId) deliberately retains timeout-only behavior. The `null` return of `requestHierarchySync` is unchanged — callers just get it fast now instead of after 10s.

**Scope note:** the mid-wait `request_hierarchy_if_stale` nudge uses an uncorrelated `stale_` id, so an error on *that* frame still waits out the timeout — outside this issue's acceptance criteria (which targets the primary `request_hierarchy` path). The iOS-parity half is tracked separately in #3022; the Kotlin async-broadcast half in #3023.

## Testing

All no-device (bun + Interface/Fake/FakeTimer), following repo conventions.

- `CtrlProxyClient.test.ts` (new `issue #3032` suite):
  - A `type:"error"` frame whose requestId matches an in-flight hierarchy request fails `requestHierarchySync` **fast** — asserted by resolving with no timer advance and `getCurrentTime() < 10000` (proves it does not sit through the 10s timeout). This test **hangs to timeout without the fix** (TDD red confirmed).
  - A `type:"error"` frame with an **unknown** requestId is a safe no-op: an in-flight hierarchy sync still resolves normally from its own `hierarchy_update`.
- Regression guard: the existing #3019 `error frame handling` suite (RequestManager path) still passes — 4 tests / 146ms.
- `bun test test/features/observe/` → 769 pass, 1 fail. The single failure (`getLatestHierarchy > should preserve SDK screen names …`) is **pre-existing** — it fails on unmodified `main` in isolation, and is untouched here.
- `bun build.ts` and `eslint` on the changed files are green.

> **Delivery note:** this is a pure TS/daemon-side change — no runner re-cut required. It takes effect against any runner that already emits the `type:"error"` frame from #3019.

### DIFF
--- a/src/features/observe/android/AndroidCtrlProxyClient.ts
+++ b/src/features/observe/android/AndroidCtrlProxyClient.ts
@@ -1966,11 +1966,17 @@ export class AndroidCtrlProxyClient extends DeviceServiceClient implements Andro
       // awaiter hang to timeout. requestId is best-effort — when null the runner could not
       // correlate it, so there is no pending request to resolve (resolveError no-ops on unknown
       // ids anyway).
+      //
+      // Fan the error out to ALL wait mechanisms, not just RequestManager: request_hierarchy does
+      // not await through RequestManager, so an error frame for a hierarchy requestId must also be
+      // routed into CtrlProxyHierarchy's bespoke wait or the hierarchy caller hangs to timeout
+      // (issue #3032). Both calls are safe no-ops on unknown ids.
       if (message.type === "error") {
         const errorText = message.error || "Runner reported an unstructured protocol error";
         logger.warn(`[CTRL_PROXY] Runner error (requestId: ${message.requestId ?? "none"}): ${errorText}`);
         if (message.requestId) {
           this.requestManager.resolveError(message.requestId, errorText);
+          this._hierarchy?.rejectPendingHierarchy(message.requestId, errorText);
         }
         return;
       }
--- a/src/features/observe/android/CtrlProxyHierarchy.ts
+++ b/src/features/observe/android/CtrlProxyHierarchy.ts
@@ -57,7 +57,11 @@ export class CtrlProxyHierarchy {
   // NOT await through RequestManager (it blocks in waitForFreshData for a hierarchy_update push), so
   // a runner type:"error" frame must be fanned into this map to unblock the correct waiter fast
   // instead of hanging to timeout. See issue #3032.
-  private readonly pendingHierarchyRejectors = new Map<string, (error: string) => void>();
+  //
+  // The hook is wrapped in an object and invoked via the fixed `.reject` property (mirroring
+  // RequestManager's `request.reject(...)`) rather than calling the map value directly — a
+  // user-controlled requestId must never drive a dynamic method-name dispatch.
+  private readonly pendingHierarchyRejectors = new Map<string, { reject: (error: string) => void }>();
 
   constructor(context: HierarchyDelegateContext) {
     this.context = context;
@@ -76,7 +80,7 @@ export class CtrlProxyHierarchy {
     if (!rejector) {
       return false;
     }
-    rejector(error);
+    rejector.reject(error);
     return true;
   }
 
@@ -598,9 +602,11 @@ export class CtrlProxyHierarchy {
       // Route a runner type:"error" frame (delivered via AndroidCtrlProxyClient) for this hierarchy
       // request into a fast rejection instead of hanging to the timeout below (issue #3032).
       if (requestId) {
-        this.pendingHierarchyRejectors.set(requestId, (error: string) => {
-          logger.warn(`[CTRL_PROXY] Hierarchy request ${requestId} failed via runner error: ${error}`);
-          settleReject(new Error(error));
+        this.pendingHierarchyRejectors.set(requestId, {
+          reject: (error: string) => {
+            logger.warn(`[CTRL_PROXY] Hierarchy request ${requestId} failed via runner error: ${error}`);
+            settleReject(new Error(error));
+          }
         });
       }
 
--- a/src/features/observe/android/CtrlProxyHierarchy.ts
+++ b/src/features/observe/android/CtrlProxyHierarchy.ts
@@ -53,8 +53,31 @@ export class CtrlProxyHierarchy {
   private recompositionTrackingConfigured: boolean = false;
   private recompositionTrackingEnabled: boolean = false;
 
+  // Outstanding hierarchy request IDs mapped to their error-rejection hook. request_hierarchy does
+  // NOT await through RequestManager (it blocks in waitForFreshData for a hierarchy_update push), so
+  // a runner type:"error" frame must be fanned into this map to unblock the correct waiter fast
+  // instead of hanging to timeout. See issue #3032.
+  private readonly pendingHierarchyRejectors = new Map<string, (error: string) => void>();
+
   constructor(context: HierarchyDelegateContext) {
     this.context = context;
+  }
+
+  /**
+   * Reject an in-flight hierarchy wait whose requestId matches a runner type:"error" frame,
+   * surfacing the runner's error text so the caller fails fast instead of hanging to the
+   * waitForFreshData timeout (issue #3032).
+   *
+   * Returns false (safe no-op) when the id is not an outstanding hierarchy request — e.g. a
+   * null/unknown requestId the runner could not correlate — preserving existing behavior.
+   */
+  rejectPendingHierarchy(requestId: string, error: string): boolean {
+    const rejector = this.pendingHierarchyRejectors.get(requestId);
+    if (!rejector) {
+      return false;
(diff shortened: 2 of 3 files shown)


kaeawc:
## Review + hardening pass

Ran the AutoMobile code-review skill plus three independent review subagents (concurrency/race auditor, regression hunter, architecture/direction reviewer). Summary:

- **Concurrency:** no rejector leak (every settle path funnels through `cleanup()`), no double-settle (synchronous `settled` guard), concurrent syncs stay isolated by unique requestId, no unhandled rejection escapes (`getLatestHierarchy` never registers a rejector; `requestHierarchySync` catches), no TDZ, and `resolveError`/`rejectPendingHierarchy` operate on disjoint id namespaces so no double-resolution.
- **Regression:** all pre-existing `waitForFreshData` paths preserved; `sendHierarchyRequest`'s `boolean`→`string|null` change is truthiness-equivalent (its one caller switched to `=== null`); the `null` return contract of `requestHierarchySync` is unchanged (callers treat runner-error-`null` identically to timeout-`null`).
- **Direction:** Option 2 (self-tracked map) is the right fit — the hierarchy success path can't use `RequestManager.resolve` because the `hierarchy_update` push carries no requestId. Convention-conformant.

### CodeQL fix (pushed)
CodeQL's `js/unvalidated-dynamic-method-call` flagged invoking the map value directly (`rejector(error)`) since the `requestId` key is runner-controlled. Reworked to store a `{ reject }` wrapper and invoke via the fixed `.reject` property — mirroring `RequestManager.request.reject(...)`, which is why that identical `Map.get(requestId)` pattern was never flagged. Behavior unchanged.

### Note on acceptance criteria
"Surfacing the runner's error text" means it reaches the **logs** and fails fast — `requestHierarchySync` still returns `null` (its existing contract), not the text, so callers can't yet distinguish handler-failure from timeout. Propagating the text upward is a larger contract change; filing as a follow-up.

### Deferred follow-ups (issues to file)
1. Correlate `request_hierarchy_if_stale` (`stale_` id) error frames — sibling command still hangs to timeout on a runner error (same class as #3032).
2. Optionally let `requestHierarchySync` return the runner error text instead of collapsing to `null`.

_Unrelated: the `getLatestHierarchy > should preserve SDK screen names` test fails on unmodified `main` too — pre-existing, not introduced here._

PR #3066, opened 2026-07-04
test(observe): deterministically preserve SDK screen names via in-memory nav DB (#3063)

### LINKED ISSUE #3063: test(observe): "getLatestHierarchy > should preserve SDK screen names when hierarchy updates follow navigation events" fails on main
Discovered while validating PR #3060 (issue #3032) — **not** caused by that change.

## Summary
`test/features/observe/CtrlProxyClient.test.ts:581` — `AndroidCtrlProxyClient > getLatestHierarchy > should preserve SDK screen names when hierarchy updates follow navigation events` — fails deterministically:

```
expect(received).toBe(expected)
Expected: "SdkHome"
Received: null
    at test/features/observe/CtrlProxyClient.test.ts:581
```

## Pre-existing, verified
Reproduced on **unmodified `origin/main`**: with all of PR #3060's changed files reverted (or the new describe block removed), the test fails identically. It fails both in-suite and in isolation (`bun test test/features/observe/CtrlProxyClient.test.ts -t "preserve SDK screen names"`), so it is a deterministic failure, not an order-dependent flake. `bun test test/features/observe/` = 769 pass / **1 fail** (this test).

It exercises `getLatestHierarchy`'s navigation-name seeding path (hierarchy update following a navigation event → expects the SDK screen name `SdkHome` to be preserved on the NavigationGraph), which the #3032 change does not touch.

## Why it matters
A red test on `main` erodes signal — future PRs (including #3060) show "1 fail" and authors must re-derive that it is pre-existing every time.

## Ask
Investigate whether the assertion or the underlying nav-name-preservation behavior regressed. Likely candidates: a change to `NavigationGraphManager` screen-naming (named-nodes-only feature — note the adjacent test at ~line 519 comments that "hierarchy updates alone don't create screens"), or the SDK-screen-name seeding path in `handleHierarchyUpdate`. Either fix the behavior or update/quarantine the test.

## Context
- Surfaced during #3060 review; independent of #3032.
- File: `test/features/observe/CtrlProxyClient.test.ts:532-581`.

kaeawc:
## Summary

Closes #3063. The test `AndroidCtrlProxyClient > getLatestHierarchy > should preserve SDK screen names when hierarchy updates follow navigation events` failed **deterministically on `main`** (`Expected "SdkHome", Received null`).

Root cause is **test determinism, not a behavior regression.** `currentScreen` is set by `recordNavigationEvent` only *after* its async DB writes commit (`setCurrentApp` → `getOrCreateApp`, then `runInTransaction`). The test drove the singleton `NavigationGraphManager.getInstance()`, which is backed by the **real file database** (`getDatabase()`), whose first-use path runs migrations + file IO on **real wall-clock time**. After awaiting the hierarchy promise the test only yielded microtasks (`advanceTimersByTimeAsync(1)`), so the navigation write had not landed when the assertion ran → `currentScreen` was still `null`.

Confirmed it is a race, not a regression:
- A direct `recordNavigationEvent({ destination: "SdkHome", applicationId: "com.example.sdk" })` on a fresh singleton sets `currentScreen = "SdkHome"` — production behavior is correct.
- Replacing the microtask yield with a real `setTimeout(1500)` makes the test pass.
- A trace showed the handler reaching `recordNavigationEvent` and blocking inside `setCurrentApp` → `getOrCreateApp` (the real-DB async write) until real time elapsed.

The sibling test `should seed navigation graph from hierarchy updates` only passed because it asserts the `null` **pre-write** state — it was racy-by-luck against the same real DB.

## What changed

- `src/features/navigation/NavigationGraphManager.ts` — added a **test-only** `static setInstanceForTesting(instance)` (documented, mirrors `resetInstance` / `createForTesting`) so a test can back the singleton with an in-memory manager. **No production behavior change.**
- `test/features/observe/CtrlProxyClient.test.ts` — added an `installInMemoryNavManager()` helper that builds a `:memory:` DB via `createTestDatabase`, wires `NavigationRepository` + `TestCoverageRepository` on the **same** connection (satisfies `assertSharedConnection`), and registers it as the singleton. Rewrote the failing test **and its racy sibling** to use it, drain deterministically (`setImmediate` + `advanceTimersByTimeAsync`), and destroy the DB in cleanup. The in-memory DB has no migration gate and no file IO, so `recordNavigationEvent`'s writes settle within a deterministic number of microtask/immediate turns.

## Verification

- `bun test test/features/observe/CtrlProxyClient.test.ts` → **23 pass / 0 fail** (was 22 pass / 1 fail). Stable across **5 consecutive runs**.
- `bun test test/features/observe/` → all pass.
- `bun test test/features/navigation/ test/db/` → 833 pass / 0 fail (new test-only helper causes no regression).
- `turbo run lint` → clean. `bun build.ts` → OK.
- No wall-clock sleeps; determinism comes from the injected in-memory DB + microtask drain.

## Non-goals / follow-ups

- Other tests in `CtrlProxyClient.test.ts` (and elsewhere) that drive the singleton `NavigationGraphManager` against the real file DB but only assert pre-write state are not affected today; a broader sweep to route all such tests through an injected in-memory DB could be a follow-up.
- The post-commit fire-and-forget `TelemetryRecorder` write still targets the real DB (unchanged, pre-existing); not awaited, so it does not affect determinism here.

### DIFF
--- a/src/features/navigation/NavigationGraphManager.ts
+++ b/src/features/navigation/NavigationGraphManager.ts
@@ -202,6 +202,20 @@ export class NavigationGraphManager implements NavigationGraphService {
   }
 
   /**
+   * Install a pre-built instance as the singleton (for testing only).
+   *
+   * Lets a test back the singleton with an in-memory database so consumers that
+   * resolve the manager via `getInstance()` (e.g. AndroidCtrlProxyClient) exercise
+   * deterministic, migration-gate-free DB writes instead of the real file DB. Using
+   * `getInstance()` alone would lazily build a manager on the shared `getDatabase()`
+   * singleton, whose first-use migration + file IO run on real wall-clock time and
+   * make async writes race the assertions. Pair with `resetInstance()` in teardown.
+   */
+  public static setInstanceForTesting(instance: NavigationGraphManager): void {
+    NavigationGraphManager.instance = instance;
+  }
+
+  /**
    * Create a new instance for testing with injected dependencies.
    *
    * Precondition for recordNavigationEvent's atomicity: `repository` and
--- a/test/features/observe/CtrlProxyClient.test.ts
+++ b/test/features/observe/CtrlProxyClient.test.ts
@@ -144,10 +144,14 @@ describe("AndroidCtrlProxyClient", function() {
   // getInstance(); the default getDatabase() singleton runs migrations + file IO on
   // real wall-clock time, so its async writes never settle within these tests'
   // microtask-only drains and race the assertions (issue #3063). An in-memory DB has
-  // no migration gate and no file IO, so recordNavigationEvent's writes commit within
-  // a deterministic number of setImmediate/microtask turns. Both repositories share
-  // the one connection to satisfy NavigationGraphManager's shared-connection
-  // precondition. Callers must destroy the returned db and resetInstance() in cleanup.
+  // no migration gate and no file IO, so recordNavigationEvent's *graph* writes — the
+  // ones that assign currentScreen, which the assertions read — commit within a
+  // deterministic number of setImmediate/microtask turns. (recordNavigationEvent's
+  // post-commit fire-and-forget TelemetryRecorder write still targets the real DB, but
+  // it runs after currentScreen is assigned and is never awaited, so it cannot affect
+  // determinism here.) Both repositories share the one connection to satisfy
+  // NavigationGraphManager's shared-connection precondition. Callers must destroy the
+  // returned db and resetInstance() in cleanup.
   const installInMemoryNavManager = async (): Promise<{
     navManager: NavigationGraphManager;
     navDb: Kysely<Database>;
--- a/test/features/observe/CtrlProxyClient.test.ts
+++ b/test/features/observe/CtrlProxyClient.test.ts
@@ -15,6 +15,11 @@ import { FakeInstalledAppsRepository } from "../../fakes/FakeInstalledAppsReposi
 import { FakeTimer } from "../../fakes/FakeTimer";
 import type { DeviceConnectionLostNotifier } from "../../../src/features/observe/DeviceConnectionLostNotifier";
 import { PortManager } from "../../../src/utils/PortManager";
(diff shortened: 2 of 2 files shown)


kaeawc:
### Multi-perspective review — all three reviewers say merge

Ran three independent review passes (determinism, API-surface/architecture, correctness/completeness). Consensus: **merge as-is**; all findings are optional/non-blocking.

- **Determinism:** deterministic — the graph write settles at the *first* `setImmediate` boundary (bun:sqlite executes synchronously; the in-memory DB has no migration gate/file IO). Could not reproduce a flake in 60+ runs, including under 8-core CPU saturation. The 10-iteration drain is large, machine-independent headroom.
- **Architecture:** `NavigationGraphManager.setInstanceForTesting()` is the minimal-surface seam and is directly precedented by `setAppFileServiceForTesting`. Alternatives (session-binding, global `getDatabase()` override, DI into the 8-arg `createForTesting`) all add *more* production surface or a larger blast radius.
- **Correctness:** production nav-name preservation is correct and unchanged — for an SDK app the `hierarchy_update` path is skipped (`sdkNavigationAppIds`) and, even if that regressed, `setCurrentApp`'s idempotent early-return protects `currentScreen`. `git blame` confirms the load-bearing lines are months old and untouched by this PR.

Addressed inline: clarified the helper comment so it no longer overclaims "no file IO" — the *graph* writes are in-memory; `recordNavigationEvent`'s post-commit fire-and-forget `TelemetryRecorder` write still targets the real DB but runs after `currentScreen` is assigned and is never awaited, so it can't affect determinism.

Filed as follow-ups (deferred, non-blocking):
- #3067 — test-infra guard so unit tests fail loudly when they resolve the real file-backed `getDatabase()` (durable fix for the whole real-DB-race flake class; also covers isolating the fire-and-forget telemetry write).
- #3068 — optionally pin the `sdkNavigationAppIds` skip mechanism in the preservation test (the test verifies the outcome but not that specific mechanism; pre-existing blind spot).

PR #3070, opened 2026-07-04
fix(android): correlate request_hierarchy_if_stale error frames into the hierarchy wait (#3061)

### LINKED ISSUE #3061: Android CtrlProxy: correlate request_hierarchy_if_stale (stale_ id) error frames into the hierarchy wait
Follow-up to #3032 / PR #3060.

## Summary
#3032 / PR #3060 routed runner `type:"error"` frames into the hierarchy wait for the **primary** `request_hierarchy` path: `CtrlProxyHierarchy.sendHierarchyRequest` now returns its `req_<ts>_<id>` requestId, `requestHierarchySync` correlates the `waitForFreshData` wait to it via `pendingHierarchyRejectors`, and `AndroidCtrlProxyClient` fans a `type:"error"` frame into `rejectPendingHierarchy`. A hierarchy handler failure on that path now fails fast instead of hanging to timeout.

**But the sibling `request_hierarchy_if_stale` nudge is still uncorrelated.**

## Detail
`CtrlProxyHierarchy.sendHierarchyIfStaleRequest` mints a `stale_<ts>_<id>` requestId and sends a `request_hierarchy_if_stale` frame mid-wait (the "no push after 2s" nudge inside `waitForFreshData`). That id is **never** inserted into `pendingHierarchyRejectors`, so a runner `type:"error"` frame whose `requestId` is the `stale_` id has nothing to reject — `rejectPendingHierarchy` returns `false` (safe no-op) and the wait sits until its timeout. This is the exact hang class #3032 set out to close, on a sibling command.

Also inherent (documented, not necessarily fixable here): the **ADB-broadcast fallback** path (when `sendHierarchyRequest` returns `null` because the WebSocket is down) has no WebSocket requestId to correlate, so it retains timeout-only behavior.

## Impact
Low/moderate: the `stale_` nudge is best-effort and only fires after a 2s no-push window, and the primary path is already fixed. But a decode/handler failure specifically on the `request_hierarchy_if_stale` command still degrades to a full `waitForFreshData` timeout rather than a fast, correlated failure.

## Proposed direction
Track the `stale_` requestId the same way as the primary path: have `sendHierarchyIfStaleRequest` return its id (currently `boolean`) and register a rejector in `pendingHierarchyRejectors` scoped to the in-flight `waitForFreshData` promise (so an error frame for the stale id rejects the same wait). Care needed because the stale nudge is sent from *inside* the `waitForFreshData` interval callback, not before it — the rejector for the stale id would need to reject the enclosing wait, so it may be simplest to register the stale id against the same `settleReject` closure already created for the primary request.

## Acceptance criteria
- [ ] A runner `type:"error"` frame whose `requestId` matches an in-flight `request_hierarchy_if_stale` nudge rejects/fails the enclosing hierarchy wait fast (no `waitForFreshData` timeout).
- [ ] Non-correlatable / null requestId frames remain a safe no-op (preserve current behavior).
- [ ] Unit test mirroring the #3032 tests in `test/features/observe/CtrlProxyClient.test.ts` but driving the `stale_` frame.

## Context
- Parent: #3032 (PR #3060). Sibling async-broadcast half: #3023. iOS parity: #3022.
- Key files: `src/features/observe/android/CtrlProxyHierarchy.ts` (`sendHierarchyIfStaleRequest`, `waitForFreshData`, `pendingHierarchyRejectors`, `rejectPendingHierarchy`), `src/features/observe/android/AndroidCtrlProxyClient.ts` (`type:"error"` branch).

kaeawc:
## Summary

Follow-up to #3032 / PR #3060. That PR routed runner `type:"error"` frames into the hierarchy wait for the **primary** `request_hierarchy` path, but left the sibling `request_hierarchy_if_stale` nudge **uncorrelated**. This closes that gap.

`CtrlProxyHierarchy.sendHierarchyIfStaleRequest` mints a `stale_<ts>_<id>` requestId and sends the nudge from **inside** `waitForFreshData`'s interval callback (the "no push after 2s" path). That id was never inserted into `pendingHierarchyRejectors`, so a runner `type:"error"` frame whose `requestId` was the `stale_` id had nothing to reject — `rejectPendingHierarchy` returned `false` (safe no-op) and the wait sat until its full timeout. This is the exact hang class #3032 set out to close, on a sibling command.

## Changes

- `sendHierarchyIfStaleRequest` now returns its `stale_` requestId (`string | null`, was `boolean`), mirroring `sendHierarchyRequest`'s contract from #3032.
- `waitForFreshData` captures that id when the nudge fires and registers a rejector against the **same** `settleReject` closure as the primary request, so an error frame for the stale id rejects **this exact** wait fast. `cleanup()` unregisters the stale id too. The hook is wrapped in a fixed `.reject` property (as the primary path is) so a runner-controlled id can't drive a dynamic method-name dispatch (CodeQL `js/unvalidated-dynamic-method-call`).
- No change to `AndroidCtrlProxyClient` — its `type:"error"` branch already fans into `rejectPendingHierarchy`, which now finds the stale id in the map.

The ADB-broadcast fallback path has no WebSocket requestId to correlate and retains its documented timeout-only behavior (called out in the issue as inherent).

## Acceptance criteria

- [x] A runner `type:"error"` frame whose `requestId` matches an in-flight `request_hierarchy_if_stale` nudge rejects/fails the enclosing hierarchy wait fast (no `waitForFreshData` timeout).
- [x] Non-correlatable / null requestId frames remain a safe no-op (preserved).
- [x] Unit tests mirroring the #3032 tests in `test/features/observe/CtrlProxyClient.test.ts`, driving the `stale_` frame (positive fast-fail + uncorrelated-id no-op).

## Testing

- `bun test test/features/observe/CtrlProxyClient.test.ts` — 27 pass (2 new).
- `turbo run lint build` — clean.
- `scripts/all_fast_validate_checks.sh` — 11/11 pass.

Closes #3061

## Related

- Parent: #3032 (PR #3060). Sibling async-broadcast half: #3023. iOS parity: #3022.

### DIFF
--- a/src/features/observe/android/CtrlProxyHierarchy.ts
+++ b/src/features/observe/android/CtrlProxyHierarchy.ts
@@ -605,15 +605,25 @@ export class CtrlProxyHierarchy {
         reject(error);
       };
 
-      // Route a runner type:"error" frame (delivered via AndroidCtrlProxyClient) for this hierarchy
-      // request into a fast rejection instead of hanging to the timeout below (issue #3032).
-      if (requestId) {
-        this.pendingHierarchyRejectors.set(requestId, {
+      // Register a fast-fail rejector for a hierarchy requestId so a runner type:"error" frame
+      // (fanned in via AndroidCtrlProxyClient.rejectPendingHierarchy) rejects THIS wait instead of
+      // hanging to the timeout below. The hook is wrapped in a fixed `.reject` property so a
+      // runner-controlled id can never drive a dynamic method-name dispatch
+      // (CodeQL js/unvalidated-dynamic-method-call). Shared by the primary request_hierarchy path
+      // (issue #3032) and the request_hierarchy_if_stale nudge (issue #3061).
+      const registerRejector = (id: string, label: string): void => {
+        this.pendingHierarchyRejectors.set(id, {
           reject: (error: string) => {
-            logger.warn(`[CTRL_PROXY] Hierarchy request ${requestId} failed via runner error: ${error}`);
+            logger.warn(`[CTRL_PROXY] ${label} ${id} failed via runner error: ${error}`);
             settleReject(new Error(error));
           }
         });
+      };
+
+      // Route a runner type:"error" frame (delivered via AndroidCtrlProxyClient) for this hierarchy
+      // request into a fast rejection instead of hanging to the timeout below (issue #3032).
+      if (requestId) {
+        registerRejector(requestId, "Hierarchy request");
       }
 
       intervalId = this.context.timer.setInterval(() => {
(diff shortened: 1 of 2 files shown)


kaeawc:
## Multi-perspective review round (addressed in ee2d116)

Ran three independent review passes (concurrency/lifecycle, test-quality, completeness/direction) plus the AutoMobile code-review skill against the branch diff. Findings and resolutions:

- **Latent regression — stale-nudge rejection could discard stale cache (concurrency pass).** `getLatestHierarchy` enters `waitForFreshData` with **no** primary `requestId`; its timeout is meant to gracefully fall through to the stale cache. Registering the stale-nudge rejector there unconditionally meant a runner `type:"error"` frame for the `stale_` id would *reject* the wait → propagate to the function-level catch → return `hierarchy:null`, discarding an available cache. **Fix:** gate stale-nudge correlation on `requestId` presence — only fast-fail-intended waits (a primary WebSocket `request_hierarchy` outstanding, i.e. `requestHierarchySync`) correlate the nudge. `getLatestHierarchy` and the sync ADB-broadcast fallback keep their documented timeout→stale-cache behavior. Added a regression test that fails if the `&& requestId` gate is removed (verified: returns `null` without the gate, cache preserved with it).

- **Duplication (direction pass).** Extracted the two structurally-identical `pendingHierarchyRejectors.set(id, { reject })` blocks (primary + stale) into a local `registerRejector(id, label)` lambda; both sites now share the CodeQL-safe fixed `.reject` wrapper by construction.

- **Test flake vector (direction pass).** The stale tests advanced a fixed amount then assumed the interval had registered — a microtask-ordering flake (one macOS first-run timeout was observed). Replaced with a bounded `driveUntilStaleNudge()` retry that advances until the nudge is actually sent, keeping total advance under the sync timeout so the "no timeout occurred" assertion stays valid.

- **No other hang-class paths missed.** `recomp_` (recomposition tracking) is fire-and-forget and never awaits `waitForFreshData`, so it's correctly excluded; `sync_` (ADB broadcast) is the inherent, documented timeout-only path.

- **Test-quality pass** empirically confirmed the positive test is a true regression guard (removing the fix makes it time out).

28 CtrlProxyClient tests pass (3× no flakes); `turbo lint build` clean; `scripts/all_fast_validate_checks.sh` 11/11.

> Note: the macOS "Node TypeScript Build and Test" failure on the prior run was an **unrelated** pre-existing flake — `IOSCtrlProxyBuilder.test.ts:734` timed out; this PR touches only Android CtrlProxy hierarchy code.

PR #3071, opened 2026-07-04
fix(db): guard against production AUTOMOBILE_DB_PATH=:memory: (#3065)

### LINKED ISSUE #3047: Move dbWriteBarrierResetOnClose barrier test to a :memory: sentinel DB (eliminate the file-handle flake class)
## Summary
Follow-up from the review of #3040 (closes #2992). #3040 stabilized `dbWriteBarrierResetOnClose.test.ts` by keeping a real temp file and awaiting migrations before close. That removes the *race*, but the test still opens a real file (temp-dir cleanup, WAL sidecars, EBUSY-tolerant retry). For **this specific test** the file is not actually needed, so a `:memory:` sentinel would eliminate the file-handle flake class for it entirely.

## Why the barrier test doesn't need a real file
The "…NOT skipped" test's assertions are purely about `getDbWriteBarrier()` **identity** and `isDraining()` after `closeDatabase()` (`resetDbWriteBarrier()` cold-start, #2896). It exercises the `getDatabase()`/`closeDatabase()` **lifecycle**, but needs neither cross-connection migration *visibility* nor even a completed migration. Precedent in-repo: the sibling `dbWriteBarrierReopenConsumers.test.ts` (#2912) tests the same reset/reopen contract with a `FakeDbWriteBarrier` and no `database.ts`, and this file's own second test opens no DB at all. (Note: #3040 additionally added a real reopen query inside the tracked closure to validate schema health — that assertion would need `:memory:` migrations to still create `tool_calls`, which they do per-connection, but the app connection wouldn't see the migration connection's private `:memory:` DB. So this migration also needs to decide whether to keep that end-to-end query or drop it for this test.)

## Blocker: sentinel path handling
`resolveDatabasePathFromEnvironment()` routes `AUTOMOBILE_DB_PATH` through `resolvePathFromDaemonLaunchWorkingDirectory()`, which would `path.resolve(":memory:")` into a bogus absolute path, and `createFileMigrationLock()` would try to create `:memory:.migrate.lock`. So this needs a small source change to pass sentinel DB paths (`:memory:`, and decide on `file::memory:?cache=shared`) through un-resolved and to use `NoOpMigrationLock` for them.

`file::memory:?cache=shared` was considered and rejected (reintroduces the cross-process migration lock concern + bun URI-filename risk for no benefit here). Prefer plain `:memory:`.

## Scope
- Add sentinel-path passthrough to `resolveDatabasePathFromEnvironment()` + `NoOpMigrationLock` selection for `:memory:`.
- Point the barrier "…NOT skipped" test at `:memory:` and reconcile the reopen-query assertion added in #3040.
- Leave `databaseLazyPath` / `databaseReset` on real files (they genuinely assert migrated schema across connections).

## References
- Surfaced by the direction/alternatives review of #3040.
- Context: `test/db/tempDbDir.ts` (why file-backed today), #2896, #2912, #2849.

### LINKED ISSUE #3065: Guard against production AUTOMOBILE_DB_PATH=:memory: (migrated-but-empty app connection footgun)
## Summary
Follow-up from PR #3058 (closed #3047) and the Codex P2 review on it.

#3058 added a `:memory:` sentinel DB path (`IN_MEMORY_DATABASE_PATH` / `isInMemoryDatabasePath` / `createMigrationLock` in `src/db/migrationLock.ts`, threaded into `src/db/database.ts`) so the barrier-reset test can be flake-free without a temp file. The sentinel is **test-only by design** and is now documented as such at the sentinel definition.

But there is no **runtime guard**: if a user sets `AUTOMOBILE_DB_PATH=:memory:` on a real daemon, `ensureMigrations()` reports success while the application connection has **none** of the migrated tables, so the first schema-dependent read/write (`tool_calls`, etc.) fails with `no such table` — a silently-broken DB rather than a clear error.

## Why this happens (root cause)
`:memory:` is private per connection. `getDatabase()` opens its own `new Database(":memory:")`, while `startMigrations()` runs migrations on a **separate** `migrationDb` Kysely instance (`src/db/database.ts` `startMigrations`). The two are distinct empty in-memory DBs. This separation is deliberate for real files (it gives migration failure-isolation), but for `:memory:` it means the app connection never sees the migrated schema.

Note this is also a subtle **behavior change**: before #3058, `AUTOMOBILE_DB_PATH=:memory:` was `path.resolve`'d into a literal file named `:memory:` (a working, if oddly-named, file DB on POSIX; illegal on Windows). After #3058 it is a private in-memory DB that reports migrated-but-empty on the app connection.

## Options (decide in this issue)
1. **Runtime guard (recommended):** reject or loudly warn when `:memory:` is resolved outside a test context — e.g. throw an `ActionableError` from `getDatabase()`/`resolveDatabasePathFromEnvironment()` for `:memory:` unless an explicit test opt-in env/flag is set. Keeps the test seam while making production misuse fail fast and legibly.
2. **Share one connection for `:memory:`:** inject the app connection into the migrator so migrations land on the same in-memory DB. Rejected in #3058 as out-of-scope — it would break the migration failure-isolation the separate `migrationDb` provides — but re-evaluate if a production `:memory:` use case ever emerges.
3. **Leave as documented-only** (status quo): accept the footgun as sufficiently unlikely (no production path sets `:memory:`) and rely on the docstring.

## Secondary nit (optional, from the architecture review)
`createMigrationLock(dbPath)` (dispatcher) vs `createFileMigrationLock(dbPath)` (one branch) don't self-signal which is the chooser. Consider renaming the dispatcher to `selectMigrationLock` / `migrationLockFor` for clarity. Cosmetic; bundle only if touching this area.

## References
- PR #3058 (merged), issue #3047, Codex P2 inline comment on `src/db/database.ts:74`.
- `src/db/migrationLock.ts` `IN_MEMORY_DATABASE_PATH` docstring (documents the test-only limitation).
- `src/db/database.ts` `startMigrations` (separate `migrationDb`), `resolveDatabasePathFromEnvironment` (`:memory:` passthrough).

kaeawc:
## What

Add a **runtime guard** that rejects a production `AUTOMOBILE_DB_PATH=:memory:` with a clear `ActionableError` instead of silently running against a migrated-but-empty database.

Closes #3065. Follow-up to PR #3058 (closed #3047) and its Codex P2 review comment on `src/db/database.ts:74`.

Affected files:
- `src/db/migrationLock.ts` — new `AUTOMOBILE_ALLOW_IN_MEMORY_DB` opt-in (`IN_MEMORY_DB_OPT_IN_ENV` + `isInMemoryDatabaseOptInEnabled`); dispatcher `createMigrationLock` → `selectMigrationLock` (issue's secondary nit).
- `src/db/database.ts` — guard in `resolveDatabasePathFromEnvironment`; import rename.
- `test/db/databasePath.test.ts`, `test/db/migrationLock.test.ts`, `test/db/dbWriteBarrierResetOnClose.test.ts` — guard + opt-in coverage; existing `:memory:` lifecycle tests opt in.

## Why

A SQLite `:memory:` DB is **private per connection**. `getDatabase()` opens its own `new Database(":memory:")`, while `startMigrations()` runs migrations on a **separate** `migrationDb` Kysely instance — two distinct empty in-memory DBs. This separation is deliberate for real files (migration failure-isolation), but for `:memory:` it means the app connection never sees the migrated schema.

Before #3058, `AUTOMOBILE_DB_PATH=:memory:` was `path.resolve`'d into a literal file named `:memory:` (a working, oddly-named file DB on POSIX). After #3058 the `:memory:` sentinel is a private in-memory DB, so `ensureMigrations()` reports **success** while the daemon's app connection has **none** of the migrated tables — the first schema-dependent read/write (`tool_calls`, …) fails with `no such table`, a silently-broken DB rather than a clear error. The sentinel is documented test-only, but there was no runtime backstop.

## How

Implemented the issue's **option 1 (runtime guard)** — the recommended choice — with the test seam preserved via an explicit opt-in:

- **Single choke point.** The guard lives in `resolveDatabasePathFromEnvironment()`, the one place `:memory:` is passed through. Both `getDatabase()` and `ensureMigrations()` reach it via `resolveDbPath()`, and the daemon-startup DB touchpoints (`directModeGuard` / `daemon.ts` `getDatabasePath()`) reach it too — so production misuse fails fast at bring-up, before any query.
- **Explicit opt-in, fail-safe.** `AUTOMOBILE_ALLOW_IN_MEMORY_DB` must be truthy (`1`/`true`/`yes`, case/whitespace-insensitive) for `:memory:` to pass. Anything else — absent, empty, `0`/`false` — is treated as disabled, so a typo'd flag never silently re-opens the footgun. An explicit env flag is preferred over auto-detecting the test runner, which a stray `NODE_ENV=test` could leak into production.
- **Actionable error.** Names the misused path, explains the private-per-connection cause and the migrated-but-empty consequence, and gives the fix (point at a real file / unset it, or set the opt-in from a test).
- **Secondary nit.** Renamed the migration-lock dispatcher `createMigrationLock` → `selectMigrationLock` so it self-signals as the chooser vs `createFileMigrationLock` (the branch). The issue explicitly invited this when touching the area.

**Preserved behavior:** real-file path resolution is untouched (verified with and without the flag). The `:memory:` sentinel remains fully usable for lifecycle tests — they now opt in explicitly. The separate-`migrationDb` failure-isolation (option 2, sharing one connection) was left out of scope, as in #3058.

## Testing

All no-device (bun, injected `env` object / `importFreshDatabaseModule`), <100ms per unit test.

- `databasePath.test.ts` — new `guards production :memory: misuse (#3065)` suite:
  - bare `AUTOMOBILE_DB_PATH=:memory:` and legacy `AUTO_MOBILE_DB_PATH=:memory:` throw `ActionableError`;
  - opt-in present-but-false-ish (`""`/`0`/`false`) still throws;
  - the error message names `:memory:`, `AUTOMOBILE_DB_PATH`, and the opt-in flag;
  - real file paths resolve unchanged with **and** without the flag.
  - Existing `:memory:` passthrough tests updated to opt in.
- `migrationLock.test.ts` — `isInMemoryDatabaseOptInEnabled` truthy/false-ish matrix; `selectMigrationLock` still returns `NoOp` for `:memory:` / `File` otherwise.
- `dbWriteBarrierResetOnClose.test.ts` — lifecycle tests opt in; new integration test proves `getDatabase()` throws `/:memory:/` when the flag is absent.
- Full `bun test test/db/` → **555 pass / 0 fail**; `turbo run lint build` green.

### DIFF
--- a/src/db/database.ts
+++ b/src/db/database.ts
@@ -4,9 +4,14 @@ import * as os from "os";
 import * as fs from "fs";
 import type { Database as DatabaseSchema } from "./types";
 import { runMigrations } from "./migrator";
-import { createMigrationLock, isInMemoryDatabasePath } from "./migrationLock";
+import {
+  isInMemoryDatabaseOptInEnabled,
+  isInMemoryDatabasePath,
+  IN_MEMORY_DB_OPT_IN_ENV,
+  selectMigrationLock,
+} from "./migrationLock";
 import { logger } from "../utils/logger";
-import { toActionableError } from "../models/ActionableError";
+import { ActionableError, toActionableError } from "../models/ActionableError";
 import { BunSqliteDialect } from "./bunSqliteDialect";
 import { resolvePathFromDaemonLaunchWorkingDirectory } from "../utils/workingDirectory";
 import {
@@ -67,11 +72,29 @@ export function resolveDatabasePathFromEnvironment(
   if (envDbPath) {
     // The `:memory:` sentinel is not a filesystem path: routing it through the
     // daemon-launch-cwd resolver would `path.resolve(":memory:")` into a bogus
-    // absolute path (and `createFileMigrationLock` would then try to create a
+    // absolute path (and `selectMigrationLock` would then try to create a
     // `:memory:.migrate.lock` file). Pass it through un-resolved (issue #3047).
-    return isInMemoryDatabasePath(envDbPath)
-      ? envDbPath
-      : resolvePathFromDaemonLaunchWorkingDirectory(envDbPath);
+    if (isInMemoryDatabasePath(envDbPath)) {
+      // `:memory:` is a test-only seam: it is private per connection, so the app
+      // connection never sees the migration connection's schema — migrations
+      // report success while the daemon queries a migrated-but-empty DB and the
+      // first schema-dependent read/write fails with `no such table`. Reject it
+      // outside an explicit opt-in so a production `AUTOMOBILE_DB_PATH=:memory:`
+      // fails fast and legibly instead of silently breaking (issue #3065).
+      if (!isInMemoryDatabaseOptInEnabled(env)) {
+        throw new ActionableError(
+          `AUTOMOBILE_DB_PATH=:memory: is not a valid production database. A ` +
+            `SQLite \`:memory:\` DB is private per connection, so startup migrations ` +
+            `run on a separate in-memory database and the daemon's connection is left ` +
+            `migrated-but-empty — the first query (e.g. \`tool_calls\`) would fail with ` +
+            `\`no such table\`. Point AUTOMOBILE_DB_PATH at a real file (or unset it to use ` +
+            `~/.auto-mobile/auto-mobile.db). The \`:memory:\` sentinel is for lifecycle ` +
+            `tests only; set ${IN_MEMORY_DB_OPT_IN_ENV}=1 to opt in from a test.`
+        );
+      }
+      return envDbPath;
+    }
+    return resolvePathFromDaemonLaunchWorkingDirectory(envDbPath);
   }
 
   const envDbDir = env.AUTOMOBILE_DB_DIR ?? env.AUTO_MOBILE_DB_DIR;
@@ -259,7 +282,7 @@ async function startMigrations(dbPath: string): Promise<void> {
       // file (override env / --no-proxy alongside a daemon) must not both enter
       // migrateToLatest() and collide on the kysely_migration PRIMARY KEY (#2794).
       // A `:memory:` DB is private per connection, so it gets a no-op lock (#3047).
-      lock: createMigrationLock(dbPath),
+      lock: selectMigrationLock(dbPath),
       backup: () => backupDatabaseFile(migrationDb, dbPath),
     });
     // `migrationsRun` is NOT set here: it is a fenced global written only by the
--- a/src/db/migrationLock.ts
+++ b/src/db/migrationLock.ts
@@ -200,7 +200,9 @@ export function migrationLockPathFor(dbPath: string): string {
  * open/close/reopen contract without querying migrated schema — not a production
  * DB configuration. (Sharing one connection across the app and the migrator would
  * break the migration failure-isolation the separate `migrationDb` provides, so
- * it is deliberately out of scope.)
+ * it is deliberately out of scope.) A production `AUTOMOBILE_DB_PATH=:memory:` is
+ * now rejected at path-resolution time unless {@link IN_MEMORY_DB_OPT_IN_ENV} is
+ * set, so this footgun fails fast instead of silently (issue #3065).
  */
 export const IN_MEMORY_DATABASE_PATH = ":memory:";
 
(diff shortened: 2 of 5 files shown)


kaeawc:
## Adversarial review summary (3 perspectives)

Ran three independent reviews — DB-reliability, API/DX, and test-completeness — against latest `origin/main`. Consensus: **no blocking findings.**

**DB reliability (regressions):** No regression found. The guard throws at the single path-resolution choke point *before* any caching (`lifecycle.resolvedDbPath` stays `null`, `dbInstance` stays `null`), so repeated calls behave identically and no partial-init state leaks. Every reachable caller (`getDatabase`/`ensureMigrations` via `guardDatabaseStartup`, `directModeGuard` via `main().catch`, `daemon.ts` pidfile *after* open) sits at a startup/first-use point where fail-fast is the intent — no swallow, no health/shutdown-path throw. Opt-in is read from the same `env` object as the path. Windows behavior for real files is unchanged (guard is platform-agnostic; the Windows-skipped tests are unrelated symlink-lock tests). Test helpers that use `:memory:` directly bypass the resolver, so they're unaffected.

- Non-issue confirmed: in daemon mode the rejection is classified `permanent` by `classifyDatabaseFailure` (the fail-safe default), so it engages the crash-loop circuit breaker rather than hot-looping — the desired behavior for a static misconfiguration.

**API/DX (direction):** Option 1 (runtime guard) is the right call vs option 2 (share-connection, breaks migration failure-isolation) and option 3 (docs-only, leaves the silent footgun). Flag name `AUTOMOBILE_ALLOW_IN_MEMORY_DB` fits repo conventions; no legacy `AUTO_MOBILE_` alias needed (brand-new, test-only → YAGNI). Truthy parsing (`1`/`true`/`yes`, case/whitespace-insensitive) matches `parseBool`, with a deliberate fail-safe divergence (unrecognized → disabled, not throw). Error message is accurate — `tool_calls` is a real migrated table and phrased as an example. Rename `createMigrationLock`→`selectMigrationLock` is clean and complete (zero stragglers).

**Test completeness:** Coverage is strong and the guard is provably load-bearing (removing it makes the `toThrow` tests fail). Both env aliases covered on accept + reject paths; real `getDatabase()` integration asserted; no env leaks; no pre-existing assertion weakened; error-message assertion robust-but-not-brittle. One minor gap flagged and **addressed in 91afcdac**: the resolver-level passthrough test now exercises all truthy opt-in spellings (`1`/`true`/`yes`/`"  Yes  "`) end-to-end, not just `"1"`.

Deferred (not this PR): the `AUTOMOBILE_DB_*` env-var family is undocumented in `docs/` — worth a consolidated env-var reference doc later. Will file a follow-up.

Issue #2992, opened 2026-07-04
Windows CI: databaseLazyPath + dbWriteBarrierResetOnClose file-DB tests still flake (EBUSY/livelock) despite #2916/#2923


kaeawc:
## Summary
The Windows-only `Node TypeScript Build and Test (windows-latest)` CI job intermittently fails on **file-backed DB lifecycle tests** despite the hardening in #2916 / PR #2923 ("harden Windows file-DB close/reopen tests against EBUSY livelock"). The failures rotate between tests run-to-run and are pure timing (9–10s durations before failing), consistent with the documented Windows file-DB EBUSY + livelock class. They are green on ubuntu-latest and macos-latest.

This blocked auto-merge of the **unrelated** PR #2987 (navigation-graph transactions), which had to be merged manually after confirming the failures were pre-existing and not caused by that change.

## Evidence (from PR #2987 CI, 2026-07-04)
Two consecutive Windows runs of the same commit failed on **different** DB-lifecycle tests, both green on ubuntu/macos:

- Run 1 (`runs/28708897964`, job `85139088997`):
  `(fail) database path lazy resolution > queries issued immediately after getDatabase wait for startup migrations [9219.00ms]`
  → `test/db/databaseLazyPath.test.ts`
- Run 2 (rerun, job `85139930542`):
  `(fail) closeDatabase cold-starts the dbWriteBarrier drain latch (issue #2896) > a tracked best-effort write after drain -> close -> reopen is NOT skipped [10313.00ms]`
  → `test/db/dbWriteBarrierResetOnClose.test.ts`

Both jobs otherwise ran all 6007 tests; only one file-DB lifecycle test failed each time, each after ~9–10s (timeout-shaped, not an assertion mismatch). The rotating target confirms an environmental/timing issue, not a deterministic bug.

## Why it matters
- It produces **false-red** CI on PRs that don't touch the DB lifecycle at all, blocking auto-merge and forcing manual review/merge (as happened for #2987).
- #2916 / #2923 already targeted this exact class (`:memory:` + temp lock file portability, per the repo's known-issue notes) but the `databaseLazyPath` and `dbWriteBarrierResetOnClose` suites still use real file-backed temp dirs (`mkdtemp` + `closeDatabase` + `removeTempDbDir`) and still flake on windows-latest.

## Proposed fix
Extend the #2916/#2923 hardening to the two suites still flaking on Windows:
- `test/db/databaseLazyPath.test.ts`
[...]

PR #3040, opened 2026-07-04
test(db): stabilize Windows file-DB lifecycle suites (#2992)

kaeawc:
## Problem

The Windows-only `Node TypeScript Build and Test (windows-latest)` CI job intermittently failed **file-backed DB lifecycle tests** with timeout-shaped errors that rotated between suites run-to-run (#2992). This produced false-red CI on PRs that don't touch the DB lifecycle at all, blocking auto-merge (it forced a manual merge of the unrelated #2987).

Two suites were still flaking despite the #2916/#2923 hardening:
- `test/db/databaseLazyPath.test.ts`
- `test/db/dbWriteBarrierResetOnClose.test.ts`

**Root cause** (from the failing run `28708897964`, job `85139930542`):
`getDatabase()` kicks off `startMigrations()` on a **separate, detached** connection (resolve-never-reject). `dbWriteBarrierResetOnClose` then `closeDatabase()`/reopened *without awaiting* that run, so on `windows-latest` the app-connection WAL checkpoint contended with the still-running migration connection — producing compounding synchronous `busy_timeout=5000ms` stalls that blew **both** the 30s body timeout (`Test ... timed out after 30000ms`) **and** the `afterEach` hook (`a beforeEach/afterEach hook timed out`, `removeTempDbDir` losing to a still-locked handle). `databaseLazyPath`'s "queries wait for migrations" test additionally had **no explicit body timeout**, so a slow cold Windows migration was killed at bun's 5s default.

## Fix

- **`dbWriteBarrierResetOnClose.test.ts`**: `await ensureMigrations()` to fully settle after each `getDatabase()` and before each `closeDatabase()`/reopen, so no detached migration connection can contend with the close-time WAL checkpoint. This models a *fully-booted* daemon before shutdown (more realistic than draining mid-migration) and leaves every barrier-reset assertion untouched. Gave `afterEach` the same generous ceiling so a slow temp-dir release cannot read as a hook timeout (verified bun honors the `afterEach(fn, timeout)` arg).
- **`databaseLazyPath.test.ts`**: gave the two file-backed tests a generous body timeout; the two pure path-resolution tests (no DB opened) are unchanged.
- **`test/db/fileBackedDbTestTimeout.ts`**: one canonical `WINDOWS_FILE_DB_TEST_TIMEOUT_MS` constant shared by both suites (was an inline literal in one).
[...]

PR #3078, opened 2026-07-04
test(db): shared withFileBackedDb harness to enforce the DB-lifecycle flake-avoidance pattern (#3046)

### LINKED ISSUE #3046: Shared withFileBackedDb test harness to enforce the DB-lifecycle flake-avoidance pattern
## Summary
Follow-up from the review of #3040 (closes #2992). This is the **third** iteration of the same Windows file-backed DB test flake class (#2916 → #2923 → #2992). Each round hardens the *instances* that are currently flaking; nothing enforces the pattern, so the next engineer who adds a file-backed close/reopen suite will re-hit the same flake.

## The pattern that must be re-applied by hand today
Every file-backed DB lifecycle suite must independently remember to:
1. `importFreshDatabaseModule()` for module-global isolation (`test/db/freshDatabaseModule.ts`),
2. track temp dirs and clean them with the bounded `removeTempDbDir()` (`test/db/tempDbDir.ts`, #2916),
3. `getDatabase()` **then `await ensureMigrations()`** before `closeDatabase()` so a detached migration connection can't contend with the close-time WAL checkpoint on Windows (#2992/#3040),
4. set the body/hook timeout to `WINDOWS_FILE_DB_TEST_TIMEOUT_MS` (`test/db/fileBackedDbTestTimeout.ts`).

This knowledge lives only in prose comments spread across ~5 files. Prose hasn't stopped the recurrence.

## Proposal
Add a shared test harness (e.g. `test/db/withFileBackedDb.ts`) exposing something like:
- `openLifecycleTestDb(prefix)` → fresh module import + tracked `mkdtemp` + `getDatabase()` + `await ensureMigrations()`, returning the module + dir + a `close()` that does `closeDatabase()` + bounded `removeTempDbDir()`.
- The suite-level timeout applied via the shared constant.

Migrate `databaseLazyPath`, `dbWriteBarrierResetOnClose`, `databaseReset`, `databaseMigrationFailure`, `databaseMigrationLockReopen`, `databaseLifecycleReset`, `databaseMigrationGenerationFence` onto it so correct ordering is the path of least resistance and a new suite gets it for free.

## References
- Surfaced by the direction/alternatives review of #3040 (highest-value durability follow-up).
- Prior rounds: #2916 / PR #2923, #2992 / PR #3040, #2849.

kaeawc:
## Summary

Closes #3046. Follow-up from the direction/alternatives review of #3040 (the third round of the same Windows file-backed DB test flake class: #2916 → #2923, #2992 → #3040). Each prior round hardened the *instances* that were flaking; nothing enforced the pattern, so the next engineer who added a file-backed close/reopen suite re-hit it. The knowledge lived only in prose comments spread across ~5 files — and prose did not stop the recurrence.

This adds a shared harness so **correct ordering is the path of least resistance** and a new suite gets the whole pattern for free.

## Sequencing (per the issue's blocker note)

#3046 was blocked by #3047, which moved `dbWriteBarrierResetOnClose` **off files onto a `:memory:` sentinel**. #3047 has landed (main `fa69e26a9`), so this PR is **re-scoped to exclude `dbWriteBarrierResetOnClose`** — it is no longer file-backed. The remaining file-backed lifecycle suites are migrated.

## What changed

- **`test/db/withFileBackedDb.ts`** (new) — `createFileBackedDbHarness(deps?)` centralizes the 4-part pattern in one place:
  1. fresh, isolated `database.ts` module import (collision-proof counter via the existing `importFreshDatabaseModule`, replacing the duplicated `Date.now()-Math.random()` cache-bust keys),
  2. tracked temp dirs cleaned with the **bounded** `removeTempDbDir` (#2916) — replacing three copies of an **unbounded** 100×50ms `removeTempDirWithRetry` that could stall `afterEach`,
  3. `openLifecycleTestDb(prefix)` — the common open path that does `getDatabase()` **then** `await ensureMigrations()` before any close (#2992/#3040 ordering),
  4. full-env snapshot/restore + the shared `WINDOWS_FILE_DB_TEST_TIMEOUT_MS` (re-exported).
  Every side effect (`mkdtemp`, `removeTempDir`, `importDatabaseModule`, `env`) is injectable so the orchestration is unit-tested with fakes.
- **`test/db/withFileBackedDb.test.ts`** (new) — fake-injected unit tests for the orchestration (tracking, env restore, the getDatabase→ensureMigrations ordering, close() untracking, env-restore-in-`finally`), **plus one real-DB integration test** that drives the un-injected `openLifecycleTestDb` end-to-end (fresh module → full startup migration set on a real temp file → queryable migrated `tool_calls` → self-cleaning `close()`). The integration test deliberately does **not** assert the temp dir is gone after close — `removeTempDbDir` is best-effort and gives up on a Windows handle livelock, so asserting removal would reintroduce the flake class; it asserts `close()`/`cleanup()` resolve and the dir is untracked.
- **Migrated onto the harness** (behavior-preserving): `databaseLazyPath`, `databaseReset`, `databaseMigrationFailure`, `databaseMigrationLockReopen`, `databaseLifecycleReset`, `databaseMigrationGenerationFence`, and the same-pattern `databaseIncompleteExtraction`. Each drops its private `importFreshDatabaseModule` / `removeTempDirWithRetry` / env-restore boilerplate. The delicate mid-flight concurrency tests (gen-fence, lock-reopen) keep driving the module directly via the harness primitives; only the simple open+migrate+close flow uses `openLifecycleTestDb`.

## Non-goals / notes

- `dbWriteBarrierResetOnClose` stays on the `:memory:` sentinel from #3047 (excluded here, per the issue).
- `backupDatabaseFile` uses `BunDatabase` directly (not the `database.ts` module), so it is not a lifecycle suite and is left alone.

## Verification

- `bun test test/db/` → **556 pass, 0 fail** (52 files).
- `turbo run lint build` → clean.
- Harness unit tests run with fakes only (no real DB/fs/clock) — well under the 100ms budget; the one real-DB integration test carries the shared file-backed timeout.
- No remaining inline `Date.now()-${Math.random()` fresh-import or unbounded `removeTempDirWithRetry` in `test/db/`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -87,6 +87,20 @@ bun test --bail        # Stop on first failure (no cache)
 bun test <file>        # Run specific test file (no cache)
 ```
 
+## File-backed DB lifecycle tests
+
+Any test suite that opens a real `auto-mobile.db` (module close/reopen, migration
+lifecycle, path resolution) MUST go through the shared
+`createFileBackedDbHarness()` in `test/db/withFileBackedDb.ts` — never hand-roll a
+`mkdtemp` + raw `import("database.ts?...")` + env restore. The harness centralizes
+the four-part Windows-flake-avoidance pattern (fresh isolated module import,
+tracked temp dirs cleaned with the bounded `removeTempDbDir`, `getDatabase()` then
+`await ensureMigrations()` before close, and the `WINDOWS_FILE_DB_TEST_TIMEOUT_MS`
+ceiling). Use `openLifecycleTestDb(prefix)` for the common open→migrate→close flow;
+use the `makeTempDbDir` / `importFreshDatabaseModule` primitives for suites that
+must control the migration lifecycle mid-flight. `:memory:`-backed suites (e.g.
+`dbWriteBarrierResetOnClose`) are deliberately exempt — they carry no file handle.
+
 # Validate Shell Scripts
 
 Shell scripts under `scripts/` are tested with [BATS](https://github.com/bats-core/bats-core) and linted with shellcheck.
--- a/test/db/databaseIncompleteExtraction.test.ts
+++ b/test/db/databaseIncompleteExtraction.test.ts
@@ -1,10 +1,9 @@
-import { afterEach, describe, expect, test } from "bun:test";
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
 import path from "path";
-import { mkdir, mkdtemp, writeFile } from "node:fs/promises";
-import { tmpdir } from "node:os";
+import { mkdir, writeFile } from "node:fs/promises";
 import { defaultTimer } from "../../src/utils/SystemTimer";
 import { isIncompleteExtractionError } from "../../src/db/migrationDependencyIntegrity";
-import { removeTempDbDir } from "./tempDbDir";
+import { createFileBackedDbHarness } from "./withFileBackedDb";
 
 /**
  * Issue #2833: the incomplete-extraction mapping is scoped to KNOWN migration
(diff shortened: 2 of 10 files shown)


kaeawc:
## Review round (self `/auto-mobile-code-review` + 3 adversarial perspectives)

Ran three independent review passes — a concurrency/regression skeptic, an API-design architect, and a Windows-flake-durability reviewer — plus the self-review.

**Correctness / flake-safety: green.** The concurrency review confirmed the migration is behavior-preserving (every `expect(...)` identical old→new; `databaseLifecycleReset` *gained* the 30s ceiling — a hardening; the collision-proof counter replaces `Date.now()-Math.random()`; bounded cleanup can't stall/throw in `afterEach`). The Windows reviewer confirmed the one real-file integration test asserts only lock-independent facts (no `existsSync`/`-wal`/`-shm`/handle-timing assertion), the body carries `WINDOWS_FILE_DB_TEST_TIMEOUT_MS` (not the hook), and the 9 fake-injected unit tests are sub-millisecond. **windows-latest TS Build & Test passed (3m1s)** — the exact flake surface this PR targets.

**Addressed before merge — the one substantive finding:** `openLifecycleTestDb` (the issue's proposed centerpiece) had **no real consumer** — all migrated suites used only the primitives. Fixed in `ad7fa2d76` by routing `databaseReset`'s healthy cold-boot phase through it (proving the flagship open→migrate→close flow + `LIFECYCLE_OVERRIDE_ENV_KEYS` against a real file-backed reopen suite) and adding a CLAUDE.md pointer so new file-backed DB lifecycle suites reach for the harness by default.

> Note on a reviewer suggestion: migrating `dbWriteBarrierResetOnClose` onto `openLifecycleTestDb` was declined — #3047 deliberately moved it to a `:memory:` sentinel (no file handle), and this issue explicitly excludes it. `databaseReset` is the correct file-backed consumer.

**Deferred to a follow-up issue** (non-blocking, per YAGNI): drop the `<M>` generic + `as unknown as` cast now that there's a single module type; consider a `bindFileBackedDbHarness()` hooks binder to remove the per-suite `beforeEach/afterEach` boilerplate; and a structural/lint guard so a new file-backed suite *cannot* bypass the harness (the enforcement the issue ultimately wants — this PR makes the harness the documented path but does not yet mechanically enforce it).

Issue #3075, opened 2026-07-04
Extract shared two-repo runInTransaction helper for NavigationGraphManager (follow-up to #2980)


kaeawc:
## Context

Follow-up from #2980 / PR #3072 (which pinned the `assertSharedConnection()`
guard in `NavigationGraphManager.recordHierarchyNavigation` Case 1 with
regression tests). During that review, all three reviewers (DB-concurrency,
test-rigor, direction) noted that the YAGNI trigger #2980 named for extracting a
shared two-repo transaction helper is **now met**.

## Problem

There are now **two** call sites that (a) enlist BOTH the navigation and
test-coverage repositories onto one transaction and (b) must call
`assertSharedConnection()` first:

- `NavigationGraphManager.recordNavigationEvent` — `src/features/navigation/NavigationGraphManager.ts:362` (guard) + `:370` (`runInTransaction`)
- `NavigationGraphManager.recordHierarchyNavigation` Case 1 — `src/features/navigation/NavigationGraphManager.ts:545` (guard) + `:547` (`runInTransaction`)

#2980 explicitly deferred this under YAGNI while there was a single site:

> If a second transactional call site lands here, consider promoting the guard
> into a shared two-repo `runInTransaction` helper that owns the invariant so each
> new site can't forget to call it.

The second site (`recordHierarchyNavigation` Case 1) landed via #2987. The
`assertSharedConnection()` + `withExecutor(trx)` bind-both-repos preamble is now
duplicated across the two methods; a third both-repos site could forget the
guard (the exact failure mode #2968/#2980 guard against).

## Proposed work

Extract a helper on `NavigationGraphManager` (or `NavigationRepository`) — e.g.
`runBothReposInTransaction(fn: (navTrx, coverageTrx) => Promise<T>)` — that:

1. Calls `assertSharedConnection()` once, internally, so no caller can forget it.
2. Opens `this.repository.runInTransaction`, binds `this.repository.withExecutor(trx)`
   and `this.testCoverageRepository.withExecutor(trx)`, and passes both bound repos
   to `fn`.
3. Leaves side-effects (in-memory field updates, `notifyGraphUpdated`, telemetry)
   to the caller, post-commit — matching the current convention.

Then refactor `recordNavigationEvent` and `recordHierarchyNavigation` Case 1 onto
it.

### Also fold in (minor, surfaced in #3072 review)

The guard's `ActionableError` message
(`src/features/navigation/NavigationGraphManager.ts:502`) names only
[...]

PR #3090, opened 2026-07-04
test(db): cover promoteSuggestion notifyGraphUpdated commit/rollback (#2981)

### LINKED ISSUE #2981: Make NavigationGraphManager.promoteSuggestion atomic (partial-write on mid-sequence throw)
## Context

Surfaced while implementing #2968 / PR #2977 (the shared-connection guard for
`recordNavigationEvent`). This is a sibling atomicity gap in the same lineage as
#2791 / PR #2959 ("the transaction pattern extended to the sequences above" — the
deferred hardening Devin Osei noted in the #2959 review).

## Problem

`NavigationGraphManager.promoteSuggestion`
(`src/features/navigation/NavigationGraphManager.ts:1392-1414`) performs two
separately-awaited navigation-repository writes with **no enclosing transaction**:

```ts
```

`NavigationRepository.promoteSuggestion` (`src/db/navigationRepository.ts`) itself
does multiple writes: a `getOrCreateFingerprint` upsert **plus** an
`updateTable("navigation_suggestions").set({ promoted_to_fingerprint_id })`.

So the full `getOrCreateNode` → `getOrCreateFingerprint` → suggestion-link sequence
is non-atomic. A throw (e.g. a UNIQUE conflict or a daemon interrupted mid-`await`)
after `getOrCreateNode` commits leaves a **named node created but the suggestion
not linked** (or a fingerprint created without the suggestion pointer) — a partial
write of exactly the class #2791 fixed for `recordNavigationEvent`.

Unlike #2968/#2980, this path touches **only** the navigation repository (no
`TestCoverageRepository`), so there is no connection-identity dimension — it needs a
transaction, not the shared-connection guard.

## Proposed work

Wrap the node-create + promote sequence in `this.repository.runInTransaction(...)`
using a `withExecutor(trx)`-bound repo (mirroring `recordNavigationEvent`), moving
the `notifyGraphUpdated()` side effect outside the transaction. Note
`getOrCreateUIElement` is already `isTransaction`-aware; `promoteSuggestion` does not
call it, so no nested-transaction concern here.

## Acceptance criteria

- [ ] `promoteSuggestion`'s DB writes run in a single transaction; a mid-sequence
      throw rolls back (no node persisted without its fingerprint/suggestion link).
- [ ] `notifyGraphUpdated()` fires only after commit, never on rollback.
- [ ] Unit test with an injected repo that throws mid-sequence asserts zero new
      `navigation_nodes` / fingerprint / suggestion-link rows (real in-memory DB +
      fake, <100ms, no device), following repo conventions.

## References
- #2968, PR #2977 (surfacing context; the shared-connection guard)
- #2980 (sibling follow-up: `recordHierarchyNavigation` coverage-enlistment path)
- #2791 / PR #2959 (the original `recordNavigationEvent` transactionalization + pattern)

kaeawc:
## Summary

Closes #2981 — the last uncovered acceptance criterion for making
`NavigationGraphManager.promoteSuggestion` atomic.

The production transaction #2981 asks for **already landed** via PR #2987 (for the
sibling #2968): `promoteSuggestion`'s node-create + promote sequence runs inside a
single `runInTransaction` with a `withExecutor(trx)`-bound repo, and
`notifyGraphUpdated()` is placed **outside** the transaction. Its test block, however,
asserted only on **AC #1/#3** (DB-row rollback on a mid-sequence throw) and never on
**AC #2** — that `notifyGraphUpdated()` fires *only after commit, never on rollback*.

This PR is **test-only** and closes that AC #2 gap.

## What changed

`test/features/navigation/NavigationGraphManager.test.ts` — three assertions added to
the existing `promoteSuggestion transaction` block:

- **Both mid-sequence-throw rollback tests** now register a graph-update listener and
  assert it **never fires** (`notifyCount === 0`) — one for the `Suggestion not found`
  path (link fails after the node is created) and one for the `ThrowingLinkPromoteRepo`
  path (link fails after the fingerprint is created).
- **New happy-path test** asserts the listener fires **exactly once, after** the final
  in-transaction write (`repository.promoteSuggestion`) completes — using the
  call-count-at-notify pattern already established for `recordNavigationEvent` (#2791):
  spy on `NavigationRepository.prototype.promoteSuggestion`, record its call count at
  every notify fire, assert `[1]`. A notification firing *before* the transaction
  completes (count still 0) would be caught, not masked by a later post-commit fire.

No production code changed — the transaction from #2987 is untouched.

## Verification

- `bun test test/features/navigation/NavigationGraphManager.test.ts` → **64 pass / 0 fail**
  (was 63; +1 new test), 843ms for the whole file.
- **Teeth proven** (TDD): temporarily misplacing `notifyGraphUpdated()` inside the
  transaction fails the new tests — at the end of the txn body the "exactly once" test
  fails (`[1,1]` ≠ `[1]`); at the start of the txn body both rollback tests fail
  (listener fires before the throw). Reverted; all green.
- `bunx turbo run lint build` → clean. `scripts/all_fast_validate_checks.sh` → 11/11 pass.
- Unit-test conventions: Interface + Fake, real in-memory DB (`createTestDatabase`), no
  device/network, <100ms.

## Non-goals / follow-ups

- The `describe` block is titled `(#2968)` because that PR seeded it; the new assertions
  are kept co-located with the rollback tests they extend rather than split into a
  near-duplicate `(#2981)` block.
- #2980 (sibling `recordHierarchyNavigation` coverage-enlistment guard) is tracked
  separately and unaffected here.

### DIFF
--- a/test/features/navigation/NavigationGraphManager.test.ts
+++ b/test/features/navigation/NavigationGraphManager.test.ts
@@ -1528,19 +1528,24 @@ describe("NavigationGraphManager - promoteSuggestion transaction (#2968)", () =>
     // prototype, so the trx-bound write is counted too.
     const promoteSpy = spyOn(NavigationRepository.prototype, "promoteSuggestion");
 
-    const notifyPromoteCounts: number[] = [];
-    manager.setGraphUpdateListener(() => {
-      notifyPromoteCounts.push(promoteSpy.mock.calls.length);
-    });
+    // try/finally so a failing assertion can never leave this prototype spy armed
+    // for sibling suites in the same process (spyOn calls through, but restore is
+    // still the hygienic contract).
+    try {
+      const notifyPromoteCounts: number[] = [];
+      manager.setGraphUpdateListener(() => {
+        notifyPromoteCounts.push(promoteSpy.mock.calls.length);
+      });
 
-    await manager.promoteSuggestion(suggestionId, "SettingsScreen");
+      await manager.promoteSuggestion(suggestionId, "SettingsScreen");
 
-    // promoteSuggestion ran exactly once, and notify fired exactly once, AFTER that
-    // final in-transaction write completed — never inside or before the transaction.
-    expect(promoteSpy).toHaveBeenCalledTimes(1);
-    expect(notifyPromoteCounts).toEqual([1]);
-
-    promoteSpy.mockRestore();
+      // promoteSuggestion ran exactly once, and notify fired exactly once, AFTER that
+      // final in-transaction write completed — never inside or before the transaction.
+      expect(promoteSpy).toHaveBeenCalledTimes(1);
+      expect(notifyPromoteCounts).toEqual([1]);
+    } finally {
+      promoteSpy.mockRestore();
+    }
   });
 });
 
--- a/test/features/navigation/NavigationGraphManager.test.ts
+++ b/test/features/navigation/NavigationGraphManager.test.ts
@@ -1466,6 +1466,12 @@ describe("NavigationGraphManager - promoteSuggestion transaction (#2968)", () =>
     const nodesBefore = await countNodes();
     const fingerprintsBefore = await countFingerprints();
 
+    // notifyGraphUpdated must fire only after commit — never on rollback (#2981 AC).
+    let notifyCount = 0;
+    manager.setGraphUpdateListener(() => {
+      notifyCount++;
+    });
+
     // A non-existent suggestion id: getOrCreateNode creates the "OrphanScreen"
     // node first, then repository.promoteSuggestion throws "Suggestion not found"
     // during the multi-write promotion. Without a transaction the node would be
@@ -1478,6 +1484,9 @@ describe("NavigationGraphManager - promoteSuggestion transaction (#2968)", () =>
     expect(await countNodes("OrphanScreen")).toBe(0);
     expect(await countNodes()).toBe(nodesBefore);
     expect(await countFingerprints()).toBe(fingerprintsBefore);
+
+    // The post-commit side effect never fired because the transaction threw.
+    expect(notifyCount).toBe(0);
   });
 
   test("rolls back BOTH the node and the fingerprint when the link fails after the fingerprint is created", async () => {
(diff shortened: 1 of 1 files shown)


kaeawc:
## Review pass (3 perspectives) + #3075 conflict check

Ran three independent reviews — DB-atomicity skeptic, test-flake auditor, scope/direction — plus the AutoMobile code-review skill. **Consensus: the PR correctly and fully closes #2981.**

- **Atomicity (verified):** `promoteSuggestion` (`NavigationGraphManager.ts:1441-1449`) rebinds the whole repo via `withExecutor(trx)`; the full `getOrCreateNode → getOrCreateFingerprint → suggestion-link UPDATE` chain resolves through the bound executor — no statement escapes to the singleton connection, no bunSqliteDialect deadlock, no partial-write window. `notifyGraphUpdated()` is outside the transaction.
- **AC coverage:** AC #1/#3 landed with #2987; this PR adds the missing AC #2 assertions (notify never fires on rollback; fires exactly once after the final in-transaction write). TDD teeth confirmed by temporarily misplacing the notify (all 3 assertions fail), then reverting.
- **Hardening applied (bc3d87a9b):** two reviewers flagged the prototype-spy `mockRestore()` should be in `try/finally` so a failing assertion can't leak the spy into sibling suites. Applied to the new test. (The pre-existing #2791 `touchSpy` has the same pattern; left for #3075's broader helper refactor to avoid widening this test-only PR.)

### No conflict with #3075
#3075 extracts a shared **two-repo** (nav + coverage) `runInTransaction` helper for `recordNavigationEvent` and `recordHierarchyNavigation` Case 1, and reworats the `assertSharedConnection()` guard message. `promoteSuggestion` is **nav-repo-only** — it uses `withExecutor(trx)` but neither `assertSharedConnection()` nor `testCoverageRepository`, so it is deliberately **not** one of the two sites #3075 refactors. This diff touches none of #3075's territory (no guard, no guard message, no coverage repo, no two-repo test block) and lives entirely in the `promoteSuggestion` describe block (lines 1466–1547), disjoint from the guard/two-repo test blocks (880+, 1087+, 1578+). No open #3075 PR exists; a future #3075 merge will be conflict-free.

CI was green on the first commit across all Node TypeScript (ubuntu/macos/windows), CodeQL, Memory Leak, and Fast Validation jobs; re-running on the hardening commit.

Issue #3017, opened 2026-07-04
Android CtrlProxy: decode failures send no wire error frame (silent drop) — mirror iOS #2986 field-attributed decode errors


kaeawc:
## Summary

The **Android** CtrlProxy runner silently drops any inbound command it fails to decode — unlike the **iOS** runner, which now surfaces an actionable, field-attributed decode error on the wire (#2965, #2986). On Android a malformed command produces **no error frame**, so the daemon-side TS client never gets a rejection; its pending request just **times out** with no diagnostic.

## Detail

`android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/WebSocketServer.kt:301-307`:

```kotlin
val request =
  try {
    protocolJson.decodeFromString<ProtocolRequest>(message)
  } catch (e: Exception) {
    Log.w(TAG, "Failed to parse client message: $message", e)
    return          // <-- logs and returns; no error frame sent back
  }
```

Compare the iOS runner: `WebSocketServer.swift` `handleMessage` catch → `buildErrorResponseData(requestId:error:)` → `wireErrorMessage(for:)`, which sends a structured `{"type":"error","success":false,"requestId":...,"error":"Malformed request: wrong type for field 'x' (...)"}` frame. iOS attributes `typeMismatch` / `valueNotFound` to the offending field and rewrites `dataCorrupted` (overflow / malformed JSON) into an actionable cause.

There is **no `wireErrorMessage` analog on Android** — the decode failure path has nothing to enrich because it emits no frame at all.

## Impact

- A client that sends a malformed/wrong-typed command against an Android device gets a **timeout**, not an immediate actionable error — harder to diagnose than the iOS behavior.
- kotlinx.serialization already rejects out-of-range / non-finite literals at decode (`allowSpecialFloatingPointValues=false`, see #2964), so the *rejection* is correct; only the **wire feedback** is missing.

## Proposed fix

In `handleClientMessage`'s decode `catch`, before `return`, best-effort extract the `requestId` from the raw JSON (mirroring iOS `extractRequestId`) and `broadcast` a structured error frame — ideally field-attributed for kotlinx `SerializationException` subtypes (e.g. `MissingFieldException`, type-mismatch decode errors expose the path/field). Keep the discriminator/unknown-command contract intact (Android's equivalent of `Unknown command type: <type>`, if any).

## Acceptance criteria

[...]

Issue #2986, opened 2026-07-04
iOS runner: extend decode-error legibility to typeMismatch/valueNotFound (field-attributed) — #2965 option 2


kaeawc:
## Summary

#2965 / PR #2983 made the iOS CtrlProxy runner's **`DecodingError.dataCorrupted`** wire message actionable (out-of-range numeric literals, malformed JSON). But the *other* common `DecodingError` cases — **`typeMismatch`** and **`valueNotFound`** — still surface the same opaque `localizedDescription` ("The data couldn't be read because it isn't in the correct format."), even though they carry a **non-empty `codingPath`** and so *can* be attributed to a specific field.

This is issue #2965's "option 2" (map common `DecodingError` cases more generally), deliberately left out of PR #2983's narrow scope.

## Detail

`ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift` `wireErrorMessage(for:)` rewrites only `.dataCorrupted`; everything else returns `error.localizedDescription`. Verified (Swift 6.2 host):

- `typeMismatch` (e.g. `{"type":"request_tap_coordinates","x":"hi","y":2}`) → `localizedDescription` = the **same opaque** "…isn't in the correct format.", but the error's `codingPath` is `[x]` — a field-attributed message like `"Malformed request: wrong type for field 'x' (expected number)."` is feasible.
- `valueNotFound` → same opaque string, also has a `codingPath`.

`keyNotFound` must stay unchanged per #2965 AC2 (relied on by `TypedRequestDecodeTests`), so it should remain excluded — but `typeMismatch` / `valueNotFound` are fair game.

## Proposed fix

Extend `wireErrorMessage(for:)` to map `typeMismatch` and `valueNotFound` to field-attributed actionable messages using `context.codingPath.last?.stringValue` and `context.debugDescription`. Keep `keyNotFound` and `unknownCommand` byte-for-byte unchanged (existing contracts). Note the nested-`dataCorrupted`-with-non-empty-`codingPath` case currently drops the `codingPath` when falling back to `context.debugDescription` — could also preserve the field there.

## Acceptance criteria

- [ ] A `typeMismatch` / `valueNotFound` inbound command surfaces a field-attributed actionable message naming the offending field, instead of the opaque decoder default.
- [ ] `keyNotFound` and `Unknown command type: <type>` messages remain unchanged (existing tests still pass).
- [ ] Unit tests pinning each mapped case.

## Context

[...]

PR #3092, opened 2026-07-04
fix(android): surface runner error text from requestHierarchySync via diagnostics (#3062)

### LINKED ISSUE #3062: observe/android: let requestHierarchySync surface the runner error text instead of collapsing to null
Follow-up to #3032 / PR #3060.

## Summary
#3032's acceptance criterion 1 says a correlated runner `type:"error"` frame should reject the hierarchy wait fast **"surfacing the runner's error text."** In PR #3060 the text reaches the **logs** (the `pendingHierarchyRejectors` hook `logger.warn`s it, and `requestHierarchySync`'s catch logs `Sync hierarchy request failed ... ${error}`), and the wait fails fast — but the public contract of `CtrlProxyHierarchy.requestHierarchySync` is `Promise<{hierarchy, perfTiming} | null>`, and it returns `null` on the rejection. So the **caller** receives `null`, indistinguishable from a plain timeout `null`.

## Why this might matter
A caller (`getAccessibilityHierarchy`, and `verifyServiceReady` at `AndroidCtrlProxyClient.ts:~1751`) currently cannot tell "the runner reported a structured handler failure" apart from "the push never arrived (timeout)". Both collapse to `null` → fallback. Distinguishing them would let higher layers log/telemetry the actual runner error, or choose not to retry on a deterministic handler failure.

## Proposed direction (optional / design-first)
Decide whether it is worth widening the contract. Options:
1. Return a discriminated result (e.g. `{ hierarchy } | { error: string } | null`) so callers can branch. Larger blast radius — every consumer of `requestHierarchySync` must handle the new shape.
2. Keep `null` but stash the last runner error on the client for telemetry/diagnostics.
3. Leave as-is (fail-fast + logged is sufficient) and just close this as wontfix.

This is intentionally deferred from #3060 because the whole method already returns `null` on every failure path and widening it is out of scope for the hang fix.

## Context
- Parent: #3032 (PR #3060). Related surfacing/legibility work: iOS #2986/#3017.
- Key file: `src/features/observe/android/CtrlProxyHierarchy.ts` (`requestHierarchySync` catch at the sync path; `waitForFreshData` rejection).

kaeawc:
## Summary

Follow-up to #3032 / PR #3060 (and sibling #3061 / PR #3070). Those routed a correlated runner `type:"error"` frame into the hierarchy wait so it fails fast instead of hanging to the `waitForFreshData` timeout — but `CtrlProxyHierarchy.requestHierarchySync` still collapses **both** a plain timeout and a runner-reported handler failure into a `null` return. Callers (`getAccessibilityHierarchy`, `verifyServiceReady`) therefore cannot tell "the runner reported a structured handler failure" apart from "the push never arrived (timeout)" — both become `null` → fallback.

This surfaces the runner's error text without widening the primary return contract.

## Approach

The issue's Option 1 (discriminated result) has a large blast radius — every caller of `requestHierarchySync` checks `if (result)` truthiness then `result.hierarchy`, so a new error shape would break them. Option 2 literal (stash on the client) is racy: a later timeout could read a **stale** runner error from a previous call and misattribute it, defeating the whole point of distinguishing the two.

Instead: a **per-call `HierarchySyncDiagnostics` out-parameter**.

- **Race-free** — per call, no shared instance state, so concurrent syncs can't misattribute one another's errors.
- **Backward compatible** — optional trailing param; the `null` return is unchanged, existing callers untouched.
- **Narrow / YAGNI** — only the callers that care opt in.
- The distinction is already made **per-call inside** `requestHierarchySync`: a correlated runner error rejects the wait (reaches the `catch`), while a timeout resolves `null` (reaches the `try` return). A typed `HierarchyRunnerError` carrier lets the `catch` tell a runner-reported failure apart from any other thrown cause (abort, connection failure) and write only the former to `diagnostics.runnerError`.

## Changes

- `types.ts`: new `HierarchySyncDiagnostics { runnerError?: string }`.
- `CtrlProxyHierarchy.ts`:
  - module-private `HierarchyRunnerError` carrier; `waitForFreshData`'s rejector now rejects with it instead of a bare `Error` (covers both the primary `request_hierarchy` and the `request_hierarchy_if_stale` nudge).
  - `requestHierarchySync(..., diagnostics?)`: on `error instanceof HierarchyRunnerError`, populates `diagnostics.runnerError`; still returns `null`.
  - `getAccessibilityHierarchy`: threads a diagnostics object and appends the runner text to its "Both WebSocket and sync methods failed" warn.
- `AndroidCtrlProxyClient.ts`: threads `diagnostics?` through the delegate wrapper and the `AndroidCtrlProxy` interface; `verifyServiceReady` captures it and includes the runner text in its per-attempt failure so the retry log/telemetry is attributable rather than an anonymous "no hierarchy".

No change to retry policy — whether a handler failure is deterministic-and-should-not-retry vs transient is intentionally out of scope (the issue framed it as optional).

## Acceptance criteria

- [x] A correlated runner `type:"error"` frame populates `diagnostics.runnerError` with the runner text while still returning `null` (fast-fail, no timeout).
- [x] A plain timeout leaves `diagnostics.runnerError` **undefined** — no misattribution.
- [x] An uncorrelated / unknown error id leaves `diagnostics.runnerError` undefined and the real push still resolves normally.
- [x] The `request_hierarchy_if_stale` nudge (#3061) runner error also populates `diagnostics.runnerError`.
- [x] Passing no diagnostics object preserves original behavior; all #3032 / #3061 tests remain green.
- [x] `verifyServiceReady` / `getAccessibilityHierarchy` attribute the runner text in their failure logs.

## Testing

- `bun test test/features/observe/CtrlProxyClient.test.ts` — 32 pass (4 new for #3062).
- `turbo run lint build` — clean.
- `tsc --noEmit` — no new type errors in touched files (only the pre-existing `moduleResolution` deprecation).
- Full `turbo run test` — the only failures are the pre-existing jimp/image-backend suites (`PerceptualHasher`, `JimpBackend`, `ScreenshotUtils`), untouched by this change.

Closes #3062

## Related

- Parent hang-class: #3032 (PR #3060). Sibling stale-nudge: #3061 (PR #3070). iOS surfacing/legibility: #2986 / #3017.

### DIFF
--- a/src/features/observe/android/AndroidCtrlProxyClient.ts
+++ b/src/features/observe/android/AndroidCtrlProxyClient.ts
@@ -126,6 +126,7 @@ import type {
   A11yLaunchIntentResult,
   InstalledPackageRecord,
   AndroidPerfTiming,
+  HierarchySyncDiagnostics,
 } from "./types";
 
 
@@ -1177,9 +1178,10 @@ export class AndroidCtrlProxyClient extends DeviceServiceClient implements Andro
     perf: PerformanceTracker = new NoOpPerformanceTracker(),
     disableAllFiltering: boolean = false,
     signal?: AbortSignal,
-    timeoutMs: number = 10000
+    timeoutMs: number = 10000,
+    diagnostics?: HierarchySyncDiagnostics
   ): Promise<{ hierarchy: AccessibilityHierarchy; perfTiming?: AndroidPerfTiming[] } | null> {
-    return this.hierarchy.requestHierarchySync(perf, disableAllFiltering, signal, timeoutMs);
+    return this.hierarchy.requestHierarchySync(perf, disableAllFiltering, signal, timeoutMs, diagnostics);
   }
 
   async requestHierarchySyncWithoutObservationStreamPush(
@@ -1748,14 +1750,20 @@ export class AndroidCtrlProxyClient extends DeviceServiceClient implements Andro
       async attempt => {
         logger.debug(`[CTRL_PROXY] Verifying service ready (attempt ${attempt}/${maxAttempts})`);
 
-        const hierarchyResult = await this.requestHierarchySync(new NoOpPerformanceTracker(), false, undefined, timeoutMs);
+        // Capture the runner's structured error text (issue #3062) so a deterministic handler
+        // failure is attributable in the retry log, not collapsed into an anonymous "no hierarchy".
+        const diagnostics: HierarchySyncDiagnostics = {};
+        const hierarchyResult = await this.requestHierarchySync(new NoOpPerformanceTracker(), false, undefined, timeoutMs, diagnostics);
 
         if (hierarchyResult && hierarchyResult.hierarchy) {
           logger.debug(`[CTRL_PROXY] Service verified ready after ${attempt} attempt(s)`);
           return true;
         }
 
-        throw new Error(`Verification attempt ${attempt} returned no hierarchy`);
+        const runnerErrorSuffix = diagnostics.runnerError
+          ? `: runner error: ${diagnostics.runnerError}`
+          : "";
+        throw new Error(`Verification attempt ${attempt} returned no hierarchy${runnerErrorSuffix}`);
       },
       {
         maxAttempts,
--- a/src/features/observe/android/AndroidCtrlProxyClient.ts
+++ b/src/features/observe/android/AndroidCtrlProxyClient.ts
@@ -650,7 +650,8 @@ export interface AndroidCtrlProxy extends CtrlProxyClient {
     perf?: PerformanceTracker,
     disableAllFiltering?: boolean,
     signal?: AbortSignal,
-    timeoutMs?: number
+    timeoutMs?: number,
+    diagnostics?: HierarchySyncDiagnostics
   ): Promise<{ hierarchy: AccessibilityHierarchy; perfTiming?: AndroidPerfTiming[] } | null>;
 
   requestHierarchySyncWithoutObservationStreamPush(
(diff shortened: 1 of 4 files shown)


kaeawc:
## Multi-perspective review summary

Ran three independent reviews (systems-correctness / architecture-direction / test-completeness). Consensus: the change is correct and fully closes #3062. One actionable finding was addressed.

**Correctness (concurrency/error-propagation):** No bug. `HierarchyRunnerError` survives `perf.track`'s `try/finally` re-throw untyped-wrapped, so `instanceof` at the catch holds. The "race-free / no-misattribution" claim verified: `diagnostics` is per-call, the rejector map is keyed by unique `req_`/`stale_` ids, and the `settled` guard makes resolve-null and reject-runner mutually exclusive — a timeout can never read a `runnerError`. Abort resolves `null` (never sets it). `cleanup()` unregisters on every settle path.

**Direction (API design):** The per-call out-parameter is a reasonable 4th option — race-free vs. option 2 (stash-on-client) and far lower blast radius than option 1 (discriminated union would force all ~9 callers to handle a new shape for a signal only 2 want). No iOS parity gap: the entire error-frame-correlation mechanism (#3032/#3060/#3061) is Android-only; iOS `CtrlProxyHierarchy` has no `pendingHierarchyRejectors` to surface.

**Completeness/tests:** Fully closes #3062. The 4 contract-level tests cover the full truth table (correlated-primary / timeout / uncorrelated-id / stale-nudge). Flagged that the *consumption* was only tested at the contract level — and, more substantively, that in `verifyServiceReady` the runner text only reached a per-attempt `logger.debug` (dropped at the default INFO level), so the default-visible failure log still collapsed a handler failure into an anonymous "no hierarchy".

**Addressed (261b419bf):** `verifyServiceReady` now tracks the last runner error across attempts and appends it to the terminal `Service not ready` **warn** (default-visible), plus a consumption test spying on `logger.warn` proving the text reaches an observable log line.

**Deferred (follow-up issue to be filed):** `verifyServiceReady` still retries on a deterministic runner handler failure — the issue framed retry-avoidance as optional/out-of-scope, so it's left for a follow-up.

PR #3094, opened 2026-07-04
test(db): mechanical guard for the file-backed DB flake-avoidance pattern + drop harness generic (#3081)

### LINKED ISSUE #3081: Follow-ups from withFileBackedDb harness (#3046/PR #3078): enforcement + generic cleanup + optional hooks binder
## Context

Follow-up from PR #3078 (closes #3046), which added the shared `test/db/withFileBackedDb.ts` harness (`createFileBackedDbHarness()`) and migrated the file-backed DB-lifecycle suites onto it. During that PR's multi-perspective review, three items were deliberately deferred as non-blocking (YAGNI / out of scope). This issue tracks them.

Background: #3046 is the third round of the same windows-latest file-backed sqlite test flake class (#2916 → #2923, #2992 → #3040). The harness makes the flake-avoidance pattern the *documented* path of least resistance (fresh isolated module import, tracked temp dirs + bounded `removeTempDbDir`, `getDatabase()` then `await ensureMigrations()` before close, `WINDOWS_FILE_DB_TEST_TIMEOUT_MS`). It does **not** yet mechanically prevent a new suite from hand-rolling the anti-pattern — that enforcement is the highest-value item below.

### 1. Structural / lint enforcement (highest value)
The harness is opt-in prose today: a new file-backed DB lifecycle suite can still `mkdtemp` + raw `import("../../src/db/database.ts?...")` + hand-restore env and never touch `withFileBackedDb.ts`. Since this is the *third* recurrence of the identical flake, "prose hasn't stopped the recurrence" applies to the harness too — it is better prose, not enforcement. Add a mechanical guard so the anti-pattern fails fast, e.g. a lint rule (or a `test/db/` meta-test) that flags, in `test/db/*.test.ts`:
- a raw `import(".../database.ts?...")` cache-busted module import outside `freshDatabaseModule.ts` / `withFileBackedDb.ts`;
- a hand-rolled unbounded `removeTempDirWithRetry` (the `attempt < 100` loop that #3046 deleted three copies of);
- a `mkdtemp` of an `auto-mobile` DB dir not funneled through the harness.

### 2. Drop the `<M>` generic type parameter (cleanup)
`test/db/withFileBackedDb.ts` threads `M extends LifecycleDatabaseModule = FreshDatabaseModule` through 4 interfaces plus an `as unknown as () => Promise<M>` cast on the default importer. There is only one real module type (`FreshDatabaseModule`); every consumer gets it. The generic exists only so the fake-injected unit tests can pass a narrow `LifecycleDatabaseModule`. Per CLAUDE.md ("grow the interface when the second consumer arrives"), simplify to a non-generic surface (`importer: () => Promise<FreshDatabaseModule>`) and have the fakes cast at the single injection site. Keep the narrow `LifecycleDatabaseModule` interface as the documented minimal contract + fake target. Weigh against the mild ergonomic cost of extra casts in `withFileBackedDb.test.ts`.

### 3. Optional `bindFileBackedDbHarness()` hooks binder (only if the boilerplate bites)
Every migrated suite repeats `let harness = createFileBackedDbHarness(); beforeEach(reassign); afterEach(cleanup)` plus local `setEnv`/`makeTempDir` aliases (~4 lines/suite). A binder that registers the hooks and returns a getter would remove it. Explicitly YAGNI for now (a fourth abstraction layer chasing ~4 lines); file only so it isn't forgotten. Do NOT build until the repetition actually causes a bug or a second concrete need appears.

## Non-goals
- `dbWriteBarrierResetOnClose` stays on the `:memory:` sentinel from #3047 — not file-backed, intentionally exempt.
- `backupDatabaseFile` uses `BunDatabase` directly (not the `database.ts` module) — not a lifecycle suite.

## References
- PR #3078 (the harness), issue #3046, and its review thread.
- Prior flake rounds: #2916 / PR #2923, #2992 / PR #3040, #3047 / PR #3058 (the `:memory:` sentinel), #2849 (Windows file-DB test portability).

kaeawc:
## Summary

Closes #3081. Follow-up to the `withFileBackedDb` harness (#3046 / PR #3078), the third round of the `windows-latest` file-backed sqlite test flake class (#2916 → #2923, #2992 → #3040). The harness made the correct pattern the *documented* path of least resistance but was still opt-in prose — a new `test/db/*.test.ts` suite could hand-roll the anti-pattern and never touch it. Since "prose hasn't stopped the recurrence" applies to the harness too, this adds the **mechanical enforcement** the issue flagged as highest value, plus the generic cleanup. Item 3 (the optional hooks binder) is intentionally left unbuilt per the issue (YAGNI).

### Item 1 — mechanical anti-pattern guard (highest value)
- **`test/db/fileBackedDbAntiPattern.ts`** (new) — a pure `(fileName, source) → violations` detector for the exact three signatures #3046 deleted:
  1. `raw-cache-busted-import` — a raw `import(".../database.ts?…")` (must go through the canonical `importFreshDatabaseModule()` / harness);
  2. `hand-rolled-temp-dir-retry` — a local `removeTempDirWithRetry` / `attempt < 100` EBUSY·EPERM·ENOTEMPTY loop (must use the one bounded `removeTempDbDir`);
  3. `unfunneled-mkdtemp` — a fresh-module lifecycle suite calling `mkdtemp`/`mkdtempSync` directly instead of the harness's tracked `makeTempDbDir` / `openLifecycleTestDb`.
- **`test/db/fileBackedDbAntiPattern.test.ts`** (new) — fixture unit tests for each signature (positive + the legitimate patterns that must NOT trip it, e.g. a long poll loop, `EBUSY` in a classification string, Node's bounded `rmSync({ maxRetries })`, the canonical `removeTempDbDir`, `makeFakeMkdtemp`), **plus a meta-test** that runs the detector over the real `test/db/*.test.ts` tree and asserts it is clean. The detector's own fixture file is excluded from the tree scan.
- **`test/db/unitTestDbGuard.test.ts`** — migrated off its ad-hoc `database.ts?guard-test=${Date.now()}-${Math.random()}` import onto the canonical `importFreshDatabaseModule()`. Removes the last raw cache-busted import in `test/db` and a repo-banned `Math.random()` randomness source.

### Item 2 — drop the `<M>` generic (cleanup)
- **`test/db/withFileBackedDb.ts`** — removes the `<M extends LifecycleDatabaseModule = FreshDatabaseModule>` generic threaded through four interfaces plus the `as unknown as () => Promise<M>` cast on the default importer. There is one real module type; the surface is now concrete on `FreshDatabaseModule` and the default importer is `importFreshDatabaseModule` with no cast. The narrow `LifecycleDatabaseModule` interface stays as the documented minimal contract + fake target.
- **`test/db/withFileBackedDb.test.ts`** — the fakes now cast at a single `asFreshModule()` injection helper (the sanctioned trade for dropping the generic).

## Non-goals / notes
- **Item 3 (`bindFileBackedDbHarness()` hooks binder)** — intentionally NOT built; the issue marks it YAGNI until the ~4-line boilerplate actually causes a bug or a second concrete need. Tracked for follow-up.
- The guard is a fast *textual* backstop for the recurring inline copy, not a full AST parser — variable-assembled specifiers, `unlink`/`rimraf` removals, or transitively-reached fresh modules can slip through (documented in the detector). The real enforcement remains the primitives it points at; a stricter AST guard is unwarranted until a variant recurs.
- `dbWriteBarrierResetOnClose` (`:memory:` sentinel, #3047) and `backupDatabaseFile` (`BunDatabase` directly) are not file-backed lifecycle suites and are untouched.

## Verification
- `bun test test/db/` → **595 pass, 0 fail** (55 files).
- **Mutation-verified**: reintroducing all three anti-patterns into a throwaway scanned file turns the meta-test red with all three flagged; removing it returns to green.
- `bun run build.ts` clean; `tsc --noEmit` reports only the pre-existing `moduleResolution` deprecation; eslint clean on all changed files.
- Detector + meta-test run purely textually (no real DB/fs/clock), well under the 100ms budget.

### DIFF
--- a/test/db/fileBackedDbAntiPattern.test.ts
+++ b/test/db/fileBackedDbAntiPattern.test.ts
@@ -88,6 +88,36 @@ describe("findFileBackedDbAntiPatterns detector (issue #3081)", () => {
       expect(rules(source)).toContain("hand-rolled-temp-dir-retry");
     });
 
+    test("flags a `while (attempt < n)` retry-shaped loop that removes a dir and gates on a lock code", () => {
+      const source = [
+        "let attempt = 0;",
+        "while (attempt < 100) {",
+        "  attempt += 1;",
+        "  try {",
+        "    await rm(dir, { recursive: true });",
+        "    break;",
+        "  } catch (error) {",
+        '    if (error.code !== "ENOTEMPTY") throw error;',
+        "  }",
+        "}",
+      ].join("\n");
+      expect(rules(source)).toContain("hand-rolled-temp-dir-retry");
+    });
+
+    test("does NOT flag a generic `while (cond)` poll even with a lock code and rm nearby", () => {
+      // A generic poll condition is not retry/attempt/backoff-shaped, so the
+      // structural conjunction must not fire — this guards tempDbDir.test.ts (the
+      // bounded remover's own test, which holds the quoted lock codes) from a
+      // false positive if it ever grows a real-fs poll + cleanup case.
+      const source = [
+        "while (!ready) {",
+        "  await rmSync(dir);",
+        '  if (lastCode === "EBUSY") ready = check();',
+        "}",
+      ].join("\n");
+      expect(rules(source)).not.toContain("hand-rolled-temp-dir-retry");
+    });
+
     test("does NOT flag a long POLL loop that never removes a dir or gates on a lock code", () => {
       const source = [
         "for (let attempt = 0; attempt < 500; attempt += 1) {",
@@ -163,6 +193,49 @@ describe("findFileBackedDbAntiPatterns detector (issue #3081)", () => {
     });
   });
 
+  describe("sanctioned primitive homes are exempted (they ARE the primitive)", () => {
+    test("freshDatabaseModule.ts may hold the raw cache-busted import", () => {
+      const source = "return import(`../../src/db/database.ts?fresh-db-module=${moduleCounter}`);";
+      const found = findFileBackedDbAntiPatterns("test/db/freshDatabaseModule.ts", source);
+      expect(found).toEqual([]);
+    });
+
+    test("tempDbDir.ts may hold the bounded attempt-loop + lock codes", () => {
+      const source = [
+        'const codes = new Set(["EBUSY", "EPERM", "ENOTEMPTY"]);',
+        "for (let attempt = 0; attempt < maxAttempts; attempt += 1) {",
+        "  try { await rm(dir); return; } catch (e) { if (!codes.has(e.code)) throw e; }",
+        "}",
+      ].join("\n");
+      const found = findFileBackedDbAntiPatterns("test/db/tempDbDir.ts", source);
+      expect(found).toEqual([]);
+    });
+
+    test("withFileBackedDb.ts may call mkdtemp for its tracked makeTempDbDir", () => {
+      const source = [
+        'import { importFreshDatabaseModule } from "./freshDatabaseModule";',
+        "const dir = await mkdtemp(path.join(tmpdir(), prefix));",
+      ].join("\n");
+      const found = findFileBackedDbAntiPatterns("test/db/withFileBackedDb.ts", source);
+      expect(found).toEqual([]);
+    });
+
+    test("the exemption is per-rule: freshDatabaseModule.ts still flags an UNfunneled mkdtemp", () => {
+      // It's exempt only for the cache-busted import it owns — a different
+      // anti-pattern in the same file must still be caught.
+      const source = [
+        "return import(`../../src/db/database.ts?fresh=${n}`);",
+        "await importFreshDatabaseModule();",
+        'const dir = await mkdtemp("am-");',
+      ].join("\n");
+      const rulesFound = findFileBackedDbAntiPatterns("test/db/freshDatabaseModule.ts", source).map(
+        v => v.rule
+      );
+      expect(rulesFound).not.toContain("raw-cache-busted-import");
+      expect(rulesFound).toContain("unfunneled-mkdtemp");
+    });
+  });
+
   test("a clean lifecycle suite that uses the harness produces zero violations", () => {
     const source = [
       'import { createFileBackedDbHarness } from "./withFileBackedDb";',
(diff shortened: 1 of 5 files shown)


kaeawc:
## Multi-perspective review (3 agents) + response

Ran three independent adversarial reviews before merge — a CI-flake veteran, a TS type-system purist, and a pragmatic YAGNI maintainer.

**Type-system purist** (item 2, drop `<M>` generic): all `non-issue-verified`. No consumer ever instantiated a narrower `M` (every call site is bare `createFileBackedDbHarness()`); suites need `FreshDatabaseModule` at full width for `getMigrationsError`/`getMigrationsPromiseForTest`, still statically visible. The `asFreshModule` cast is test-only, single-seam, and actually *removes* a prior production-path cast. `LifecycleDatabaseModule` remains load-bearing as the fake's contract.

**YAGNI maintainer**: "ship it". Meta-test (vs eslint plugin) is the right-sized call; runs under `turbo run test` on macos + windows CI so it has real teeth, not just locally. Item 3 correctly not built. Notable: the PR eats its own dogfood — `unitTestDbGuard.test.ts` was itself a live instance of the banned import, migrated here.

**CI-flake veteran**: real enforcement for the shape that actually recurred (verbatim copy-paste every round). Raised two `[should-fix]` findings, both **addressed in `3a9d6f2`**:
- **Sibling-`.ts`-helper blind spot** — the meta-test scanned only `*.test.ts`, but this harness *promotes* extracting helpers into `.ts` files; a hand-rolled loop moved into a sibling helper was invisible. The scan now covers every `test/db/*.ts`, with the sanctioned primitive homes exempted per-rule. Mutation-verified: a `removeTempDirWithRetry` in a non-test helper now fails the meta-test.
- **`while (…)` over-match** — `MANUAL_RETRY_LOOP` matched any `while(`, a latent false-positive on `tempDbDir.test.ts`. Tightened to require a retry/attempt/backoff-shaped condition; added fixtures both ways.

Remaining textual-guard escapes (variable-assembled import specifier; `unlink`/`rimraf` removal) are documented in the detector as accepted trades — the recurrence has always been an inline copy, and an AST guard is unwarranted until a variant appears (CLAUDE.md YAGNI).

`bun test test/db/` → **601 pass / 0 fail**; build + eslint + tsc clean.

Issue #3051, opened 2026-07-04
actions-diff-observe: sign off the diff format on real-device output (#3026 acceptance)


kaeawc:
## Context
Follow-up to the consolidated output-reduction spec **#3026**, delivered by merged **PR #3034** (`--actions-diff-observe` + `--actions-no-observe`, commit `09e49a71d`). Supersedes #2761/#2762.

Both #2761 and #3026 list an explicit acceptance item that a unit test cannot satisfy:

> Diff output format signed off on first real output before finalizing.

The diff shape shipped is `{ isDiff: true, added, removed, changed, fields }`, where `added`/`removed` are `{ key, attributes }` (readable leaf key = `resource-id + bounds + text + sibling index`, matched on a globally-positional ancestor path key), `changed` is `{ key, changes: { <attr>: { from, to } } }`, and `fields` are changed top-level scalars. See `src/features/observe/output/ObserveResultOutput.ts` (`diffObserveResult`) and `src/server/finalizeToolResponse.ts`.

## Task
Run `--actions-diff-observe` against a **real device/emulator** (not the synthetic + frozen-fixture unit tests) through a representative interaction loop (e.g. `tapOn` a checkbox, `swipeOn`/scroll, `inputText`) and confirm:
1. The diff is actually agent-consumable — an LLM can act on `added`/`removed`/`changed`/`fields` without the full tree.
2. `changed` fires cleanly on state toggles (checked/selected/focused) with the expected `{from,to}`.
3. Scroll/insert behavior (positional-key cascade — see the follow-up on stable node identity) is acceptable in practice or motivates the stable-ID work.
4. The byte/token reduction on a real screen matches the fixture measurement (a one-node change diffed to <10% of the full observation on `android-home.json`).

If the shape needs revision, adjust before recommending the flag for general use. The flag is **off by default**, so the merged plumbing is safe to sit ahead of this sign-off.

## Pointers
- Impl: `src/features/observe/output/ObserveResultOutput.ts`, `src/server/finalizeToolResponse.ts`, baseline in `src/daemon/sessionManager.ts` (`lastRenderedObservation`).
- Tests: `test/features/observe/output/diffObserveResult.test.ts`, `test/server/finalizeToolResponse.test.ts`.
- Local override to test runner changes if needed: see `AUTOMOBILE_CTRL_PROXY_*` env vars.

kaeawc:
## Partial sign-off from PR #3096

[...]

PR #3096, opened 2026-07-04
feat(observe): internal no-diff for nav/setup + element-diff size guard + structural-id prototype (#3087, #3064, #3088)

### LINKED ISSUE #3064: actions-diff-observe: add a byte-size regression guard for element-field diffs (follow-up to #3059)
## Context

Follow-up to merged **PR #3059** (closes #3052), which added diffing of the top-level Element mirror fields (`focusedElement`, `accessibilityFocusedElement`, `awaitedElement`) + `awaitDuration` to `diffObserveResult` (`src/features/observe/output/ObserveResultOutput.ts`), emitting `{from,to}` into `ObserveDiff.fields`.

During the adversarial review of #3059, the architecture reviewer flagged that emitting the *full* `Element` as `{from,to}` could re-embed the element's `node` child subtree and undercut the token-reduction purpose of `--actions-diff-observe`. This was verified real (`parseNodeBounds` shallow-copies the source node; CtrlProxy nodes have no `.$` wrapper, so a focused *container* retains its children). PR #3059 **addressed** it structurally via `leanElementForDiff`, which strips the `node` subtree from both the compare and the emitted value.

## Deferred work

The strip is proven by unit tests that assert `.node` is absent on the emitted `{from,to}` (`test/features/observe/output/diffObserveResult.test.ts`: "the emitted focusedElement is stripped of its `node` child subtree", "a mirror element whose only change is inside its `node` subtree is not reported"). What is **not** yet guarded is a **byte-size regression** assertion: nothing fails CI if a future change reintroduces subtree re-embedding (or otherwise bloats the element-field diff).

The repo already has the harness for this — the `measureValue` helper (`test/fixtures/observe/observeFixture.ts`, from #2755) and the existing size-proof test ("on a real observation, a one-node change diffs far smaller than the full observation", which asserts `< 10%` of the full observation).

### Proposed

Add a size-proof test in `diffObserveResult.test.ts` for the **focus-change** case: build a baseline/next from `loadAndroidHomeObserve()` where `focusedElement` is set to a *rich* element carrying a deep `node` subtree, change the focus target, and assert the resulting diff's `measureValue(diff).bytes` stays a small fraction of the full observation (i.e. the subtree is not re-embedded). This turns the current structural guarantee into a measured, regression-proof one.

## Pointers
- `leanElementForDiff` + the element-field loop in `diffObserveResult` — `src/features/observe/output/ObserveResultOutput.ts`.
- `measureValue` + `loadAndroidHomeObserve` — `test/fixtures/observe/observeFixture.ts`.
- Existing size-proof pattern — `test/features/observe/output/diffObserveResult.test.ts` ("diffs far smaller than the full observation").

## Non-goals / decided in #3059
- Whether to diff the two *focus* mirrors at all (redundant with node `focused`/`accessibility-focused` flips): **decided to keep** all three, now that the subtree strip bounds their cost, for consumer ergonomics (the mirror exists so a reader gets a direct answer without correlating node flips).
- A separate `elements` section on `ObserveDiff` vs. reusing the `fields` map: **decided to reuse `fields`** (uniform `{from,to}` shape; element values are bounded after the strip).

### LINKED ISSUE #3087: Mark navigation/setup internal tool calls no-diff (DefaultUIStateSetup, NavigateTo) — baseline pollution under --actions-diff-observe
## Context
Follow-up to **PR #3080** (issue #3053), which added `FinalizeToolResponseContext.internal` and threaded `__internalNoDiff` from `PlanExecutor`'s step calls so a plan step's finalized envelope is never diffed/stripped and never advances the agent-facing diff baseline (`SessionCacheData.lastRenderedObservation`).

During that PR's multi-perspective review it was confirmed the guard was applied to **all three** `PlanExecutor` internal wrapped-handler call sites (sequential step, parallel-exec step, failure-recovery `observe`). But `PlanExecutor` is **not** the only internal caller of the *wrapped* `tool.handler` (the one that runs `finalizeToolResponse`).

## Problem
These navigation/setup paths also call the wrapped `tool.handler` with an injected `sessionUuid`, so under `--actions-diff-observe` they run through finalize's diff/baseline machinery:

- `src/features/navigation/DefaultUIStateSetup.ts:161` (`swipeOnTool.handler`), `:272`, `:356`, `:459` (`tapTool.handler` / `swipeTool.handler`)
- `src/features/navigation/NavigateTo.ts:317` (`tool.handler(interaction.args)`)

None of them read `.observation.viewHierarchy` off the result today (`DefaultUIStateSetup` reads only top-level `success`/`found`; `NavigateTo` ignores the result), so there is **no correctness bug**. But because these are non-`observe` actions with the flag on, each one:
1. computes a diff that no one consumes, and
2. calls `baselineStore.set`, **advancing the agent-facing baseline** with a post-action observation the agent never saw — the exact baseline-pollution class fixed for the failure-recovery `observe` in #3080.

So the agent's next real action diffs against an internal navigation step's observation instead of the last thing the agent actually saw.

## Task
Mark these internal navigation/setup tool calls internal too (set `__internalNoDiff = true` on the params after `schema.parse`, mirroring `PlanExecutor.ts`), OR introduce a single shared "internal tool call" helper so every internal wrapped-handler invocation opts in uniformly (better: removes the per-call-site footgun). Then a future internal `.observation` reader on any of these paths stays correct, and internal navigation never pollutes the agent baseline.

### Evaluation criteria
- With `--actions-diff-observe` on, a `DefaultUIStateSetup` / `NavigateTo` internal tool call does not advance `lastRenderedObservation`.
- `success`/`found` reads in `DefaultUIStateSetup` still work (full observation, not stripped).
- A unit/integration test pins it (mirror `test/plan/planExecutorInternalNoDiffE2E.test.ts`).

## Pointers
- Guard: `src/server/finalizeToolResponse.ts` (`ctx.internal`), `src/server/toolRegistry.ts` (`args.__internalNoDiff`), `INTERNAL_PARAMS` in `src/features/record/McpCallRecorder.ts`.
- Prior art threading: `src/utils/plan/PlanExecutor.ts` (three marked sites).

### LINKED ISSUE #3088: Explore structural-id content identity for in-place edits + optional pre-move key in diff (follow-up to #3053 part 1)
## Context
Follow-up to **PR #3080** (issue #3053 part 1), which added content-hash node identity to `diffObserveResult` (`src/features/observe/output/ObserveResultOutput.ts`). After positional `pathKey` matching, a leftover removed+added node is re-paired into one `changed` delta when they share a **stable content key** — `resource-id / view-id / content-desc / text` (NUL-joined, excludes bounds & sibling index) — that is **unique among the leftovers on both sides**. This collapses scroll/insert bounds churn into `changed` deltas while staying false-merge-safe.

### 1. In-place content edits still cascade to remove+add
Because `text` and `content-desc` are part of the identity key, a node whose **label changes in place** (same position, edited text) gets a different content key on each side and reads as **remove+add**, not `changed`. This is correct-and-safe but misses a real compaction opportunity (a list row whose text updates in place). Pinned by the test `a content-key field change (text edited in place) reads as remove+add, not changed`.

**Explore:** keying identity on the **structural** id only (`resource-id` / `view-id`) with `text`/`content-desc` demoted to *changed attributes*. Weigh the compaction win against the false-merge risk from **reused/recycled resource-ids** (RecyclerView rows commonly share a resource-id — the uniqueness-on-both-sides guard mitigates but reused ids are far more likely to be non-unique, so most would fall back to positional anyway). Quantify the actual token win on a scroll+edit fixture before committing; it may not be worth the added false-merge surface.

### 2. Re-paired `changed` entry carries only the next-side key
`repairByContentIdentity` emits `key: addedNode.key` (post-move positional key). A consumer cannot recover the **pre-move** key from a re-paired `changed` entry. No reader needs it today (no native Kotlin/Swift consumer maps diff `changed` by key), so this is informational. If a future consumer needs to locate the pre-move node, add an optional `fromKey` to `ObserveDiffNodeChange` for re-paired entries only.

## Task
Prototype limitation 1 behind a config toggle, measure token savings vs. false-merge incidence on real scroll+edit fixtures, and decide whether to adopt. Address limitation 2 only if/when a consumer needs the pre-move key.

## Pointers
- `contentIdentityKey`, `repairByContentIdentity`, `diffObserveResult` in `src/features/observe/output/ObserveResultOutput.ts`.
- Existing edge tests + `measureValue` size harness in `test/features/observe/output/diffObserveResult.test.ts`.
- Parent spec: #3053 part 1 ("weigh token savings vs. added complexity and false-merge risk").

kaeawc:
Closes #3087. Closes #3064. Closes #3088.

_#3051 (real-device sign-off) stays open — see its thread. Follow-ups filed: #3106 (pre-existing `getTool("swipe")` dead code), #3107 (structural-id adoption revisit, gated on #3051), #3108 (optional `callInternal` seam)._

## Summary

Three follow-ups to the `--actions-diff-observe` output-reduction work, all in the observe-diff / finalize path:

- **#3087** — Internal navigation/setup tool calls were advancing the agent-facing diff baseline.
- **#3064** — No CI guard against re-embedding an element's `node` subtree in an element-field diff.
- **#3088** — Prototype + measure + decide on structural-id content identity for in-place edits.

A fourth follow-up, **#3051** (real-device diff-format sign-off), is addressed with findings in the Non-goals section — it needs a fully live flag-enabled run that this shared/worktree setup can't provide.

### #3087 — Mark navigation/setup internal tool calls no-diff
`DefaultUIStateSetup` and `NavigateTo` call the **wrapped** `tool.handler` (the one that runs `finalizeToolResponse`) on behalf of an internal caller. Under `--actions-diff-observe` each such call ran through finalize's diff machinery and — the actual bug — called `baselineStore.set`, **advancing the agent-facing baseline** (`SessionCacheData.lastRenderedObservation`) with a post-action observation the agent never saw. So the agent's next real action diffed against an internal setup step instead of the last thing it actually saw (same baseline-pollution class fixed for the failure-recovery `observe` in #3080).

- New `src/server/internalToolCall.ts`: `INTERNAL_NO_DIFF_PARAM` constant + non-mutating `markInternalToolCall(args)` helper (single source of truth for the marker).
- `toolRegistry.ts` (the wrapper reader) and `McpCallRecorder.INTERNAL_PARAMS` now reference the constant instead of a bare literal.
- Routed the **5** navigation/setup sites through the helper — `DefaultUIStateSetup` (setup scroll swipe, setup tap, bottom-sheet dismiss swipe, close-button tap) and `NavigateTo` (tool-call replay).
- Refactored PlanExecutor's **3** existing sites (sequential step, parallel-exec step, failure-recovery `observe`) to the same helper — removing the per-call-site footgun the issue called out.
- `markInternalToolCall` returns a **shallow copy**, so `NavigateTo`'s shared nav-graph `interaction.args` is never mutated.

Result: a `DefaultUIStateSetup` / `NavigateTo` internal call no longer advances `lastRenderedObservation`, and the caller's own `success`/`found` reads still see the full (unstripped) observation.

### #3064 — Byte-size regression guard for element-field diffs
The `node`-subtree strip (`leanElementForDiff`, #3059) was proven only structurally (`.node` absent on the emitted `{from,to}`). Added a **measured** guard: a focus change whose `focusedElement` carries the entire hierarchy as its `node` subtree must diff to `< 10%` of the full observation. If the strip ever regresses, both `{from,to}` re-embed ~the whole tree and the assertion fails (verified red by temporarily disabling the strip).

### #3088 — Structural-id content identity (prototype, off by default)
Added `DiffObserveConfig.structuralIdentity` (default **off**, unwired from any CLI flag). When on, the re-pair identity key is `resource-id`/`view-id` only, with `text`/`content-desc` demoted to ordinary `changed` attributes — so a node whose **label changes in place** collapses to one `changed` delta instead of remove+add. Tests measure the compaction win on a scroll+edit fixture and pin the trade-off: the uniqueness-on-both-sides guard still blocks a merge on ambiguous recycled ids, but a lone reused id unique on both sides **will** merge two logically-distinct nodes.

**Decision:** keep the flag off/unwired. The additional win over the shipped content-identity diff is marginal (only in-place edits) while the false-merge surface on recycled `resource-id`s is real; revisit with the real-device measurement in #3051. Limitation 2 (pre-move `fromKey`) stays deferred — no consumer needs it (YAGNI).

`finalizeToolResponse` calls `diffObserveResult` with **no cfg**, so `structuralIdentity` is unreachable in production and this PR is behaviorally inert there except the intended #3087 baseline fix.

## Verification

- New/changed tests: `test/server/internalToolCall.test.ts` (4), `test/features/navigation/navigationInternalNoDiff.test.ts` (2), `test/features/observe/output/structuralIdentityDiff.test.ts` (10), and the #3064 size guard in `diffObserveResult.test.ts`.
- `bun test test/features/observe/output/ test/features/navigation/ test/plan/ test/server/ test/features/record/` -> **1347 pass / 0 fail**.
- The #3064 guard verified **red** with the strip disabled, **green** with it restored.
- `bun run build.ts` -> OK. `bunx eslint` on all changed files -> clean. `tsc --noEmit` -> no new errors in touched files.
- Full `bun test`: the only failures (39) are the pre-existing `@jimp/wasm-webp`-missing image failures on this machine (JimpBackend / ContrastChecker / memory-leak), independent of this change — none of my files touch image/jimp code.
- Live emulator `observe` confirmed real device output is shape-identical to what `diffObserveResult` consumes.

## Non-goals / follow-ups

- **#3051 (real-device diff-format sign-off):** the shared MCP daemon runs the **main** build with the flag off, so a true end-to-end `--actions-diff-observe` run against this branch isn't possible here. The real-captured `android-home.json` fixture already drives the byte-reduction (`< 10%`) and `changed {from,to}` assertions, and live `observe` output was confirmed shape-compatible. The residual — a local flag-enabled daemon driven through a real `tapOn`/`swipeOn`/`inputText` loop with an LLM acting on the diff — stays open on #3051.
- **#3088 adoption / CLI wiring:** deliberately deferred pending the #3051 real-device token-win vs false-merge data.
- **#3088 limitation 2** (optional pre-move `fromKey` on re-paired `changed` entries): deferred — no diff `changed`-by-key consumer exists (YAGNI).

### DIFF
--- a/src/features/navigation/DefaultUIStateSetup.ts
+++ b/src/features/navigation/DefaultUIStateSetup.ts
@@ -4,6 +4,7 @@ import { AdbClient } from "../../utils/android-cmdline-tools/AdbClient";
 import { logger } from "../../utils/logger";
 import { AndroidCtrlProxyClient } from "../observe/android";
 import { ToolRegistry } from "../../server/toolRegistry";
+import { markInternalToolCall } from "../../server/internalToolCall";
 import { NavigationEdge, UIState } from "./NavigationGraphManager";
 import { ModalState, ScrollPosition } from "../../utils/interfaces/NavigationGraph";
 import { UIStateExtractor } from "./UIStateExtractor";
@@ -157,8 +158,11 @@ export class DefaultUIStateSetup implements UIStateSetup {
         swipeOnArgs.speed = scrollPosition.speed;
       }
 
-      // Execute swipeOn with lookFor
-      const result = await swipeOnTool.handler(swipeOnArgs);
+      // Execute swipeOn with lookFor. Marked internal (#3087) so that under
+      // `--actions-diff-observe` this setup scroll neither diffs its observation
+      // nor advances the agent-facing diff baseline — the `found` read below still
+      // sees the full (unstripped) result.
+      const result = await swipeOnTool.handler(markInternalToolCall(swipeOnArgs));
 
       // `swipeOn` pre-serializes into an MCP envelope via createStructuredToolResponse,
       // which hoists only `success`/`error` to the top level — `found` lives under
@@ -269,7 +273,8 @@ export class DefaultUIStateSetup implements UIStateSetup {
         args.elementId = element.resourceId;
       }
 
-      await tapTool.handler(args);
+      // Internal setup tap (#3087): no diff/strip, no baseline advance.
+      await tapTool.handler(markInternalToolCall(args));
 
       // Small delay for UI to update
       await this.sleep(100);
@@ -352,12 +357,13 @@ export class DefaultUIStateSetup implements UIStateSetup {
       try {
         const swipeTool = ToolRegistry.getTool("swipe");
         if (swipeTool) {
-          // Swipe down from middle of screen to dismiss bottom sheet
-          await swipeTool.handler({
+          // Swipe down from middle of screen to dismiss bottom sheet.
+          // Internal setup swipe (#3087): no diff/strip, no baseline advance.
+          await swipeTool.handler(markInternalToolCall({
             action: "swipe",
             direction: "down",
             platform
-          });
+          }));
           await this.sleep(200);
 
           // Verify dismissal
@@ -456,11 +462,12 @@ export class DefaultUIStateSetup implements UIStateSetup {
 
     for (const text of closeTexts) {
       try {
-        await tapTool.handler({
+        // Internal close-button tap (#3087): no diff/strip, no baseline advance.
+        await tapTool.handler(markInternalToolCall({
           action: "tap",
           text,
           platform
-        });
+        }));
         logger.debug(`[UI_STATE_SETUP] Tapped close button: "${text}"`);
         return true;
       } catch (error) {
--- a/src/features/navigation/NavigateTo.ts
+++ b/src/features/navigation/NavigateTo.ts
@@ -5,6 +5,7 @@ import { logger } from "../../utils/logger";
 import { AndroidCtrlProxyClient } from "../observe/android";
 import { createGlobalPerformanceTracker } from "../../utils/PerformanceTracker";
 import { ToolRegistry } from "../../server/toolRegistry";
+import { markInternalToolCall } from "../../server/internalToolCall";
 import {
   NavigationGraphManager,
   ToolCallInteraction,
@@ -313,8 +314,11 @@ export class NavigateTo {
 
     logger.info(`[NAVIGATE_TO] Replaying tool call: ${interaction.toolName}`);
 
-    // Call the tool handler with the original args
-    await tool.handler(interaction.args);
+    // Call the tool handler with the original args, marked internal (#3087):
+    // `markInternalToolCall` returns a copy, so the stored edge `interaction.args`
+    // is never mutated. Under `--actions-diff-observe` this replay neither diffs
+    // its observation nor advances the agent-facing diff baseline.
+    await tool.handler(markInternalToolCall(interaction.args as Record<string, unknown>));
   }
 
   /**
--- a/src/features/observe/output/ObserveResultOutput.ts
+++ b/src/features/observe/output/ObserveResultOutput.ts
@@ -315,6 +315,22 @@ export interface DiffObserveConfig {
    * by the path key. Set `false` for exact positional-only behavior.
    */
   contentIdentity?: boolean;
+  /**
+   * Structural-only node identity (issue #3088 — **prototype**, default **off**).
+   * Requires `contentIdentity` (has no effect when that is off). When on, the
+   * re-pair identity key is keyed on the *structural* id alone (`resource-id` /
+   * `view-id`), with `text` / `content-desc` demoted to ordinary *changed*
+   * attributes rather than identity. This lets a node whose **label changes in
+   * place** (same position, edited text) collapse to a single `changed` delta
+   * instead of remove+add — a real compaction win for list rows whose text
+   * updates. The trade-off is a larger false-merge surface: recycled/reused
+   * `resource-id`s (common for RecyclerView rows) are far more likely to be
+   * non-unique among the leftovers, so most fall back to positional matching, but
+   * a lone reused id that is unique-on-both-sides *can* mis-merge two unrelated
+   * nodes. Off by default and not wired to any CLI flag pending a real-device
+   * token-win vs. false-merge measurement (see the #3088 tests + issue #3051).
+   */
+  structuralIdentity?: boolean;
 }
 
 /**
@@ -548,11 +564,33 @@ function contentIdentityKey(attrs: Record<string, unknown>): string | null {
   return [resourceId, viewId, contentDesc, text].join(" ");
 }
 
-/** Index leftover diff nodes by their content-identity key, dropping keyless nodes. */
-function indexByContentKey(nodes: ObserveDiffNode[]): Map<string, number[]> {
+/**
+ * A node's *structural* identity key (issue #3088 prototype): `resource-id /
+ * view-id` only, NUL-joined — `text` and `content-desc` are deliberately excluded
+ * so a node keeps this key across an in-place label edit (they become ordinary
+ * `changed` attributes instead). Returns `null` when the node carries no
+ * structural id (both empty) — the same "empty key is not identity" rule as
+ * `contentIdentityKey`. Distinct key space from `contentIdentityKey`; only one is
+ * ever used per diff (selected by `DiffObserveConfig.structuralIdentity`).
+ */
+function structuralIdentityKey(attrs: Record<string, unknown>): string | null {
+  const resourceId = String(attrs["resource-id"] ?? "");
+  const viewId = String(attrs["view-id"] ?? "");
+  if (resourceId === "" && viewId === "") {
+    return null;
+  }
+  // NUL-joined, matching `contentIdentityKey`'s convention.
+  return [resourceId, viewId].join(" ");
+}
+
+/** Identity-key function for a diff node's attributes; `null` = no stable identity. */
+type IdentityKeyFn = (attrs: Record<string, unknown>) => string | null;
+
+/** Index leftover diff nodes by an identity key, dropping keyless nodes. */
+function indexByContentKey(nodes: ObserveDiffNode[], keyFn: IdentityKeyFn): Map<string, number[]> {
   const byKey = new Map<string, number[]>();
   nodes.forEach((node, index) => {
-    const key = contentIdentityKey(node.attributes);
+    const key = keyFn(node.attributes);
     if (key === null) {
       return;
     }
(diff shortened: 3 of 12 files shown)


kaeawc:
## Multi-perspective review — findings addressed

Ran three independent review passes (adversarial regression / architecture-direction / test-completeness). Summary and how each was handled:

**Adversarial regression pass — no real bugs.** Traced the marker end-to-end (`toolRegistry.ts:190` reads `args[INTERNAL_NO_DIFF_PARAM]` at wrapper entry → `internal:true` → `finalizeToolResponse.ts:112-116` forces `canDiff=false` → `baselineStore.set` skipped). Confirmed: no import cycle (`internalToolCall.ts` is a zero-import leaf), recording is byte-identical (`__internalNoDiff` was already in `INTERNAL_PARAMS`; only the literal→constant swap changed), `structuralIdentity` is production-inert (`finalizeToolResponse` calls `diffObserveResult` with no cfg), all 5 nav + 3 PlanExecutor `.handler(` sites marked, and the NUL (`\x00`) key joins are correct (verified by byte inspection, not the blank-rendering Read tool).

**Architecture pass — ship it.** Shared-helper approach is the right call; the two-step `tool.handler(markInternalToolCall(args))` is an acceptable stopping point given the 8 heterogeneous call sites. Optional future nicety noted below.

**Test-completeness pass — one real gap, now closed.** The nav test proved a *proxy* (marker reaches the handler), not the issue's end-to-end "baseline not advanced" outcome. **Added `test/features/navigation/navigateToInternalNoDiffE2E.test.ts`**, mirroring `test/plan/planExecutorInternalNoDiffE2E.test.ts`: drives a real `NavigateTo` replay through the wrapped handler + `finalizeToolResponse` + baseline store under `--actions-diff-observe` and asserts `getLastRenderedObservation` is unchanged. Verified it goes **red** when the marker is removed (baseline advances) and green with it — so it catches a future nav-unmarking regression the plan E2E can't. Also fixed the CI lint failure (unused `beforeEach` import).

### Follow-ups filed separately (out of scope here)
- **Latent dead-code bug (pre-existing):** `DefaultUIStateSetup.dismissTopModal` calls `ToolRegistry.getTool("swipe")`, but only `swipeOn` is registered — so bottom-sheet dismissal-by-swipe always falls through to the back-button path. Unchanged by this PR; filing an issue.
- **Optional `ToolRegistry.callInternal(name, args)` seam** to remove the residual "forgot to wrap" possibility — deferred as a nicety.

## #3051 (real-device diff sign-off) — findings
The shared MCP daemon runs the **main** build with `--actions-diff-observe` off, so a true end-to-end flag run against this branch can't happen from this environment. What I confirmed: a live emulator `observe` returns output shape-identical to what `diffObserveResult` consumes, and the real-captured `android-home.json` fixture already drives the `< 10%` byte-reduction and `changed {from,to}` assertions. The residual — a local flag-enabled daemon driven through a real `tapOn`/`swipeOn`/`inputText` loop with an LLM acting on the diff — stays open on #3051 with these notes.

Issue #3109, opened 2026-07-05
ci(ios): XCTestRunner videoRecording step fails deterministically — unit-test DB guard (#3082) trips the integration test


kaeawc:
## Summary

The **XCTestRunner Simulator Tests** CI job (`ios-xctest-runner-simulator-tests` in `.github/workflows/pull_request.yml`) fails **deterministically** at the "Run videoRecording MP4 integration test" step on every run of a branch containing `e4d893ba7` (PR #3082). Observed on 6/6 runs from 2026-07-04 21:56Z onward; zero passes since.

```
Error: Failed to start video recordings: Error: Unit test resolved the real file-backed
database at /Users/runner/.auto-mobile/auto-mobile.db ...
```

## Root cause (traced to file:line)

PR #3082 (issue #3067) added a bun-test preload (`test/setup/unitTestDbGuard.ts`, wired via `bunfig.toml` `[test].preload`) that sets `AUTOMOBILE_UNIT_TEST_DB_GUARD=1`, making `getDatabase()` throw when a test resolves the **default** `~/.auto-mobile/auto-mobile.db` without an explicit override.

The job's video step runs `bun test test/integration/iosVideoRecordingStartStop.integration.test.ts` (via `scripts/ios/video-recording-start-stop-integration.sh:52`), so the preload arms the guard. The test drives the real `videoRecording` start handler, whose first DB touch trips the guard:

- `startVideoRecording` (`src/server/videoRecordingManager.ts:533`) → `getVideoRecordingDependencies()` builds `new VideoRecordingRepository()` with no injected db
- `initializeVideoRecordingState` (`videoRecordingManager.ts:114`) → `VideoRecordingRepository.getDb()` (`src/db/videoRecordingRepository.ts:138`) → `ensureMigrations()` → `getDatabase()` → `resolveDbPath()` → `assertUnitTestDbAccessAllowed()` (`src/db/database.ts:257`) **throws** (no `AUTOMOBILE_DB_DIR`/`AUTOMOBILE_DB_PATH` is set anywhere in the job/script/composite action).

## Why it landed latent

`bunfig.toml` and `test/**` are **not** in the workflow's `native_integration` path filter (`pull_request.yml:253-284`), so PR #3082's own CI **skipped** this job. The break shipped undetected and was discovered by the next unrelated PR that triggered the job. On-merge (main) runs never execute this job at all, so there is no main-branch signal.

## Timeline

[...]

Issue #3067, opened 2026-07-04
test-infra: guard against unit tests resolving the real file-backed getDatabase()


kaeawc:
## Summary

Surfaced during review of #3066 (fix for #3063). Several unit tests silently exercise the **real file-backed database** (`getDatabase()`) instead of an injected in-memory DB. Because `getDatabase()`'s first-use path runs migrations + file IO on **real wall-clock time**, any test that asserts on the *result* of an async DB write (rather than the pre-write state) races that write and flakes — which is exactly how #3063 failed deterministically:

- The nav-graph write in `recordNavigationEvent` only assigns `currentScreen` after the DB commit; a test that only yielded microtasks read `null`.
- The sibling `should seed navigation graph from hierarchy updates` test only "passed" because it asserts the `null` pre-write state — racy-by-luck.

#3066 fixed the two affected tests by injecting an in-memory DB via `createTestDatabase` + a new test-only `NavigationGraphManager.setInstanceForTesting()`. But nothing *prevents* the next test from reaching for the real DB again.

## Additional evidence

Even after #3066, `recordNavigationEvent`'s post-commit **fire-and-forget** `TelemetryRecorder.getInstance().recordNavigationEvent(...)` (`src/features/navigation/NavigationGraphManager.ts:472-476`) still writes to the **real `~/.auto-mobile` file DB** on a floating promise. A reviewer proved this empirically: running the two `CtrlProxyClient.test.ts` tests under `AUTOMOBILE_DB_DIR=$tmp` creates `auto-mobile.db` + a ~2.9 MB WAL. It does not affect determinism (the assertion reads `currentScreen`, assigned before the telemetry call) and produced no open-handle warnings, but unit tests should not touch the user's real DB.

## Ask

Add a unit-test guard that **fails loudly when the real `getDatabase()` / file-backed DB is resolved during unit tests**, forcing tests to inject an in-memory DB (`createTestDatabase`) instead. Consider:
- A test-mode flag (e.g. env var set by the bun test preload) that makes `resolveDbPath()` / `getDatabase()` throw unless an explicit in-memory override was installed.
- Or a shared test harness that resets/mocks the DB + `TelemetryRecorder` singletons so fire-and-forget telemetry writes are captured, not persisted (mirrors the pattern already in `test/features/navigation/NavigationGraphManager.test.ts`, which `spyOn`s `TelemetryRecorder.getInstance().recordNavigationEvent`).

[...]

PR #3114, opened 2026-07-05
fix(ci): inject an in-memory test DB into the XCTestRunner videoRecording integration test (#3109)

kaeawc:
## Summary

Fixes the deterministic failure of the **"XCTestRunner Simulator Tests"** CI job at the *"Run videoRecording MP4 integration test"* step:

```
Error: Failed to start video recordings: Error: Unit test resolved the real file-backed
database at /Users/runner/.auto-mobile/auto-mobile.db ...
```

**Root cause.** The bun-test preload `test/setup/unitTestDbGuard.ts` (wired via `bunfig.toml` `[test].preload`) arms `AUTOMOBILE_UNIT_TEST_DB_GUARD`, so `assertUnitTestDbAccessAllowed` in `src/db/database.ts:149` throws whenever a test resolves the **default** `~/.auto-mobile/auto-mobile.db` path (issues #3067 / #3082). The integration test `test/integration/iosVideoRecordingStartStop.integration.test.ts` drives the real `videoRecording` start handler, which reaches the DB through the manager's default repositories:

- `src/server/videoRecordingManager.ts:142-143` construct `new VideoRecordingRepository()` / `new VideoRecordingConfigRepository()` with no injected db,
- `src/db/videoRecordingRepository.ts:138-144` (and the config repo's `getDb`) then call `ensureMigrations()` + `getDatabase()`,
- which resolves the default path and trips the guard at `src/db/database.ts:156`.

## This fix (option **A2** — inject an in-memory DB via the existing DI seam)

The manager already exposes a DI seam: `setVideoRecordingManagerDependencies(...)` / `resetVideoRecordingManagerDependencies()` (`src/server/videoRecordingManager.ts:154,191`), and both repositories accept an injected Kysely instance in their constructors (`videoRecordingRepository.ts:134`, `videoRecordingConfigRepository.ts:12`).

Before the start handler runs, the test now injects repositories backed by `createTestDatabase()` (`test/db/testDbHelper.ts`), which opens a fresh **migrated** `:memory:` `bun:sqlite` connection directly — it never routes through `resolveDbPath()`, so `getDatabase()`'s default path (and therefore the guard) is never reached. `resetVideoRecordingManagerDependencies()` remains in the existing `finally` block to restore global state.

Why A2: the DI seam and injectable-db constructors already exist, so the fix is a few lines in the test with no production-code change and no new helper.

## TRADEOFF (honest, vs option A1)

This makes the integration test run against an **in-memory** DB instead of the real **file-backed** one the daemon actually uses at runtime. It therefore **stops exercising real file-DB semantics** on this path — WAL mode / `busy_timeout` pragmas, on-disk migration lifecycle, and the daemon's real `getDatabase()` path resolution are no longer touched by this test. The alternative option **A1** keeps a real temp-dir file DB (via `AUTOMOBILE_DB_DIR`), preserving file-DB semantics at the cost of temp-dir setup/cleanup. This PR is the **A2** option among competing PRs for #3109; pick A1 if preserving real file-DB coverage on this test matters more than minimal test-only wiring. File-DB semantics remain covered by the dedicated file-backed DB harness suites (`test/db/withFileBackedDb.ts`), so the loss here is scoped to this one MP4 pipeline test whose actual purpose is the start -> stop -> non-empty playable MP4 assertion (unchanged).

## Testing

The integration test body itself is gated on `AUTOMOBILE_IOS_VIDEO_RECORDING_INTEGRATION=1` and needs a booted simulator + ffmpeg, so it cannot run to completion in a plain CI/dev shell.

- `bunx eslint test/integration/iosVideoRecordingStartStop.integration.test.ts --fix` — passes.
- `bunx turbo run build` — passes.
- **DI-wiring validation** (throwaway, not committed): with the guard armed by the bunfig preload, injecting `createTestDatabase()`-backed repos via `setVideoRecordingManagerDependencies` and then driving `listActiveVideoRecordings()` / `listVideoRecordings()` (which run `getVideoRecordingDependencies` -> `initializeVideoRecordingState` -> `recordingRepository.listRecordings()`, the exact read that resolves `getDatabase()`) does **not** throw. A negative control confirmed that **without** the injection the same call **does** throw `/real file-backed database/`, proving the fix is load-bearing rather than a no-op.
- Did not run the full `bun test` suite (~39 pre-existing unrelated image-backend failures — JimpBackend/ContrastChecker/etc.).

Refs #3082, #3067, #2632.

Refs #3109

### DIFF
--- a/test/integration/iosVideoRecordingStartStop.integration.test.ts
+++ b/test/integration/iosVideoRecordingStartStop.integration.test.ts
@@ -5,9 +5,15 @@ import path from "node:path";
 import { promisify } from "node:util";
 import { registerVideoRecordingTools } from "../../src/server/videoRecordingTools";
 import { ToolRegistry } from "../../src/server/toolRegistry";
-import { resetVideoRecordingManagerDependencies } from "../../src/server/videoRecordingManager";
+import {
+  resetVideoRecordingManagerDependencies,
+  setVideoRecordingManagerDependencies,
+} from "../../src/server/videoRecordingManager";
 import { serverConfig } from "../../src/utils/ServerConfig";
 import { defaultTimer } from "../../src/utils/SystemTimer";
+import { createTestDatabase } from "../db/testDbHelper";
+import { VideoRecordingRepository } from "../../src/db/videoRecordingRepository";
+import { VideoRecordingConfigRepository } from "../../src/db/videoRecordingConfigRepository";
 
 const execFileAsync = promisify(execFile);
 const RUN_INTEGRATION = process.env.AUTOMOBILE_IOS_VIDEO_RECORDING_INTEGRATION === "1";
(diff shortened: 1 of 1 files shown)


kaeawc:
Superseded by **A1 (#3111)**, which the maintainer chose to ship for #3109 (now merged to main, verified green in CI). A1 preserves the integration test's real file-backed DB semantics via an `AUTOMOBILE_DB_DIR` temp dir, whereas this DI-inject approach switched the test to `:memory:` and couldn't exercise its own CI job until A3 (#3113) landed. Closing to keep the option set clean; reopen if the DI approach is preferred later.

PR #3118, opened 2026-07-05
fix(navigation): dismissTopModal bottom-sheet uses swipeOn, not unregistered swipe (#3106)

### LINKED ISSUE #3106: DefaultUIStateSetup.dismissTopModal calls getTool("swipe") — never registered, so bottom-sheet swipe-dismiss is dead code
## Context
Surfaced during the multi-perspective review of PR #3096 (follow-ups to `--actions-diff-observe`). Pre-existing, unrelated to that PR — filing separately.

## Problem
`DefaultUIStateSetup.dismissTopModal` (`src/features/navigation/DefaultUIStateSetup.ts`, the `modal.type === "bottomsheet"` branch) resolves the swipe tool via:

```ts
const swipeTool = ToolRegistry.getTool("swipe");
if (swipeTool) {
  await swipeTool.handler(markInternalToolCall({ action: "swipe", direction: "down", platform }));
  ...
}
```

But there is **no tool registered under the name `"swipe"`** — the registered interaction tool is `"swipeOn"` (see `src/server/interactionTools.ts` and `ToolRegistry`). So `getTool("swipe")` always returns `undefined`, the `if (swipeTool)` block never executes, and **bottom-sheet dismissal-by-swipe is dead code** — it silently falls through to the back-button fallback below it.

This means a bottom sheet that can only be dismissed by a swipe-down (not by back) will fail to dismiss during navigation setup.

## Fix
Use `getTool("swipeOn")` with the correct `swipeOn` arg shape (`direction`, optional `lookFor`/`container`), or route through the same seam the rest of the class uses. Add a unit test that pins the bottomsheet branch actually invokes the swipe tool (a spy on the registered `swipeOn` handler), so a wrong tool name regresses loudly.

## Pointers
- `src/features/navigation/DefaultUIStateSetup.ts` — `dismissTopModal`, bottomsheet branch.
- Registered names: `src/server/interactionTools.ts` (`swipeOn`, not `swipe`).

kaeawc:
Closes #3106.

## Summary

`DefaultUIStateSetup.dismissTopModal` resolved the swipe tool in its bottom-sheet branch via `ToolRegistry.getTool("swipe")`, but **no tool is registered under the name `"swipe"`** — the registered interaction tool is `"swipeOn"` (`src/server/interactionTools.ts`). So `getTool("swipe")` always returned `undefined`, the `if (swipeTool)` block never ran, and swipe-down dismissal of bottom sheets was **dead code** that silently fell through to the back-button fallback below it. A bottom sheet that only dismisses by swipe-down (not by back) would fail to dismiss during navigation setup.

Surfaced during the multi-perspective review of #3096; pre-existing and unrelated to that PR, so filed and fixed separately.

### Fix the tool name + arg shape (`src/features/navigation/DefaultUIStateSetup.ts`)
- Resolve `ToolRegistry.getTool("swipeOn")` instead of the never-registered `"swipe"`.
- Pass the correct `swipeOn` arg shape — `{ direction: "down", platform }`, dropping the bogus `action: "swipe"` field that `swipeOn`'s schema does not accept.
- Keep the `markInternalToolCall(...)` wrapper (#3087) so this setup swipe neither diffs/strips its observation nor advances the agent-facing diff baseline.

### Inject a `Timer` (test speed)
- `DefaultUIStateSetup` now takes an optional `Timer` (default `defaultTimer`), used by the private `sleep()`. This lets the new unit tests drive an auto-advancing `FakeTimer` so the `dismissTopModal` `sleep(200)` delays resolve immediately, keeping them <100ms and non-flaky. Backward-compatible — no production call site changes.

### Regression tests (`test/features/navigation/DefaultUIStateSetup.test.ts`)
- Spy on the registered `swipeOn` handler and assert the bottom-sheet branch invokes it exactly once with `direction: "down"`, `platform`, the `__internalNoDiff` marker, and **no** `action` field — so a wrong tool name regresses loudly.
- A second test registers a tool under the old name `"swipe"` and asserts it is **never** resolved.

## Verification

- `bun test test/features/navigation/DefaultUIStateSetup.test.ts` -> **7 pass / 0 fail**; both new tests verified **red** against the pre-fix `getTool("swipe")` and **green** after the fix.
- `bun test test/features/navigation/` -> **303 pass / 0 fail**.
- `bunx turbo run lint build` -> clean.

## Non-goals / follow-ups

- No production call site passes a custom `Timer`; the injection exists purely as a test seam (YAGNI-sized to the one consumer).

### DIFF
--- a/src/features/navigation/DefaultUIStateSetup.ts
+++ b/src/features/navigation/DefaultUIStateSetup.ts
@@ -11,7 +11,7 @@ import { UIStateExtractor } from "./UIStateExtractor";
 import { RealObserveScreen } from "../observe/ObserveScreen";
 import { getStructuredField } from "../../utils/toolUtils";
 import { UIStateSetup } from "./interfaces/UIStateSetup";
-import { defaultTimer } from "../../utils/SystemTimer";
+import { defaultTimer, Timer } from "../../utils/SystemTimer";
 
 /**
  * Default implementation of UIStateSetup that handles UI state alignment
@@ -29,14 +29,17 @@ export class DefaultUIStateSetup implements UIStateSetup {
   private device: BootedDevice;
   private adb: AdbClient;
   private observeScreenProvider: () => ObserveScreenLike;
+  private timer: Timer;
 
   constructor(
     device: BootedDevice,
     adb: AdbClient,
-    observeScreenProvider?: () => ObserveScreenLike
+    observeScreenProvider?: () => ObserveScreenLike,
+    timer: Timer = defaultTimer
   ) {
     this.device = device;
     this.adb = adb;
+    this.timer = timer;
     // The setup holds a resolved AdbClient (not a factory), so wrap it in a
     // trivial factory to satisfy ObserveScreen's factory-only contract (matches
     // the AndroidCtrlProxyClient.getInstance call below).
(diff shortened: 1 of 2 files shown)


kaeawc:
### Multi-perspective review pass (3 reviewers) — addressed

Ran three independent reviews (runtime/regression, test-quality, direction). Summary:

- **Regression review:** no substantive issues — the revived branch mirrors the existing `setupScrollPosition` handler pattern, args are schema-valid, timer injection is safe.
- **Test-quality review:** tests solid, red-before-green verified, branch genuinely pinned (not vacuous). One *optional* gap noted (fallback-to-back-button coverage) — deferred to a follow-up issue since that path is pre-existing and unchanged here.
- **Direction review — substantive, fixed in 2dd3a88a6:** with `autoTarget` defaulting `true` and no `lookFor`/`container`, `swipeOn` targets a scrollable child (`SwipeOn.execute`), so a bare `{ direction: "down" }` would scroll the sheet's inner list instead of dragging the sheet down to dismiss it. Now passes `autoTarget: false` to force `executeScreenSwipe` (the mid-screen downward drag a dismissal needs), and the regression test pins `autoTarget === false`. This makes the branch not just *live* but *correct*.

`bun test test/features/navigation/` → all pass; `turbo run lint build` → clean.

kaeawc:
Deferred coverage from the review (optional fallback-to-back-button test for the bottom-sheet branch) filed as #3125 with full context.

Issue #3045, opened 2026-07-04
Silent-hang when a broadcast helper's own broadcast() throws (pre-existing, #3023-adjacent)


kaeawc:
## Context

Surfaced during the #3023 / PR #3039 multi-agent review (concurrency reviewer, confirmed independently by the architecture reviewer). **Pre-existing, out of #3023's scope** — filing for tracking.

## The gap

#3023 / PR #3039 fixed throws in the **action body** of a launched coroutine. But the `broadcast*Result` / `broadcast*Error` helpers in `CtrlProxy.kt` (e.g. `broadcastPreferenceFilesResult`, `broadcastSwipeResult`, `broadcastHighlightResponse`) each wrap their `webSocketServer.broadcast(...)` in an internal `try/catch` that **logs and swallows**. If that inner `broadcast()` itself throws, the helper emits nothing and the daemon awaiter for that `requestId` hangs to its `RequestManager` timeout — the same silent-hang class, one layer down.

This affects **both** the handlers wired to `AsyncActionRunner` and the ones deliberately excluded (gestures/focus/traversal/highlight), equally, and predates PR #3039. AsyncActionRunner does not help here because the throw is caught-and-swallowed inside the helper, never propagating to the runner.

## Impact

Narrow tail case (broadcast/serialization failure while sending a result), but it means the "no silent hang" guarantee isn't fully closed for any result-broadcasting handler.

## Proposed direction

Options:
- Have the `broadcast*` helpers, on an internal broadcast failure, fall back to broadcasting a correlated `ErrorResponse` (best-effort, guarded against recursive failure) instead of swallowing.
- Or centralize result-broadcasting through a helper that guarantees *some* correlated frame is sent per requestId.

## Acceptance criteria
- [ ] A `broadcast()` failure inside a result-broadcast helper yields a correlated failure frame (or is otherwise guaranteed not to hang the awaiter), not a swallowed log.
- [ ] Robolectric/unit test simulating a broadcast throw asserts a correlated frame is still emitted.

PR #3120, opened 2026-07-05
feat(android): close raw serviceScope.launch silent-hang gap via ServiceScopeGuard (#3104)

### LINKED ISSUE #3085: Enforce broadcast-helper guard adoption (ResultBroadcaster/AsyncActionRunner) so #3045/#3023 can't silently regress
## Context

Follow-up from PR #3073 (closed #3045), which introduced `ResultBroadcaster.guard(requestId, action) { … }` and routed all 31 requestId-correlated `broadcast*Result` / `broadcast*Error` / `broadcast*Response` helpers in `CtrlProxy.kt` through it, so a throw while *sending* a result emits a correlated `type:"error"` frame instead of being logged and swallowed. Squash-merged as `6cff3833e`.

Raised by the **design/direction reviewer** in the PR's multi-agent review (finding #3).

## The gap

The "no silent hang for any result-broadcasting handler" guarantee is currently enforced **only by the author having converted all 31 helpers by hand**. Nothing structurally prevents regression:

1. A **new** `broadcast*Result` helper added later can forget to wrap its body in `resultBroadcaster.guard(...)` and reintroduce the exact silent-hang bug #3045 fixed.
2. An existing guarded helper can regress by reintroducing an inner `try { … } catch (e: Exception) { Log.e(TAG, "Error broadcasting …", e) }` swallow *inside* the guard block — the guard only fires on a **throw** that escapes the block, so a stray inner swallow silently defeats it.

This is the **same latent weakness the sibling `AsyncActionRunner` already has** (from #3023) — so PR #3073 is not a regression, but neither fix is structurally enforced.

## Proposed direction

A cheap CI/test backstop, e.g. one or more of:
- A unit/Robolectric (or plain source-scanning) test asserting no `Log.e(TAG, "Error broadcasting …", e)` swallow remains in a `requestId`-carrying `broadcast*` helper.
- A test asserting every `private suspend fun broadcast*Result/Error/Response(requestId: …)` helper's body routes through `resultBroadcaster.guard(...)` (and symmetrically that fire-and-forget action launches route through `asyncActionRunner.launch(...)`).
- Consider generalizing to also cover `AsyncActionRunner` adoption.

## Acceptance criteria
- [ ] A newly-added requestId-correlated result-broadcast helper that forgets the guard (or reintroduces an inner log-and-swallow) fails CI.
- [ ] The check covers both `ResultBroadcaster.guard` and `AsyncActionRunner.launch` adoption, or is clearly scoped with the other tracked separately.

## References
- PR #3073 (merge `6cff3833e`), issue #3045
- Sibling helper: `android/control-proxy/.../AsyncActionRunner.kt` (#3023)
- Guard helper: `android/control-proxy/.../ResultBroadcaster.kt`

### LINKED ISSUE #3104: Enforce/close the raw serviceScope.launch fire-and-forget silent-hang gap (AsyncActionRunner adoption)
## Context

Follow-up from PR #3095 (closes #3085), which added `BroadcastGuardAdoptionTest` — a source-scanning backstop enforcing that requestId-correlated `broadcast*Result/Error/Response` helpers in `CtrlProxy.kt` route through `resultBroadcaster.guard(...)` (#3045) and that neither those helpers nor `asyncActionRunner.launch { … }` blocks reintroduce an inner log-and-swallow catch (#3023).

Raised by the **adversarial regression-hunter** in that PR's multi-agent review (finding FN-3), and explicitly scoped out of #3085 per its acceptance criterion 2 ("or is clearly scoped with the other tracked separately").

## The gap

The #3085 backstop enforces `AsyncActionRunner` adoption only for the **swallow-regression** (an inner log-and-swallow inside an existing `asyncActionRunner.launch` block). It **cannot** catch a brand-new fire-and-forget action that forgets `asyncActionRunner.launch` entirely and instead dispatches via a raw `serviceScope.launch { … }`:

```kotlin
serviceScope.launch {
  val r = doExpensiveActionAsync()      // throws
  broadcastActionResult(requestId, r)   // never reached -> client hangs to RequestManager timeout
}
```

A throw here escapes to the scope's `SupervisorJob` with only a log line — no correlated `type:"error"` frame is emitted and the daemon request awaiting that `requestId` hangs. This is the exact #3023 symptom, on the more-traveled dispatch path.

This is **not structurally detectable by a source-scanning test without false positives**: `CtrlProxy.kt` has ~30 raw `serviceScope.launch { … }` sites that are legitimate non-action work (e.g. line 846 logcat streaming, lines 5274/5288/5345 wrapping already-guarded `broadcast*Result` calls). A blanket "no raw `serviceScope.launch`" rule would fire on all of them. So the #3085 test deliberately does not attempt it.

## Proposed directions (pick one)

1. **Install a `CoroutineExceptionHandler` on `serviceScope`** that emits a best-effort correlated `ErrorResponse` (or at least a `type:"error"` frame) when a launched action coroutine throws. This closes the gap *by construction* for every raw launch, making the source-scanning approach unnecessary for this half. The tricky part is recovering the originating `requestId` at the handler boundary (may require threading it via `CoroutineName` or a context element).
2. **Refactor action dispatch to a single choke point** so a new action *cannot* be dispatched except through `asyncActionRunner.launch` (which already provides the correlated-error-on-throw guarantee). Then the #3085 test's swallow-regression check is sufficient.
3. **Accept the residual risk** and rely on the `asyncActionRunner` convention + review, documenting that raw `serviceScope.launch` for a requestId-correlated action is disallowed.

Direction 1 is likely the cheapest robust fix and mirrors how `AsyncActionRunner`/`ResultBroadcaster` already centralize the behavior.

## References
- PR #3095 (closes #3085); guard helpers `ResultBroadcaster.kt` (#3045/PR #3073), `AsyncActionRunner.kt` (#3023)
- Test: `android/control-proxy/src/test/kotlin/dev/jasonpearson/automobile/ctrlproxy/BroadcastGuardAdoptionTest.kt` (scope note in the class KDoc)

kaeawc:
## Summary

Closes #3104 — the raw-`serviceScope.launch` half of the silent-hang gap that PR #3095 (closing #3085) explicitly scoped out.

`CtrlProxy.kt` dispatches almost every command fire-and-forget on `serviceScope`. Two seams already close the silent-hang class on their own paths: `AsyncActionRunner` (a throw inside a launched action coroutine, #3023) and `ResultBroadcaster` (a throw while *sending* a result, #3045). But a **brand-new** requestId-correlated action dispatched via a **raw** `serviceScope.launch { … }` that throws *before* reaching its guarded broadcast is invisible to both:

```kotlin
serviceScope.launch {
  val r = doExpensiveActionAsync()      // throws
  broadcastActionResult(requestId, r)   // never reached -> client hangs to RequestManager timeout
}
```

The throw escapes to the scope's `SupervisorJob` with only a log line — no correlated `type:"error"` frame — and the daemon awaiter hangs. (Worse: an uncaught throw in a scope-root `launch` is otherwise delivered to Android's default handler, which crashes the process.) The #3085 source-scanning backstop deliberately can't flag this: `CtrlProxy.kt` has ~30 legitimate raw `serviceScope.launch` sites (logcat streaming, wrapping already-guarded `broadcast*Result` calls), so a blanket rule would be all false positives.

This implements **Direction 1** from the issue — close the gap *by construction* with a scope-level `CoroutineExceptionHandler`.

## What changed

**Production (`android/control-proxy/.../ctrlproxy/`):**

- **`ServiceScopeGuard.kt`** (new) — builds the `CoroutineExceptionHandler` installed on `serviceScope`. On any uncaught throwable other than a cooperative `CancellationException`, it logs the failure and re-launches a best-effort `ErrorResponse` broadcast, recovering the correlation id from a `RequestIdContext` element on the failed coroutine's context. A secondary failure while broadcasting the fallback is logged-and-swallowed (never re-enters the handler → no loop). Injectable broadcast sink + `logError` → Android-free and unit-tested, exactly like `AsyncActionRunner`/`ResultBroadcaster`.
- **`RequestIdContext.kt`** (new) — a `CoroutineContext.Element` (mirrors `CoroutineName`) that threads a request's correlation id into a raw launch: `serviceScope.launch(RequestIdContext(requestId)) { … }`. When absent, the guard still emits a best-effort frame with a **null** requestId — a safe no-op on the daemon (`AndroidCtrlProxyClient` only resolves when the id is present) — plus a log trace, so a raw launch that forgot to thread the id never dies silently.
- **`CtrlProxy.kt`** — construct `serviceScopeGuard` and install its handler: `serviceScope = CoroutineScope(Dispatchers.IO + SupervisorJob() + serviceScopeGuard.handler)`. `emitScope` resolves `serviceScope` lazily to break the scope↔handler construction cycle.

**Tests:**

- **`ServiceScopeGuardTest.kt`** (new) — fast, Android-free (`runTest` + `StandardTestDispatcher`): correlated error frame on throw, null requestId when no `RequestIdContext`, no frame on `CancellationException`, secondary-failure swallow (no re-entrant loop), and `RequestIdContext` key recovery.
- **`BroadcastGuardAdoptionTest.kt`** — added `serviceScope installs the ServiceScopeGuard CoroutineExceptionHandler`, a source-scan **enforcement** test so the wiring can't silently regress; updated the class KDoc scope note (the raw-launch half is now closed by construction here, not "tracked separately").

## Acceptance criteria

- [x] A new fire-and-forget action dispatched via a raw `serviceScope.launch` that throws before broadcasting no longer hangs the awaiter silently — the scope handler emits a correlated (or best-effort null-id) `type:"error"` frame and logs. *(Verified by `ServiceScopeGuardTest`.)*
- [x] Correlation id recoverable at the handler boundary — via the `RequestIdContext` context element (the issue's suggested `CoroutineName`/context-element approach).
- [x] Cooperative cancellation is never converted into a client-facing error frame.
- [x] The wiring is enforced against silent regression (source-scan test), replacing the source-scanning approach the issue noted was infeasible for this half.
- [x] No regression to the existing seams: the handler only fires on a genuinely *uncaught* throw; existing raw launches wrap guarded broadcasts or catch-and-broadcast their own error, so none double-emit.

## Testing

- `./gradlew -p android :control-proxy:testDebugUnitTest --tests "*ServiceScopeGuardTest" --tests "*BroadcastGuardAdoptionTest"` → **26 pass / 0 fail** (<1s; no Robolectric, no device).
- Full `:control-proxy:testDebugUnitTest` → green. `:control-proxy:assembleDebug` → compiles. ktfmt `--google-style` clean.

## Non-goals / follow-ups

- **Structural correlation for raw `serviceScope.launch` (Direction 2) — tracked in #3128.** `AsyncActionRunner.launch` now tags its coroutine with `RequestIdContext(requestId)`, so correlation is *live* for the recommended dispatch path (including the non-`Exception` `Throwable` tail that escapes AsyncActionRunner's own catch). Correlation for a *raw* `serviceScope.launch` is still opt-in via `RequestIdContext`; today no raw-launch site is a live silent-hang path (they wrap already-guarded `broadcast*Result` helpers or are legitimately non-correlated), so a choke point / lint that makes it structural is future-proofing, filed as #3128.

## Related

- Follow-up from PR #3095 (closes #3085); raised by the adversarial regression-hunter (finding FN-3).
- Sibling seams: `AsyncActionRunner` (#3023), `ResultBroadcaster` (#3045 / PR #3073).

### DIFF
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/AsyncActionRunner.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/AsyncActionRunner.kt
@@ -53,32 +53,39 @@ class AsyncActionRunner(
     requestId: String?,
     action: String,
     block: suspend CoroutineScope.() -> Unit,
-  ): Job = scope.launch {
-    try {
-      block()
-    } catch (e: CancellationException) {
-      // Cooperative cancellation means the service scope is shutting down — never convert it into
-      // a client-facing error frame; let it propagate so the coroutine unwinds cleanly.
-      throw e
-    } catch (e: Exception) {
-      val cause = e.message ?: e::class.simpleName ?: "unknown error"
-      logError("Async action '$action' failed (requestId=$requestId)", e)
+  ): Job =
+    // Tag the launched coroutine with [RequestIdContext] so that a throwable this handler does
+    // *not*
+    // catch — a non-[Exception] [Throwable] such as an [Error]/[AssertionError] — still escapes
+    // with
+    // the correlation id attached, letting the scope-level [ServiceScopeGuard] emit a *correlated*
+    // error frame instead of a null-id one. The common `Exception` case is handled below directly.
+    scope.launch(RequestIdContext(requestId)) {
       try {
-        broadcastResponse(
-          ErrorResponse(requestId = requestId, error = "Action '$action' failed: $cause")
-        )
-      } catch (broadcastError: CancellationException) {
-        throw broadcastError
-      } catch (broadcastError: Exception) {
-        // A failure while reporting the error must not escape and crash the launch (which would
-        // defeat the whole point). Log and swallow — the client will fall back to its timeout.
-        logError(
-          "Failed to broadcast async error for action '$action' (requestId=$requestId)",
-          broadcastError,
-        )
+        block()
+      } catch (e: CancellationException) {
+        // Cooperative cancellation means the service scope is shutting down — never convert it into
+        // a client-facing error frame; let it propagate so the coroutine unwinds cleanly.
+        throw e
+      } catch (e: Exception) {
+        val cause = e.message ?: e::class.simpleName ?: "unknown error"
+        logError("Async action '$action' failed (requestId=$requestId)", e)
+        try {
+          broadcastResponse(
+            ErrorResponse(requestId = requestId, error = "Action '$action' failed: $cause")
+          )
+        } catch (broadcastError: CancellationException) {
+          throw broadcastError
+        } catch (broadcastError: Exception) {
+          // A failure while reporting the error must not escape and crash the launch (which would
+          // defeat the whole point). Log and swallow — the client will fall back to its timeout.
+          logError(
+            "Failed to broadcast async error for action '$action' (requestId=$requestId)",
+            broadcastError,
+          )
+        }
       }
     }
-  }
 
   companion object {
     private const val TAG = "AsyncActionRunner"
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
@@ -111,7 +111,28 @@ class CtrlProxy : AccessibilityService(), CtrlProxyActions {
     const val ACTION_OPERATION_RESULT = "dev.jasonpearson.automobile.OPERATION_RESULT"
   }
 
-  private val serviceScope = CoroutineScope(Dispatchers.IO + SupervisorJob())
+  /**
+   * Emits a correlated `type:"error"` frame when *any* raw `serviceScope.launch { … }` throws
+   * uncaught — closing the raw-launch half of the silent-hang gap by construction (issue #3104).
+   * Its [ServiceScopeGuard.handler] is installed on [serviceScope] below. `emitScope` resolves
+   * [serviceScope] lazily to break the scope↔handler construction cycle (the handler re-launches
+   * its fallback broadcast on the same scope); the broadcast sink resolves [webSocketServer] lazily
+   * because it is `lateinit` (assigned in [onServiceConnected]). See [AsyncActionRunner] /
+   * [ResultBroadcaster] for the sibling seams on the dispatch and result-send paths.
+   */
+  private val serviceScopeGuard: ServiceScopeGuard =
+    ServiceScopeGuard(
+      emitScope = { serviceScope },
+      broadcastResponse = { response ->
+        if (::webSocketServer.isInitialized && webSocketServer.isRunning()) {
+          webSocketServer.broadcast(response)
+        }
+      },
+      logError = { message, error -> Log.e(TAG, message, error) },
+    )
+
+  private val serviceScope: CoroutineScope =
+    CoroutineScope(Dispatchers.IO + SupervisorJob() + serviceScopeGuard.handler)
 
   /**
    * Wraps fire-and-forget action launches so a throw inside the launched coroutine broadcasts a
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/RequestIdContext.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/RequestIdContext.kt
@@ -1,0 +1,31 @@
+package dev.jasonpearson.automobile.ctrlproxy
+
+import kotlin.coroutines.AbstractCoroutineContextElement
+import kotlin.coroutines.CoroutineContext
+
+/**
+ * A [CoroutineContext] element carrying the originating request's correlation id into a
+ * fire-and-forget `serviceScope.launch { … }`, so [ServiceScopeGuard]'s [CoroutineExceptionHandler]
+ * can emit a *correlated* error frame when that launch throws uncaught.
+ *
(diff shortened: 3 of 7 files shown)


kaeawc:
## Multi-agent review round — addressed

Ran three independent reviewers (coroutines/concurrency specialist, design/direction, test-quality/regression). **No correctness bugs** were found — nested-launch propagation, shutdown race, field-init order, re-entrancy, and the "Android crashes the process without a handler" claim were all verified sound. All three independently flagged the same real gap:

**`RequestIdContext` shipped with zero production call sites** → the guard always recovered a `null` requestId → for AsyncActionRunner-routed actions the guard downgraded a silent hang to a *logged* hang rather than a correlated fail-fast.

**Fix (pushed in `a20d10a38`):** `AsyncActionRunner.launch` now tags its coroutine with `RequestIdContext(requestId)`. AsyncActionRunner's own `catch (e: Exception)` still handles the common case; the tag ensures a non-`Exception` `Throwable` (`Error`/`AssertionError`) that escapes that catch still lands in the scope guard **with the correlation id**, so the guard emits a correlated frame instead of a null-id (daemon no-op) one. This gives `RequestIdContext` a real consumer and makes the two seams compose.

**Tests added:**
- `AsyncActionRunnerTest` — the launched coroutine carries `RequestIdContext(requestId)`.
- `ServiceScopeGuardTest` — end-to-end: an `Error` escaping `AsyncActionRunner` is correlated by the guard; and a re-entrancy test with `emitScope` = the **guarded** scope (production shape), closing the test-quality gap that the prior swallow test used a handler-free scope.

**Scoped to a follow-up (#3128):** making correlation *structural* for raw `serviceScope.launch` sites (Direction 2 choke point, or a discoverability lint). Today no raw-launch site is a live silent-hang path — they all wrap already-guarded `broadcast*Result` helpers or are legitimately non-correlated — so this is future-proofing, not an active bug.

PR #3126, opened 2026-07-05
fix(android): correlate the ADB-broadcast fallback hierarchy request into waitForFreshData (#3089)

### LINKED ISSUE #3089: Android CtrlProxy: correlate the ADB-broadcast fallback hierarchy request (sync_ uuid) into the wait
Follow-up to #3032 (PR #3060) and #3061 (PR #3070).

## Summary

The `waitForFreshData` "hang class" — a runner `type:"error"` frame that no-ops in `RequestManager` while the hierarchy caller blocks in `waitForFreshData` for a `hierarchy_update` push, hanging to the full timeout — has now been closed on the **WebSocket** paths:

- **#3032 / PR #3060**: primary `request_hierarchy` (`req_<ts>_<id>`) — correlated.
- **#3061 / PR #3070**: `request_hierarchy_if_stale` nudge (`stale_<ts>_<id>`) — correlated (gated on a primary WebSocket requestId being outstanding).

**Still uncorrelated: the ADB-broadcast fallback.** When `sendHierarchyRequest` returns `null` (WebSocket down / send failed), `requestHierarchySync` falls back to an `am broadcast` and then waits with **no** correlatable requestId:

`src/features/observe/android/CtrlProxyHierarchy.ts` (as of merge `30826cde6`):
```ts
// Fall back to ADB broadcast if WebSocket failed
if (hierarchyRequestId === null) {
  const uuid = `sync_${this.context.timer.now()}_${generateSecureId()}`;   // ~line 378
  await this.context.adb.executeCommand(
    `shell "am broadcast -a dev.jasonpearson.automobile.EXTRACT_HIERARCHY --es uuid ${uuid} --ez disableAllFiltering ${disableAllFiltering}"`,
    ...
  );
}
// hierarchyRequestId is null here, so waitForFreshData registers no rejector:
const freshData = await this.waitForFreshData(effectiveTimeoutMs, startTime, false, signal, hierarchyRequestId ?? undefined);
```

The broadcast already mints a `sync_<ts>_<id>` `uuid` and passes it to the runner via `--es uuid`, but that uuid is **not** threaded into `waitForFreshData`, so a runner `type:"error"` frame carrying it has nothing to reject — it degrades to a full `waitForFreshData` timeout. This is the exact hang class #3032/#3061 closed, on the last remaining hierarchy path.

Both #3060 and #3070 documented this as inherent-for-now; this issue tracks actually closing it.

## Proposed direction

1. Thread the broadcast `uuid` (the `sync_` id) into `waitForFreshData` as its `requestId` when the WebSocket path failed, so `pendingHierarchyRejectors` registers a rejector for it (same mechanism as `req_`/`stale_`).
2. Confirm the Android runner (`android/control-proxy`) emits a `type:"error"` frame whose `requestId` **echoes the broadcast `uuid`** when the `EXTRACT_HIERARCHY` broadcast handler fails/decodes badly. If it does not (broadcasts may not carry a WebSocket response channel back), this reduces to a runner-side change (see #2985 / PR #3019 for the WebSocket error-envelope precedent) and may be **not feasible over the broadcast transport** — in which case document that conclusively and close.
3. Add a unit test in `test/features/observe/CtrlProxyClient.test.ts` mirroring the #3032/#3061 tests but forcing the WebSocket-send failure so the ADB-broadcast fallback is taken, then injecting a `type:"error"` frame with the `sync_` uuid and asserting fast failure.

## Acceptance criteria

- [ ] A runner `type:"error"` frame whose `requestId` matches an in-flight ADB-broadcast fallback hierarchy request rejects/fails the enclosing wait fast (no `waitForFreshData` timeout), **or** a definitive finding + doc that the broadcast transport cannot carry a correlatable error frame back (then close as won't-fix with rationale).
- [ ] Non-correlatable / null requestId frames remain a safe no-op.
- [ ] Unit test driving the WebSocket-send-failure → ADB-broadcast fallback path.

## Context

- Parent chain: #3032 (PR #3060) → #3061 (PR #3070, merged `30826cde6`).
- Related runner error-envelope work: #2985 / PR #3019 (Android `type:"error"` ErrorResponse), #3017 (Android silent decode drop), iOS parity #3022.
- Key files: `src/features/observe/android/CtrlProxyHierarchy.ts` (`requestHierarchySync` ADB-fallback branch, `waitForFreshData`, `pendingHierarchyRejectors`), `src/features/observe/android/AndroidCtrlProxyClient.ts` (`type:"error"` branch), `android/control-proxy` (broadcast `EXTRACT_HIERARCHY` handler error emission).

## Note on scope (from PR #3070 review)

PR #3070 deliberately **gated** stale-nudge correlation on a primary WebSocket `requestId` being present, precisely so the ADB-broadcast fallback (and the best-effort `getLatestHierarchy` path) retain their graceful timeout→stale-cache behavior rather than rejecting to `null`. Any fix here must preserve that: a correlated fast-fail on the broadcast path should still return the stale cache where one exists, not discard it. See the `getLatestHierarchy` regression test added in PR #3070.

kaeawc:
Closes #3089.

Follow-up to #3032 (PR #3060) and #3061 (PR #3070). This closes the **last member** of the `waitForFreshData` "hang class" — a runner `type:"error"` frame that no-ops while the hierarchy caller blocks in `waitForFreshData`, degrading to the full timeout — on the remaining uncorrelated path: the ADB-broadcast fallback.

## Feasibility (issue step 2 — resolved: FEASIBLE)

Both #3060 and #3070 documented the broadcast fallback as "inherent-for-now" on the premise that the broadcast transport may carry no correlatable channel back. It does: the `EXTRACT_HIERARCHY` broadcast arrives over ADB, but the runner's handler runs with `webSocketServer` available and already pushes the success-path `hierarchy_update` over that **same WebSocket**. So a failure can be reported back over WebSocket, correlated by the broadcast's `sync_` uuid — exactly like the `req_`/`stale_` paths. This is option (A), not the won't-fix documentation option.

## Coverage scope (review clarification)

The correlated error frame reaches the daemon only when the WebSocket is still **readable** at the moment the fallback runs — i.e. the transient-send-failure / WS-flapping case (socket OPEN but `ws.send()` threw). When the socket is **fully down**, the daemon receives neither the `hierarchy_update` push nor the error frame, so that subset still degrades to the existing timeout. This PR therefore closes the **flapping subset** of the hang class on the broadcast path; the socket-down subset is inherently timeout-bound (there is no channel back). The change is harmless in the socket-down case (an undelivered frame simply never arrives) and safe against an un-re-cut older runner.

### TS — thread the `sync_` uuid into the wait (`src/features/observe/android/CtrlProxyHierarchy.ts`)
When the WebSocket `request_hierarchy` send fails, `requestHierarchySync` falls back to `am broadcast ... EXTRACT_HIERARCHY --es uuid sync_<ts>_<id>`. That uuid was minted for the runner but never threaded into `waitForFreshData`, so `pendingHierarchyRejectors` registered no rejector for it. Now the fallback's `sync_` uuid is passed as the wait's correlation `requestId`, so a runner `type:"error"` frame echoing it rejects the wait fast — the same mechanism as `req_` (#3032) and `stale_` (#3061).

A fast fail still returns `null`, so the caller (`getAccessibilityHierarchy`) keeps its stale-cache fallback — **nothing is discarded**, honoring the PR #3070 scope note. `getLatestHierarchy` is untouched and still enters `waitForFreshData` with no `requestId`, so its timeout→stale-cache behavior is preserved.

### Runner — emit a correlated error frame on extraction failure (`android/control-proxy/.../CtrlProxy.kt`)
`handleCommand`'s `ACTION_EXTRACT_HIERARCHY` branch now, on a null-hierarchy result **or** a throw (uuid in scope), emits a correlated `ErrorResponse(requestId = uuid)` over the WebSocket via a new `broadcastHierarchyExtractError` helper — in addition to the legacy ADB `ACTION_OPERATION_RESULT` (kept for non-WebSocket consumers). The helper is routed through `resultBroadcaster.guard` so a throw while *sending* the frame degrades to the daemon's timeout rather than escaping the receiver coroutine (#3045 / #3085 compliance — auto-enforced by `BroadcastGuardAdoptionTest`). `CancellationException` is rethrown, never converted to an error frame.

## Verification

- **New TS tests** (`test/features/observe/CtrlProxyClient.test.ts`, new #3089 describe block, 4 tests) driving the WebSocket-send-failure → ADB-broadcast fallback path (a capturing socket that stays OPEN but throws on `request_hierarchy` sends):
  - a `type:"error"` frame echoing the broadcast `sync_` uuid fails `requestHierarchySync` **fast** (no timeout);
  - the correlated error populates `diagnostics.runnerError` (parity with #3062);
  - an uncorrelated id during the fallback is a safe no-op and the real push still resolves;
  - a `sync_`-shaped error frame does **not** disturb the `getLatestHierarchy` stale-cache path (locks in that the correlation is scoped to `requestHierarchySync`).
  - Confirmed TDD red first (the two fast-fail tests hung before the impl), green after.
- Full `test/features/observe/` suite: **824 pass**.
- `:control-proxy:testDebugUnitTest` (incl. `BroadcastGuardAdoptionTest`, `ResultBroadcasterTest`): **BUILD SUCCESSFUL** — the new guarded helper is recognized and compliant.
- `turbo run lint build` clean; ktfmt applied.

## Delivery note

Per the runner-release convention, the Android `control-proxy` change lands in source here but the delivered runner is **not** re-cut in this feature PR — the daemon downloads a pinned release runner, and `RELEASE_CHECKSUM_REGISTRY` is bumped only by the periodic `chore: bump versions` PRs. The TS-side correlation is inert-but-safe against an older runner (an error frame that never arrives just falls through to the existing timeout); the end-to-end fast-fail activates once the runner is re-cut.

### DIFF
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
@@ -46,6 +46,7 @@ import dev.jasonpearson.automobile.protocol.BroadcastEventResponse
 import dev.jasonpearson.automobile.protocol.CrashData
 import dev.jasonpearson.automobile.protocol.CrashEvent
 import dev.jasonpearson.automobile.protocol.DeviceInfo
+import dev.jasonpearson.automobile.protocol.ErrorResponse
 import dev.jasonpearson.automobile.protocol.HandledExceptionData
 import dev.jasonpearson.automobile.protocol.HandledExceptionEvent
 import dev.jasonpearson.automobile.protocol.LifecycleEventData
@@ -79,6 +80,7 @@ import java.io.ByteArrayOutputStream
 import java.io.File
 import java.security.MessageDigest
 import kotlin.coroutines.resume
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
@@ -1871,29 +1873,46 @@ class CtrlProxy : AccessibilityService(), CtrlProxyActions {
       ACTION_EXTRACT_HIERARCHY -> {
         val uuid = intent.getStringExtra("uuid")
         if (uuid.isNullOrBlank()) {
+          // No uuid to correlate a WebSocket error frame to, so only the legacy ADB result is sent.
           sendResult(success = false, error = "UUID parameter is required")
           return
         }
 
-        val textFilter = intent.getStringExtra("text")
-        val disableAllFiltering = intent.getBooleanExtra("disableAllFiltering", false)
-        val hierarchy = extractHierarchy(textFilter, disableAllFiltering)
-        if (hierarchy != null) {
-          val filename = "hierarchy_$uuid.json"
-          writeHierarchyToFile(hierarchy, filename)
+        // The daemon's ADB-broadcast hierarchy fallback awaits this uuid over the WebSocket in
+        // waitForFreshData. On extraction failure we send a correlated error frame keyed by the
+        // uuid so that wait fails fast rather than hanging to timeout, closing the last member of
+        // the #3032/#3061 hang class (issue #3089). The legacy ADB result broadcast is retained.
+        try {
+          val textFilter = intent.getStringExtra("text")
+          val disableAllFiltering = intent.getBooleanExtra("disableAllFiltering", false)
+          val hierarchy = extractHierarchy(textFilter, disableAllFiltering)
+          if (hierarchy != null) {
+            val filename = "hierarchy_$uuid.json"
+            writeHierarchyToFile(hierarchy, filename)
 
-          // Broadcast to WebSocket clients
-          broadcastHierarchyUpdate(hierarchy)
+            // Broadcast to WebSocket clients
+            broadcastHierarchyUpdate(hierarchy)
 
-          val message =
-            if (textFilter != null) {
-              "Hierarchy extracted with text filter: '$textFilter', saved as $filename"
-            } else {
-              "Hierarchy extracted successfully, saved as $filename"
-            }
-          sendResult(success = true, data = message)
-        } else {
-          sendResult(success = false, error = "Failed to extract hierarchy")
+            val message =
+              if (textFilter != null) {
+                "Hierarchy extracted with text filter: '$textFilter', saved as $filename"
+              } else {
+                "Hierarchy extracted successfully, saved as $filename"
+              }
+            sendResult(success = true, data = message)
+          } else {
+            sendResult(success = false, error = "Failed to extract hierarchy")
+            broadcastHierarchyExtractError(uuid, "Failed to extract hierarchy")
+          }
+        } catch (e: CancellationException) {
+          // Cooperative cancellation (service scope shutting down) must never be converted into an
+          // error frame — let it propagate so the coroutine unwinds cleanly.
+          throw e
+        } catch (e: Exception) {
+          val cause = e.message ?: e::class.simpleName ?: "unknown error"
+          Log.e(TAG, "Error extracting hierarchy for uuid=$uuid", e)
+          sendResult(success = false, error = cause)
+          broadcastHierarchyExtractError(uuid, "Hierarchy extraction failed: $cause")
         }
       }
     }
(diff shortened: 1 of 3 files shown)


kaeawc:
Review complete (3 persona subagents: concurrency/race, Android runtime/protocol, test-quality). **No blocking bugs found.** Addressed inline:
- Fixed a stale comment that miscited the ADB-broadcast fallback as an uncorrelated (no-requestId) example — it now threads the `sync_` uuid (commit b0311ce45).
- Qualified the coverage narrative: this closes the **WS-flapping / transient-send-failure** subset of the hang class; when the socket is fully down the daemon has no readable channel and still degrades to timeout (harmless, and safe against an un-re-cut older runner).

Reviewers verified: no rejector leak/double-fire, unique per-wait `sync_` scoping (no cross-rejection), stale-cache preserved on fast-fail, mutation-check red (reverting the correlation line times both fast-fail tests out), deterministic <100ms tests.

Follow-ups filed: #3130 (pre-existing outer `commandReceiver` catch swallows `CancellationException`), #3131 (behavioral Robolectric test for the runner's `EXTRACT_HIERARCHY` error-frame emission).

Issue #3141, opened 2026-07-05
Reduce MCP output token cost for agentic runs (inert --actions-no-observe, element subtree bloat, pretty-printed JSON)


RyeBread1026:
Three things inflate MCP output token cost, which compounds significantly in agentic runs since the full observation is re-read from cache every turn:

1. **`--actions-no-observe` is a no-op** - it parses into `ServerConfig` but the interaction-tool response path never reads it, so `tapOn`/`inputText`/etc. always return the full post-action observation. Stripping `elements`/`viewHierarchy` from action results (leaving `observe` intact) cut a `tapOn` result from ~5.6k to ~0.4k tokens.

2. **`elements` includes each node's full child subtree** - `observe` results are ~65KB where ~24KB carries the same interactive info.

3. **Tool results are pretty-printed** (`JSON.stringify(…, 2)`) - ~35% of an element-heavy payload is indentation whitespace, pure token cost with no value to the model.

Impact: together, roughly half of our per-run token cost (MCP mode).

RyeBread1026:
Update: `--actions-no-observe` is already implemented on `main` (output-only at `finalizeToolResponse`) - that first item was written against an older base, so disregard it.

The other two — flat `elements` (drop the nested `node` subtrees) and a new `--tool-results-compact-json` flag — are in #3142.

kaeawc:
**Scope/lineage note:** Item 1 here (`--actions-no-observe` is a no-op) is the **strip-flag wiring** that was nominally owned by #3026 (and originally #2762) but never actually shipped — PR #3034 only wired the `--actions-diff-observe` diff half. #3026 is now closed with the diff half done+signed-off (#3051) and the strip half delegated here, so **#3141 is the single live tracker** for it. No duplicate work outstanding.

**Impact:** this is likely the highest-ROI output-reduction work in the backlog — the three items together are ~half the per-run MCP token cost in agentic mode. Suggest splitting into 3 deliverables so they can land independently:
1. Wire `--actions-no-observe` to strip `elements`/`viewHierarchy` from action-tool results (leave `observe` intact) — ~5.6k→~0.4k per `tapOn`.
2. Trim each node's full child subtree from `elements` (~65KB→~41KB observe).
3. Drop the pretty-print indentation on tool results (`JSON.stringify(…,2)` → no-indent) — ~35% of element-heavy payloads.
[...]

PR #3142, opened 2026-07-05
perf(mcp): reduce observe/tool-result token cost; name recordings by outputName

RyeBread1026:
Three MCP output-size / usability improvements for agentic runs, where each turn re-reads the whole transcript so tool-result size compounds.

## 1. Flatten `observe` `elements`

Each entry in `elements.{clickable,scrollable,text,media}` carried its full nested `node` subtree, so a single scrollable container serialized its entire list (10s of KB per element). `elements` are meant to be flat descriptors (id/text/bounds/…); callers that need the tree already have `viewHierarchy`.

Root cause: `ElementParser.parseNodeBounds` spread the whole node, and `extractNodeProperties` returns attributes-only for the xml2js format but the entire node (including children) for the current CtrlProxy JSON hierarchy.

Measured on a dense Android screen: `elements` **52.8KB → 24.9KB**, whole observe **74KB → 46KB** - back under the ~50KB limit where MCP clients (e.g. Claude Code) persist the result to disk and hand the agent a truncated stub, effectively blinding it.

## 2. `--tool-results-compact-json` flag

Drop the 2-space pretty-printing in `stringifyToolResponse`. ~35% fewer characters on element-heavy payloads, identical data (parses back the same; no effect on `tapOn`/text matching). Off by default; composes with the other output-reduction flags.

## 3. Name recordings by `outputName`

`VideoRecorderService` named recording folders/files by the recordingId UUID; the caller-supplied `outputName` (e.g. a test-case id) was only stored as metadata. Now the dir/file is `<sanitized outputName>-<recordingId>`, falling back to the UUID when no name is given. The recordingId suffix keeps paths unique per recording, so delete/evict (which removes the per-recording directory) stays safe.

## Tests
- compact-json serialization + `ServerConfig` toggle
- recording naming: `outputName` path, sanitization, UUID fallback

Refs #3141 (the `--actions-no-observe` piece from that issue is already implemented on main).


RyeBread1026:
Superseded by https://github.com/kaeawc/auto-mobile/pull/3143 — moved to a branch on the repo (normal contributor flow) instead of a fork.

PR #3169, opened 2026-07-05
fix(ios): correct storage_changed recorder mapping (valueType/changeType) (#3001)

### LINKED ISSUE #3001: fix(ios): recordStorageEvent call passes undeclared 'operation' and omits required valueType/changeType (masked by Bun no-typecheck)
## Summary
The iOS SDK event ingestor calls `recorder.recordStorageEvent(...)` with an object shape that **does not match** the `TelemetryRecorder.recordStorageEvent` input type: it passes an undeclared `operation` field and **omits the required `valueType` and `changeType`** fields.

```ts
// src/features/observe/ios/IosSdkEventIngestor.ts:~245
case "storage_changed":
  await recorder.recordStorageEvent({
    timestamp: ts, applicationId,
    fileName: (p.suiteName as string) ?? "", key: (p.key as string) ?? "",
    value: (p.value as string) ?? "", operation: (p.operation as string) ?? "write",
  });
  break;
```

The recorder's input shape (`src/features/telemetry/TelemetryRecorder.ts`, `recordStorageEvent`) declares `valueType: string | null` and `changeType: string` (both required) and has no `operation` field.

## Why it hasn't surfaced
Bun's bundler skips type-checking (project memory: "Bun bundler skips type checking"), so this strict-TS error never fails the build. At runtime the iOS storage events are recorded with `value_type = undefined` and `change_type = undefined` (persisted as NULL/undefined), and the intended `operation` is dropped on the floor.

## What to do
- Decide the mapping: `operation` ("write"/"delete"/…) → `changeType`, and derive/pass `valueType`.
- Fix the call site to pass `valueType` and `changeType` and stop passing `operation` (or add `operation` to the input shape and map it inside the recorder).
- Consider a lightweight `tsc --noEmit` type-check gate for `src/` so this class of masked type error is caught (currently nothing does).

## Context
Found by the regression reviewer during #2798 / PR #2999 review. Pre-existing on `main`, unrelated to and unaffected by that PR — filed separately so it isn't lost.

kaeawc:
## Summary

Fixes #3001. The iOS `storage_changed` SDK-telemetry call site in
`src/features/observe/ios/IosSdkEventIngestor.ts` passed an **undeclared
`operation`** field and **omitted the required `valueType` and `changeType`**
fields, violating `TelemetryRecorder.recordStorageEvent`'s input type. Bun's
bundler skips type-checking, so this strict-TS error (`TS2353`) never failed the
build — at runtime iOS storage rows were persisted with `value_type` /
`change_type` = NULL and the intended change semantics dropped on the floor.

Tracing the **real iOS SDK wire format** (`SdkStorageChangedEvent` in
`ios/auto-mobile-sdk/Sources/AutoMobileSDK/Events/SdkEvent.swift`) surfaced a
second latent bug: the payload serializes as
`suiteName` / `key` / `newValue` / `valueType` / `sequenceNumber` — there is **no
`value` and no `operation`**. The old call site read `p.value` (a key the wire
never sends), so the recorded value was **always empty**.

## Mapping (decided)

| recorder field | source |
| --- | --- |
| `fileName` | `suiteName ?? ""` |
| `key` | `key ?? null` |
| `value` | `newValue ?? value ?? null` (was reading the nonexistent `value`) |
| `valueType` | `valueType ?? null` (wire carries it; passed through) |
| `changeType` | `operation ?? "modify"` |

`changeType` defaults to `"modify"` — matching the Android call site
(`AndroidCtrlProxyClient.ts:2570`), since KVO cannot distinguish set vs remove so
the SDK emits no operation. An explicit `operation` is honored for forward-compat.

## Tests

Rewrote the `storage_changed` ingestor tests to lock the recorder contract:
required `valueType`/`changeType` present, no `operation` leak into recorder
input, `value` sourced from `newValue`, `valueType` passthrough of the real
`"unknown"` KVO value, and sensible defaults on a sparse payload. 21/21 pass.

## Deferred (follow-ups filed)

- **`tsc --noEmit` gate (issue #3001 bullet 3):** a repo-wide gate is infeasible
  today — `tsc --noEmit` reports ~542 pre-existing errors (incl. the known
  TS2300 `CtrlProxyHierarchy` collision and TS2305 missing exports). Filed a
  follow-up for a scoped/baseline (changed-files, fail-on-new) typecheck gate.
- **iOS SDK never captures real values:** `UserDefaultsInspector.swift` hardcodes
  `newValue: nil`, `valueType: "unknown"`, `key: nil`. This fix is a necessary
  prerequisite (stops silently dropping fields) but iOS storage rows stay
  near-empty until the SDK snapshots the changed key/value/type. Follow-up filed.
- **Cross-platform `value_type` normalization:** Android emits `"STRING"`
  (uppercase), iOS `"string"`/`"unknown"` — DB queries on `value_type` are
  case-inconsistent. Follow-up filed.

## Validation

- `bunx turbo run lint build` — green
- `bun test` (ingestor + TelemetryRecorder + storageEventRepository) — 61 pass
- Self code-review + 3 independent persona reviews (regression / completeness /
  architecture) — all "ship it", no blocking findings.

### DIFF
--- a/src/features/observe/ios/IosSdkEventIngestor.ts
+++ b/src/features/observe/ios/IosSdkEventIngestor.ts
@@ -242,10 +242,20 @@ export class DefaultIosSdkEventIngestor implements IosSdkEventIngestor {
             });
             break;
           case "storage_changed":
+            // The iOS SDK's SdkStorageChangedEvent serializes as suiteName/key/newValue/
+            // valueType/sequenceNumber (ios/auto-mobile-sdk/.../SdkEvent.swift). It carries
+            // no `value` and no `operation`, and the recorder REQUIRES valueType + changeType
+            // (issue #3001). Map: suiteName→fileName, newValue→value, pass valueType through,
+            // and derive changeType. KVO can't distinguish set vs remove so the SDK emits no
+            // operation — default to "modify" (matching the Android call site) while honoring
+            // an explicit `operation` if a future SDK build adds one.
             await recorder.recordStorageEvent({
               timestamp: ts, applicationId,
-              fileName: (p.suiteName as string) ?? "", key: (p.key as string) ?? "",
-              value: (p.value as string) ?? "", operation: (p.operation as string) ?? "write",
+              fileName: (p.suiteName as string) ?? "",
+              key: (p.key as string) ?? null,
+              value: (p.newValue as string) ?? (p.value as string) ?? null,
+              valueType: (p.valueType as string) ?? null,
+              changeType: (p.operation as string) ?? "modify",
             });
             break;
           default:
--- a/test/features/observe/ios/IosSdkEventIngestor.test.ts
+++ b/test/features/observe/ios/IosSdkEventIngestor.test.ts
@@ -141,11 +141,65 @@ describe("DefaultIosSdkEventIngestor", () => {
     expect((recorder.logs[0].event as { message: string }).message).toContain("sku");
   });
 
-  test("storage_changed routes to recordStorageEvent (fileName from suiteName)", async () => {
+  test("storage_changed maps the real iOS SDK wire shape to the recorder contract", async () => {
+    // The iOS SDK emits SdkStorageChangedEvent → JSON keys suiteName/key/newValue/valueType/
+    // sequenceNumber (see ios/auto-mobile-sdk/.../SdkEvent.swift). It carries no `operation`
+    // and no `value` field. The recorder REQUIRES valueType + changeType (issue #3001).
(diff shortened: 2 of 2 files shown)


kaeawc:
Follow-ups filed from review (non-blocking, out of scope for this fix):
- #3171 — iOS SDK never captures real key/value/type (KVO limitation)
- #3172 — scoped tsc --noEmit gate for src/ (issue #3001 bullet 3; ~542 pre-existing errors block a repo-wide gate)
- #3173 — normalize storage_events.value_type vocabulary across iOS/Android

PR #3170, opened 2026-07-05
perf(telemetry): emit prior storage value from CtrlProxy runners to skip recordStorageEvent lookup (#3000)

### LINKED ISSUE #3000: perf(telemetry): emit prior storage value from CtrlProxy runners to skip recordStorageEvent lookup
## Summary
PR #2999 (closes #2798) added an optional `previousValue` field threaded through `TelemetryRecorder.recordStorageEvent` → `RecordStorageEventInput`. When a source supplies it, the repository's `recordStorageEvent` skips its per-insert previous-value `SELECT` entirely (and closes the mutex-released read-then-insert stale-read window). **Today this hook is dormant**: no runner emits a prior value on the wire, so nothing supplies the field and every storage telemetry event still pays for the lookup (now a fast prefix seek via `idx_storage_events_key_lookup`, but still a read).

## What to do
Have the Android and iOS CtrlProxy runners include the **prior value** in their `storage_changed` push payloads, then pass it through at the two ingest call sites so the repository lookup is skipped.

- Android wire message currently carries only the new `value` (no `oldValue`): `src/features/observe/android/AndroidCtrlProxyClient.ts` (~line 2515, `storage_changed` handler; `WsStorageChangedMessage` shape).
- iOS ingest: `src/features/observe/ios/IosSdkEventIngestor.ts` (~line 245, `storage_changed` case).
- Repository already honors it: `src/db/storageEventRepository.ts` — guard is `input.previousValue !== undefined`, so an explicit `null` ("no prior value") is also honored verbatim without a lookup.

## Constraints / notes
- This is a **runner protocol change**, which is release-gated: runner changes don't reach default installs until the pinned release runner + checksum registry (`src/constants/release.ts`) are re-cut. Plan for that delivery gate (see project memory on iOS/Android runner release delivery).
- The SharedPreferences/NSUserDefaults change callbacks on-device generally expose the old value; confirm the accessibility-service / SDK hook can capture it cheaply before over-reading.
- Acceptance: with a runner that emits the prior value, `recordStorageEvent` issues **no** `SELECT` on the hot path (assert via a query-counting fake or EXPLAIN-free unit path), and stored `previous_value` matches the runner-supplied value.

## Context
Split out of #2798 / PR #2999 as explicitly-scoped follow-up. The `previousValue` skip is the "bigger, race-relevant win" Mira Kessler flagged in the #2798 review; the index in #2999 is the near-term lever, this is the correctness-adjacent completion.

kaeawc:
## What

Activates the dormant `previousValue` skip path added in #2999 by having the **Android CtrlProxy runner capture and emit the prior SharedPreferences value** in `storage_changed` push payloads, then threading it through both TS ingest sites so the repository's per-insert previous-value `SELECT` is skipped.

Advances #3000 (Android runner + full TS ingest for both platforms). The iOS **UserDefaults old-value capture** is genuinely follow-up-sized (the iOS SDK inspector is currently a stub that emits `newValue: nil`), so #3000 stays open and the iOS work is tracked in #3174.

## Why

#2999 threaded an optional `previousValue` through `TelemetryRecorder.recordStorageEvent` → `RecordStorageEventInput`; when a caller supplies it, the repository skips the per-insert lookup (and closes the mutex-released read-then-insert stale-read window). **No runner emitted it on the wire**, so every storage telemetry insert still paid for the lookup. This wires a source to emit it.

The guard is `input.previousValue !== undefined` (not `?? null`), so:
- **legacy runners** that omit the field → `undefined` → auto-lookup preserved (back-compat),
- **updated runners** → skip the lookup,
- an explicit **`null`** ("no prior value") → honored verbatim, still skips.

**Semantic note:** for updated runners the stored `previous_value` now reflects the **on-device value at change time** (from the SDK's per-file snapshot), rather than the last-recorded `storage_events` row that the auto-lookup returned. This is strictly more correct (ground-truth device state) and closes the mutex-released read-then-insert stale-read window, but consumers of `previous_value` should know its meaning shifted from "last recorded value" to "prior on-device value."

## How

**Android chain (source of the wire value):**
- `SharedPreferencesDriverImpl` maintains a per-file value **snapshot**, seeded on `startListening` from `prefs.all`, so the change listener — which fires only *after* the write commits — can report the value present *before* the change. `PreferenceChange` carries `previousValue` **and** `previousValueType`.
- `previousValueType` is threaded because on a **remove** (new value null ⇒ type UNKNOWN) or a **type-changing write**, the prior value must be serialized/quoted by its **own** type. Quoting it by the new type would emit an unquoted STRING and break the whole message's JSON. `SharedPreferencesInspectorProvider` serializes with `change.previousValueType`.
- Protocol `StorageChangeEvent` + control-proxy `PreferenceChangeEvent` gain `previousValue` / `previousValueType` (default null ⇒ older SDKs still deserialize).
- The wire JSON build was extracted into a unit-tested `buildStorageChangedMessage` (`StorageChangeWireEncoder.kt`) that quotes `value`/`previousValue` each by its own type.

**TS ingest:**
- `WsStorageChangedMessage.previousValue`; a pure, unit-tested `storageTelemetryInputFromWire` helper threads it **only when present**.
- `IosSdkEventIngestor` threads `payload.previousValue` only when present.

**iOS:** `SdkStorageChangedEvent.previousValue` added for struct/protocol symmetry (UserDefaults old-value capture is deferred — see follow-up).

## Not in scope (follow-ups filed)

- **iOS UserDefaults old-value capture** — the iOS SDK's `notifyChangeListeners` currently emits `newValue: nil` (never captures new *or* old value); real capture is a separate effort. The TS/struct plumbing here is ready for it.
- **release.ts checksum re-cut** — this is a runner protocol change; the pinned release runner + checksum registry (`src/constants/release.ts`) are auto-generated and re-cut by periodic version-bump PRs, so an updated runner reaches default installs only after that.
- Pre-existing iOS `recordStorageEvent` type mismatch (`operation`/missing `valueType`,`changeType`) tracked in #3001.

## Testing

- **`test/db/storageEventRepository.test.ts`** — new **query-counting** proof: a runner-supplied `previousValue` issues **zero** `SELECT` against `storage_events` on the hot path (the acceptance criterion), while an omitted field still performs the lookup.
- **`test/features/observe/android/storageTelemetryInputFromWire.test.ts`** — wire→telemetry mapping: present threads, explicit null honored, omitted field absent (auto-lookup).
- **`test/features/observe/android/storageChangedWireToDb.test.ts`** — end-to-end guard: a realistic wire payload (shaped as `buildStorageChangedMessage` emits) is parsed → mapped → persisted; pins the Kotlin-encoder ↔ TS-wire field contract for modify / remove / legacy-no-field.
- **`test/features/observe/ios/IosSdkEventIngestor.test.ts`** — iOS ingest threads present/null, omits when absent.
- **`StorageProtocolTest.kt`** — `previousValue`/`previousValueType` round-trip + backward-compat decode of a payload omitting them.
- **`StorageChangeWireEncoderTest.kt`** — the wire payload is valid JSON across add/modify/**remove**/type-change; the removed-STRING case (prior value quoted by its own type) is the regression guard.
- **`SharedPreferencesPreviousValueTest.kt`** (Robolectric) — snapshot capture across add/modify/remove/successive-writes/re-listen, asserting `previousValueType`.
- `bun test` (TS storage/telemetry/observe + db) green; `:protocol:test`, `:auto-mobile-sdk:testDebugUnitTest`, `:control-proxy:testDebugUnitTest` green; iOS `swift build` green; ktfmt clean.

## Reviewed

Ran `/auto-mobile-code-review` plus three independent adversarial reviewers (DB-correctness, Android-runner-regression, architecture/completeness). Substantive findings addressed:
- **Wire JSON regression** (code-review): `previousValue` was quoted by the **new** value's type, so removing/type-changing a STRING pref emitted invalid JSON and dropped the entire `storage_changed` message on the TS side. Fixed by threading `previousValueType`; the `StorageChangeWireEncoder` unit test reproduces the broken case.
- **Stale Android API dump** (Android reviewer): the public `PreferenceChange` data class changed but `android/auto-mobile-sdk/api/auto-mobile-sdk.api` wasn't regenerated (iOS's was) → `:auto-mobile-sdk:apiCheck` would fail. Regenerated (minimal `PreferenceChange`-only hunk).
- **Query-count test robustness** (DB reviewer): narrowed the SELECT filter to the exact lookup shape so it can't be confounded by the retention `count(*)`/`offset` probes.
- **Missing end-to-end coverage** (architecture reviewer): added `storageChangedWireToDb.test.ts` to pin the cross-language field contract.
- **iOS scope** (architecture reviewer): confirmed iOS capture is genuinely follow-up-sized; this PR does not auto-close #3000; iOS capture is tracked in #3174.

### DIFF
--- a/android/auto-mobile-sdk/api/auto-mobile-sdk.api
+++ b/android/auto-mobile-sdk/api/auto-mobile-sdk.api
@@ -1259,21 +1259,26 @@ public interface dev.jasonpearson.automobile.sdk.storage.OnPreferenceChangeListe
 Compiled from "KeyValueModels.kt"
 public final class dev.jasonpearson.automobile.sdk.storage.PreferenceChange {
   public static final int $stable;
-  public dev.jasonpearson.automobile.sdk.storage.PreferenceChange(java.lang.String, java.lang.String, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType, long, long);
+  public dev.jasonpearson.automobile.sdk.storage.PreferenceChange(java.lang.String, java.lang.String, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType, long, long, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType);
+  public dev.jasonpearson.automobile.sdk.storage.PreferenceChange(java.lang.String, java.lang.String, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType, long, long, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType, int, kotlin.jvm.internal.DefaultConstructorMarker);
   public final java.lang.String getFileName();
   public final java.lang.String getKey();
   public final java.lang.Object getNewValue();
   public final dev.jasonpearson.automobile.sdk.storage.KeyValueType getType();
   public final long getTimestamp();
   public final long getSequenceNumber();
+  public final java.lang.Object getPreviousValue();
+  public final dev.jasonpearson.automobile.sdk.storage.KeyValueType getPreviousValueType();
   public final java.lang.String component1();
   public final java.lang.String component2();
   public final java.lang.Object component3();
   public final dev.jasonpearson.automobile.sdk.storage.KeyValueType component4();
   public final long component5();
   public final long component6();
-  public final dev.jasonpearson.automobile.sdk.storage.PreferenceChange copy(java.lang.String, java.lang.String, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType, long, long);
-  public static dev.jasonpearson.automobile.sdk.storage.PreferenceChange copy$default(dev.jasonpearson.automobile.sdk.storage.PreferenceChange, java.lang.String, java.lang.String, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType, long, long, int, java.lang.Object);
+  public final java.lang.Object component7();
+  public final dev.jasonpearson.automobile.sdk.storage.KeyValueType component8();
+  public final dev.jasonpearson.automobile.sdk.storage.PreferenceChange copy(java.lang.String, java.lang.String, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType, long, long, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType);
+  public static dev.jasonpearson.automobile.sdk.storage.PreferenceChange copy$default(dev.jasonpearson.automobile.sdk.storage.PreferenceChange, java.lang.String, java.lang.String, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType, long, long, java.lang.Object, dev.jasonpearson.automobile.sdk.storage.KeyValueType, int, java.lang.Object);
   public java.lang.String toString();
   public int hashCode();
   public boolean equals(java.lang.Object);
--- a/android/auto-mobile-sdk/src/debug/kotlin/dev/jasonpearson/automobile/sdk/storage/SharedPreferencesInspectorProvider.kt
+++ b/android/auto-mobile-sdk/src/debug/kotlin/dev/jasonpearson/automobile/sdk/storage/SharedPreferencesInspectorProvider.kt
@@ -371,6 +371,12 @@ class SharedPreferencesInspectorProvider : ContentProvider() {
         type = change.type.name,
         timestamp = change.timestamp,
         sequenceNumber = change.sequenceNumber,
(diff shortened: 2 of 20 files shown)


kaeawc:
Good catch — fixed in 7a76bf7c5. The updated encoder now gates `previousValue` on `previousValueType != null`: a real SDK capture always carries a type (`UNKNOWN` even for a genuine no-prior-value add, which stays emitted as `null`), whereas a legacy SDK that predates #3000 leaves `previousValueType` null through the protocol + `StorageSubscriptionManager`, so the field is now **omitted entirely** on the wire. That leaves `message.previousValue === undefined` on the TS side, preserving the DB auto-lookup for legacy-SDK changes that land after an existing row. Added `StorageChangeWireEncoderTest.legacy SDK without previousValueType omits the previousValue field entirely` as a regression guard.

PR #3180, opened 2026-07-05
perf(db): truncate request/response bodies in the getNetworkEvents list projection (#2801)

### LINKED ISSUE #2801: perf(db): truncate request/response bodies in the getNetworkEvents list projection
## Summary
`getNetworkEvents` in `src/db/networkEventRepository.ts` returns **untruncated** `request_body`/`response_body` strings, unlike its single-row sibling `getNetworkEventById`, which caps each body at `BODY_TRUNCATION_LIMIT` (10&nbsp;KB). The telemetry push backfill ships up to 100 network rows per subscriber over a Unix socket on every subscribe (`src/daemon/telemetryPushSocketServer.ts`), so a burst of large-payload requests can inflate a single backfill by megabytes of body text the dashboard already caps to 10&nbsp;KB when rendering. Found via a DB concurrency/migration audit (2026-07-01), finding P7bodies, priority P2, verified against source.

## Background & history
- The `network_events` table and its repository were introduced in commit `d20e1f0e` (2026-03-16, [#1535](https://github.com/kaeawc/auto-mobile/pull/1535)) — network/log/OS telemetry SDK with the IDE dashboard. At that point `mapRow` already returned raw body columns and there was no truncation anywhere.
- Body truncation was added in commit `73130e51` (2026-03-28, [#1624](https://github.com/kaeawc/auto-mobile/pull/1624)) — "add network capture, mocking, and graph MCP tools". That change introduced `BODY_TRUNCATION_LIMIT = 10_240` and applied it **only** in `getNetworkEventById` (the detail-view lookup). The list projection `getNetworkEvents` (same commit) was left returning raw bodies.
- The backfill consumer was wired in commit `fdb94336` (2026-03-23, [#1571](https://github.com/kaeawc/auto-mobile/pull/1571)); the parallel `Promise.all` shape it has today is from `9006fa11` (2026-04-02). It requests `limit = 100` per category.
- Why it's shaped this way: truncation was added reactively for the single-event detail view (which is the one screen that displays a full body), and the list path was simply never revisited — an easy omission since both paths share `mapRow` but only the detail path post-processes the result.

## The problem
`getNetworkEventById` truncates; `getNetworkEvents` does not:

```ts
// src/db/networkEventRepository.ts:110  (detail lookup — TRUNCATES)
export async function getNetworkEventById(id, db) {
  const row = await getDb(db).selectFrom("network_events").selectAll()
    .where("id", "=", id).executeTakeFirst();
  if (!row) { return null; }
  const event = mapRow(row);
  // Truncate bodies to 10KB
  if (event.requestBody && event.requestBody.length > BODY_TRUNCATION_LIMIT) {
    event.requestBody = event.requestBody.slice(0, BODY_TRUNCATION_LIMIT);
  }
  if (event.responseBody && event.responseBody.length > BODY_TRUNCATION_LIMIT) {
    event.responseBody = event.responseBody.slice(0, BODY_TRUNCATION_LIMIT);
  }
  return event;
}
```

```ts
// src/db/networkEventRepository.ts:137  (list projection — NO truncation)
export async function getNetworkEvents(query, db) {
  let q = getDb(db).selectFrom("network_events").selectAll();
  // ...filters...
  q = q.orderBy("timestamp", "desc").limit(query.limit ?? 100);
  const rows = await q.execute();
  return rows.map(mapRow);   // raw request_body / response_body, uncapped
}
```

The uncapped list result is exactly what the backfill fans out to each subscriber:

```ts
// src/daemon/telemetryPushSocketServer.ts:91
const [networkRows, logRows, osRows] = await Promise.all([
  shouldInclude("network") ? getNetworkEvents({ deviceId, limit }) : [],  // limit = 100
  // ...
]);
for (const r of networkRows) {
  events.push({ category: "network", ..., data: r });  // full bodies serialized to the socket
}
```

Each of the 100 rows is JSON-serialized (`this.sendJson(socket, msg)`) with its full body. A device under a load/upload workload can store multi-hundred-KB (or MB) bodies, so a single subscribe can serialize and stream tens of MB even though the dashboard renders at most `BODY_TRUNCATION_LIMIT` bytes per body — the extra bytes are pure waste (SQLite read amplification, JSON.stringify CPU, socket bandwidth, and subscriber memory). This is a projection-efficiency issue, not a correctness bug; it doesn't corrupt data and there's no concurrency interleaving involved (it's a pure read on the single daemon connection).

## Impact
- Backfill payload for the network category is bounded only by stored body size × 100, instead of by 10&nbsp;KB × 100 (~1&nbsp;MB). With large upload/download traffic captured, backfill can balloon from ~1&nbsp;MB to tens of MB, adding latency and memory pressure on every dashboard reconnect/subscribe.
- Also affects any other caller of `getNetworkEvents` (e.g. MCP list tooling) that assumes the same 10&nbsp;KB contract the detail path advertises — today the two paths silently disagree.
- Severity is narrow: it only bites when captured bodies are large and a subscriber (re)connects; steady-state single-event views already truncate. No data loss, no crash.

## Proposed fix

> **Scope is INCOMPLETE for its own goal — cap ALL large text at the backfill boundary (per review).** Sam verified that `telemetryPushSocketServer` fans out `limit = 100` for SIX categories plus failure occurrences, and several carry unbounded `text` columns no path truncates — so truncating only network fixes the largest offender but leaves the same "backfill balloons to tens of MB" vector open:
> - **log** (`getLogEvents`, line 93) — `message` (`text notNull`, migration line 74) and `properties_json` (line 112).
> - **os** (`getOsEvents`, line 94) — `details_json` (line 150).
> - **storage** (`getStorageEvents`, line 204) — `value` (`text`, `2026_03_19_000_storage_events.ts:14`).
> - **failure occurrences** (line 168) — `tool_args_json`, plus the group's `stack_trace_json` / `tool_call_info_json`.
>
> **Preferred fix: a single cap at the backfill boundary** in `src/daemon/telemetryPushSocketServer.ts` — bound any serialized large-text field before `sendJson`, so the contract is enforced in ONE place rather than re-omitted per repo. This also avoids the regression below: `getNetworkEvents` (and the other list projections) are also called by MCP tooling that may WANT the raw stored body (e.g. replay/mock generation) — grep callers before truncating at the repo source; if any need raw bodies, truncate at the backfill projection, not in `mapRow`, so the DB read contract is unchanged and only the socket payload is bounded.
>
> **Fix the UTF-8 hazard (per review):** slicing a body by JS string length (`.slice(0, LIMIT)` on UTF-16 code units) can split a multibyte character or produce lone surrogates. The existing `getNetworkEventById` slice already has this latent mojibake bug. Use a codepoint-/byte-safe truncation.

Move the truncation into the shared `mapRow` path (or a tiny helper it calls) so **both** `getNetworkEventById` and `getNetworkEvents` return capped bodies, then drop the now-redundant post-processing block in `getNetworkEventById`. Minimal sketch:

```ts
// src/db/networkEventRepository.ts
function truncateBody(body: string | null): string | null {
  return body && body.length > BODY_TRUNCATION_LIMIT
    ? body.slice(0, BODY_TRUNCATION_LIMIT)
    : body;
}

function mapRow(r: any): NetworkEventWithId {
  return {
    // ...unchanged fields...
    requestBody: truncateBody(r.request_body ?? null),
    responseBody: truncateBody(r.response_body ?? null),
    contentType: r.content_type ?? null,
  };
}
```

With truncation centralized in `mapRow`, the explicit slice block in `getNetworkEventById` becomes dead code and should be removed so there is exactly one truncation site (mirrors the repo's "one canonical primitive per concern" convention).

Must NOT:
- Do NOT switch to lazy-load-on-expand / stripping bodies out of the list result — the dashboard relies on inline (capped) body content in the backfill stream. Truncate, don't remove.
- Do NOT change the 10&nbsp;KB constant or the byte semantics (`.slice` on the string, matching the existing detail-view behavior; the existing test asserts `length === 10_240`).
- Do NOT alter filtering, ordering, or row count — this is projection-only.

- [ ] `getNetworkEvents` returns each row with `requestBody`/`responseBody` capped at `BODY_TRUNCATION_LIMIT` (10&nbsp;240): inserting a row with a body longer than 10&nbsp;KB and reading it back through `getNetworkEvents` yields `event.responseBody.length === 10_240` (and same for `requestBody`).
- [ ] **(per review) ALL large text is capped at the backfill boundary, not just network.** The `limit = 100` backfill in `telemetryPushSocketServer` bounds log (`message`, `properties_json`), os (`details_json`), storage (`value`), and failure occurrence (`tool_args_json`, `stack_trace_json`, `tool_call_info_json`) large-text fields — OR explicitly state those are out of scope with a named follow-up issue. Add parallel truncation tests for at least log `message` and storage `value`.
- [ ] **(per review) UTF-8-safe truncation.** Truncation does not split a multibyte character or emit lone surrogates (fixes the latent mojibake bug in the existing `getNetworkEventById` UTF-16-code-unit slice). Add a test with a body whose Nth code unit lands mid-multibyte-sequence.
- [ ] **(per review) Grep callers of `getNetworkEvents` (and sibling list projections) before truncating at the repo source;** if any MCP caller needs raw bodies (replay/mock generation), truncate at the backfill projection so the DB read contract is unchanged.
- [ ] **(per review) Add an invariant test that the two paths AGREE:** `getNetworkEvents` and `getNetworkEventById` return the same truncated body for the same row (prevents the two paths silently diverging again — the actual invariant this issue is about).
- [ ] `getNetworkEventById` still truncates identically (no behavior regression) after the truncation is centralized in `mapRow`; the existing test at `test/db/networkEventRepository.test.ts:73` ("getNetworkEventById truncates bodies over 10KB") stays green.
- [ ] New unit test added to `test/db/networkEventRepository.test.ts` (e.g. "getNetworkEvents truncates bodies over 10KB") mirroring the existing `getNetworkEventById` truncation test; sub-10&nbsp;KB bodies pass through unchanged; null bodies stay null. Follows repo conventions (in-memory/fake DB already used in this file, no real device, runs in <100ms, non-flaky).
- [ ] Result set is otherwise byte-identical for all non-body fields and for bodies ≤10&nbsp;KB (query results unchanged): the win is payload size only, not different rows.
- [ ] Measurement note: with a synthetic fixture of 100 rows each carrying a 100&nbsp;KB body, the serialized backfill payload for the network category drops from ~10&nbsp;MB to ~1&nbsp;MB (100 × 10&nbsp;KB). Capture before/after `JSON.stringify(events).length` in the test or a scratch script.
- [ ] `bun test test/db/networkEventRepository.test.ts` green; `turbo run lint build test` green.

## References
- `src/db/networkEventRepository.ts:137` — `getNetworkEvents` list projection (no truncation)
- `src/db/networkEventRepository.ts:110` — `getNetworkEventById` (truncates; the contract to mirror)
- `src/db/networkEventRepository.ts:83` — `BODY_TRUNCATION_LIMIT = 10_240`
- `src/db/networkEventRepository.ts:85` — `mapRow` (shared projection, ideal truncation site)
- `src/daemon/telemetryPushSocketServer.ts:91` — backfill calls `getNetworkEvents({ deviceId, limit })` with `limit = 100`
- `test/db/networkEventRepository.test.ts:73` — existing detail-view truncation test to mirror
- Finding P7bodies — DB concurrency/migration audit, 2026-07-01

**Regressions & testability (Sam Whitfield)**
- The fix as scoped is **incomplete for its own "bound backfill size" goal**: the same `limit = 100` backfill in `telemetryPushSocketServer` ships unbounded large text from log (`message`, `properties_json`), os (`details_json`), storage (`value`), and failure occurrences (`tool_args_json`, `stack_trace_json`, `tool_call_info_json`). Cap ALL of these — preferably with a **single cap at the backfill boundary** (`telemetryPushSocketServer`) rather than re-omitted per repo — or explicitly scope them out with a follow-up.
- Truncating in `mapRow` risks regressing MCP callers that want raw bodies (replay/mock generation) — grep callers first; if any need raw data, truncate at the backfill projection so the DB read contract is unchanged.
- **UTF-8 hazard:** the existing `getNetworkEventById` `.slice(0, LIMIT)` on UTF-16 code units can split a multibyte char / emit lone surrogates. Use a codepoint-/byte-safe truncation.
- Add a test that `getNetworkEvents` and `getNetworkEventById` AGREE on the truncated body for the same row (the real invariant), plus parallel truncation cases for the newly-scoped columns.

**Cross-links:** shares the backfill payload-bloat surface with #2799 (a shared batched ingestion + `EventRetentionPolicy` path could centralize this cap) and #2800 (both are `telemetryPushSocketServer`/dashboard read-path efficiency).

kaeawc:
## Summary

Closes #2801.

`getNetworkEvents` (the list projection) returned **untruncated** `request_body`/`response_body` strings, while its sibling `getNetworkEventById` capped each body at `BODY_TRUNCATION_LIMIT` (10 KB). The telemetry push backfill fans out up to 100 network rows per subscribe over a Unix socket on every subscribe, so a burst of large-payload requests could inflate a single backfill by megabytes of body text the dashboard already caps to 10 KB when rendering.

## What changed

- **New canonical primitive** `src/utils/truncateBodyText.ts` — exports `BODY_TRUNCATION_LIMIT` and a **surrogate-safe** `truncateBodyText()`. A naive `.slice(0, LIMIT)` on UTF-16 code units can leave a lone high surrogate when an emoji/astral-plane character straddles the boundary (the latent mojibake bug in the old `getNetworkEventById` slice); the helper drops a trailing high surrogate so the result is always well-formed. ASCII bodies still cap to exactly 10_240, so byte semantics are unchanged. Dedupes the two previously-copied `BODY_TRUNCATION_LIMIT` constants.
- **`src/db/networkEventRepository.ts`** — apply `truncateBodyText` inside the shared `mapRow`, so **both** `getNetworkEvents` and `getNetworkEventById` return capped bodies from one truncation site. Removed the now-redundant slice block in `getNetworkEventById`. Original sizes remain in `request_body_size`/`response_body_size`, so the detail view's `*Truncated` flags are unaffected.
- **`src/server/networkResources.ts`** — import the shared `BODY_TRUNCATION_LIMIT` (unify the duplicate).
- **`src/daemon/telemetryPushSocketServer.ts`** — single **backfill-boundary cap** (`boundBackfillEventText`) for the remaining opaque plain-text fields fanned out by the same `limit = 100` backfill: log `message` and storage `value`/`previousValue`. Network bodies are already capped upstream (idempotent). Applied once before `sendJson`.

### Why truncate network at the repo, not the boundary
Grepped every `getNetworkEvents` caller: `eventToSummary` omits bodies; `eventToDetail` is fed only by the already-capped `getNetworkEventById`; `buildNetworkGraph` ignores bodies; stats/bucketing use only scalars. No caller needs raw bodies, so capping in `mapRow` is safe and satisfies the issue's read-back acceptance criterion.

### Deferred (follow-up issue)
Structured-JSON columns fanned out by the same backfill — os `details`, layout `detailsJson`, and failure `stack_trace_json`/`tool_call_info_json`/`tool_args_json` — are **not** capped here because a blind string slice would corrupt the JSON. Tracked as a follow-up for JSON-aware/size-guarded handling.

## Tests

- `test/utils/truncateBodyText.test.ts` — null/short/exact/over-limit, custom limit, surrogate-pair straddle (both boundary orientations), well-formedness.
- `test/db/networkEventRepository.test.ts` — `getNetworkEvents` truncation (req + resp), sub-10KB/null passthrough, **list-vs-detail agree-invariant**, surrogate-safety, and a **~10× payload-shrink measurement** (100×100 KB bodies → serialized bytes < raw/5, each body == 10_240).
- `test/daemon/telemetryPushSocketServer.test.ts` — `boundBackfillEventText` caps log `message`, storage `value`/`previousValue`; leaves small fields and unrelated categories untouched; surrogate-safe.

## Validation
- `bun test test/utils test/db test/daemon test/server` green (2000+ tests).
- `turbo run lint build` green.

## Acceptance criteria (from #2801)
- [x] `getNetworkEvents` caps req/resp bodies at 10_240 (read-back test).
- [x] Large text at the backfill boundary bounded for network + log + storage; os/layout/failure JSON blobs explicitly deferred to a named follow-up.
- [x] UTF-8/surrogate-safe truncation (fixes latent lone-surrogate bug) + test.
- [x] Grepped `getNetworkEvents` callers; none need raw bodies → safe to cap in `mapRow`.
- [x] List and detail paths **agree** on the truncated body (invariant test).
- [x] `getNetworkEventById` behavior unchanged; existing truncation test stays green.
- [x] Non-body fields and ≤10 KB bodies byte-identical.
- [x] Measurement: ~10 MB → ~1 MB for 100×100 KB.
- [x] `bun test` + `turbo run lint build test` green.

### DIFF
--- a/src/daemon/telemetryPushSocketServer.ts
+++ b/src/daemon/telemetryPushSocketServer.ts
@@ -21,8 +21,8 @@ import { truncateBodyText } from "../utils/truncateBodyText";
  * (#2801). Network bodies are already capped upstream in the repository
  * `mapRow`; this bounds the remaining plain-text columns at the backfill
  * boundary so the DB read contract for those repos is unchanged. Only opaque
- * text is listed — structured JSON columns (os `details_json`, failure blobs)
- * would be corrupted by a blind slice and are tracked as a follow-up.
+ * text is listed — structured JSON columns (os `details`, layout `detailsJson`,
+ * failure blobs) would be corrupted by a blind slice and are tracked in #3182.
  */
 const BOUNDED_BACKFILL_TEXT_FIELDS: Record<string, readonly string[]> = {
   log: ["message"],
--- a/src/daemon/telemetryPushSocketServer.ts
+++ b/src/daemon/telemetryPushSocketServer.ts
@@ -13,6 +13,47 @@ import { getDatabase } from "../db/database";
 import type { Database } from "../db/types";
 import type { Kysely } from "kysely";
 import { TELEMETRY_PUSH_SOCKET_CONFIG } from "./daemonFiles";
+import { truncateBodyText } from "../utils/truncateBodyText";
+
+/**
+ * Opaque plain-text fields that the backfill fans out (limit=100) and that can
+ * balloon a single subscribe by megabytes the dashboard already caps at 10KB
+ * (#2801). Network bodies are already capped upstream in the repository
+ * `mapRow`; this bounds the remaining plain-text columns at the backfill
+ * boundary so the DB read contract for those repos is unchanged. Only opaque
+ * text is listed — structured JSON columns (os `details_json`, failure blobs)
+ * would be corrupted by a blind slice and are tracked as a follow-up.
+ */
+const BOUNDED_BACKFILL_TEXT_FIELDS: Record<string, readonly string[]> = {
+  log: ["message"],
+  storage: ["value", "previousValue"],
+};
+
+/**
+ * Cap known large plain-text fields on a backfilled telemetry event to
+ * BODY_TRUNCATION_LIMIT, returning a shallow copy when anything changed so the
+ * caller's source rows are not mutated. Surrogate-safe (see truncateBodyText).
+ */
+export function boundBackfillEventText(event: TelemetryEvent): TelemetryEvent {
+  const fields = BOUNDED_BACKFILL_TEXT_FIELDS[event.category];
+  if (!fields || event.data === null || typeof event.data !== "object") {
+    return event;
+  }
+  const data = event.data as Record<string, unknown>;
+  let bounded: Record<string, unknown> | null = null;
+  for (const field of fields) {
+    const value = data[field];
+    if (typeof value !== "string") {
+      continue;
+    }
+    const capped = truncateBodyText(value);
+    if (capped !== value) {
+      bounded = bounded ?? { ...data };
+      bounded[field] = capped;
+    }
+  }
+  return bounded ? { ...event, data: bounded as TelemetryEvent["data"] } : event;
+}
 
 interface TelemetryFilter {
   category: string | null; // "network", "log", "os", "navigation", "crash", "anr", "nonfatal", "storage", "layout", or null for all
@@ -215,7 +256,9 @@ export class TelemetryPushSocketServer extends PushSubscriptionSocketServer<
     events.sort((a, b) => a.timestamp - b.timestamp);
 
     for (const event of events) {
-      const msg = this.createPushMessage(event);
+      // Single boundary cap for large plain-text fields (#2801): network bodies
+      // are already capped in the repository mapRow; this bounds log/storage.
+      const msg = this.createPushMessage(boundBackfillEventText(event));
       this.sendJson(socket, msg);
     }
 
--- a/src/db/networkEventRepository.ts
+++ b/src/db/networkEventRepository.ts
@@ -2,6 +2,7 @@ import type { Kysely } from "kysely";
 import type { Database } from "./types";
 import { getDatabase } from "./database";
 import { logger } from "../utils/logger";
+import { truncateBodyText } from "../utils/truncateBodyText";
 
 export interface RecordNetworkEventInput {
   deviceId: string | null;
@@ -86,8 +87,6 @@ export interface NetworkEventQuery {
   minStatusCode?: number;
 }
 
-const BODY_TRUNCATION_LIMIT = 10_240; // 10KB
-
 function mapRow(r: any): NetworkEventWithId {
   return {
     id: r.id,
@@ -107,8 +106,13 @@ function mapRow(r: any): NetworkEventWithId {
     error: r.error,
     requestHeaders: r.request_headers_json ? JSON.parse(r.request_headers_json) : null,
     responseHeaders: r.response_headers_json ? JSON.parse(r.response_headers_json) : null,
-    requestBody: r.request_body ?? null,
-    responseBody: r.response_body ?? null,
+    // Truncate bodies to 10KB here so both getNetworkEventById and the
+    // getNetworkEvents list projection (fanned out 100-at-a-time by the
+    // telemetry backfill) share one cap — the dashboard renders at most this
+    // many bytes anyway (#2801). Original sizes are preserved in
+    // request_body_size / response_body_size for truncation-flag callers.
+    requestBody: truncateBodyText(r.request_body ?? null),
+    responseBody: truncateBodyText(r.response_body ?? null),
     contentType: r.content_type ?? null,
   };
 }
@@ -127,17 +131,9 @@ export async function getNetworkEventById(
     return null;
   }
 
-  const event = mapRow(row);
-
-  // Truncate bodies to 10KB
-  if (event.requestBody && event.requestBody.length > BODY_TRUNCATION_LIMIT) {
-    event.requestBody = event.requestBody.slice(0, BODY_TRUNCATION_LIMIT);
-  }
-  if (event.responseBody && event.responseBody.length > BODY_TRUNCATION_LIMIT) {
-    event.responseBody = event.responseBody.slice(0, BODY_TRUNCATION_LIMIT);
-  }
-
-  return event;
+  // mapRow already caps bodies at BODY_TRUNCATION_LIMIT (surrogate-safe), so
+  // there is exactly one truncation site shared with getNetworkEvents (#2801).
+  return mapRow(row);
 }
 
 export async function getNetworkEvents(
--- a/src/server/networkResources.ts
+++ b/src/server/networkResources.ts
@@ -6,8 +6,7 @@ import {
   type NetworkEventWithId,
 } from "../db/networkEventRepository";
 import { NetworkState } from "./NetworkState";
-
-const BODY_TRUNCATION_LIMIT = 10_240; // 10KB
+import { BODY_TRUNCATION_LIMIT } from "../utils/truncateBodyText";
 
 const NETWORK_RESOURCE_URIS = {
   REQUEST: "automobile:network/request/{requestId}",
--- a/src/utils/truncateBodyText.ts
+++ b/src/utils/truncateBodyText.ts
@@ -1,0 +1,38 @@
+/**
+ * Canonical body/text truncation for telemetry read paths (#2801).
+ *
+ * The network detail view (`getNetworkEventById`) has capped request/response
+ * bodies at 10&nbsp;KB since #1624, but the list projection (`getNetworkEvents`)
+ * and the telemetry backfill never did — so a burst of large payloads could
+ * inflate a single backfill by megabytes the dashboard already caps to 10&nbsp;KB.
+ * Centralize the cap here so every read path shares one primitive rather than
+ * re-omitting it per repository ("one canonical primitive per concern").
+ */
+
+/** Maximum retained body length, in UTF-16 code units (10&nbsp;KB). */
+export const BODY_TRUNCATION_LIMIT = 10_240;
+
+/**
+ * Truncate `text` to at most `limit` UTF-16 code units without splitting a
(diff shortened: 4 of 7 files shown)


kaeawc:
Deferred structured-JSON backfill columns (os `details`, layout `detailsJson`, failure `stack_trace_json`/`tool_call_info_json`/`tool_args_json`) are tracked in #3182 — blind-slicing those would corrupt JSON, so they need JSON-aware/size-guarded handling. This PR's acceptance criteria explicitly permit the "scope out with a named follow-up" path.

kaeawc:
**Scope note (from review):** the cap is applied on the *backfill* read path only (repository `mapRow` + `boundBackfillEventText`), matching #2801's scope ("backfill fans out large payloads"). The live single-event push path (`pushTelemetryEvent`) is intentionally not bounded here — live frames carry a single event, not 100 rows. **Review fixes pushed** (`02a48c683`): hardened the surrogate guard to also drop a trailing lone *low* surrogate (malformed input), added orphan-low/limit≤1 tests, and tightened the payload-shrink assertion to raw/8.

PR #3202, opened 2026-07-05
refactor(ios): finish control-proxy typed-model cleanups (#2859)

### LINKED ISSUE #2859: refactor(ios): finish control-proxy typed-model cleanups (dead CommandError cases, enum-derived responseType, WebSocketServer test)
## Summary

Maintainability follow-ups from the typed-command-model migration #2846 / #2854 (merged `aa25414a`) and its 3-perspective review. None block; grouped here so the "add a command" surface shrinks and the remaining stringly-typed/dead pieces the migration set out to retire are finished off. Fits the repo-wide stale-code cleanup catalog.

## 1. Dead `CommandError` cases

`ios/control-proxy/Sources/CtrlProxy/CommandHandler.swift` — `CommandError.notSupported` and `CommandError.elementNotFound` have **no constructors** anywhere (verified: the only `.notSupported(`/`.elementNotFound(` throws are on the *separate* `GestureError` enum in `GesturePerformer.swift`). `.unknownCommand` became live again in #2854; `.missingParameter` / `.invalidParameter` / `.executionFailed` are used.

- [ ] Remove `CommandError.notSupported` and `CommandError.elementNotFound` (and their `errorDescription` arms).

## 2. `responseType(for:)` is the stringly-typed table the migration targeted

`CommandHandler.swift` still has a ~38-arm `responseType(for requestType: String) -> String` mapping `RequestType.rawValue → ResponseType.rawValue`, with a `default: "error"` fallback. It's now used only on the error path (`handle`'s `catch`). Every case maps 1:1 to a `ResponseType`, so this is exactly the "kitchen-sink stringly-typed switch" #2846 set out to retire — just relocated.

- [ ] Convert to a compiler-enforced computed property (e.g. `var errorResponseType: ResponseType` on `WebSocketRequest` or `RequestType`) with an exhaustive `switch` and no `default`. Note: it currently omits `getCurrentFocus`/`getTraversalOrder`/`addHighlight` (they fall to `default:"error"` because those handlers don't throw) — make the mapping total.

## 3. Mechanical per-command accessor switches

Adding a new command currently touches ~7 sites: `RequestType`, payload struct, the `init(from:)` arm, the `requestType` accessor arm, the `requestId` accessor arm, `responseType(for:)`, and the `handle` arm. Only `handle` is compiler-enforced.

- [ ] Introduce a `protocol CommandPayload { var requestId: String? { get } }` that all payload structs conform to, so the `requestId` accessor collapses to `payload.requestId` (and future common fields come free). Combined with #2 this cuts the silent (non-compiler-checked) arms toward zero.

## 4. No `WebSocketServer` tests

There is no `WebSocketServerTests`. #2854 moved unknown-command handling *out* of `CommandHandler.handle` and *into* `WebSocketServer.handleMessage`'s `catch` → `extractRequestId` → `sendErrorResponse` path (`WebSocketServer.swift:151-155`, `:281-320`), but that path is asserted only by a comment. `extractRequestId` is `private static`.

- [ ] Add a small `WebSocketServerTests` locking: a decode failure (unknown type / missing required field) yields a `type:"error"`, `success:false` response that preserves the `requestId` from the raw JSON and carries the error text. (May require loosening `extractRequestId`/`buildErrorResponseData` access for `@testable`.)

## Risk

Low. #1 is pure deletion; #2/#3 are refactors covered by the existing 290-test suite + the decode/dispatch tests; #4 is additive.

kaeawc:
Closes #2859.

Completes the maintainability follow-ups from the typed-command-model migration (#2846 / #2854). Pure refactor + tests; wire format unchanged.

## 1. Dead `CommandError` cases (deletion)
Removed `CommandError.notSupported` and `.elementNotFound` (+ their `errorDescription` arms). Both had **zero constructors** anywhere in `CtrlProxy` — every `.notSupported(` / `.elementNotFound(` throw is on the *separate* `GestureError` (`GesturePerformer.swift`) / `LocatorError` (`ElementLocator.swift`) enums.

## 2. Enum-derived `responseType` (kill the stringly-typed table)
Replaced the ~38-arm `responseType(for requestType: String) -> String` (with its `default:"error"` fallback) with a compiler-enforced, exhaustive `RequestType.responseType: ResponseType` — **no `default`**. `CommandHandler.handle`'s catch now uses `request.requestType.responseType.rawValue`.

The mapping is now **total**: `getCurrentFocus` → `current_focus_result`, `getTraversalOrder` → `traversal_order_result`, `addHighlight` → `highlight_response` — the three the old table silently dropped to `"error"` (their handlers return an error response directly and never throw, so this is unobservable on the wire but removes the silent gap). Adding a new `RequestType` now fails to compile until mapped.

## 3. `CommandPayload` protocol collapses the `requestId` accessor
Introduced `protocol CommandPayload { var requestId: String? { get } }`, conformed by every payload struct (marker-only — they already declare it). Added `WebSocketRequest.payload: CommandPayload` (one exhaustive switch) and collapsed `requestId` to `payload.requestId`, so future common fields come free.

## 4. `WebSocketServerTests`
New suite locking the decode-failure → error-response path #2854 moved out of `CommandHandler.handle` and into `WebSocketServer.handleMessage`'s catch (`extractRequestId(from:)` → `buildErrorResponseData`): an unknown type / missing required field yields a `type:"error"`, `success:false` envelope that preserves the raw-JSON `requestId` and carries the error text — plus direct `extractRequestId` unit tests (present / absent / non-string / malformed). Loosened `extractRequestId` from `private` to internal for `@testable`.

## Validation
- **TDD**: each part's test written first (confirmed red), then implemented to green.
- 364 XCTest tests pass (347 baseline + 17 new). `swift build` green.
- `swiftformat --lint` clean on changed files (`--extensionacl on-declarations`); no new swiftlint violations.
- Regenerated `CtrlProxy.xcodeproj` for the 4 new test files (16-line pbxproj diff).

Risk: low. #1 is pure deletion; #2/#3 are refactors covered by the existing 347-test decode/dispatch suite; #4 is additive.

### DIFF
--- a/ios/control-proxy/CtrlProxy.xcodeproj/project.pbxproj
+++ b/ios/control-proxy/CtrlProxy.xcodeproj/project.pbxproj
@@ -50,12 +50,15 @@
 		8E00E330C839DBF9AC5DCC8E /* ObjCExceptionCatcher.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FFAE974582B8CDFFBA64AFB5 /* ObjCExceptionCatcher.framework */; };
 		96ED92C8F268021CE148C4FF /* CtrlProxy.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 056D5DCA8003C7DFA37244C4 /* CtrlProxy.framework */; };
 		9E975A4E407F93A21BD80EB4 /* SdkDatabaseClient.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0EDA90BC9E26458BCE47381F /* SdkDatabaseClient.swift */; };
+		9EFF1197D83979B0FC0E0BCE /* CommandPayloadTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8A702D259D2A37D9AA088EF8 /* CommandPayloadTests.swift */; };
 		9FE4564D1819768FEFD62606 /* ElementLocator.swift in Sources */ = {isa = PBXBuildFile; fileRef = 95565999DC108319DEED894A /* ElementLocator.swift */; };
 		9FED7BF37AFE34ADDF4E730A /* PinchFallback.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0FEB521226F4C469CBFE7296 /* PinchFallback.swift */; };
 		A603343B94D35DDF48F7AA4D /* Fakes.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9A1B2931E58B158EE51C5F4F /* Fakes.swift */; };
 		A9D79278D7A9C7ED3E7E4F8C /* CtrlProxy.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 056D5DCA8003C7DFA37244C4 /* CtrlProxy.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		A9E8300EE35B13DED1A0E579 /* MainThreadRunner.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6B8E0FBA4D04484920DA1D03 /* MainThreadRunner.swift */; };
+		AE108A9093B6E046855BD678 /* CommandErrorTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2DDC1D66210157A64C7AC76F /* CommandErrorTests.swift */; };
 		AF0C3A37F5AD66261DF7ED24 /* MainThreadRunner.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6B8E0FBA4D04484920DA1D03 /* MainThreadRunner.swift */; };
+		B33435D31AF41A6C03499842 /* WebSocketServerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = AC68B1D8C55DA187332B555E /* WebSocketServerTests.swift */; };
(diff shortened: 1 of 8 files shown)


kaeawc:
## Adversarial review (3 perspectives) — findings addressed

Ran three independent reviewers over the diff (Swift API-design, wire/regression, test-quality/completeness).

**No bugs or regressions found:**
- **Wire/regression:** the new `RequestType.responseType` was compared case-by-case against the deleted `responseType(for:)` table (`git show origin/main`) — **byte-identical for every reachable case**. The 3 formerly-`default:"error"` cases (`getCurrentFocus`/`getTraversalOrder`/`addHighlight`) now map to their real result types, but their handlers never `throw`, so the mapping is unreachable on the error path — no observable change. `"Unknown command type: <type>"` contract preserved (TS `rewriteUnknownCommandError` still matches). No iOS runner re-cut needed (identical wire output).
- **API design:** `responseType` on `RequestType` and the grouped `CommandPayload` conformances are correct/idiomatic; `public` is consistent with the surrounding surface. Both new switches are `default`-free (compiler-total).

**One real gap addressed (test-quality, AC-4):** the original `WebSocketServerTests` reproduced `handleMessage`'s catch via the static helpers instead of driving the real method, so a wiring regression in the catch block itself (e.g. dropping `extractRequestId`) would have shipped green. Fixed in 2ab50d3: introduced a `WebSocketResponding` seam, split the connection lookup from an internal `handleMessage(_:responder:)`, and drove the tests end-to-end through the **real** `handleMessage` with a capturing fake responder (+ a success-path assertion). Verified by mutation: dropping `extractRequestId` in the real catch now fails the suite.

365 XCTest tests green; `swiftformat --lint`/swiftlint clean on changed files.

PR #3018, opened 2026-07-04
fix(observe): finish observeResultCompact follow-ups (#2990)

kaeawc:
## What

Completes the three follow-up riders from #2990 (named in #2978 but not landed by the merged PR #2976, which only did the "compact everywhere" bounds work):

1. **Composition guard** for `--observe-result-compact` × `--actions-diff-observe`.
2. **Wire-schema advertisement** of the compact `bounds` tuple.
3. **Second desktop hierarchy parser** taught the tuple form (real regression fix).

Closes #2990.

## Why

`--observe-result-compact` (PR #2976) flattens every `bounds` object `{left, top, right, bottom}` to the positional tuple `[left, top, right, bottom]` at the `finalizeToolResponse` serialization chokepoint. Three loose ends remained:

- **Investigation finding:** `actionsDiffObserve` is a **dormant flag** — fully plumbed (CLI/env → `outputReductionFlags` → `ServerConfig` → daemon startup) but **no production code reads** `serverConfig.isActionsDiffObserveEnabled()`. So there is no diff implementation to break today; the risk is a *future* differ that reads `bounds.left`-style fields colliding with tuple-shaped bounds. The guard tests pin the ordering invariant now so that future work can't regress it silently.
- **Investigation finding:** the `observe` tool advertises **no** `outputSchema`, but `tapOn`/`accessibility`/`accessibilityFocus`/etc. do, all routing `bounds` through the shared `elementBoundsSchema` (object-only). Under `--observe-result-compact` those served bounds become tuples, so the advertised schema misdescribed the response — a strict MCP client would reject it, and an external consumer couldn't decode `[l,t,r,b]` from `tools/list`. This mirrors the `suppressOutputSchema` consistency precedent already in `toolRegistry.ts` (#2899).
- **Investigation finding (real bug):** PR #2976 taught `RealLayoutDataSource.parseBounds` the tuple form but **missed the sibling `HierarchyParser.parseBoundsElement`**, which has **3 live call sites** (`AutoMobileContent`, `LayoutInspectorDashboard`, `TelemetryDetailPanel` via `parseHierarchyFromJson`). With the flag on it returned `null` for a `JsonArray`, falling through to `ElementBounds(0,0,0,0)` — every element rendered at a zero rect.

## How

**1. `test/server/finalizeToolResponse.test.ts`** — new `compact × actions-diff-observe composition (#2990)` describe block:
[...]

PR #3205, opened 2026-07-05
feat(mcp): advertise ObserveResult outputSchema for observe (#3025)

### LINKED ISSUE #3025: observe tool has no outputSchema — advertise ObserveResult (incl. compact bounds tuple) on the wire
## Summary

The headline `observe` MCP tool declares **no `outputSchema`**, so it advertises nothing machine-readable on the wire — including, now, the `--observe-result-compact` bounds tuple. This surfaced during #2990 / PR #3018.

## Background

- `--observe-result-compact` (env `AUTOMOBILE_OBSERVE_RESULT_COMPACT`) flattens **every** `bounds` object `{left, top, right, bottom}` in an observe result to the positional tuple `[left, top, right, bottom]` at the `finalizeToolResponse` serialization chokepoint (PR #2976).
- PR #3018 (#2990 task 2) taught the shared `elementBoundsSchema` (`src/server/toolOutputSchemas.ts`) to advertise the tuple as a `bounds-object | [l,t,r,b]` union, gated on the flag via `advertiseBoundsForCompact` (`src/server/compactBoundsAdvertisement.ts`) so advertised shape == emitted shape.
- **But `observe` has no `outputSchema`** (`src/server/observeTools.ts` registers it without one; verified: not in the `outputSchema:`-carrying tool set — only `accessibility`, `accessibilityFocus`, `executePlan`, `exportPlan`, `recordSteps`, `setUIState`, `startTestRecording`, `tapOn` declare one). So the union only reaches a client via the **summary-element** bounds on schema-declaring tools like `tapOn`/`accessibilityFocus`. A client hitting bare `observe` — which produces the bulk of compacted bounds (hierarchy nodes, window/root/region) — learns the tuple order only from prose (`docs/design-docs/mcp/feature-flags.md`).

## Task

Give `observe` a real `outputSchema` (an `ObserveResult` schema) so its hierarchy/window/element bounds are advertised machine-readably, and route its `bounds` fields through `elementBoundsSchema` so the compact tuple is flag-advertised there too (via the existing `advertiseBoundsForCompact` hook in `getToolDefinitions`).

## Considerations / risks

- `observe`'s payload is large and somewhat dynamic (the hierarchy `node` field is polymorphic single-object-or-array; `elements` duplicates the tree). A faithful Zod schema is non-trivial — the issue #2990 already flagged this as non-trivial. A `.passthrough()` top level with typed `viewHierarchy`/`elements`/`bounds` sub-schemas may be the pragmatic middle ground.
- Adding an `outputSchema` means the tool now advertises `structuredContent` expectations; confirm it composes with `--tool-results-no-structured-content` (which suppresses advertised output schemas via `suppressOutputSchema` in `src/server/toolRegistry.ts`) and with the compact gating.
- Regenerate `schemas/tool-definitions.json` and confirm no unintended drift.

## Provenance
- Parent: #2990 (observeResultCompact follow-ups), PR #3018.
- Related: #2951 (compact wiring), #2976 (compact everywhere), #2899/PR #2933 (suppressOutputSchema precedent).

kaeawc:
## Summary

Closes #3025. The headline `observe` MCP tool declared **no `outputSchema`**, so it advertised nothing machine-readable on the wire — including the `--observe-result-compact` bounds tuple, of which `observe` (its hierarchy nodes, window/root/region, and `elements` arrays) produces the bulk. A client hitting bare `observe` learned the tuple order only from prose (`docs/design-docs/mcp/feature-flags.md`).

This gives `observe` a real `ObserveResult` `outputSchema` and routes its `bounds` fields through the shared `elementBoundsSchema`, so the compact tuple is flag-advertised there via the existing `advertiseBoundsForCompact` hook — exactly as on the schema-declaring action tools (`tapOn`, `accessibilityFocus`).

## What changed

- **`src/server/toolOutputSchemas.ts`** — new `observeResultSchema`: a `.passthrough()` top level with typed `viewHierarchy`/`elements`/`bounds` sub-schemas (the pragmatic middle ground the issue calls for, given observe's large, polymorphic payload). Adds `viewHierarchyNodeSchema` (recursive; the polymorphic single-object-or-array `node`) and `viewHierarchyResultSchema`. **Every** `bounds` field — hierarchy nodes, window/root/region, `contentHiddenRegions`, `elements`, and the focused/awaited element fields — routes through `elementBoundsSchema`.
- **`src/server/observeTools.ts`** — register `observe` with `{ outputSchema: observeResultSchema }`.
- **`schemas/tool-definitions.json`** — regenerated (observe now carries `outputSchema`; bounds collapsed to the object arm by default, no drift elsewhere).
- **`docs/design-docs/mcp/feature-flags.md`** — the note that "observe declares no outputSchema" is now stale; updated to say observe advertises the union.
- **`scripts/typecheck-baseline.txt`** — +10, all the same `zod`-stub-incompleteness class (`passthrough`/`lazy`/`nullish` "does not exist on type 'ZodSchema'") that this file's existing schemas already carry (the hand-rolled `declare module "zod"` in `src/types/mcp-sdk.d.ts` is incomplete). Only the `toolOutputSchemas.ts` lines were added — no unrelated base-drift.
- **`test/server/observeOutputSchema.test.ts`** — 14 tests: parses the frozen `android-home` fixture in both object and compacted (tuple) forms, polymorphic node recursion, passthrough, and the advertisement gating through the real `getToolDefinitions` (object arm off / tuple on, and suppression under `--tool-results-no-structured-content`).

## Composition (verified)

| Flag state | observe `outputSchema` in `tools/list` | `bounds` arm |
|---|---|---|
| default | advertised | object only (no `prefixItems`) |
| `--observe-result-compact` | advertised | object \| tuple union (`prefixItems`) |
| `--tool-results-no-structured-content` | **suppressed** | n/a |

## ⚠️ Known tradeoff for reviewers (most important thing to weigh)

Adding an `outputSchema` means the wire-boundary strip (`stripToolResultStructuredContent`) now **retains** observe's `structuredContent` duplicate by default, where previously (no schema) it was stripped **unconditionally** (#2759). Reproduced: for the `android-home` fixture that is a **~50 KB duplicate** re-added to every default observe response. This is:

- **Spec-correct** — MCP (2025-06-18) says a tool that advertises `outputSchema` should return matching `structuredContent`; the existing design deliberately couples the two via `toolHasOutputSchema`, and decoupling would be non-compliant.
- **Consistent** — every other schema tool (`tapOn`, …) already behaves this way.
- **Correctness-neutral** — Swift/embedded consumers read only `content[0].text` (#2759), so the duplicate is pure bytes, not a behavior change.
- **Opt-out preserved** — output-reduction deployments set `--tool-results-no-structured-content`, which now suppresses **both** the schema advertisement and the duplicate.

It is nonetheless a **default wire-size increase** in a tool under the "Output Size Reduction" milestone. Flagging explicitly so the milestone owner can confirm the tradeoff is intended (the issue anticipates the `--tool-results-no-structured-content` composition).

## Test plan

- `bun test test/server/ test/features/observe/ test/plan/ test/features/navigation/ test/daemon/` → green
- `bun run build` → green
- `bun run typecheck` → no new errors beyond the recorded baseline
- `schemas/tool-definitions.json` regenerates clean (pre-commit hook confirms in-sync)


kaeawc:
## Adversarial review (3 independent perspectives) + resolutions

Ran three independent reviewers — correctness/defensive, regression/output-size, and convention/completeness. All concluded **merge-ready**; actionable findings addressed in 31ce054.

### Correctness (iOS bounds faithfulness) — fixed
The advertised schema must not reject data the server really emits (it's advertisement-only, but strict clients validate against it).
- **iOS root `Hierarchy.bounds`** is `{left?, top?, right, bottom}` (points) and `compactObserveBounds` can emit `[null, null, r, b]` for a partial root. Routing it through `elementBoundsSchema` (all-four-int object \| int tuple) made the schema reject a real iOS observation. **Fix:** that niche site now rides the hierarchy object's `.passthrough()` (honest by omission). ✅
- **`elements.media[].bounds`** IS compacted like every other bounds (prior comment claimed otherwise). **Fix:** routed through `elementBoundsSchema` via a MediaView passthrough schema, so its tuple is advertised too. ✅
- Verified clean: `advertiseBoundsForCompact` correctly reaches the recursive union inside `$defs`/`$ref`; `z.lazy` terminates on the full real hierarchy; `accessibilityState`/`freshness`/`selectedElements` enums/shapes faithful.

### Regression / output size — accepted tradeoff (verified)
Giving observe a schema means the wire strip now **retains** `structuredContent` by default (reproduced: **+50 KB / ~1.9×** on the fixture). This is spec-mandated (advertise `outputSchema` ⇒ return `structuredContent`), correctness-neutral (Swift/embedded read only `content[0].text`), and opt-out-preserved (`--tool-results-no-structured-content` suppresses **both** schema and duplicate). Decoupling would be MCP-non-compliant and break the honest-by-default coupling (#2899). No other regressions (finalize, internal tool-to-tool, plan capture, nav baseline, daemon proxy cache all unaffected).

### Convention / completeness — fully closes #3025
All 4 task items done; both compositions tested through the real `getToolDefinitions`; per-tool JSON diff shows **only `observe` changed** (no drift); baseline growth confined to the known incomplete-`zod`-stub class.

### Deferred to follow-ups
- Shared `elementBoundsSchema` uses `.int()` while iOS coordinates are fractional **points** — pre-existing (#3018), also affects `tapOn`; unreproduced (no iOS fixture in tree).
- Complete the hand-rolled `declare module "zod"` stub (`src/types/mcp-sdk.d.ts`) that forces the whole file into the typecheck baseline; note the latent `z.lazy` TS7022 risk if the stub is later completed.

### CI note
The typecheck gate was red on 6 pre-existing `src/db/migrations/*` errors from #3176/#2904 (base drift, untouched by this PR); recorded in the baseline (f6b0ca9) to unblock the gate.

PR #3210, opened 2026-07-05
feat(server): debug-log intentional structuredContent omission (#2962)

### LINKED ISSUE #2962: Output reduction: debug-log intentional structuredContent omission (follow-up to #2759)
## Context

Follow-up from #2759 / PR #2958 (gate `structuredContent` duplication). Raised by reviewer HarperZ9 on #2759 and echoed in multi-agent review of PR #2958.

## Problem

When a client consumes both the text and `structuredContent` paths, it should be able to distinguish an **intentional** `structuredContent` omission from an **accidental missing** one. PR #2958 gates `structuredContent`:
- No-schema tools: absence is unambiguous (per MCP 2025-06-18 spec, `structuredContent` was never meaningful without an `outputSchema`).
- Flag-on schema tools: the omission is already signaled by *also* dropping the `outputSchema` advertisement in `tools/list`, so a spec-compliant client won't expect `structuredContent`.

So the distinction is carried by the advertisement, not a separate signal. But for **field/server-side debuggability**, a lightweight trace of when/why the strip fired would help.

## Proposed

Add a `logger.debug` at the strip site (`src/server/index.ts`, where `stripToolResultStructuredContent` is called) recording that `structuredContent` was intentionally omitted for tool `<name>` and the reason (no-schema vs flag). Debug level (dropped unless the user opts into debug logging), so no default noise. Aligns with the repo's "log-and-continue with a why" convention (CLAUDE.md error-handling §3).

## Acceptance

- With debug logging enabled, a stripped tool call emits one debug line naming the tool and the omission reason.
- No log at default (INFO) level; no per-call overhead beyond the guard.

kaeawc:
## Problem

Follow-up from #2759 / PR #2958. When a client consumes both the text and `structuredContent` paths, it should be able to distinguish an **intentional** `structuredContent` omission from an **accidental missing** one. PR #2958 gates `structuredContent` at the MCP CallTool wire boundary (`src/server/index.ts`) — but there was no server-side trace of *when* or *why* the strip fired, which hurts field/server-side debuggability.

## Fix

- **New `structuredContentOmissionReason(hasOutputSchema, flagEnabled?)`** in `src/server/stripToolResultStructuredContent.ts` is the single source of truth for the omission decision, returning the reason the field is omitted:
  - `"no-schema"` — no `outputSchema`, stripped unconditionally (#2759). Takes precedence over the flag.
  - `"flag"` — schema tool stripped because `--tool-results-no-structured-content` is on (#2899).
  - `null` — retained (schema tool, flag off).
  `stripToolResultStructuredContent` now derives its `shouldStrip` from this helper, so the strip and the trace can never disagree.
- **`src/server/index.ts`** emits one `logger.debug` line at the strip site naming the tool and the reason when the omission fires. Debug level → dropped at the default INFO level, so no per-call noise beyond the reason guard. Aligns with the "log-and-continue with a why" convention (CLAUDE.md error-handling §3).

## Tests

- `stripToolResultStructuredContent.test.ts`: full `structuredContentOmissionReason` matrix (explicit-flag-arg + live-`serverConfig` paths) and a consistency check that `reason === null` iff `structuredContent` is kept. Existing strip tests remain green (behavior unchanged).
- `stripToolResultWireBoundary.test.ts`: drives the real CallTool boundary and spies `logger.debug`/`info`/`warn` — no-schema tool emits exactly one `no-schema` trace (and none at info/warn), schema-tool+flag-on emits one `flag` trace, schema-tool+flag-off emits none. Matcher is an anchored regex on the exact message so a format regression fails the test.
- 19 tests pass across the two files. eslint clean, `bun build.ts` OK.

## Acceptance (from #2962)

- [x] With debug logging enabled, a stripped tool call emits one debug line naming the tool and the omission reason.
- [x] No log at default (INFO) level; no per-call overhead beyond the guard.

Closes #2962

### DIFF
--- a/src/server/index.ts
+++ b/src/server/index.ts
@@ -13,7 +13,7 @@ import { createDefaultPlanExecutionLock, type PlanExecutionLock } from "./PlanEx
 
 // Import the tool registry
 import { ToolRegistry, toolHasOutputSchema } from "./toolRegistry";
-import { stripToolResultStructuredContent } from "./stripToolResultStructuredContent";
+import { stripToolResultStructuredContent, structuredContentOmissionReason, responseCarriesStructuredContent } from "./stripToolResultStructuredContent";
 
 // Import the resource registry
 import { ResourceRegistry } from "./resourceRegistry";
@@ -347,7 +347,20 @@ export const createMcpServer = (options: McpServerOptions = {}): McpServer => {
       // #2899). Applied here — not inside a tool handler — so internal handler
       // callers keep reading `structuredContent`, and so plain-registered tools
       // are covered.
-      return stripToolResultStructuredContent(result, toolHasOutputSchema(tool));
+      // Field-debuggability trace (issue #2962): when the wire boundary actually
+      // drops a `structuredContent` tree, emit one debug line naming the tool and
+      // WHY, so a client that reads both paths can tell an intentional omission
+      // from an accidental miss. The reason is resolved once here and passed into
+      // the strip (single source of truth), and the trace is gated on the same
+      // "a field is actually present" condition as the strip, so it never
+      // over-reports on non-envelope responses. Debug level → dropped at the
+      // default INFO level, so no per-call noise beyond the guard. Aligns with the
+      // "log-and-continue with a why" convention.
+      const omissionReason = structuredContentOmissionReason(toolHasOutputSchema(tool));
+      if (omissionReason !== null && responseCarriesStructuredContent(result)) {
+        logger.debug(`[MCP] Omitted structuredContent for tool "${name}" (reason: ${omissionReason})`);
+      }
+      return stripToolResultStructuredContent(result, omissionReason);
     } finally {
       executionTracker.endExecution(execution.id);
     }
--- a/src/server/stripToolResultStructuredContent.ts
+++ b/src/server/stripToolResultStructuredContent.ts
@@ -1,24 +1,70 @@
 import { serverConfig } from "../utils/ServerConfig";
 
 /**
+ * Why `structuredContent` is intentionally omitted for a tool result:
+ * - `"no-schema"` — the tool declares no `outputSchema`, so `structuredContent`
+ *   is dropped unconditionally (#2759).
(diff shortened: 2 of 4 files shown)


kaeawc:
## Multi-agent review + follow-up

Ran `/auto-mobile-code-review` plus three independent adversarial reviewers (correctness/regression, conventions/maintainability, test-adversary/flake). Both correctness and test-adversary verdicted the change correct, behavior-preserving, and non-vacuous (mutation-verified); no blockers. Addressed the actionable findings in the amended commit:

- **Single source of truth (double-eval smell).** The omission reason is now resolved **once** in `index.ts` and passed into `stripToolResultStructuredContent(response, reason)`, instead of being re-derived inside the strip. Removes the redundant `serverConfig` read.
- **Log on actual deletion, not policy (over-report edge).** A shared `responseCarriesStructuredContent` predicate now gates **both** the strip's delete and the debug trace, so the trace fires only when a `structuredContent` tree is really dropped — never on a non-envelope response (image/plain-text). Added a CallTool-boundary test pinning this: a no-schema tool that produces no `structuredContent` emits **no** trace.
- **House logging convention.** Prefixed the trace with `[MCP]` to match the surrounding traces in this handler.
- **Test convention.** Logger spies now use `.mockImplementation(() => {})` (matches sibling suites), and the matcher is an anchored regex on the exact message so a format regression fails the test.

Behavior preserved: `structuredContentOmissionReason` is exactly the old `!hasOutputSchema || flag` gate (verified across the full matrix + an end-to-end agreement test). The only new observable behavior is the debug line.

Local: eslint clean, `bun build.ts` OK, 21/21 tests in the two touched suites, 10/10 determinism on the wire-boundary suite. The only `test/server/` failures are the pre-existing `observeOutputSchema.test.ts` (#3025) cases, untouched by this PR.

kaeawc:
Filed two low-priority deferred follow-ups from the review (optional, no behavior change): #3215 (directly test the real INFO-level suppression of the trace) and #3216 (consider structured log fields vs the formatted string). Auto-merge (squash) is enabled.

Issue #3053, opened 2026-07-04
actions-diff-observe: positional node key cascades on scroll/insert; explore stable node identity + internal-consumer diff exposure


kaeawc:
## Context
Follow-up to merged **PR #3034** (`--actions-diff-observe`, tracking #3026). Consolidates two lower-priority items raised in the multi-perspective review of the PR.

### 1. Positional-key cascade on scroll / mid-list insert
Nodes carry no stable id, so `diffObserveResult` identifies each node by a globally-positional **path key** (ancestor chain of `resource-id + bounds + text + sibling index`). This correctly fixed cross-subtree collisions (the ship-blocker in review), but it is inherently sensitive to reindexing:
- Inserting a row mid-list shifts every following sibling's `sibling index` (and usually `bounds`), so they surface as **remove+add** rather than a small "one row inserted" delta.
- A scroll shifts every visible node's `bounds` → large remove+add churn.

This is acceptable for same-screen state toggles (the common case), but erodes the token savings on scroll-heavy loops.

**Explore:** a content-hash node identity. AutoMobile already computes a `view-id` content-hash in the TS layer (see `ViewHierarchy` processing / the `view-id` field). Keying the diff on a stable content hash (falling back to the positional path key) could turn a scroll into "these N nodes moved" instead of "N removed + N added". Weigh token savings vs. added complexity and false-merge risk.

### 2. Internal tool-to-tool calls receive diffs when the flag is on
`PlanExecutor` invokes the **wrapped** `tool.handler` (so `finalizeToolResponse` runs) and injects the agent's `sessionUuid` (`src/.../PlanExecutor.ts`), so plan step envelopes DO get diffed and update `lastRenderedObservation` when `--actions-diff-observe` is on. Today this is safe: `PlanExecutor` reads only `success`/`error`/`failureObservation`, never `.observation.viewHierarchy`; `criticalSection` uses the raw `deviceAwareHandler` (bypasses finalize); the one internal `.observation` reader (`systemTrayHelpers.ts`) reads a raw `observeScreen.execute` result, not a finalized envelope.

**Task:** add a regression guard (or an explicit `internal`/no-diff context flag threaded from internal callers) so a future internal consumer that reads `.observation.viewHierarchy` off a finalized step envelope cannot silently break when the flag is on. At minimum, a test asserting plan-step logic is unaffected under the flag.

### 3. (minor) Double session-activity write per diffed action
[...]

PR #2929, opened 2026-07-03
fix(daemon): typed SessionCacheData slots as observe-cache source of truth (#2917)

kaeawc:
## What

Make the typed `SessionCacheData` slots the canonical source of truth for the
observe cache (issue #2917, **Option A** — the issue's preferred fix).

`SessionCacheData` declared typed slots that **nothing wrote**, while the real
observe cache landed in the untyped `customData` map under same-named keys but a
**different type**:

```ts
// before — src/daemon/sessionManager.ts
lastHierarchy?: string;   // typed string, never written
lastScreenshot?: string;  // typed string, never written
lastObserveTime?: number; // never written
customData?: Record<string, any>; // where the data ACTUALLY landed
```

- `customData.lastHierarchy` held a **`ViewHierarchyResult` object**, not the declared `string`.
- `customData.lastScreenshot` held the base64 string.

The typed `lastHierarchy: string` was a **type lie**: any consumer trusting the
interface (e.g. the #2761 hierarchy-diff baseline) reads `undefined`, or worse
assumes the wrong shape.

Closes #2917.

## Why

The observe-cache write was repaired in #2758 / PR #2891 so the cache is finally
populated — but into `customData` under a decoy-shadowed key. A reader coded
against `session.cacheData.lastHierarchy` (the typed slot) still sees nothing.
The keying is correct (per-session `cacheData.customData`); only the type/slot
modelling was wrong.

## How

- **Retype** `SessionCacheData.lastHierarchy?: ViewHierarchyResult`. `lastScreenshot?: string`
  and `lastObserveTime?: number` were already correct and are kept.
- **Add first-class setters** on `SessionManager`:
  - `setLastHierarchy(sessionId, hierarchy: ViewHierarchyResult)` — writes the typed
    top-level slot and stamps `lastObserveTime`.
  - `setLastScreenshot(sessionId, b64: string)` — writes the typed top-level slot.
  Both route through the existing `updateSessionCache`, so session activity/heartbeat
  bookkeeping is unchanged and a missing session is a logged no-op (not a throw).
- **Route** `src/server/toolRegistry.ts` observe caching through the setters instead of
  the generic `updateSessionCache(context, "lastHierarchy"/"lastScreenshot", …)` that
  wrote to `customData`. Still runs before `finalizeToolResponse`, so the cache keeps
  the **full untrimmed** hierarchy while the wire copy is sanitized (#2758 behavior intact).
- `lastActionTime` stays ad-hoc in `customData` (not part of the typed interface).
[...]

PR #3214, opened 2026-07-05
refactor(daemon): promote well-known customData keys to typed SessionCacheData slots (#2973)

### LINKED ISSUE #2973: Promote well-known `customData` keys (keep-awake, device-label) to typed SessionCacheData slots (follow-up to #2917)
## Summary

Follow-up to #2917 / PR #2929. That PR promoted the observe-cache fields to typed
`SessionCacheData` slots, but the `customData?: Record<string, any>` escape hatch
remains — and it still holds **well-known, fixed-type keyed state**, not
genuinely-ad-hoc data. The exact #2917 bug class (a typed expectation vs an
untyped `customData` key of a different type) can recur for any of these keys.

## The remaining untyped keyed state

Despite living in the `Record<string, any>` bag, these keys have fixed types and
are accessed via named constants — i.e. the same situation `lastHierarchy` was in:

- `KEEP_SCREEN_AWAKE_STATE_KEY` → `KeepScreenAwakeState`
  - written: `src/server/ToolExecutionContext.ts` (`ensureKeepScreenAwake`)
  - read: `src/daemon/sessionManager.ts` (`restoreKeepScreenAwake`),
    `src/server/ToolExecutionContext.ts`
- `DEVICE_LABEL_CACHE_KEY` → device-label map
  - written/read: `src/server/deviceLabelMapping.ts`, `src/daemon/socketServer.ts`

Each is retrieved with an unchecked `as SomeType` cast, so a type drift between
writer and reader is invisible at compile time — precisely the failure #2917
addressed for the observe cache.

## Proposed work

1. Promote `keepScreenAwake?: KeepScreenAwakeState` and `deviceLabels?: <map type>`
   to typed top-level `SessionCacheData` slots (or a typed nested `customData`
   shape), with dedicated setter/getter helpers, removing the `as` casts.
2. Once all well-known keys are typed, evaluate whether `customData` can be
   dropped entirely or narrowed to a closed discriminated shape, eliminating the
   `Record<string, any>` escape hatch that enables this bug class.

### Optional cheap safeguard (interim)
Add a lint/grep guard (or a review-checklist note) forbidding direct
`.lastHierarchy` / `.lastScreenshot` writes outside `SessionManager` so callers
must use the setters — the retyped tests currently write the typed slot directly
via the public `updateSessionCache(Partial<SessionCacheData>)`, which TS now
type-checks but doesn't force through the setter.

## Acceptance
- No well-known session-cache key is accessed via an unchecked `as` cast out of
  `customData`; each has a typed slot + helper.
- `SessionCacheData` has no remaining `Record<string, any>` escape hatch, or it
  is explicitly documented as the last ad-hoc-only surface with no fixed-type keys.

## Context
- Design-review finding on PR #2929 (#2917 fix). Same bug class as #2917.

kaeawc:
## Summary

Follow-up to #2917 / PR #2929. That PR promoted the observe-cache fields to typed
`SessionCacheData` slots, but the `customData?: Record<string, any>` escape hatch
remained — and it still held **well-known, fixed-type keyed state** (keep-awake
`KeepScreenAwakeState`, the device-label map) accessed via unchecked `as` casts.
A writer/reader type drift on any such key was invisible at compile time: the exact
#2917 decoy bug class.

This PR promotes those two keys to typed top-level slots with dedicated helpers and
removes the `customData` bag entirely (they were its only consumers).

- Add `keepScreenAwake?: KeepScreenAwakeState` and `deviceLabels?: DeviceLabelMap`
  to `SessionCacheData`; **remove `customData`** (Acceptance #2 — the escape hatch is
  fully eliminated, not just narrowed).
- Add dedicated `set/getKeepScreenAwake` and `set/getDeviceLabels` helpers on
  `SessionManager`. Getters route through `getSession` so they record **no session
  activity**, mirroring the side-effect-free `getLastRenderedObservation` reader (#3053).
- Rewire every consumer to the typed slots/helpers, dropping the `as` casts:
  `ensureKeepScreenAwake` / `restoreKeepScreenAwake`, `registerDeviceLabelMap` /
  `getDeviceLabelMap`, and `socketServer.resolveDeviceLabelSession` (the narrow
  `DaemonStateAccess` interface gains `getDeviceLabels`).
- Remove the now-orphaned generic `updateSessionCache(context, key, value)` free
  function and the `KEEP_SCREEN_AWAKE_STATE_KEY` constant.

### Acceptance (from the issue)
- ✅ No well-known session-cache key is accessed via an unchecked `as` cast out of
  `customData`; each has a typed slot + helper.
- ✅ `SessionCacheData` has no remaining `Record<string, any>` escape hatch.

Closes #2973.

## Tests

- `test/daemon/sessionCacheTypedSlots.test.ts` — new: round-trip for both slots,
  getters record no session activity (contrast `getSessionCache`), and the
  `releaseSession` keep-awake-restore read path.
- `test/daemon/sessionCacheNoEscapeHatch.test.ts` — new source-scan guard: fails if
  the `customData` bag or an unchecked `as KeepScreenAwakeState | undefined` cast is
  reintroduced in any former consumer.
- `test/server/ToolExecutionContext.test.ts` — new: `ensureKeepScreenAwake` writes the
  typed `keepScreenAwake` slot (not `customData`) on setup.
- Migrated existing seeding/decoy tests (`sessionManager`, `socketServer*`,
  `toolRegistry.*`) off `customData` to the typed slots.
- `NODE_ENV=test bun test test/daemon/ test/server/` → all green (608 daemon, 878
  daemon+server pass, 0 fail).
- `bunx turbo run lint build` → clean. `bun run typecheck` → no NEW errors from this
  diff (the remaining adm-zip/ajv baseline errors are pre-existing, in untouched files).

## Review

- Ran `/auto-mobile-code-review` against the rebased branch; no actionable findings.
  Verified the removed runtime validation in `resolveDeviceLabelSession` is safe —
  `Session.cacheData` is in-memory only (never hydrated from the DB), so the typed
  slot is the only write path; empty/absent map still falls back to the base session.

### DIFF
--- a/src/daemon/daemonRequestHandlers.ts
+++ b/src/daemon/daemonRequestHandlers.ts
@@ -1,10 +1,11 @@
 import { DaemonRequest } from "./types";
-import { Session } from "./sessionManager";
+import { DeviceLabelMap, Session } from "./sessionManager";
 
 export interface DaemonStateAccess {
   isInitialized(): boolean;
   getSessionManager(): {
     getSession(sessionId: string): Session | null;
+    getDeviceLabels(sessionId: string): DeviceLabelMap | undefined;
     releaseSession(sessionId: string): Promise<string | null>;
   };
   getDevicePool(): {
--- a/src/daemon/sessionManager.ts
+++ b/src/daemon/sessionManager.ts
@@ -15,6 +15,16 @@ import type { ObserveResult } from "../models/ObserveResult";
 export type DeviceLabelMap = Record<string, string>;
 
 /**
+ * Narrow seam for restoring keep-awake state on session release. Production uses
+ * `KeepScreenAwakeManager`; tests inject a fake to assert (without a real device)
+ * that `releaseSession` reads the typed `keepScreenAwake` slot and passes its full
+ * payload to `restore` — the behavioral half of the #2973 typed-slot round trip.
+ */
+export interface KeepScreenAwakeRestorer {
+  restore(state: KeepScreenAwakeState): Promise<void>;
+}
+
+/**
  * Session Cache Data
  *
  * Stores data that can be reused across multiple tool calls
@@ -93,6 +103,7 @@ export class SessionManager {
   private releaseCallbacks: SessionReleaseCallback[] = [];
   private deviceSessionRepository: DeviceSessionRepository;
   private readonly getBarrier: () => DbWriteBarrier;
+  private readonly keepScreenAwakeRestorerFactory: (device: BootedDevice) => KeepScreenAwakeRestorer;
 
   // Session timeout: 30 minutes
   private readonly SESSION_TIMEOUT_MS = 30 * 60 * 1000;
@@ -109,10 +120,15 @@ export class SessionManager {
     // same-process DB reopen (resetDbWriteBarrier swaps in a fresh barrier) is
     // seen instead of a pinned drained instance (issue #2912).
     getBarrier: () => DbWriteBarrier = getDbWriteBarrier,
+    // Seam for keep-awake restore (issue #2973): defaults to the real manager;
+    // tests inject a fake to assert the typed slot's payload reaches `restore`.
+    keepScreenAwakeRestorerFactory: (device: BootedDevice) => KeepScreenAwakeRestorer =
+    device => new KeepScreenAwakeManager(device),
   ) {
     this.timer = timer;
     this.deviceSessionRepository = deviceSessionRepository;
     this.getBarrier = getBarrier;
+    this.keepScreenAwakeRestorerFactory = keepScreenAwakeRestorerFactory;
     // Start periodic cleanup of expired sessions
     this.startCleanupTimer();
   }
@@ -542,7 +558,7 @@ export class SessionManager {
       platform: session.platform,
       deviceId: session.assignedDevice
     };
-    const manager = new KeepScreenAwakeManager(device);
+    const manager = this.keepScreenAwakeRestorerFactory(device);
     await manager.restore(state);
   }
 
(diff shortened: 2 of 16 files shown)


kaeawc:
## Review pass (multi-perspective) — addressed

Ran `/auto-mobile-code-review` plus three independent reviewers (adversarial-correctness, principal-architecture, test-quality). Summary:

- **Correctness:** no grounded bugs. The change is behavior-preserving — `Session.cacheData` is in-memory only (never hydrated from the DB), so the removed runtime `typeof/Array.isArray` validation in `resolveDeviceLabelSession`/`getDeviceLabelMap` guarded only garbage the old untyped bag allowed; the typed setter is now the sole write path. null-vs-undefined, missing-session no-op, and getter side-effects all verified equivalent to the old paths.
- **Architecture:** ship — dropping `customData` entirely (vs. narrowing) is the right call since both keys are well-known/fixed-type; `DeviceLabelMap`'s home + import direction, helper design, and naming all match repo convention.
- **Test-quality:** two real gaps fixed in the follow-up commit:
  1. The keep-awake restore round trip (the #2917 writer/reader-drift bug class) was only proven structurally. Added a narrow injected `KeepScreenAwakeRestorer` seam so `releaseSession` can be asserted to read the **full** typed-slot payload and hand it to `restore` — with the applied:false / no-state no-restore branches.
  2. `deviceLabelMapping.ts` had no behavioral coverage (socketServer tests use a fake `getDeviceLabels`). Added `test/server/deviceLabelMapping.test.ts` exercising the real SessionManager typed-slot round trip + null branches.
  Also hardened the escape-hatch guard regex to survive an `interface`→`type` refactor.

One optional low-priority follow-up (extend the guard to forbid setter-bypass writes in production code) is filed separately.

PR #3229, opened 2026-07-05
refactor(server): add ToolRegistry.callInternal seam for internal tool calls (#3108)

### LINKED ISSUE #3108: Optional: ToolRegistry.callInternal(name, args) seam to remove the residual mark-then-call footgun
## Context
Nicety surfaced in the architecture review of **PR #3096** (which introduced `markInternalToolCall` for #3087). The PR routed all 8 internal wrapped-handler call sites (5 navigation/setup + 3 PlanExecutor) through the shared `markInternalToolCall` helper, removing the mistype/forget-the-key footgun.

## Residual gap
The call sites still write the two-step `tool.handler(markInternalToolCall(args))`. A **new** internal caller could still forget the `markInternalToolCall` wrap entirely and silently pollute the agent-facing diff baseline (the #3087 bug class).

## Proposal (weigh, don't over-engineer)
A single seam — e.g. `ToolRegistry.callInternal(name, args, progress?, signal?)` — that resolves the tool (`getToolForPlan`/`getTool`), marks the args, and invokes `.handler()` in one call, returning the raw response. This centralizes the `getTool` null-check the sites repeat and makes the internal path the path of least resistance.

## Caveat
The sites are heterogeneous: `DefaultUIStateSetup.setupScrollPosition` reads `found` off the result, `NavigateTo.executeToolCall` discards it, and `PlanExecutor.captureFailureObservation` races the handler against a timeout. The seam must return the raw response and let callers keep their own result handling. With only 8 sites and strong test coverage today, this is a low-priority ergonomics improvement, not a correctness fix.

## Pointers
- `src/server/internalToolCall.ts`, `src/server/toolRegistry.ts` (`getTool`/`getToolForPlan`, `registerDeviceAware` wrapper).

kaeawc:
## Summary

Adds `ToolRegistry.callInternal(tool, args, progress?, signal?, { forPlan? })` as the single internal-caller entry point for tool-to-tool calls. In one step it resolves the tool (by name via `getTool`/`getToolForPlan`, or from an already-resolved `RegisteredTool`), marks the args via `markInternalToolCall`, invokes the wrapped `.handler()`, and returns the **raw** response — callers keep their own result handling.

This removes the residual mark-then-call footgun surfaced in PR #3096's architecture review: a **new** internal caller could still forget the `markInternalToolCall` wrap and silently advance the agent-facing diff baseline (the #3087 bug class). The seam makes the marked internal path the path of least resistance.

## Linked Issue

Closes #3108

## Changes

- **New seam** `ToolRegistry.callInternal` in `src/server/toolRegistry.ts`. The `string | RegisteredTool` overload handles the heterogeneous sites the issue's caveat calls out: pass a **name** to also centralize the resolve-and-null-check (`forPlan` selects `getToolForPlan` for tools hidden from MCP discovery but valid in plans); pass an **already-resolved tool** for sites that must resolve it themselves first (e.g. to run `tool.schema.parse` before the call). Throws `ActionableError` when a name does not resolve.
- **Migrated all 7 live internal sites** to the seam, preserving each caller's own resolution / null-check / result handling:
  - `NavigateTo.executeToolCall` (string form)
  - `DefaultUIStateSetup`: `setupScrollPosition` (reads `found`), `tapOnElement`, bottomsheet swipe (has a fallback branch), `tapCloseButton` (tool form)
  - `PlanExecutor`: `captureFailureObservation` (races handler against a 3s timeout), `executeStep` (pre-parses against `tool.schema`, passes `signal`)
- `markInternalToolCall` now lives only inside the seam; its import is dropped from all three migrated files.

## Validation

- [x] `bun test` — full suite green (6651 pass, 2 skip, 0 fail)
- [x] `bun run typecheck` reports no new errors from this change (the 4 flagged errors are pre-existing local `node_modules` `adm-zip`/`ajv` noise in untouched files — identical with these changes stashed; resolve on CI)
- [x] New unit tests use no real DB / no timers; run in <100ms
- [x] No Android/iOS/shell/schema surfaces touched (TS only)

## Acceptance criteria (inferred — the issue is an optional ergonomics nicety)

- [x] **AC1** — single `callInternal` seam marks args + invokes `.handler()`, returning the raw response
- [x] **AC2** — resolves by name via `getTool` and `getToolForPlan` (`forPlan`)
- [x] **AC3** — throws `ActionableError` when the tool is unresolved
- [x] **AC4** — all 7 internal sites route through the seam, each keeping its own result handling
- [x] **AC5** — internal calls stay opted out of diff/baseline (marker still reaches the wrapped handler)

Pinned by new `test/server/toolRegistry.callInternal.test.ts` (AC1–AC3) plus existing regression suites that stay green (AC4/AC5): `toolRegistry.internalNoDiff`, `planExecutorInternalNoDiff`, `navigationInternalNoDiff`, `navigateToInternalNoDiffE2E`, `DefaultUIStateSetup`.

## Follow-ups

None identified.

### DIFF
--- a/src/features/navigation/DefaultUIStateSetup.ts
+++ b/src/features/navigation/DefaultUIStateSetup.ts
@@ -4,7 +4,6 @@ import { AdbClient } from "../../utils/android-cmdline-tools/AdbClient";
 import { logger } from "../../utils/logger";
 import { AndroidCtrlProxyClient } from "../observe/android";
 import { ToolRegistry } from "../../server/toolRegistry";
-import { markInternalToolCall } from "../../server/internalToolCall";
 import { NavigationEdge, UIState } from "./NavigationGraphManager";
 import { ModalState, ScrollPosition } from "../../utils/interfaces/NavigationGraph";
 import { UIStateExtractor } from "./UIStateExtractor";
@@ -177,7 +176,7 @@ export class DefaultUIStateSetup implements UIStateSetup {
       // scroll-position and #2758 lastHierarchy fixes). With `result` typed,
       // `result.found` is now a compile error and the `getStructuredField` keys
       // are checked against the payload.
-      const result = asToolEnvelope<SwipeOnToolPayload>(await swipeOnTool.handler(markInternalToolCall(swipeOnArgs)));
+      const result = asToolEnvelope<SwipeOnToolPayload>(await ToolRegistry.callInternal(swipeOnTool, swipeOnArgs));
 
       if (getStructuredField(result, "success") && getStructuredField(result, "found")) {
         logger.info(`[UI_STATE_SETUP] Successfully scrolled to target element`);
@@ -282,8 +281,9 @@ export class DefaultUIStateSetup implements UIStateSetup {
         args.elementId = element.resourceId;
       }
 
-      // Internal setup tap (#3087): no diff/strip, no baseline advance.
-      await tapTool.handler(markInternalToolCall(args));
+      // Internal setup tap (#3087) via the callInternal seam (#3108): no
+      // diff/strip, no baseline advance.
+      await ToolRegistry.callInternal(tapTool, args);
 
       // Small delay for UI to update
       await this.sleep(100);
@@ -377,13 +377,13 @@ export class DefaultUIStateSetup implements UIStateSetup {
           // swipeOn targets a scrollable child and would scroll the sheet's inner
           // list instead of dragging the sheet itself down (SwipeOn.execute). We
           // want the full-screen downward swipe (executeScreenSwipe) that a
-          // dismissal needs. Internal setup swipe (#3087): no diff/strip, no
-          // baseline advance.
-          await swipeTool.handler(markInternalToolCall({
+          // dismissal needs. Internal setup swipe (#3087) via the callInternal
+          // seam (#3108): no diff/strip, no baseline advance.
+          await ToolRegistry.callInternal(swipeTool, {
             direction: "down",
             autoTarget: false,
             platform
-          }));
+          });
           await this.sleep(200);
 
           // Verify dismissal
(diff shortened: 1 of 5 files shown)


kaeawc:
## Review triage (3 orthogonal lenses + auto-mobile-code-review)

Ran three review lenses — **regression/behavioral**, **test adequacy**, **API-contract/interface** — plus the AutoMobile code-review skill.

**Regression lens:** no genuine regressions. The four behavioral deltas were each verified safe:
- `NavigateTo.executeToolCall` now throws `ActionableError` (was plain `Error`) after its `logger.info` — caught and stringified by the enclosing handler; no test depends on the type/ordering; `ActionableError` is the repo-preferred error.
- PlanExecutor's debug-log preview no longer shows `__internalNoDiff` (marker now applied inside the seam) — nothing reads `parsedParams` post-call and no test asserts on the logged string; the marker still reaches the wrapped handler.
- `captureFailureObservation` Promise.race identity + `clearTimeout` cleanup unchanged (seam returns the handler promise untouched).
- `DefaultUIStateSetup` sites keep their own `getTool` null-check and pass the resolved tool — control flow (incl. the bottomsheet-swipe fallback) unchanged.

**API-contract lens:** approved, no must-fix. Signature/options shape matches existing conventions; `string | RegisteredTool` union is type-guarded; correctly scoped to the concrete singleton (the DI interface has no consumer needing it); no tool-schema/`tool-definitions.json` changes; `markInternalToolCall` still exported and now used only inside the seam.

**Test adequacy lens:** AC1–AC3 fully pinned by the new suite; AC4/AC5 (marker delivery) well-covered by the existing internal-no-diff regression suites that exercise the migrated sites end-to-end.

### Deferred (pre-existing, out of scope)
The test-adequacy lens flagged three coverage gaps — `tapOnElement` error path, `tapCloseButton` loop termination, and the `captureFailureObservation` Promise.race **timeout** path. All three predate this PR: the migration preserved the exact try/catch, loop, and race structure (only `tool.handler(markInternalToolCall(x))` → `callInternal(tool, x)`), so no behavior this PR changed is left uncovered. Growing these tests here would expand scope beyond the issue; noting them rather than filing since they're minor and orthogonal to the seam.

PR #3232, opened 2026-07-05
perf(db): cache Bun SQLite prepared statements

### LINKED ISSUE #2797: perf(db): add a bounded prepared-statement cache to the Bun SQLite dialect
## Summary
`BunSqliteConnectionState.executeQuery` calls `db.prepare(sql)` on **every** query (`src/db/bunSqliteDialect.ts:162`), so SQLite re-parses and re-plans identical SQL on every invocation. Because the daemon runs the same handful of Kysely-compiled statements over and over (upserts, node-visit increments, telemetry inserts), this is pure repeated work — measured at roughly 2.5x waste on hot serial inserts (~4.29us vs ~1.72us/op). Adding a **bounded** LRU cache of `Statement` objects removes the re-plan cost without changing any query results. Found via a DB concurrency/migration audit (2026-07-01), finding P2stmt, priority P2, verified against source.

## Background & history
- The `stmt = db.prepare(sql)` + `.all(...)` / `.run(...)` pattern was introduced in commit `79591d4e` (2025-12-31, [PR #97](https://github.com/kaeawc/auto-mobile/pull/97), "Migrate test runner from Mocha to Bun") when the dialect was first written as a bun:sqlite port of Kysely's better-sqlite3 `SqliteDialect`.
- The line was last touched in commit `1670cf174` (2026-06-29, [PR #2632](https://github.com/kaeawc/auto-mobile/pull/2632)), which only changed how the `db` handle is obtained (`this.#getDatabase()` lazy init) — the per-query `prepare` was left in place.
- Shaped this way because it mirrors upstream Kysely's reference dialect literally (prepare-per-query is the simplest correct implementation). It was never revisited for the daemon's steady-state hot path, where the same compiled SQL repeats indefinitely.

## The problem
Every logical query re-prepares from scratch:

```ts
// src/db/bunSqliteDialect.ts:157-187
    try {
      await this.#beforeQuery?.();
      const db = this.#getDatabase();

      // Prepare statement
      const stmt = db.prepare(sql);

      // Check if this is a SELECT query or query with RETURNING clause
      const sqlLower = sql.trim().toLowerCase();
      const isSelect = sqlLower.startsWith("select");
      const hasReturning = sqlLower.includes("returning");

      if (isSelect || hasReturning) {
        const rows = stmt.all(...(parameters as any[])) as R[];
        ...
      } else {
        const result = stmt.run(...(parameters as any[]));
        ...
      }
    }
```

`db.prepare(sql)` parses the SQL text and builds a fresh query plan each time. The concurrency model makes this a per-logical-query cost: a SINGLE bun:sqlite connection is guarded by a hand-rolled async mutex in this file (`#enterQuery`/`#reserveTransaction`) that serializes individual queries, and everything runs inside ONE singleton daemon per uid. So there is no cross-connection reason to re-prepare — the exact same `sql` string flows through `executeQuery` repeatedly (e.g. `navigationRepository.updateNodeVisit`, `installedAppsRepository.upsertInstalledApp`, and the high-frequency telemetry inserts), and each pass pays the parse+plan tax again.

Note bun:sqlite's `db.query()` *does* memoize prepared statements, but its cache is **unbounded**. Kysely's `SqliteQueryCompiler` emits variable-arity `IN (?, ?, ...)` SQL, so the set of distinct SQL strings is unbounded — a naive switch to `db.query()` (or an unbounded `Map`) would grow the statement cache without limit and leak. That is why the cache must be bounded.

## Impact
- Steady-state DB throughput inside the daemon is ~2.5x slower than necessary on hot repeated statements (~4.29us vs ~1.72us/op measured on serial inserts). This is latent, not a correctness bug — nothing breaks, results are identical.
- Blast radius is every query the daemon runs; most visible under bursty workloads (navigation graph updates, telemetry/failure-event ingestion) where the same SQL fires many times in a tight window.
- Narrow severity: it is a perf-only cleanup. No user-visible failure, no data loss.

## Proposed fix

**Sequence this AFTER the batched-ingestion direction in #2799 (per review).** This is the lowest-leverage item in the perf cluster: bun:sqlite is synchronous, so the prepare cost is saved on a background write path and amounts to microseconds. If #2799's batched ingestion lands (one multi-row `INSERT` per flush prepares one statement and commits once), the prepare-per-query cost largely evaporates and this issue is mostly moot. Only build the cache if batching is rejected *and* profiling on a real device still shows prepare cost dominating steady-state. If built, the **bounded size + finalize-on-evict design below is non-negotiable** (an unbounded `db.query()` cache leaks against Kysely's variable-arity `IN(?,?,…)` SQL).

Add a bounded LRU `Map<string, Statement>` on `BunSqliteConnectionState`, keyed by SQL text, with a hard cap and LRU eviction. Finalize evicted statements and finalize all on `close()`.

```ts
// sketch — src/db/bunSqliteDialect.ts
const MAX_CACHED_STATEMENTS = 200;
#statementCache = new Map<string, Statement>(); // Map preserves insertion order -> LRU

#getStatement(db: BunDatabase, sql: string): Statement {
  const cached = this.#statementCache.get(sql);
  if (cached) {
    // refresh LRU position
    this.#statementCache.delete(sql);
    this.#statementCache.set(sql, cached);
    return cached;
  }
  const stmt = db.prepare(sql);
  this.#statementCache.set(sql, stmt);
  if (this.#statementCache.size > MAX_CACHED_STATEMENTS) {
    const oldestKey = this.#statementCache.keys().next().value as string;
    const evicted = this.#statementCache.get(oldestKey);
    this.#statementCache.delete(oldestKey);
    evicted?.finalize();
  }
  return stmt;
}
```

Then replace `const stmt = db.prepare(sql);` with `const stmt = this.#getStatement(db, sql);`, and in `close()` finalize every cached statement before clearing the map and closing `#db`.

Constraints the fix must respect:
- Do **not** switch to bun:sqlite's `db.query()` — its cache is unbounded (see above).
- Do **not** change the SELECT-vs-RETURNING branching or the `.all`/`.run` result shape; query results and `numAffectedRows`/`insertId` must be byte-for-byte identical.
- Finalize on eviction and on `close()` to avoid leaking prepared statements (mirror the deterministic-cleanup discipline of `performanceAuditRepository.startPeriodicPruning`'s `stopPeriodicPruning`).
- Cache is keyed by SQL text only; parameters are still bound per call via `.all(...)` / `.run(...)`, so variable-arity `IN (...)` SQL stays correct (distinct arities are distinct keys, bounded by eviction).

- [ ] Repeated identical SQL reuses a cached `Statement`: executing the same compiled query N times calls `db.prepare` exactly once for that SQL string.
- [ ] Bounding holds: after executing more than `MAX_CACHED_STATEMENTS` distinct SQL strings, `#statementCache.size` never exceeds the cap, and the least-recently-used entry is evicted.
- [ ] Evicted statements and all statements on `close()` are finalized (assert via a fake `Statement` whose `finalize()` is recorded).
- [ ] Results are unchanged: SELECT rows, RETURNING rows, `numAffectedRows`, and `insertId` match the pre-change behavior for INSERT/UPDATE/DELETE and SELECT paths.
- [ ] New unit test (e.g. `test/db/bunSqliteDialect.test.ts`) uses an in-memory fake `BunDatabase`/`Statement` per CLAUDE.md conventions (interfaces + fakes, no real device, no filesystem), runs in <100ms, and is non-flaky.
- [ ] `bun test` green; `turbo run lint build test` green.
- [ ] Perf win measured with a before/after micro-benchmark on hot serial inserts (target ~2.5x on repeated statements), with an explicit note that query results are identical before and after.

### Direction (Devin Osei) — defer / sequence after #2799
- **Lowest-leverage item in the perf cluster.** The daemon runs one `bun:sqlite` connection and `executeQuery` calls `stmt.all()`/`stmt.run()` **synchronously** (`bunSqliteDialect.ts:170,180`), blocking the event loop for the query's full duration. The mutex serializes logical queries but cannot make them concurrent — bun:sqlite is synchronous and single-threaded, so there is no "parallelism" to reclaim and adding connections wouldn't help. This (and #2787/#2788) are micro-opts *underneath* that ceiling. The cited benchmark (~4.29us → ~1.72us) is real but microseconds on a background write path already off the interactive tool path.
- **The real lever is batched ingestion (#2799):** one multi-row `INSERT` per flush prepares one statement and commits once, which dwarfs prepare-caching, *also* retires the per-commit fsync (#2787) and shrinks mutex hold time — without adding an LRU to the query hot path. If batching lands, prepare-per-query cost mostly evaporates.
- **Recommendation: defer #2797 behind the batching decision** (a "sequence after the thing that makes it mostly moot," not a hard no). If batching is rejected and real-device profiling still shows prepare cost dominating steady-state, build the bounded cache exactly as specified.
- **If built, bounded + finalize is non-negotiable.** Kysely's `SqliteQueryCompiler` emits variable-arity `IN (?, ?, …)`, so distinct SQL strings are unbounded — bun's own `db.query()` memoization would leak. The bounded-LRU-with-finalize-on-evict/close is the only acceptable shape, but it's net-new machinery on the single hottest code path (new surface for a use-after-finalize bug) in exchange for microseconds.

## References
- `src/db/bunSqliteDialect.ts:162` — per-query `db.prepare(sql)` (the hot path)
- `src/db/bunSqliteDialect.ts:152-196` — `executeQuery` body
- `src/db/bunSqliteDialect.ts:198-203` — `close()` where statements must be finalized
- `src/db/database.ts:117` — `BunSqliteDialect` wiring
- `src/db/performanceAuditRepository.ts:372` — `startPeriodicPruning`/deterministic-cleanup precedent
- Finding P2stmt, DB concurrency/migration audit, 2026-07-01

kaeawc:
## Summary

Adds a bounded LRU prepared-statement cache to `BunSqliteConnectionState` so repeated Kysely SQL strings reuse a Bun SQLite statement instead of preparing on every query. Cached statements are keyed by SQL text, capped at 200 entries, refreshed on access, finalized on eviction, and finalized on connection close.

## Linked Issue

Closes #2797

## Bug / Regression Context

- **Prior attempts:** none
- **Observed symptom:** hot repeated daemon SQL paid `db.prepare(sql)` on every execution even when the compiled SQL text was identical.
- **Repro steps / conditions:** repeated serial inserts/selects through `BunSqliteConnectionState.executeQuery`.

## Validation

- [x] `bun test test/db/bunSqliteDialect.test.ts` passes
- [x] `bun run lint` passes
- [x] `bun run build` passes
- [x] `bun test --bail` passes: 6651 pass, 2 skip, 0 fail
- [x] `bun run turbo:validate` passes (`turbo run lint build test`)
- [x] `bun run typecheck` reports no new errors (519 known baseline errors)
- [x] `git diff --check` passes
- [x] New code uses fakes; focused dialect tests run in <100ms-class timing locally

Micro-benchmark note for 50k in-memory hot serial inserts:

- prepare each iteration: 1.789 us/op
- reuse one statement: 0.860 us/op

Query result-shape coverage was added for SELECT rows, RETURNING rows and affected-row count, and write `numAffectedRows` / `insertId`.

## Device Verification

N/A. Database dialect-only change; no on-device behavior changed.

## Follow-ups

None.

### DIFF
--- a/src/db/bunSqliteDialect.ts
+++ b/src/db/bunSqliteDialect.ts
@@ -182,7 +182,8 @@ export class BunSqliteConnectionState {
         }
         const db = this.#getDatabase();
 
-        const stmt = this.#getStatement(db, sql);
+        const schemaChanging = this.#isSchemaChangingSql(sql);
+        const stmt = schemaChanging ? db.prepare(sql) : this.#getStatement(db, sql);
 
         // Check if this is a SELECT query or query with RETURNING clause
         const sqlLower = sql.trim().toLowerCase();
@@ -192,24 +193,38 @@ export class BunSqliteConnectionState {
         // word char, so `\breturning\b` correctly rejects `returning_items`.
         const hasReturning = /\breturning\b/.test(sqlLower);
 
-        if (isSelect || hasReturning) {
-          // For SELECT queries or queries with RETURNING, return all rows
-          const rows = stmt.all(...(parameters as any[])) as R[];
-          return {
-            rows,
-            numAffectedRows: hasReturning ? BigInt(rows.length) : undefined,
-          };
-        } else {
-          // For INSERT/UPDATE/DELETE queries without RETURNING, execute and return changes
-          const result = stmt.run(...(parameters as any[]));
-          return {
-            rows: [],
-            numAffectedRows: BigInt(result.changes),
-            insertId:
-              result.lastInsertRowid !== undefined
-                ? BigInt(result.lastInsertRowid)
-                : undefined,
-          };
+        try {
+          if (isSelect || hasReturning) {
+            // For SELECT queries or queries with RETURNING, return all rows
+            const rows = stmt.all(...(parameters as any[])) as R[];
+            const result = {
+              rows,
+              numAffectedRows: hasReturning ? BigInt(rows.length) : undefined,
+            };
+            if (schemaChanging) {
+              this.#clearStatementCache();
+            }
+            return result;
+          } else {
+            // For INSERT/UPDATE/DELETE queries without RETURNING, execute and return changes
+            const writeResult = stmt.run(...(parameters as any[]));
+            const result = {
+              rows: [],
+              numAffectedRows: BigInt(writeResult.changes),
+              insertId:
+                writeResult.lastInsertRowid !== undefined
+                  ? BigInt(writeResult.lastInsertRowid)
+                  : undefined,
+            };
+            if (schemaChanging) {
+              this.#clearStatementCache();
+            }
+            return result;
+          }
+        } finally {
+          if (schemaChanging) {
+            stmt.finalize();
+          }
         }
       } catch (error) {
         // BigInt-safe: Kysely can bind BigInt params, and a bare
@@ -280,6 +295,10 @@ export class BunSqliteConnectionState {
       }
     }
     return statement;
+  }
+
+  #isSchemaChangingSql(sql: string): boolean {
+    return /^(?:create|alter|drop)\b/i.test(sql.trim());
   }
 
   #clearStatementCache(): void {
--- a/src/db/bunSqliteDialect.ts
+++ b/src/db/bunSqliteDialect.ts
@@ -154,6 +154,7 @@ export class BunSqliteConnectionState {
   async rollbackTransaction(owner: symbol): Promise<void> {
     try {
       await this.executeQuery(CompiledQuery.raw("rollback"), owner);
+      this.#clearStatementCache();
     } finally {
       this.#clearTransactionOwner(owner);
     }
--- a/src/db/bunSqliteDialect.ts
+++ b/src/db/bunSqliteDialect.ts
@@ -15,6 +15,10 @@ import {
 import { CompiledQuery } from "kysely";
 import { ActionableError } from "../models/ActionableError";
 
+const MAX_CACHED_STATEMENTS = 200;
+
+type BunStatement = ReturnType<BunDatabase["prepare"]>;
+
 /**
  * Kysely dialect for Bun's built-in SQLite.
  * Based on SqliteDialect but uses bun:sqlite instead of better-sqlite3.
@@ -111,6 +115,7 @@ export class BunSqliteConnectionState {
   #pendingTransactions = 0;
   #activeQueries = 0;
   #waiters: Array<() => void> = [];
+  #statementCache = new Map<string, BunStatement>();
   #closed = false;
 
   constructor(database: BunDatabase | (() => BunDatabase), beforeQuery?: () => Promise<void>) {
@@ -177,8 +182,7 @@ export class BunSqliteConnectionState {
         }
         const db = this.#getDatabase();
 
-        // Prepare statement
-        const stmt = db.prepare(sql);
+        const stmt = this.#getStatement(db, sql);
 
         // Check if this is a SELECT query or query with RETURNING clause
         const sqlLower = sql.trim().toLowerCase();
@@ -238,6 +242,7 @@ export class BunSqliteConnectionState {
     // observes the closed state and rejects instead of running against a dead
     // handle. Then wake waiters so queued queries settle promptly.
     this.#closed = true;
+    this.#clearStatementCache();
     if (this.#db) {
       this.#db.close();
       this.#db = null;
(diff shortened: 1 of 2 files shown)


kaeawc:
Addressed the Codex review finding in 102bf23ea: CREATE/ALTER/DROP now use single-use prepared statements, and the reusable statement cache is cleared after successful DDL so cached SELECT metadata cannot survive schema changes. Added a focused test for cache invalidation after schema-changing SQL.

kaeawc:
Addressed the rollback cache-invalidation review in f5be62f5d: successful transaction rollback now clears the statement cache before releasing the transaction owner, and  covers cached statement finalization/re-prepare after rollback.

kaeawc:
Follow-up note for the rollback review: addressed in f5be62f5d. Successful transaction rollback now clears the statement cache before releasing the transaction owner, and the dialect test suite covers cached statement finalization and re-prepare after rollback.

PR #3217, opened 2026-07-05
feat(types): type-checker-enforced ToolHandler envelope boundary (#2932)

kaeawc:
## Summary

Closes #2932.

Delivers the **type-checker-enforced** arm of the envelope-vs-`structuredContent`
dead-read fix (the by-convention arm shipped in #2921 / issue #2907). Payload
fields live only under `structuredContent`; `createStructuredToolResponse` hoists
only `success`/`error` to the envelope top level, so a raw `response.found` read
silently yields `undefined`. Before this PR the compiler couldn't catch that: the
handler boundary erased the payload type back to `any` and `getStructuredField`
was stringly-typed (a `"founded"` typo relocated the same silent-`undefined` bug
from `.found` to `"found"`).

- **Typed payload interfaces** — `SwipeOnToolPayload` ([SwipeOnResult.ts](src/models/SwipeOnResult.ts)) and `ObserveToolPayload` ([ObserveResult.ts](src/models/ObserveResult.ts)) are the single source of truth for the two tool payloads read back internally.
- **`getStructuredField` is now fully typed** ([toolUtils.ts](src/utils/toolUtils.ts)): `<TPayload, K extends keyof TPayload & string>` over `StructuredToolResponse<TPayload>`. The `key` is constrained to `keyof TPayload` (a typo is a compile error) and the return type is `TPayload[K]` (a wrong `<T>` assertion is impossible). The own-key runtime guard is unchanged (defense-in-depth against prototype keys).
- **Producers annotated** — `swipeOnHandler` / `observeHandler` now declare `Promise<StructuredToolResponse<…Payload>>`, tying each payload type to its actual producer (drift breaks at the source).
- **Read sites migrated** — the toolRegistry scroll-position + lastHierarchy/lastScreenshot reads ([toolRegistry.ts](src/server/toolRegistry.ts)) and `DefaultUIStateSetup` scroll setup ([DefaultUIStateSetup.ts](src/features/navigation/DefaultUIStateSetup.ts)) narrow the `any` the heterogeneous registry hands back to the concrete typed envelope. `response.found` off the envelope top level is now a **compile error** at these sites.

Pure type tightening — no runtime behavior change. The shared `ToolHandler` /
`DeviceAwareToolHandler` registry map stays `Promise<any>` (it is intentionally
heterogeneous); the realized durability lives at the typed producers + typed read
sites + typed accessor, exactly the gap the three adversarial reviewers of #2921
converged on.

## Evaluation Criteria

[...]

Issue #2932, opened 2026-07-03
Durable typed-envelope: re-type the ToolHandler boundary so envelope payload dead-reads are a compile error


kaeawc:
## Summary

Follow-up to #2907 / PR #2921. PR #2921 landed the **by-convention** arm of the envelope-vs-`structuredContent` dead-read fix: a typed `StructuredToolResponse<T>` envelope, `getStructuredField<T>` / `getStructuredPayload<T>` accessors, and a `no-restricted-syntax` lint rule banning raw `x.structuredContent.<field>` reads. This issue tracks the **stronger, type-checker-enforced** arm that #2907 called "the more durable fix."

## Why this is still open

Adversarial review of PR #2921 (three independent reviewers) converged on one gap: the typed envelope's payoff — making `response.found` a **compile error** — is not realized, because the handler boundary erases the type back to `any`:

- `ToolHandler` / `DeviceAwareToolHandler` both return `Promise<any>` (`src/server/toolRegistry.ts:41,46`).
- The call-site variable is `let response: any` (`src/server/toolRegistry.ts:312`).
- `DefaultUIStateSetup` reads `result` from `swipeOnTool.handler(...)`, also `any` (`src/features/navigation/DefaultUIStateSetup.ts:161`).

So `createStructuredToolResponse<T>` infers `T` inside the handler and immediately discards it at the `Promise<any>` return. At every read site `response`/`result` are `any`, so `response.found` compiles clean (silent `undefined`). The lint rule + accessor prevent the mistake by convention, but the compiler still can't.

A secondary consequence (contrarian-architect + type-skeptic finding): `getStructuredField(response, "found")` is **stringly-typed** — a typo `"founded"` or a wrong `<T>` param is unchecked and reproduces the same silent-`undefined` dead branch, just relocated from `.found` to `"found"`. A typed handler boundary would make both the typo and the wrong-type a compile error (`response.structuredContent.found` against a concrete payload type).

## Proposed direction

1. Thread `StructuredToolResponse<TPayload>` through the `ToolHandler` return type instead of `Promise<any>`, at least for the tools whose results are read back internally (`swipeOn`, `observe`).
2. Give the read sites (`toolRegistry` scroll-position/lastHierarchy/screenshot, `DefaultUIStateSetup`) a concrete payload type so `getStructuredField`'s `key`/`T` are checked, or so `response.structuredContent.found` type-checks directly and `response.found` is a compile error.
[...]

PR #3234, opened 2026-07-05
feat(server): thread StructuredToolResponse<TPayload> through the tool registry (#3222)

### LINKED ISSUE #3222: Thread StructuredToolResponse<TPayload> through the tool registry (remove the asToolEnvelope unchecked cast)
## Summary

Follow-up to #2932 / PR #3217. That PR delivered the type-checker-enforced envelope
boundary at the point of consumption: typed producers (`swipeOnHandler`/`observeHandler`
return `Promise<StructuredToolResponse<…Payload>>`), a `keyof`-constrained
`getStructuredField`, and typed read sites. But the shared registry boundary is still
`any`, so the guarantee is enforced by a **local assertion**, not an end-to-end typed flow.

## Why this is still open

- `ToolHandler` / `DeviceAwareToolHandler` return `Promise<any>` and
  `RegisteredTool.handler: ToolHandler` (`src/server/toolRegistry.ts`).
- `ToolRegistry.getTool("swipeOn")?.handler(...)` is therefore `Promise<any>`.
- At the read sites the concrete envelope type is asserted via `asToolEnvelope<T>()`
  (an explicit, greppable UNCHECKED cast introduced in PR #3217, review feedback). It
  names the boundary crossing but does not verify the handler actually produces that
  shape — the producer's own return annotation is the only thing keeping them honest,
  and the two are never cross-checked by the compiler.

So `response.found` is a compile error at the migrated sites *because a type was asserted*,
not because the registry threads the payload type through.

## Proposed direction

Thread the payload generic through the registry so the boundary is checked, not asserted:

- Parameterize `RegisteredTool<TPayload>` / a typed `getTool<TPayload>(name)` (or a small
  typed-lookup helper for the handful of internally-consumed tools: `swipeOn`, `observe`)
  so `getTool("observe")` yields a handler typed `Promise<StructuredToolResponse<ObserveToolPayload>>`.
- Then the `asToolEnvelope` casts at the 3 read sites
  (`toolRegistry` scroll-position + lastHierarchy/lastScreenshot, `DefaultUIStateSetup`
  scroll setup) can be removed — the flow is checked end-to-end.
- Consider a runtime shape assertion inside `asToolEnvelope` as an interim safety net if
  full genericization is deferred.

Scope note: a fully generic registry touches ~130 register sites and a heterogeneous
dispatch pipeline; a name→payload typed-lookup map for just the internally-read tools is a
smaller, high-value first step.

## Acceptance

- `getTool("observe")`/`getTool("swipeOn")` (or the internal-lookup helper) return a
  handler whose result type is the concrete `StructuredToolResponse<…Payload>`.
- The `asToolEnvelope<T>()` unchecked casts at the internal read sites are eliminated (or
  backed by a runtime assertion), so the compile-error guarantee is enforced by the
  boundary rather than by a local assertion.

Provenance: deferred from PR #3217 (issue #2932) as out-of-scope; issue #2907 lineage.

kaeawc:
## Summary

Follow-up to #2932 / PR #3217, **composed with the `callInternal` seam (#3108)** that
landed on main during this work. That PR typed the envelope *producers* and the *read*
helper, but the shared registry boundary was still `any`, so the internal read sites
bridged the gap with `asToolEnvelope<T>()` — an explicit **unchecked** cast never
cross-checked against the producer's annotation.

This threads the concrete payload type through the registry for the **internally-consumed**
tools (`swipeOn`, `observe`) — the sanctioned smaller step (a fully generic registry over
~130 register sites is explicitly out of scope).

## Linked Issue

Closes #3222

## Changes

- **New `src/server/internalToolPayloads.ts`** — `InternalToolPayloads` name→payload map
  (single source of truth) + `narrowInternalToolEnvelope(name, response)`, a
  **runtime-validated** narrowing (bad shape → `undefined`, matching `getStructuredField`).
- **`ToolRegistry.callInternalTyped<K>(name, args)`** — a typed variant of `callInternal`
  (#3108): it **delegates to `callInternal`** (so the `markInternalToolCall` guarantee is
  preserved) and narrows the result to `StructuredToolResponse<…Payload> | undefined`.
  **No unchecked cast anywhere.**
- **Rewired the 3 read sites** — the two `DefaultAfterToolCallHandler` sites (in-flight
  pipeline `any`) use `narrowInternalToolEnvelope`; `DefaultUIStateSetup` scroll setup uses
  `callInternalTyped`.
- **Removed the unused `asToolEnvelope` helper + its test** — the unchecked-cast footgun.
- **Type-only AC1 guard** in `toolRegistry.ts` fails the `bun run typecheck` gate if
  `callInternalTyped`'s return widens to `any` (the gate compiles only `src`, so a pin in
  `test/` would be invisible — verified the guard fires on an `any` regression).

## Acceptance criteria

- ✅ **AC1** — `callInternalTyped("observe"|"swipeOn")` resolves to the concrete
  `StructuredToolResponse<…Payload>` envelope (the internal-lookup seam threads the payload
  type through the registry). Pinned by the runtime tests + the CI-visible compile-time guard.
- ✅ **AC2** — the `asToolEnvelope<T>()` unchecked casts at the internal read sites are
  eliminated (all three now runtime-validated), enforced by a source-scan test (identifier
  gone from `src/`) plus the existing scroll-position / lastHierarchy / DefaultUIStateSetup
  suites staying green (behavior preserved).

## Validation

- [x] `bun test` — 6660 pass, 0 fail
- [x] `bun run lint` clean
- [x] `bun run typecheck` — no new errors from these files (the 4 flagged are pre-existing
  local `node_modules` adm-zip/ajv noise, absent in CI's clean install)
- [x] Rebased onto latest `main`; composes with `callInternal` (#3108) rather than duplicating it

## Follow-ups

- The full generic registry across ~130 register sites remains the #2932 core; this is the
  named-tool typed-seam first step. No new follow-up filed.

### DIFF
--- a/src/features/navigation/DefaultUIStateSetup.ts
+++ b/src/features/navigation/DefaultUIStateSetup.ts
@@ -8,8 +8,7 @@ import { NavigationEdge, UIState } from "./NavigationGraphManager";
 import { ModalState, ScrollPosition } from "../../utils/interfaces/NavigationGraph";
 import { UIStateExtractor } from "./UIStateExtractor";
 import { RealObserveScreen } from "../observe/ObserveScreen";
-import { asToolEnvelope, getStructuredField } from "../../utils/toolUtils";
-import { SwipeOnToolPayload } from "../../models";
+import { getStructuredField } from "../../utils/toolUtils";
 import { UIStateSetup } from "./interfaces/UIStateSetup";
 import { defaultTimer, Timer } from "../../utils/SystemTimer";
 
@@ -124,7 +123,8 @@ export class DefaultUIStateSetup implements UIStateSetup {
     );
 
     try {
-      // Get the swipeOn tool from the registry
+      // Resolve swipeOn first so a missing tool degrades gracefully (return null)
+      // rather than throwing out of the `callInternalTyped` seam below.
       const swipeOnTool = ToolRegistry.getTool("swipeOn");
       if (!swipeOnTool) {
         logger.warn(`[UI_STATE_SETUP] swipeOn tool not found, skipping scroll setup`);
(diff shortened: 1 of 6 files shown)


kaeawc:
## Review triage (Phase 6)

Ran `/auto-mobile-code-review` plus three orthogonal review lenses (regression, test-adequacy, API-contract). Triage:

**Fixed**
- **AC1 not CI-enforced** (test-adequacy lens): AC1 is a type-level guarantee, but the `bun run typecheck` gate compiles only `src` (`tsconfig.json` include), so the compile-time pin in the test file was invisible to CI and would pass against an `any` regression. Added type-only `AssertTrue`/`IsAny` guards in `src/server/internalToolPayloads.ts` that fail to compile if `getInternalTool`'s handler-result type widens to `any` or is no longer the concrete envelope. Verified they fire (TS2344 ×2) when the return type is regressed to `any`, clean otherwise. Corrected the overclaiming test comment. (commit 920243e68)

**Accepted as-is (no change)**
- Regression lens: no regressions — narrowing is behavior-preserving at all 3 read sites (every consumer goes through `getStructuredField`, which already treated `undefined` and a `structuredContent`-less envelope identically).
- API-contract lens: `getInternalTool` double-cast is the correct single home for the crossing (runtime shapes match; `RegisteredTool` ⊇ `TypedInternalTool`); `import type` cycle erases (acyclic runtime graph); `asToolEnvelope` removal breaks no importer; `narrowInternalToolEnvelope(_name,…)` phantom-param is a standard type-witness idiom.

**Deferred**
- `InternalToolPayloads` map drift vs producers: same exposure that existed with `asToolEnvelope<SwipeOnToolPayload>`; the map reuses the canonical producer types by reference, and the real fix is the full generic registry (#2932), explicitly out of scope here. No new follow-up filed.

kaeawc:
Rebased onto latest main and reconciled with the `callInternal` seam (#3108) that landed during this work: replaced the standalone `getInternalTool` accessor with `ToolRegistry.callInternalTyped`, which **delegates to `callInternal`** (preserving the `markInternalToolCall` guarantee) and returns the concrete `StructuredToolResponse<…Payload> | undefined` via runtime-validated narrowing — no unchecked cast. AC1's compile-time guard now targets `callInternalTyped` and is verified to fire on an `any` regression.

PR #3336, opened 2026-07-06
Fix startDevice readiness checks

### LINKED ISSUE #3334: startDevice reports a device as booted before the OS has finished booting
# `startDevice` reports a device as booted before the OS has finished booting

## Summary

`startDevice` can return a `"booted"` response for a device whose OS has not finished booting. A client that immediately follows `startDevice` with `installApp` (or `launchApp`, `tapOn`, etc.) then fails against a device auto-mobile has reported as ready. On Android this surfaces as:

```
adb: failed to install <apk>: Error: device is still booting.
```

The device finishes booting a short time later, so the failure is a race, not a genuinely broken device.

There are three contributing causes. The most impactful — the "device is already running" fast paths in `startDevice` performing **no** boot-completion check — is **platform-agnostic and affects both Android and iOS**. The other two are Android-specific gaps in boot detection; the iOS readiness path is implemented correctly and is a useful reference for fixing them.

## Environment

- auto-mobile: 0.0.38 (also present on `main` at time of writing)
- Platforms: Android emulator (AVD) and iOS simulator
- Flow: MCP `startDevice` immediately followed by `installApp`

### Android
1. Have an emulator that is in the process of booting, or start one and let it register with adb (it appears in `adb devices`) before its package manager is ready. This window is easy to hit on a cold or resource-constrained host.
2. Call `startDevice` for that device (by `deviceId`, by matching criteria against running devices, or matching an image whose `isRunning` is true).
3. `startDevice` returns a `"booted"` response immediately.
4. Call `installApp`.
5. `installApp` fails: `adb: failed to install ...: Error: device is still booting.`
6. Wait ~30–60s and the same install succeeds — the device was mid-boot when it was reported ready.

### iOS
1. Start a simulator and let `simctl list devices` report it as `Booted` (this state is set early, before SpringBoard is up and the sim is usable).
2. Call `startDevice` matching that already-running simulator.
3. `startDevice` returns a `"booted"` response immediately, **without calling `simctl bootstatus`** (see cause #1).
4. A follow-up install/launch can fail because the simulator is not yet fully booted, even though auto-mobile reported it ready.

### 1. The "already running" fast paths in `startDevice` do no readiness check (both platforms)

`src/server/deviceTools.ts` — `startDeviceHandler` is platform-agnostic and has three early returns that call `buildBootedResponse(..., "booted", ...)` as soon as a device appears in the booted-device list, none of which wait for readiness:

- Direct `deviceId` lookup — `deviceTools.ts:205-206`
- Criteria match against booted devices — `deviceTools.ts:247-251`
- Matched image with `isRunning === true` — `deviceTools.ts:262-272`

`buildBootedResponse` sets `isReady: true` and returns. Only the cold-boot path (`bootAndRespond`) calls `waitForDeviceReady` (`deviceTools.ts:316`). So any device merely *present in the booted-device list* — but not finished booting — is reported as ready and handed straight to the caller.

This affects **both platforms**:

- **Android** — a device visible in `adb devices` (see cause #2) is returned without waiting; its package manager may not be up.
- **iOS** — `simctl` sets a simulator's state to `Booted` early, before SpringBoard is up and the sim can install/launch apps. That is precisely why the cold-boot path bothers to call `simctl bootstatus`. But the fast paths return such a simulator with `isReady: true` and **never call `bootstatus`** — skipping the one check that would catch a not-yet-usable sim. (`buildBootedResponse` does call `ensureCtrlProxyReady` for iOS, but that only verifies the automation proxy over WebSocket, its failures are non-fatal/swallowed, and it is not a boot-complete gate — so it does not substitute for `bootstatus`.)

### 2. `getBootedAndroidDevices` discards the adb device-state column (Android-specific)

`src/utils/android-cmdline-tools/AdbClient.ts:673-680` parses `adb devices` but keeps only the serial, ignoring the state:

```ts
const lines = result.stdout.split("\n").slice(1);
const devices = lines
  .filter(line => line.trim().length > 0)
  .map(line => {
    const parts = line.split("\t");
    return { name: parts[0], platform: "android", deviceId: parts[0] } as BootedDevice;
  });
```

`adb devices` reports a state per line (`device`, `offline`, `booting`, `unauthorized`, …). Because the state is dropped, a device that adb itself reports as `offline`/`booting` is treated as booted, and feeds directly into the fast paths in cause #1.

The iOS discovery path does not have this problem — `getBootedSimulatorsChecked` (`src/utils/ios-cmdline-tools/SimCtlClient.ts:756`) strictly filters `device.isAvailable && device.state === "Booted"`, so transient states (`Booting`, `Shutting Down`, `Shutdown`, …) are excluded. The Android path should filter on state the same way.

### 3. The cold-boot readiness gate never checks the boot-complete system properties (Android-specific)

`src/utils/android-cmdline-tools/AndroidEmulatorClient.ts:1251-1275` gates readiness on:

- `adb get-state` output containing `device`, and
- `adb shell pm list packages` returning at least one `package:` line.

It never checks the canonical Android boot-complete signals — `sys.boot_completed`, `dev.bootcomplete`, or `init.svc.bootanim`. (A repo-wide search for these property names returns no matches, and `adb wait-for-device` is not used anywhere.) `pm list packages` can return output transiently before the package manager is ready to service an `install`, leaving a residual race even on the cold-boot path.

By contrast, the iOS readiness gate is implemented correctly: `waitForSimulatorReady` (`src/utils/ios-cmdline-tools/SimCtlClient.ts:637`) blocks on `simctl bootstatus <udid> -b`, the canonical "wait until fully booted" command. The Android gate should have an equivalent boot-complete check.

## Suggested fixes

In rough order of impact:

1. **Verify readiness on the "already running" fast paths** (`deviceTools.ts:205`, `247`, `262`) — *both platforms.* Before returning a `"booted"` response, route through `waitForDeviceReady` (or run a lightweight boot-complete check). For iOS this means calling `bootstatus`; for Android, the boot-complete check from fix #3. This is the single change that closes the cross-platform hole.

2. **Filter on device state in `getBootedAndroidDevices`** (`AdbClient.ts:673-680`) — *Android.* Parse the second tab-separated field and only treat `device` as booted; exclude `offline`, `booting`, `unauthorized`, etc. This prevents not-yet-ready devices from ever reaching the fast paths. (iOS already does this at `SimCtlClient.ts:756`.)

3. **Add a boot-complete check to the Android readiness gate** (`AndroidEmulatorClient.ts:1251-1275`) — *Android.* Require `getprop sys.boot_completed` == `1` (optionally also `init.svc.bootanim` == `stopped`) in addition to the existing state and `pm` checks. This mirrors what `simctl bootstatus` already provides on iOS.

## Platform summary

| Cause | Android | iOS |
|---|---|---|
| #1 — fast paths skip readiness | affected | affected (same handler) |
| #2 — discovery leaks transient states | affected | not affected (strict `Booted` filter) |
| #3 — readiness gate lacks boot-complete check | affected | not affected (uses `bootstatus`) |

The iOS readiness path (`bootstatus` + strict `Booted` discovery filter) is a good reference implementation for the Android fixes in #2 and #3.

## Impact

Any client that calls `startDevice` and then acts on the device (install, launch, tap) without its own boot-completion polling will intermittently fail against a device auto-mobile has reported as ready. The failure is transient and timing-dependent, which makes it look like a flaky install rather than a readiness-reporting bug. Because cause #1 bypasses readiness entirely for already-running devices, it is most likely to bite in setups that reuse warm devices rather than cold-booting one per run.

kaeawc:
## Summary

Fixes `startDevice` readiness reporting so already-running devices go through the platform readiness gate before returning `isReady: true`. Android discovery now ignores `adb devices` rows that are not in `device` state, and Android emulator readiness waits for boot-complete properties in addition to adb state and package-manager responsiveness.

## Linked Issue

Closes #3334

## Bug / Regression Context

- **Prior attempts:** None.
- **Observed symptom:** `startDevice` could return a booted/ready response for an Android emulator or iOS simulator whose OS was still finishing startup, causing immediate `installApp`, `launchApp`, or interaction calls to fail.
- **Repro steps / conditions:** Reuse an already-running-but-not-ready device, or match an Android emulator visible to adb before package installation is safe.

## Validation

- [x] `bun run turbo:validate` passes (`bun run lint` + build + `bun test`)
- [x] Android/iOS changes: covered by focused TypeScript tests with fakes/FakeTimer; no shell/Swift/Docker/schema validation needed
- [x] New code uses interfaces + fakes + FakeTimer; focused unit tests run in <100ms per test
- [ ] Rebased onto latest `main`, conflicts resolved

## Device Verification

- [ ] Verified on a real Android device / iOS simulator via `observe`
- [ ] Screenshots or before/after attached

## Follow-ups

- [x] Follow-up issues filed for gaps not addressed here: none needed

### DIFF
--- a/src/server/deviceTools.ts
+++ b/src/server/deviceTools.ts
@@ -203,7 +203,8 @@ export function registerDeviceTools() {
         perf.endOperation("directLookup");
 
         if (found) {
-          return await buildBootedResponse(found, "booted", perf, args);
+          const readyDevice = await waitForAlreadyRunningDevice(found, deviceUtils, perf, args);
+          return await buildBootedResponse(readyDevice, "booted", perf, args);
         }
 
         // Check images — need to boot
@@ -248,7 +249,8 @@ export function registerDeviceTools() {
           if (progress) {
             await progress(100, 100, "Found matching running device");
           }
-          return await buildBootedResponse(match, "booted", perf, args);
+          const readyDevice = await waitForAlreadyRunningDevice(match, deviceUtils, perf, args);
+          return await buildBootedResponse(readyDevice, "booted", perf, args);
         }
       }
 
@@ -263,8 +265,9 @@ export function registerDeviceTools() {
           const booted = await deviceUtils.getBootedDevices(args.platform);
           const running = booted.find(d => d.name === imageMatch.name || d.deviceId === imageMatch.deviceId);
           if (running) {
+            const readyDevice = await waitForAlreadyRunningDevice(running, deviceUtils, perf, args);
             return await buildBootedResponse(
-              { ...running, osVersion: imageMatch.osVersion, formFactor: imageMatch.formFactor, screenWidth: imageMatch.screenWidth, screenHeight: imageMatch.screenHeight },
+              { ...readyDevice, osVersion: imageMatch.osVersion, formFactor: imageMatch.formFactor, screenWidth: imageMatch.screenWidth, screenHeight: imageMatch.screenHeight },
               "booted",
               perf,
               args,
@@ -292,6 +295,21 @@ export function registerDeviceTools() {
       throw new ActionableError(`Failed to start ${args.platform} device: ${error}`);
     }
   };
+
+  async function waitForAlreadyRunningDevice(
+    device: BootedDevice,
+    deviceUtils: PlatformDeviceManager,
+    perf: ReturnType<typeof createPerformanceTracker>,
+    args: StartDeviceArgs,
+  ): Promise<BootedDevice> {
+    perf.startOperation("waitForReady");
+    const readyDevice = await deviceUtils.waitForDeviceReady(
+      { ...device, isRunning: true },
+      args.timeoutMs,
+    );
+    perf.endOperation("waitForReady");
+    return { ...device, ...readyDevice };
+  }
 
   async function bootAndRespond(
     image: DeviceInfo,
--- a/src/utils/android-cmdline-tools/AdbClient.ts
+++ b/src/utils/android-cmdline-tools/AdbClient.ts
@@ -35,6 +35,11 @@ function getAdbPathCache(): TTLCache<string, string> {
   return adbPathCache;
 }
 
+export function resetAdbClientCaches(): void {
+  deviceListCache = null;
+  adbPathCache = null;
+}
+
 
 // Enhance the standard execFileAsync result to implement the ExecResult interface
 const execFileAsync: ExecFileAsync = async (
@@ -674,9 +679,14 @@ export class AdbClient implements AdbExecutor {
 
     const devices = lines
       .filter(line => line.trim().length > 0)
-      .map(line => {
+      .flatMap(line => {
         const parts = line.split("\t");
-        return { name: parts[0], platform: "android", deviceId: parts[0] } as BootedDevice;
+        const deviceId = parts[0]?.trim();
+        const state = parts[1]?.trim();
+        if (!deviceId || state !== "device") {
+          return [];
+        }
+        return [{ name: deviceId, platform: "android", deviceId } as BootedDevice];
       });
 
     // Cache the result
--- a/src/utils/android-cmdline-tools/AndroidEmulatorClient.ts
+++ b/src/utils/android-cmdline-tools/AndroidEmulatorClient.ts
@@ -1242,23 +1242,41 @@ export class AndroidEmulatorClient implements AndroidEmulator {
             if (emulator && emulator.deviceId) {
               logger.debug(`Target emulator found: ${emulator.name} (${emulator.deviceId}) - starting readiness checks`);
 
-              // Check if the device is online and ready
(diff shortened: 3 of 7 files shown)


kaeawc (commented):
AutoMobile code-review pass: no confirmed findings. Reviewed regression behavior, test adequacy, and API/interface contracts against issue #3334. Local validation is green: focused readiness tests, `bun run typecheck`, `git diff --check`, and `bun run turbo:validate`. Residual risk is live-device/emulator behavior for Android boot properties, which is not verified in this PR.

kaeawc:
Second AutoMobile code-review pass used a selected-device identity perspective. Confirmed one false-negative risk: Android readiness could fall back to the first local emulator when an already-running device had been selected by deviceId but AVD name matching failed. Addressed in 990453019 by threading the selected adb deviceId into waitForEmulatorReady and adding a regression test that fails on the old behavior. Local validation after the fix: focused tests, typecheck, git diff --check, and bun run turbo:validate all pass.

PR #3357, opened 2026-07-07
fix(ci): make iOS simulator boot idempotent

kaeawc:
## Summary

Make simulator boot paths tolerate the states that show up in CI when one XCTestRunner phase shuts down simulators and the next phase immediately boots the same UDID. Both the CI script and `SimCtlClient.startSimulator` now use `simctl bootstatus -b`, which boots shutdown simulators, accepts already-booted simulators, and waits for readiness while preserving the script's stdout / `GITHUB_OUTPUT` UDID contract.

## Bug / Regression Context

- **Prior attempts:** The workflow already runs `xcrun simctl shutdown all || true` between the Xcode 26.2 and 26.3 simulator phases.
- **Observed symptom:** `simctl boot` failed with `Unable to boot device in current state: Booted` when CoreSimulator still considered the selected iPhone simulator booted.
- **Repro steps / conditions:** Run the XCTestRunner simulator CI job where the second boot phase selects the same iOS runtime/UDID shortly after the first phase shuts simulators down.

## Validation

- [x] `bats test/bats/boot-simulator.bats`
- [x] `bun test test/utils/ios-cmdline-tools/simctl.test.ts`
- [x] `shellcheck scripts/ios/boot-simulator.sh`
- [x] `git diff --check`
- [ ] `scripts/all_fast_validate_checks.sh` passes (`bun run lint` + `bun test`)
- [ ] `bun run typecheck` reports no new errors (baseline gate)
- [x] Android/iOS changes: ran the matching `scripts/` validation
- [ ] New code uses interfaces + fakes + FakeTimer; unit tests run in <100ms
- [ ] Rebased onto latest `main`, conflicts resolved

## Device Verification

N/A. This is CI simulator boot orchestration covered by mocked BATS regression tests.

## Follow-ups

None.


src/utils/ios-cmdline-tools/SimCtlClient.ts
```diff
@@ -595,9 +595,11 @@ export class SimCtlClient implements SimCtl {
     logger.debug(`Starting iOS simulator ${udid}`);
     const perf = createGlobalPerformanceTracker();
 
-    perf.startOperation("simctlBoot");
-    await this.executeCommand(`boot ${udid}`);
-    perf.endOperation("simctlBoot");
+    // `bootstatus -b` is idempotent: it boots shutdown simulators, accepts
+    // already-booted simulators, and waits until CoreSimulator reports ready.
+    perf.startOperation("bootstatus");
+    await this.executeCommand(`bootstatus ${udid} -b`);
```
kaeawc: This moves the readiness wait into `startSimulator()`, but this method does not receive the caller's `timeoutMs`. In the `startDevice` path, `deviceTools.bootAndRespond()` calls `deviceUtils.startDevice(image)` first and only passes `args.timeoutMs` to `waitForDeviceReady()` afterward, so a stuck `bootstatus -b` here can block before the requested timeout is enforced. The CI shell script can keep using `bootstatus -b`, but the TS path should either keep `startSimulator()` as a quick boot trigger and rely on `waitForSimulatorReady(..., timeoutMs)`, or thread a timeout into `startSimulator()` and test that `startDevice --timeoutMs` still bounds the first wait.

kaeawc (commented):
Review pass focused on caller contracts and timeout behavior found one risk.

PR #3358, opened 2026-07-07
[codex] Make iOS simulator boot idempotent

kaeawc:
## Summary

- Make the CI iOS simulator boot script use `simctl bootstatus -b` instead of raw `simctl boot`.
- Wait out transient `Shutting Down` simulator state before invoking `bootstatus -b`.
- Harden AutoMobile's iOS `startDevice` cold-boot path by making `SimCtlClient.startSimulator()` use the same idempotent CoreSimulator primitive.
- Add regression coverage for already-booted and shutting-down simulators in the BATS CI script tests plus focused SimCtl unit coverage.

## Root Cause

`simctl boot <udid>` fails when CoreSimulator reports the target simulator is already `Booted`. That made CI fail even when the simulator was already in the desired state. Raw `boot` also leaves a race window after state inspection; `bootstatus -b` is the safer primitive because it boots if needed and waits until readiness.

## Validation

- `bun run bootstrap:worktree`
- `bun test test/utils/ios-cmdline-tools/simctl.test.ts`
- `bats test/bats/boot-simulator.bats`
- `shellcheck scripts/ios/boot-simulator.sh`
- `bun run build`
- `bun run lint`
- `git diff --check`
- `bun test --bail` (`7036 pass, 2 skip, 0 fail`)

### DIFF
--- a/scripts/ios/boot-simulator.sh
+++ b/scripts/ios/boot-simulator.sh
@@ -86,6 +86,42 @@ fi
 
 echo "Using runtime: ${RUNTIME_ID}" >&2
 
+get_device_state() {
+  xcrun simctl list devices -j \
+    | jq -r --arg udid "$1" '
+        .devices | to_entries[] | .value[]
+        | select(.udid == $udid)
+        | .state // empty
+      ' \
+    | head -1
+}
+
+wait_for_shutdown_if_needed() {
+  local udid="$1"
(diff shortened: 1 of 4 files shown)


kaeawc:
Closing as duplicate; the bootstatus-based fix has been folded into #3357.

Issue #2811, opened 2026-07-01
Flaky: RemindersAddPlanTests.testAddReminderPlan (UI 'Done' tap race at plan step 5)


kaeawc:
## Context

Surfaced during #2744 / PR #2749 CI. `RemindersAddPlanTests.testAddReminderPlan` (`ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersIntegrationTests.swift`) is **flaky** in the `XCTestRunner Simulator Tests` job.

It failed once on PR #2749 and **passed on a plain re-run** (no code change), and it fails intermittently locally with the same signature:

```
executeOnce ERROR: Plan execution failed at step 5 (tapOn):
  Error: Failed to perform tap on element: Element not found with provided text 'Done'
  Executed: 4/6 steps
```

The plan reaches the daemon and executes 4/6 steps (daemon + handshake are fine); the failure is a **UI-state race** — the Reminders "Done" button isn't present/hit at step 5. This is unrelated to the daemon handshake work; it's pre-existing UI-integration flakiness.

## Deferred work

Make the test deterministic, e.g.:
- Wait/observe for the "Done" (or Add) control before tapping instead of a bare `tapOn` at a fixed step.
- Reset Reminders app state in `setUp` so the add-flow starts from a known screen.
- Or quarantine/retry-wrap it if it can't be made reliable, so it stops red-flagging otherwise-green PRs.

## Acceptance
- [ ] `RemindersAddPlanTests.testAddReminderPlan` passes reliably across repeated CI runs (or is quarantined with a tracking note).

Refs #2749.

PR #3367, opened 2026-07-07
Fix Reminders add plan label variants

### LINKED ISSUE #2855: Reminders add-reminder plan is pinned to CI's iOS Reminders labels (Done/New Reminder, English)
Follow-up from #2811 / PR #2853.

`add-reminder.yaml` drives the **system Reminders app**, whose control labels and layout vary across iOS versions. During #2853 this was confirmed on live simulators:

- **iOS 18.6** — the new-reminder commit control is labelled **"Add"** (not "Done"); the accounts-home toolbar shows "New Reminder".
- **iOS 26.2 (with `clearAppData`)** — the accounts-home showed an **empty toolbar with no "New Reminder"**, and the commit control was not "Done" either.
- **CI (iOS 26, no `clearAppData`)** — per #2811 the plan reaches the "Done" tap, i.e. the commit control **is** "Done" on the CI image.

So PR #2853 deliberately pins the plan to CI's reality: mandatory `observe waitFor "New Reminder"` and `waitFor "Done"` guards, English UI. This de-flakes CI but is **not portable** across iOS versions/locales, and there is **no reliable post-save verification** (after saving, the accounts-home shows list summaries, not the created reminder's title), so a version-tolerant "tap Done or Add, then verify" plan can't cheaply prove the save happened.

### Work to consider
- [ ] Make the create/save taps tolerant of the known label variants ("Done"/"Add") with a **reliable** save verification (needs a version-stable signal — e.g. navigating into the list, or a resource-id rather than a localized label).
- [ ] Evaluate `addUIInterruptionMonitor` in the XCTest runner to auto-dismiss system alerts globally (the iCloud "Enable Syncing?" prompt), instead of the per-plan optional "Not Now" dismissal — noting AutoMobile drives gestures via its own daemon, so the monitor may not fire.
- [ ] Remove the `RemindersAddPlanTests` retry (added in #2853) once the guards are proven sufficient over N green CI runs.
- [ ] Pin the CI simulator iOS version explicitly if the plan must assume specific labels.

Refs #2811, #2853.

kaeawc:
## Summary

Adds ordered text-variant selectors for `observe.waitFor` and `tapOn`, then uses them to make the iOS Reminders add plan tolerate the observed `Done`/`Add`, `Not Now`/`Close`, and default-list startup variants. The add plan now uses a unique reminder title and verifies the saved reminder in deterministic plan-content coverage.

## Linked Issue

Closes #2855

## Bug / Regression Context

- **Prior attempts:** #2811 and #2853 made the Reminders launch/add flows less flaky, but the add plan was still pinned to specific CI Reminders labels/state.
- **Observed symptom:** The add flow could fail when iOS exposed `Done` instead of `Add`, showed an iCloud/location prompt, or launched on the accounts home/restored-list screen.
- **Repro steps / conditions:** Run `Plans/add-reminder.yaml` on iOS simulator states where Reminders labels or launch state differ from the original CI baseline.

## Validation

- [x] `bun run typecheck`
- [x] `bun run validate:yaml`
- [x] `cd ios/XCTestRunner && swift test --filter RemindersPlanContentTests`
- [x] `bun test test/server/observeWaitForSchema.test.ts test/server/tools/tapOnTapAnySchema.test.ts test/features/action/TapOnElement.extendedSelectors.test.ts test/utils/plan/PlanMigrator.test.ts`
- [x] `git diff --check`
- [x] New code uses interfaces/fakes; focused unit tests run in milliseconds.
- [x] Rebased onto latest `origin/main`, conflicts resolved.

## Device Verification

- [x] Verified iOS simulator hierarchy variants via `observe`; local live add-plan execution reached a simulator state where Reminders exposed no accessible `New Reminder` node, so deterministic coverage remains in plan-content tests.
- [x] PR CI confirmed the live add-plan gate is not yet reliable: the required `XCTestRunner Simulator Tests` job timed out waiting for `New Reminder` after the default-list row selector missed the current iOS 26.2 accounts-home state. This PR therefore no longer adds the live `RemindersAddPlanTests` workflow gate until a follow-up can seed or identify a stable list entry.

## Follow-ups

- [ ] Add a reliable live Reminders add-plan CI gate after the plan can seed or identify a usable default list on current CI iOS images.

### DIFF
--- a/src/server/observeTools.ts
+++ b/src/server/observeTools.ts
@@ -34,17 +34,17 @@ const waitForSchema = z.union([
     elementId: z.string().describe("Element resource ID / accessibility identifier"),
     timeout: z.number().optional().describe("Wait timeout ms (default: 5000)"),
     container: waitForContainerField
-  }),
+  }).strict(),
   z.object({
     text: z.string().describe("Element text"),
     timeout: z.number().optional().describe("Wait timeout ms (default: 5000)"),
     container: waitForContainerField
-  }),
+  }).strict(),
   z.object({
     textAny: z.array(z.string().min(1)).min(1).describe("Ordered text variants; first visible match wins"),
     timeout: z.number().optional().describe("Wait timeout ms (default: 5000)"),
     container: waitForContainerField
-  })
+  }).strict()
 ]);
 
 export const observeSchema = addDeviceTargetingToSchema(z.object({
--- a/src/server/observeTools.ts
+++ b/src/server/observeTools.ts
@@ -86,7 +86,21 @@ const waitForContainerForFinder = (
     : { text: waitFor.container.text };
 };
 
-const findWaitForElement = (
+const isElementCenterOffScreen = (
+  element: Element,
+  viewHierarchy: ViewHierarchyResult
+): boolean => {
+  if (!viewHierarchy.screenWidth || !viewHierarchy.screenHeight || !element.bounds) {
+    return false;
+  }
+
+  const centerX = (element.bounds.left + element.bounds.right) / 2;
+  const centerY = (element.bounds.top + element.bounds.bottom) / 2;
+  return centerX < 0 || centerX > viewHierarchy.screenWidth ||
+    centerY < 0 || centerY > viewHierarchy.screenHeight;
+};
+
+export const findWaitForElement = (
   finder: ElementFinder,
   waitFor: ObserveWaitForOptions,
   viewHierarchy: ViewHierarchyResult
@@ -121,6 +135,9 @@ const findWaitForElement = (
         false
       );
       if (element) {
+        if (isElementCenterOffScreen(element, viewHierarchy)) {
+          continue;
+        }
         return element;
       }
     }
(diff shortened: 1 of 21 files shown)


src/server/observeTools.ts
```diff
@@ -39,6 +39,11 @@ const waitForSchema = z.union([
     text: z.string().describe("Element text"),
     timeout: z.number().optional().describe("Wait timeout ms (default: 5000)"),
     container: waitForContainerField
+  }),
+  z.object({
```
kaeawc, line 43: Confirmed bug: the `waitForSchema` union branches are not strict, so mixed selectors are accepted and Zod silently strips the later selector. I reproduced this locally with `observeSchema.safeParse({ platform: "ios", waitFor: { text: "Add", textAny: ["Done"] } })`, which succeeds as `{ text: "Add" }`; `{ elementId, textAny }` similarly succeeds as `{ elementId }`. That creates a plan-validation false-negative exactly on the new `textAny` surface. Please make these `waitFor` branch objects `.strict()` (matching `tapOnSelectorSchema` and `createElementIdTextSelectorSchema`) and add rejection tests for mixed `text`/`elementId`/`textAny` selectors.

kaeawc (commented):
Review pass: compatibility/schema false-negative lens.

PR #3378, opened 2026-07-07
feat(ios): generate XCTestRunner version constant from package.json (#2814)

### LINKED ISSUE #2814: iOS: bake a version constant into the XCTestRunner SPM package
## Summary
The iOS XCTestRunner SPM package has **no version constant** — it can neither self-identify a version nor pin/verify a daemon version at the handshake. This mirrors the Android jar's `Implementation-Version` (stamped from `package.json`), which the iOS side lacks.

Deferred from #2746 / #2748 (native-side, not unit-testable in the TS harness).

## Evidence
`ios/XCTestRunner/Package.swift` has no version field; the package is pinned by git tag only. There is no analogue to the Android jar manifest `Implementation-Version`.

## Proposed direction
Bake a generated version constant into the XCTestRunner SPM package (source-generated at build time from the canonical `package.json` version, matching how Android stamps its jar). Wire it into the runner so it can:
- self-identify its version in logs / the `connected` handshake, and
- participate in a version-equality check against the daemon (unblocks the client-handshake work).

## Acceptance
- The built XCTestRunner reports a concrete version that matches the release it was cut from.
- The value is generated (not hand-edited) so it can't drift from `package.json`.

Refs #2746, #2732. Follow-up to #2748. Companion to the iOS Swift daemon-pin follow-up.

kaeawc:
## Summary

The XCTestRunner's baked `AutoMobileVersion.current` was kept in sync by a regex edit in `scripts/versioning/bump-versions.sh`. That mechanism uses `replace_optional_single_match`, which is a **no-op on zero matches** — so a refactor of the Swift file could silently leave the constant stale and drift from `package.json`. This makes `package.json` the single source of truth for the iOS runner version and adds a drift gate, mirroring how the Android jar stamps `Implementation-Version` from `package.json`.

- Add `scripts/versioning/generate-ios-version.sh` — renders `AutoMobileVersion.swift` from `package.json`, with a `--check` mode that fails on drift.
- `AutoMobileVersion.swift` is now a generated artifact (regenerated in place; header marks it DO NOT EDIT). The value (`0.0.40`) and public API are unchanged.
- `bump-versions.sh` regenerates the constant via the generator instead of regex-editing it.
- Wire the `--check` drift gate into `scripts/ci/validate-ios-swift.sh` so stale constants fail PR CI.

## Linked Issue

Closes #2814

## Acceptance Criteria

- [x] "The built XCTestRunner reports a concrete version that matches the release it was cut from." — pinned by Swift `AutoMobileVersionTests.testCurrentVersionMatchesCanonicalPackageJson` (reads canonical `package.json` via `#filePath`, asserts equality with `AutoMobileVersion.current`).
- [x] "The value is generated (not hand-edited) so it can't drift from `package.json`." — pinned by `test/bats/generate-ios-version.bats` (generation, generated marker, `--check` pass in-sync / fail on drift) + the CI drift gate.

## Validation

- [x] `bats test/bats/generate-ios-version.bats test/bats/bump-versions.bats` — green
- [x] `shellcheck` on all three touched scripts — clean
- [x] `cd ios/XCTestRunner && swift test -Xswiftc -warnings-as-errors --filter AutoMobileVersionTests` — green
- [x] New Swift/BATS tests run <100ms
- Note: an unrelated `RemindersAddPlanTests` integration test (22s, requires simulator Reminders state) fails independently of this change.

## Follow-ups

- [ ] None required; the release-integrity gate (`verify-release-integrity.sh`) already checks `package.json` at release time and could optionally also assert the generated Swift constant — captured as a possible follow-up.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -598,6 +598,13 @@ jobs:
         shell: bash
         run: bats test/bats/
 
+      # Drift gate for the generated iOS runner version constant (#2814). Pure
+      # bash + python3, so it runs on this Ubuntu job rather than the macOS Swift
+      # lane — the committed AutoMobileVersion.swift must match package.json.
+      - name: "Check iOS version constant is in sync with package.json"
+        shell: bash
+        run: bash scripts/versioning/generate-ios-version.sh --check
+
       - name: "Run hadolint"
         uses: hadolint/hadolint-action@v3.1.0
         with:
--- a/ios/XCTestRunner/Sources/XCTestRunner/AutoMobileVersion.swift
+++ b/ios/XCTestRunner/Sources/XCTestRunner/AutoMobileVersion.swift
@@ -1,3 +1,9 @@
+// GENERATED FILE — DO NOT EDIT.
+//
+// Rendered from package.json by scripts/versioning/generate-ios-version.sh.
+// To change the version, bump package.json (scripts/versioning/bump-versions.sh
+// does this and regenerates), then re-run the generator. CI runs
+// `generate-ios-version.sh --check` and fails on drift.
 import Foundation
 
 /// Baked client version the XCTestRunner declares to the AutoMobile daemon.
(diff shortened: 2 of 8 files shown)


kaeawc:
## Multi-lens review triage (3 orthogonal review agents)

**Regression lens · Test-adequacy lens · Interface/contract lens.**

### Fixed
- **HIGH — drift gate was dead in CI.** The `--check` gate lived only in `scripts/ci/validate-ios-swift.sh`, which no `.github/workflows/` file invokes (and which early-exits off macOS). Drift would not have failed any PR — defeating AC2's enforcement. Wired `generate-ios-version.sh --check` into the Ubuntu fast-validation job (pure bash + python3) so it runs on every PR. This also resolves the MEDIUM "gate behind macOS+swift early-exit" finding (same root cause).
- **LOW — test could not catch a regression to in-place regex editing.** `bump-versions.bats` now asserts the generated markers (`GENERATED FILE — DO NOT EDIT`, `public enum AutoMobileVersion`), so reverting `bump-versions.sh` to a regex edit fails the test.
- **LOW — untested generator failure branches.** Added bats cases: missing `package.json`, no `version` field, invalid JSON, non-semver.
- **LOW — malformed JSON → raw Python traceback.** Generator now emits a clean `ERROR: package.json is not valid JSON` line.
- **LOW (hardening) — partial-write on bad `--new-version`.** `bump-versions.sh` now validates the version shape before any file is written (the generator's late semver reject would otherwise abort mid-bump with a half-updated tree); pinned by a new bats case.

### Confirmed adequate (no change)
- AC1 pinned by Swift `testCurrentVersionMatchesCanonicalPackageJson` (both lenses verified the `#filePath`→repo-root resolution and proved the red state by mutation); AC2 pinned by the bats drift tests + the (now live) CI gate.
- API surface (`AutoMobileVersion.current`) unchanged; handshake consumers untouched. swiftformat/swiftlint clean; convention-consistent with `generate-release-constants.sh`.

### Not addressed (out of scope, no follow-up filed)
- Optional: `verify-release-integrity.sh` could additionally assert the generated Swift constant at release time. The new PR-CI drift gate already covers this on every PR, so it's redundant belt-and-suspenders — noted here rather than filed.

PR #3484, opened 2026-07-09
Fix hermetic CtrlProxy pin hardening

### LINKED ISSUE #2815: Supply-chain hardening for hermetic pinning: acceptedPreinstalled mismatch + mirror URL validation
## Summary
Two supply-chain edges in the CtrlProxy download/verify path were surfaced during the #2748 review but intentionally left unchanged (pre-existing behavior / out of the documented contract). Filing so they aren't lost.

## 1. `acceptedPreinstalled` accepts a known-SHA-mismatched APK non-blocking
`src/utils/CtrlProxyManager.ts` (`ensureCompatibleVersion`, ~`acceptedPreinstalled` path): when the installed APK's SHA differs from a **known** expected checksum, the daemon accepts the preinstalled CtrlProxy non-blocking (`status: "skipped", acceptedPreinstalled: true`) and queues a background refresh, rather than rejecting it. #2748 closed the *unknown-pin* hole (fail-closed) and documented this path as a trust boundary, but a strict/hermetic mode could fail closed here too.

- Consider: a strict flag (or reuse the hermetic pin signal) that rejects a known-SHA mismatch instead of accepting it non-blocking, for CI that must guarantee the exact release.

## 2. `AUTOMOBILE_ASSET_BASE_URL` with a query string yields a malformed URL
`src/constants/release.ts` (`resolveAssetBaseUrl` → `buildReleaseAssetUrl`): a base URL containing a query string (e.g. `https://mirror/am?token=abc`) composes to `https://mirror/am?token=abc/<version>/<file>`, which is malformed. Out of the documented mirror contract (path-only base), but worth validating/normalizing or documenting explicitly.

- Consider: validate the base URL shape (reject/normalize query strings + fragments) or document the path-only contract in `docs/using/hermetic-ci-pinning.md`.

## Acceptance
- (1) A hermetic/strict mode fails closed on a known-SHA mismatch (or the non-blocking behavior is explicitly confirmed as intended and documented).
- (2) A malformed `AUTOMOBILE_ASSET_BASE_URL` is rejected with an actionable error, or the contract is documented.

Refs #2746. Follow-up to #2748.

kaeawc:
## Summary

Harden the hermetic pinning trust boundary by failing closed when a concrete `AUTOMOBILE_VERSION` points to a registry-known CtrlProxy APK but the already-installed APK has a different SHA. Also reject `AUTOMOBILE_ASSET_BASE_URL` values with query strings or fragments before they compose malformed asset URLs.

## Linked Issue

Closes #2815

## Bug / Regression Context

- **Prior attempts:** #2748 added daemon-side hermetic pinning and fail-closed behavior for unknown pins, but intentionally left these two pre-existing edges as follow-ups.
- **Observed symptom:** A known checksum mismatch on a preinstalled Android CtrlProxy APK could still be accepted non-blocking, and mirror bases like `https://mirror/am?token=abc` composed to malformed asset paths.
- **Repro steps / conditions:** Set a concrete known `AUTOMOBILE_VERSION` with a mismatched installed APK SHA, or set `AUTOMOBILE_ASSET_BASE_URL` to a value containing `?` or `#`.

## Validation

- [x] `bun test test/constants/release.test.ts test/utils/CtrlProxyManager.test.ts`
- [x] New code uses existing fakes + `FakeTimer`; unit tests run in <100ms

## Follow-ups

None.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/constants/release.ts
+++ b/src/constants/release.ts
@@ -305,7 +305,7 @@ export function resolveAssetBaseUrl(env: EnvLike = process.env): string {
         `use a path-only base URL such as ${parsed.origin}${parsed.pathname.replace(/\/+$/, "")}.`
       );
     }
-    return trimmed.replace(/\/+$/, "");
+    return `${parsed.origin}${parsed.pathname}`.replace(/\/+$/, "");
   }
   return DEFAULT_ASSET_BASE_URL;
 }
--- a/src/constants/release.ts
+++ b/src/constants/release.ts
@@ -12,6 +12,8 @@
  * scripts/generate-release-constants.sh
  */
 
+import { ActionableError } from "../models/ActionableError";
+
 export const LATEST_RELEASE_VERSION = "latest";
 
 export interface ReleaseChecksumEntry {
@@ -289,6 +291,20 @@ export function isPinnedVersionKnown(
 export function resolveAssetBaseUrl(env: EnvLike = process.env): string {
   const trimmed = env[AUTOMOBILE_ASSET_BASE_URL_ENV]?.trim();
   if (trimmed && trimmed.length > 0) {
+    let parsed: URL;
+    try {
+      parsed = new URL(trimmed);
+    } catch {
+      throw new ActionableError(
+        `${AUTOMOBILE_ASSET_BASE_URL_ENV} must be an absolute URL, for example https://mirror.example/auto-mobile.`
+      );
+    }
+    if (parsed.search || parsed.hash) {
+      throw new ActionableError(
+        `${AUTOMOBILE_ASSET_BASE_URL_ENV} must not include a query string or fragment; ` +
+        `use a path-only base URL such as ${parsed.origin}${parsed.pathname.replace(/\/+$/, "")}.`
+      );
+    }
     return trimmed.replace(/\/+$/, "");
   }
   return DEFAULT_ASSET_BASE_URL;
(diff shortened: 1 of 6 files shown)


kaeawc:
Codex review triage: confirmed and fixed the strict-pin readiness shortcut gap.

- Completed-prefetch fallback fixed in `4a50e2e8`.
- `AUTOMOBILE_SKIP_ACCESSIBILITY_DOWNLOAD_IF_INSTALLED` now yields to known explicit pins in `a8a74391`, so the installed APK SHA is still read and mismatches fail closed.
- Added regression coverage for both shortcut paths.

Validation after the latest fix:
- `bun test test/doctor/automobile.test.ts test/constants/release.test.ts test/utils/CtrlProxyManager.test.ts`
- `bun run turbo:validate`
- `bun run typecheck`

kaeawc:
Follow-up review triage: confirmed and fixed the remaining doctor/URL contract issues found by the parallel review passes.

- `AUTOMOBILE_ASSET_BASE_URL` now returns the parser-normalized path-only base before composing APK/IPA URLs.
- `checkCtrlProxy()` validates asset URL config before no-device skips, so malformed mirror config fails doctor even on hosts without Android devices.
- Explicit `AUTOMOBILE_VERSION` provisioning failures (`versionStatus=failed`, including checksum verification failures from a bad mirror/local asset) now return doctor `fail`, preserving the CI gate.

Validation after this update:
- `bun test test/constants/release.test.ts test/doctor/automobile.test.ts test/utils/CtrlProxyManager.test.ts`
- `bun run turbo:validate`
- `bun run typecheck`

PR #3521, opened 2026-07-09
Skip same-window occlusion so overlapping Compose elements are not stripped

jstewart5000:
## Summary
Skips within-window occlusion filtering so overlapping Compose layouts (e.g. Slack's channel header) are no longer falsely marked `hidden` and stripped from the processed view hierarchy.

### Why
- `optimizeHierarchy` promotes children of bounds-only wrapper nodes, flattening the tree. After that restructuring, path-based relationship detection in `determineNodeRelationship` no longer reflects visual relationships: visual siblings can land at different tree depths.
- An interactive element overlaid on a larger content area in a `Box`/`Scaffold` (the Slack `channel_header_toolbar` over the messages fragment) gets classified `UNRELATED` to the content that covers it, exceeds the 95% coverage threshold, and is dropped — so the header is absent from `observe` output and untappable. The pattern is pervasive across Compose, not specific to the channel header.

### What Changed
- **`buildOcclusionInfo`**: same-window occluder pairs now unconditionally `continue` (skip) instead of consulting `determineNodeRelationship`. Only cross-window occlusion (system bars, IME, dialogs in separate windows) still filters. This also makes the multi-window path consistent with the existing `windowEntries.size == 1` guard that already skipped within-window occlusion.
- **`extractFromActiveWindow`**: no longer calls `applyOcclusionFilteringSingleWindow`; that now-unused private function was removed.
- **`determineNodeRelationship`** (defense-in-depth, no production callers after the above): added the reverse "nephew" check and a root-level rule (empty parent path ⇒ related). Kept semantically correct for a future direction that runs occlusion on the pre-optimization tree. Documented that it is currently test-only.

### Test Plan
- `ViewHierarchyExtractorTest` — **57 tests pass** (`:control-proxy:testDebugUnitTest`).
- Added negative tests that fail without the fix (verified by temporarily reverting the same-window skip → 2 failures):
  - `same-window occlusion skip rescues an asymmetric-depth cousin that fix 3 cannot` — the real channel-header shape (shallow header vs deeply-nested content in a different branch); only the full skip retains it.
  - `determineNodeRelationship leaves mismatched-depth cousins UNRELATED - justifies full skip` — characterizes why the relationship patch alone is insufficient.
- Corrected the existing relationship unit test that used a backwards path example / misleading name.
- Verified end-to-end on `emulator-5554` against Slack (debug CtrlProxy APK): `observe` now returns `channel_header_toolbar` (`clickable=true`); tapping it opens the channel menu, and "Settings & details" navigates to `ChannelDetailsActivity`.


kaeawc (approved):
Great change, thanks!

PR #3523, opened 2026-07-09
tapOn: target a specific control among repeated rows (nested-label sibling + instance)

### LINKED ISSUE #3522: tapOn --sibling can't reach a row's action control when the row's label is nested
# `tapOn --sibling` can't reach a row's action control when the row's label is nested

## Summary

`tapOn` with `sibling: true` (via `selectClickableSiblingOfText`) only finds a clickable element that is a **direct sibling of the text node**. Many list rows don't lay out that way: the label is nested one level down (e.g. a name/subtitle pair inside a leading avatar/content container) while the row's action control (a remove button, a row checkbox) sits as a **sibling of that container** — a "cousin" of the text, not a sibling. On those rows `--sibling` fails with `No clickable sibling found`, and there is no other reliable way to target that specific row's control:

- the control's `resource-id` is shared across rows, so an id tap resolves to the **first** row;
- `--container` scoped by the label text resolves to the label's narrow subtree, which **excludes** the control;
- there is no coordinate tap or nth-match selector.

Net: you cannot reliably tap "the action button in the row labeled X".

## Repro (generic layout)

A vertical list where each row nests its label under a leading container and puts the per-row action as a sibling of that container:

```
row (clickable)
├── avatar
│   ├── nameLabel   "Alice Adams"     ← label nested under the leading container
│   └── subtitle
└── removeButton    (clickable)        ← sibling of the AVATAR, not of the label
```

With two rows (Alice Adams, Bob Brown), each with an identically-id'd `removeButton`:

- `tapOn --selector '{"text":"Bob Brown"}' --sibling true` → `ERROR: No clickable sibling found next to element with text 'Bob Brown'`
- `tapOn --selector '{"elementId":".../removeButton"}' --container '{"text":"Bob Brown"}'` → `ERROR: Element not found ... within container text 'Bob Brown'`
- `tapOn --selector '{"elementId":".../removeButton"}'` → hits the **first** row (shared id), not Bob's.

## Root cause

`ElementFinder.findClickableSiblingsInNode` decides a level is a "row" using `nodeHasText` (shallow — the direct child's own text). When the label is a grandchild (nested under the avatar), no direct child of the row bears the text, so the row's clickable children are never collected. The shallow check was deliberate — to stop a list/recycler ancestor collecting every row's control — but it also blocks the nested-label case entirely.

## Fix

Match on a child whose **subtree** contains the text (deep), but collect the clickable, non-text children at the **deepest** qualifying node. Recurse first and bail if a descendant matched, so the nearest (row) level wins. To keep a list ancestor from collecting sibling rows' controls, only collect at a node whose text-bearing child is NOT itself clickable (a clickable text-bearing child means that child is the row and this node is the list); a matched row with no control of its own then resolves to a clean null instead of a wrong tap. The direct-sibling case is preserved.

The branch also adds a complementary `instance` field to `tapOn` (a 0-based "tap the Nth on-screen match" selector) for repeated controls that have no unique text to anchor on — the general escape hatch when `--sibling` doesn't apply.

## Impact

Any list with a nested label and a trailing per-row action (member/attendee lists, contact rows, settings rows, etc.). Without this, an automation either can't act on a specific row or silently hits the wrong (first) row.

### LINKED ISSUE #3524: tapOn: no way to target the Nth of several matching elements (repeated controls with no unique text)
# tapOn: no way to target the Nth of several matching elements

## Summary

When a selector matches multiple elements, `tapOn` can only choose one via `selectionStrategy`, which is `first` or `random`. There is no way to deterministically target the **2nd / 3rd / Nth** match.

This is a gap for **repeated controls that have no unique text to anchor on** — e.g. a list where every row has the same icon-only action button (a remove "×", a checkbox) sharing one `resource-id`, and the rows carry no per-row text you could match. `selectClickableSiblingOfText` covers the case where the row *does* have a unique label to anchor on, but when it doesn't, there's no reliable way to hit a specific one — an id/`text` match resolves to the first, and `random` isn't deterministic.

## Proposal

Add an `instance` field to `tapOn`: a 0-based index that taps the Nth on-screen match (in hierarchy order — top-to-bottom for a vertical list) instead of applying `selectionStrategy`. Out of range returns no match (rather than silently hitting another element).

```
tapOn --selector '{"elementId":"app:id/removeButton"}' --instance 1   # the 2nd on-screen match
```

Indexes into the on-screen (visible) matches so it lines up with what a user sees. Works alongside the existing selectors, including `sibling`.

## Notes

Implemented in PR #3523 (bundled with a related fix for reaching a row's control when the row's label is nested). Filing this so the capability has its own tracking issue.

RyeBread1026:
## What

Two related improvements for reliably tapping the right control when a list has many repeated controls (a remove button / checkbox per row):

1. **`selectClickableSiblingOfText` now reaches a row's control when the label is nested.** Common list rows nest the label under a leading container (e.g. a name/subtitle pair inside an avatar/content group) while the per-row action sits as a *sibling of that container* — a "cousin" of the text, not a direct sibling. Previously `tapOn --sibling` returned `No clickable sibling found` on such rows, and there was no other reliable way to target that row's control (the control's `resource-id` is shared across rows → an id tap hits the first row; `--container` by the label scopes to the label's narrow subtree and excludes the control).

2. **New `instance` field on `tapOn`.** When a selector matches multiple elements, tap the Nth on-screen match (0-based, in hierarchy order) instead of applying `selectionStrategy` — the general way to target a repeated control that has no unique text to anchor on. Out of range returns no match.

Fixes #3522 (reaching a nested-label row's control). Closes #3524 (the `instance` selector).

## Implementation

- `ElementFinder.findClickableSiblingsInNode`: match on a child whose **subtree** contains the text (deep), and collect the clickable, non-text children at the **deepest** qualifying node. To keep a list/recycler ancestor from collecting a *sibling row's* control, only collect at a node whose text-bearing child is **not itself clickable** (a clickable text-bearing child means that child is the row and this node is the list). A matched row with no control of its own therefore resolves to a clean `null` rather than silently hitting another row's control. The prior shallow (direct-text-child) behavior is preserved.
- `instance`: added to `tapOnSchema` + `TapOnElementOptions`/`TapOnArgs`, threaded through `selectByText`/`selectByResourceId` and the sibling selectors into `pickMatch` (0-based index into the on-screen matches; overrides `selectionStrategy`; out of range → no match).

## Tests

- New unit coverage: nested-label rows resolve **each row's own** control; a control-less matched row returns `null` (regression for the wrong-tap case); the direct-sibling case still works; `instance` selects the Nth match and is out-of-range safe.
- Full selector/finder/tapOn/interaction suite green (351 tests, 0 fail).
- `typecheck` gate clean (no new errors), `eslint` clean.
- Documented `instance` in `docs/design-docs/mcp/tools.md`.

## Notes

- Empirically validated on a real device: `--sibling` removes the named row's control (not the first row's), and `--instance N` removes the Nth.
- Open question for reviewers: the external field is named `instance`; happy to rename to `index` (matching the internal param / Maestro's equivalent) if that's preferred.

### DIFF
--- a/docs/design-docs/mcp/tools.md
+++ b/docs/design-docs/mcp/tools.md
@@ -8,7 +8,7 @@ Almost all other tool calls have built-in observation via the [interaction loop]
 
 #### Interactions
 
-- 👆 `tapOn` supports tap, double-tap, long press, and long-press drag actions. Selector strategies include `text`, `elementId`, `clickable` (first clickable element), `siblingOfText` (clickable sibling of a text element), and `tapClickableParent` (nearest clickable ancestor of a text match).
+- 👆 `tapOn` supports tap, double-tap, long press, and long-press drag actions. Selector strategies include `text`, `elementId`, `clickable` (first clickable element), `siblingOfText` (clickable sibling of a text element), and `tapClickableParent` (nearest clickable ancestor of a text match). When multiple elements match, `instance` (0-based) taps the Nth on-screen match instead of applying `selectionStrategy`.
 - 👉 `swipeOn` handles directional swipes and scrolling within container bounds.
 - ↔️ `dragAndDrop` for element-to-element moves.
 - 🔍 `pinchOn` for zoom in/out gestures.
--- a/docs/design-docs/mcp/tools.md
+++ b/docs/design-docs/mcp/tools.md
@@ -8,7 +8,7 @@ Almost all other tool calls have built-in observation via the [interaction loop]
 
 #### Interactions
 
-- 👆 `tapOn` supports tap, double-tap, long press, and long-press drag actions. Selector strategies include `text`, `elementId`, `clickable` (first clickable element), `siblingOfText` (clickable sibling of a text element), and `tapClickableParent` (nearest clickable ancestor of a text match). When multiple elements match, `instance` (0-based) taps the Nth on-screen match instead of applying `selectionStrategy`.
+- 👆 `tapOn` supports tap, double-tap, long press, and long-press drag actions. Selectors include `selector.text`, `selector.textAny`, and `selector.elementId`; `sibling: true` taps a clickable sibling of the selector match. When multiple elements match, `index` (0-based) taps the Nth on-screen match instead of applying `selectionStrategy`.
 - 👉 `swipeOn` handles directional swipes and scrolling within container bounds.
 - ↔️ `dragAndDrop` for element-to-element moves.
 - 🔍 `pinchOn` for zoom in/out gestures.
--- a/src/features/utility/DefaultElementSelector.ts
+++ b/src/features/utility/DefaultElementSelector.ts
@@ -35,7 +35,8 @@ export class DefaultElementSelector implements ElementSelector {
       text,
       options?.container ?? null,
       options?.partialMatch ?? true,
-      options?.caseSensitive ?? false
+      options?.caseSensitive ?? false,
+      options?.index !== undefined
     );
     return this.pickMatch(matches, strategy, viewHierarchy, options?.index);
   }
@@ -55,7 +56,8 @@ export class DefaultElementSelector implements ElementSelector {
       viewHierarchy,
       resourceId,
       options?.container ?? null,
-      options?.partialMatch ?? false
+      options?.partialMatch ?? false,
+      options?.index !== undefined
     );
     return this.pickMatch(matches, strategy, viewHierarchy, options?.index);
   }
--- a/src/features/utility/DefaultElementSelector.ts
+++ b/src/features/utility/DefaultElementSelector.ts
@@ -26,6 +26,7 @@ export class DefaultElementSelector implements ElementSelector {
       partialMatch?: boolean;
       caseSensitive?: boolean;
       strategy?: ElementSelectionStrategy;
+      index?: number;
     }
   ): ElementSelectionResult {
     const strategy = options?.strategy ?? "first";
@@ -36,7 +37,7 @@ export class DefaultElementSelector implements ElementSelector {
       options?.partialMatch ?? true,
       options?.caseSensitive ?? false
     );
-    return this.pickMatch(matches, strategy, viewHierarchy);
+    return this.pickMatch(matches, strategy, viewHierarchy, options?.index);
   }
 
   selectByResourceId(
@@ -46,6 +47,7 @@ export class DefaultElementSelector implements ElementSelector {
       container?: { elementId?: string; text?: string } | null;
       partialMatch?: boolean;
       strategy?: ElementSelectionStrategy;
+      index?: number;
     }
   ): ElementSelectionResult {
     const strategy = options?.strategy ?? "first";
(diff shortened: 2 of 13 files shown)


docs/design-docs/mcp/tools.md
```diff
@@ -8,7 +8,7 @@ Almost all other tool calls have built-in observation via the [interaction loop]
 
 #### Interactions
 
-- 👆 `tapOn` supports tap, double-tap, long press, and long-press drag actions. Selector strategies include `text`, `elementId`, `clickable` (first clickable element), `siblingOfText` (clickable sibling of a text element), and `tapClickableParent` (nearest clickable ancestor of a text match).
+- 👆 `tapOn` supports tap, double-tap, long press, and long-press drag actions. Selector strategies include `text`, `elementId`, `clickable` (first clickable element), `siblingOfText` (clickable sibling of a text element), and `tapClickableParent` (nearest clickable ancestor of a text match). When multiple elements match, `instance` (0-based) taps the Nth on-screen match instead of applying `selectionStrategy`.
```
kaeawc: Yeah I think `index` makes more sense than `instance`

kaeawc (approved):
Change looks good, i'll merge this and then there can be a follow up to rename the field to `index`

src/features/utility/DefaultElementSelector.ts
```diff
@@ -150,21 +154,33 @@ export class DefaultElementSelector implements ElementSelector {
...
       .filter(match => !this.isElementCenterOffScreen(match.element, viewHierarchy));
 
     if (visibleMatches.length === 0) {
       return { element: null, indexInMatches: -1, totalMatches, strategy };
     }
 
+    // Explicit 0-based index overrides strategy: pick the Nth on-screen match, or return
+    // no match if out of range (so a caller asking for "the 3rd" of 2 fails, not silently
+    // grabbing another element).
+    if (index !== undefined) {
```
kaeawc, line 178: Verdict: Confirmed bug. instance is documented as selecting the Nth on-screen match in hierarchy/top-to-bottom order, but this indexes the already-sorted matches array. findElementsByText and findElementsByResourceId sort matches by area before this point (ElementFinder.ts sorts text matches at lines 201-206 and resource-id matches at lines 236-238), so the new index is area order, not hierarchy order. I reproduced it locally with two app:id/remove controls where the top one was larger: instance: 0 selected the lower smaller control and instance: 1 selected the top control.

Fix by preserving traversal order for explicit instance selection, or by adding a separate hierarchy-ordered finder path for this option; add a regression test with differently sized repeated controls so equal-area rows do not mask it.

src/server/interactionTools.ts
```diff
@@ -149,6 +149,11 @@ export const tapOnSchema = addDeviceTargetingToSchema(z.object({
   selectionStrategy: elementSelectionStrategySchema.optional().describe(
     "Selection strategy when multiple match (default: first)"
   ),
+  instance: z.number().int().nonnegative().optional().describe(
```
kaeawc: Verdict: Partial. Runtime accepts the new tapOn.instance field here, but the generated schemas/tool-definitions.json was not updated. rg for instance in schemas/tool-definitions.json returns no matches, so generated consumers/IDE completion will not expose the new field even though the server schema does. Fix by regenerating schemas/tool-definitions.json with the repo generator (bun scripts/generate-tool-definitions.ts or scripts/update-tool-definitions.sh) and committing the result.

kaeawc (commented):
Code review found two actionable issues: instance currently indexes area-sorted matches rather than hierarchy/top-to-bottom order, and the generated tool definitions were not updated for the new tapOn.instance field.

PR #3655, opened 2026-07-11
fix(swipeOn): clamp container swipe bounds to safe inset region (#3587)

### LINKED ISSUE #3587: swipeOn system-inset math over-subtracts on right/bottom edges (under-scroll, inverted bounds)
Found by a `/find-bugs` sweep of `src/`. Confidence: **high** (verified against the canonical inset helper).

## Symptom
When `insets.right`/`insets.bottom` are nonzero (a gesture/nav bar is almost always present) and the swipe container does **not** span the full screen, the computed swipe region is shortened/shifted toward the top-left by the full inset amount — even when the container is nowhere near the inset region.

- List at `top=100, bottom=400` on a 2000px screen with `insets.bottom=126` → `effective.bottom = min(400,2000) - 126 = 274`, losing 126px and under-scrolling.
- Extreme case (large right inset + narrow left-positioned container, e.g. `left=300, right=350, insets.right=200`) → computed `right` (150) drops below `left` (300): **inverted bounds**, and `getSwipeWithinBounds` then produces a swipe with negative width/height.

## Root cause
Left/top edges correctly clamp with `Math.max(edge, inset)`, but right/bottom subtract the inset unconditionally:

```ts
right:  Math.min(container.right,  screenW) - insets.right,   // wrong
bottom: Math.min(container.bottom, screenH) - insets.bottom,  // wrong
```

instead of clamping to the safe boundary:

```ts
right:  Math.min(container.right,  screenW - insets.right),
bottom: Math.min(container.bottom, screenH - insets.bottom),
```

The canonical helper `src/utils/screenBounds.ts:21-22` (`getScreenBounds`) shows the intended form (`right: screenSize.width - insets.right`). The default (no-overlay) path returns these bounds straight to `getSwipeWithinBounds` with no later clamping.

## Locations (fix both — identical buggy expression)
- `src/features/action/swipeon/resolveContainerSwipeCoordinates.ts:20-21`
- `src/features/action/swipeon/ScrollUntilVisible.ts:191-194`

## Fix direction
Clamp right/bottom to `screenW - insets.right` / `screenH - insets.bottom` inside the `Math.min`, mirroring the left/top `Math.max` form and `getScreenBounds`.

## Test strategy
Unit-test `resolveContainerSwipeCoordinates` with a partial-height container and nonzero bottom/right insets; assert `effective.bottom == min(container.bottom, screenH - insets.bottom)` and that `right > left`, `bottom > top` in the extreme narrow-container case. Add a regression for the previously-inverted-bounds input.

## Related
- `src/utils/screenBounds.ts:18-23` (canonical form)
- `src/features/utility/ElementGeometry.ts` (`getSwipeWithinBounds` consumer)

kaeawc:
Fixes #3587.

## Problem
The effective swipe-bounds computation subtracted the system inset unconditionally on the **right/bottom** edges:

```ts
right:  Math.min(container.right,  screenW) - insets.right,   // wrong
bottom: Math.min(container.bottom, screenH) - insets.bottom,  // wrong
```

while the **left/top** edges correctly clamped with `Math.max(edge, inset)`. For any container that doesn't span the full screen and sits away from the inset region, this shortened the swipe by the full gesture-bar inset (under-scroll). In an extreme case (narrow left-positioned container + large right inset) the computed `right` dropped below `left` → **inverted bounds → negative-size swipe**.

## Fix
Clamp right/bottom to the safe boundary, mirroring the left/top form and the canonical `getScreenBounds` (`src/utils/screenBounds.ts`):

```ts
right:  Math.min(container.right,  (screenW ?? container.right) - insets.right),
bottom: Math.min(container.bottom, (screenH ?? container.bottom) - insets.bottom),
```

Applied to both sites with the identical bug:
- `src/features/action/swipeon/resolveContainerSwipeCoordinates.ts`
- `src/features/action/swipeon/ScrollUntilVisible.ts` (the `reverseBounds` path)

The screen-size fallback is preserved: when `screenSize` is unknown the result matches the previous conservative behavior.

## Tests
New `test/features/action/swipeon/resolveContainerSwipeCoordinates.test.ts` covers:
- partial-height container away from the bottom inset is **not** shortened;
- narrow container + large right inset does **not** invert bounds;
- container overlapping the inset **is** still clamped to the safe region.

All 100 existing swipeon tests pass; lint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/features/action/swipeon/ScrollUntilVisible.ts
+++ b/src/features/action/swipeon/ScrollUntilVisible.ts
@@ -190,8 +190,8 @@ export class ScrollUntilVisible {
       reverseBounds = {
         left: Math.max(containerElement.bounds.left, insets.left),
         top: Math.max(containerElement.bounds.top, insets.top),
-        right: Math.min(containerElement.bounds.right, lastObservation.screenSize?.width ?? containerElement.bounds.right) - insets.right,
-        bottom: Math.min(containerElement.bounds.bottom, lastObservation.screenSize?.height ?? containerElement.bounds.bottom) - insets.bottom
+        right: Math.min(containerElement.bounds.right, (lastObservation.screenSize?.width ?? containerElement.bounds.right) - insets.right),
+        bottom: Math.min(containerElement.bounds.bottom, (lastObservation.screenSize?.height ?? containerElement.bounds.bottom) - insets.bottom)
       };
     }
     const reverseSwipeCoords = this.computeHalfScreenReverseCoords(reverseDirection, reverseBounds);
--- a/src/features/action/swipeon/resolveContainerSwipeCoordinates.ts
+++ b/src/features/action/swipeon/resolveContainerSwipeCoordinates.ts
@@ -17,8 +17,8 @@ export function resolveContainerSwipeCoordinates(
     effectiveBounds = {
       left: Math.max(containerElement.bounds.left, insets.left),
       top: Math.max(containerElement.bounds.top, insets.top),
-      right: Math.min(containerElement.bounds.right, observeResult.screenSize?.width ?? containerElement.bounds.right) - insets.right,
-      bottom: Math.min(containerElement.bounds.bottom, observeResult.screenSize?.height ?? containerElement.bounds.bottom) - insets.bottom
+      right: Math.min(containerElement.bounds.right, (observeResult.screenSize?.width ?? containerElement.bounds.right) - insets.right),
+      bottom: Math.min(containerElement.bounds.bottom, (observeResult.screenSize?.height ?? containerElement.bounds.bottom) - insets.bottom)
     };
   }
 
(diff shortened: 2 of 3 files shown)


kaeawc:

**Result: PASS** — container swipe bounds clamp correctly.

`swipeOn --direction up --autoTarget false` on the playground produced a real scroll rather than a short or inverted one:

| | value |
|---|---|
| coordinates | `(540,1768) -> (540,290)` |
| screen / insets | 1080x2400, top=63 bottom=63 (safe band 63..2337) |
| hierarchy signature | `ff2867cc75` -> `d9dce0ded0` |
| text nodes | 36 -> 31 |

The swipe is positive-length, stays inside the inset-safe band, and content actually moved — no negative-size or inverted bounds from the `min(container.right, screenW - inset)` clamp.

PR #3742, opened 2026-07-11
feat(tools): gate barrier + criticalSection as plan-only (hidden from MCP discovery)

kaeawc:
Hides the `barrier` and `criticalSection` tools from the advertised MCP `tools/list` while keeping them runnable inside plans.

## Why
Both are multi-device coordination primitives — a single direct MCP call (`deviceCount: 2`) just blocks until a second device arrives. An interactive agent has no use for them standalone; they only appear as **plan steps**, which `PlanExecutor` resolves via `getToolForPlan`, not discovery.

## How
Reuses the existing `planExecutable` / `getToolForPlan` seam (the same one `setUIState` uses):
- New `planOnly` tool flag + `PLAN_ONLY_GATE_REASON` gate reason in `ToolRegistry`.
- `planOnly` tools are excluded from discovery (`getTool` / `getToolDefinitions` / `registerWithServer`) but returned by `getToolForPlan` when also `planExecutable`.
- `getToolForPlan` suppresses its "gated tool used in plan" warning for the plan-only reason (it's by design) while still warning for other reasons (e.g. a debug-only tool in a plan).
- `barrier` + `criticalSection` marked `{ planOnly: true, planExecutable: true }`.

## Audit result (the "are there others?" question)
Audited all 71 registered tools: **no other genuinely plan-only tools exist** — `barrier`/`criticalSection` are the only two. 50 are agent-facing, 16 already gated (`debugOnly`/`embeddedSdkOnly`), and 3 recording/authoring meta-tools (`startTestRecording`, `exportPlan`, `recordSteps`) are *agent-invoked entry points*, not plan steps — so the `planExecutable` mechanism can't reach them; hiding those would remove the interactive record/export workflow (a product decision, not done here).

## Honest scope on savings
The ~450-token reduction applies to the **runtime daemon-mode `tools/list`** an agent connected to the daemon actually sees. It does **not** move the context benchmark's 99% figure — the benchmark models the stdio surface and never registered these daemon-only tools in the first place. So this is a correctness/cleanliness win (these shouldn't be interactively callable) with a modest daemon-only context trim, not a big benchmark drop.

## Testing
- Full `test/server` (1036) and plan suites (217) pass; typecheck clean.
- Registration test reworked to pin the plan-only gate: absent from `getTool`/`getToolDefinitions`, present via `getToolForPlan`, plus daemon-only registration.
- `daemonModeTools.test` updated to the new behavior.
- Static `schemas/tool-definitions.json` unchanged (generated with `includeUnavailable: true`).

### DIFF
--- a/src/server/barrierTools.ts
+++ b/src/server/barrierTools.ts
@@ -83,7 +83,11 @@ export function registerBarrierTools(): void {
     "barrier",
     "Synchronize multiple devices at a barrier, then let all proceed concurrently (no serialized section).",
     barrierSchema,
-    barrierHandler
+    barrierHandler,
+    // Plan-only: a multi-device coordination primitive that only makes sense as
+    // a plan step (a single direct call would just block). Hidden from tools/list
+    // discovery, still runnable in plans via getToolForPlan.
+    { planOnly: true, planExecutable: true }
   );
 
   logger.info("Barrier tools registered");
--- a/src/server/criticalSectionTools.ts
+++ b/src/server/criticalSectionTools.ts
@@ -218,7 +218,16 @@ const criticalSectionHandler = async (
  * Register the criticalSection tool.
  */
 export function registerCriticalSectionTools(): void {
-  ToolRegistry.registerDeviceAware("criticalSection", "Synchronize multiple devices at a barrier, then run steps serially.", criticalSectionSchema, criticalSectionHandler);
+  ToolRegistry.registerDeviceAware(
+    "criticalSection",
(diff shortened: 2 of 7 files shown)


kaeawc:

**Result: PASS** — `barrier` and `criticalSection` gated out of the live tool list.

Checked the live list rather than the schema file, as intended: `schemas/tool-definitions.json` holds 71 definitions, while `tools/list` over MCP stdio returns **69** — exactly these two absent. `executePlan` remains present, so the plan-only invocation path is intact.

PR #3750, opened 2026-07-11
perf(observe): compact the advertised waitFor input schema (99% -> 92% tools budget)

kaeawc:
Cuts the biggest single consumer of the MCP `tools/list` context budget.

## Profiling first (the "which 50 tools" answer)
I profiled all 42 advertised tools. Two facts drove this change:
1. The tracked context budget (`benchmark-context-thresholds.ts`) **strips output schemas before counting**, so the 99% figure is input-schemas + descriptions only.
2. In that view, **`observe` is the #1 consumer at 2,433 tokens — and 2,064 of those are the single `waitFor` param.** Everything else is a long, even tail.

`waitFor`'s cost is pure JSON-schema *expansion* of the runtime zod union/intersection/alias machinery, which the agent doesn't need in `tools/list`.

## The change
Swap in a hand-authored **compact advertised schema** for `waitFor`, via observe's existing `withJsonSchemaOverride` seam. **Runtime validation is unchanged** — the full zod `waitForSchema` stays the source of truth; only the advertised hint shrinks:
- `waitFor`: 2,064 → **879** tokens
- `observe` inputSchema: 2,433 → **1,248** tokens

It preserves the published/runtime **reject-contract** that `observeWaitForSchema.test.ts` pins (and which I verified): at least one predicate required, `textAny` mutually exclusive with the element predicates / `matchType` / `textMatch`, `activeWindow` must carry an app id or `activityName` (incl. `packageName`/`bundleId` aliases), and the exact `textMatch` description string.

## Result
- **Context benchmark: Tools 14,910 → 13,725 (99% → 92%)**, TOTAL 86% → 80%.
- 2,075 server + observe tests pass; typecheck clean; lint clean.
- `schemas/tool-definitions.json` regenerated.

## Follow-up (not in this PR)
`tapOn`'s `selector`/`container` (~410 tokens) and the shared interaction targeting params are smaller, more-even candidates for the same treatment — a modest additional trim if wanted.

### DIFF
--- a/schemas/tool-definitions.json
+++ b/schemas/tool-definitions.json
@@ -3176,447 +3176,182 @@
           ]
         },
         "waitFor": {
-          "description": "Wait for element to appear before returning observation",
-          "anyOf": [
-            {
+          "type": "object",
+          "additionalProperties": false,
+          "description": "Wait for a predicate before returning the observation. Provide at least one of: elementId, text, textAny, className, contentDescription, or activeWindow. textAny is mutually exclusive with the element predicates.",
+          "properties": {
+            "elementId": {
+              "type": "string",
+              "description": "Element resource ID / accessibility identifier"
+            },
+            "text": {
+              "type": "string",
+              "description": "Element text"
+            },
+            "textAny": {
+              "type": "array",
+              "items": {
+                "type": "string"
+              },
+              "description": "Ordered text variants; first visible match wins"
+            },
+            "className": {
+              "type": "string",
+              "description": "Element class name"
+            },
+            "contentDescription": {
+              "type": "string",
+              "description": "Element content description / accessibility label"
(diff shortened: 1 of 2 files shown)


kaeawc:

**Result: PASS** — compact `waitFor` schema, runtime validation unchanged.

Live `tools/list` shows `observe.waitFor` as a flat object with 10 properties (`activeWindow, className, container, contentDescription, elementId, matchType, text, textAny, textMatch, timeout`) plus an exclusion guard, rather than the expanded zod union.

Mutual exclusion is still enforced at runtime — passing `textAny` together with `matchType` is rejected:

```
Invalid parameters for tool observe: waitFor.matchType expected never, received string;
waitFor.textAny expected never, received array
```

Issue #3137, opened 2026-07-05
chore(db): converge telemetry-event retention onto the canonical offset(N-1)+id-tiebreak idiom


kaeawc:
## Summary

The six telemetry event repositories prune retention with `offset(maxRows)` +
`WHERE timestamp < cutoff` (keeps `maxRows + 1` rows in steady state, and keeps
**all** rows sharing the cutoff timestamp). The three "config" retention repos —
`performanceAuditRepository.ts` / `testExecutionRepository.ts` /
`failureAnalyticsRepository.ts` — use the canonical `offset(maxRows - 1)` +
compound `id`-tiebreak delete:

```ts
.orderBy("timestamp", "desc").orderBy("id", "desc").limit(1).offset(RETENTION_MAX_ROWS - 1)
// delete WHERE timestamp < threshold.timestamp OR (timestamp = threshold.timestamp AND id < threshold.id)
```

which trims to **exactly** `maxRows` and breaks cutoff-timestamp ties deterministically.

## Why not done in #2799 / PR #3124

#2799 explicitly constrains the fix to "must NOT change query results." Converging
onto the canonical form **does** change pruned output (trims to N instead of N+1,
and can prune same-timestamp rows the current form retains), so it legitimately
could not ride that PR. It is filed here so the divergence is intentional-on-record.

## Motivation

CLAUDE.md "one canonical primitive per concern": `src/db` currently has two
retention idioms. The telemetry form is also subtly less robust — a burst of
same-millisecond events at the cutoff can retain more than `maxRows`, since the
`timestamp < cutoff` delete can never remove rows equal to the cutoff timestamp.

## Acceptance
- [ ] Six telemetry repos use the canonical `offset(N-1)` + `id`-tiebreak delete.
- [ ] Behavior tests updated for exact trim-to-N + same-timestamp tie handling.
- [ ] Note the intended pruned-output change in the PR (it is a deliberate semantics change).

## Provenance
Architecture review of PR #3124 (issue #2799), finding #6. Compare
`src/db/performanceAuditRepository.ts:313-333` vs `src/db/networkEventRepository.ts` `cleanupIfNeeded`.

Issue #2799, opened 2026-07-01
perf(db): amortize per-insert retention COUNT(*) in event repositories


kaeawc:
## Summary
Six event-telemetry repositories run a full-table `SELECT count(*)` on **every single insert** to decide whether to trim to their 10,000-row retention cap. The `cleanupInProgress` boolean only prevents *concurrent* cleanups; it does not throttle the count, so the scan fires once per recorded event. Under bursty telemetry (network capture, log/OS event streaming), this doubles the query count per insert for a decision that is only meaningful once every ~10k rows. Found via a DB concurrency/migration audit (2026-07-01), finding P5count, priority P2, verified against source.

## Background & history
- Introduced: commit `d20e1f0e` (2026-03-16), [PR #1535](https://github.com/kaeawc/auto-mobile/pull/1535) — "feat: network/log/OS/custom telemetry SDK with IDE dashboard". This landed the `recordX` + `cleanupIfNeeded` retention pattern for the telemetry repos.
- The `cleanupInProgress` re-entrancy guard was added the next day in `3a7cc7417` (2026-03-17) — it guards against overlapping cleanups but does nothing to reduce how often the count runs.
- `networkEventRepository.ts` picked up the same shape via [PR #1624](https://github.com/kaeawc/auto-mobile/pull/1624) (2026-03-28).
- Why it looks like this: retention was written as a simple "check-then-trim after each write." That is correct and cheap to reason about, but it puts a scan on the hot insert path. The repo already has a better precedent for amortized retention (see below).

## The problem
Each `recordX` fires `cleanupIfNeeded` unawaited, which unconditionally runs a `count(*)` over the whole table:

```ts
// src/db/networkEventRepository.ts:34
export async function recordNetworkEvent(input, db?): Promise<number> {
  const result = await getDb(db).insertInto("network_events").values({...})
    .returning("id").executeTakeFirstOrThrow();
  cleanupIfNeeded(db);   // fire-and-forget, unawaited
  return result.id;
}
```

```ts
// src/db/networkEventRepository.ts:173
async function cleanupIfNeeded(db?): Promise<void> {
  if (cleanupInProgress) { return; }          // prevents concurrent cleanup, NOT the count
  cleanupInProgress = true;
  try {
    const d = getDb(db);
    const count = await d.selectFrom("network_events")
      .select(d.fn.countAll().as("count"))     // full-table scan on EVERY insert
      .executeTakeFirstOrThrow();
[...]

PR #3786, opened 2026-07-12
perf(db): amortize per-insert offset-probe retention (#3435/#3436/#3440)

### LINKED ISSUE #3435: perf(db): performanceAuditRepository.cleanupRetention runs an offset-probe on every insert
**Location:** `src/db/performanceAuditRepository.ts:125` (call) -> `:304-341` (impl, offset at `:319`)

**Current:** `recordAudit` calls `cleanupRetention()` on **every** insert, which unconditionally runs `ORDER BY timestamp DESC ... LIMIT 1 OFFSET 9999` (`RETENTION_MAX_ROWS - 1`) with no `count(*)` gate and no amortization — exactly the offset-probe pattern already known to be ~57x slower than `count(*)`. Audits can arrive at live-metric cadence (per observation).

**Fix:** Adopt the `eventRetention.pruneEventTableByCount` idiom used by the six event tables — an amortized per-insert counter (fire every ~256 inserts) plus a cheap `count(*)` gate that only runs the offset probe when actually over cap.

**Impact:** HIGH — removes a 10k-row index walk from the hot write path ~255/256 of the time.

---
_Filed from a parallel algorithmic/data-structure performance audit of `src/`. Each finding was verified against the actual code; complexity and fix are concrete and implementable._

### LINKED ISSUE #3436: perf(db): failureAnalyticsRepository.cleanupRetention — per-failure offset-probe + NOT IN(DISTINCT) orphan scan
**Location:** `src/db/failureAnalyticsRepository.ts:346` (call) -> `:1124-1174` (offset at `:1139`, orphan cleanup `:1161-1168`)

**Current:** Fired via the write barrier on **every** `recordFailure`. (a) Same unconditional `LIMIT 1 OFFSET 9999` threshold probe with no `count(*)` gate/amortization; (b) orphan-group cleanup uses `WHERE id NOT IN (SELECT group_id FROM failure_occurrences DISTINCT)`, materializing the full distinct set and doing a NOT-IN scan every time.

**Fix:** Amortize with the counter + `count(*)`-gate pattern. Replace the orphan delete with `NOT EXISTS (SELECT 1 FROM failure_occurrences fo WHERE fo.group_id = failure_groups.id)` (correlated, index-served, early-exit), and only run it when the occurrence delete actually removed rows.

**Impact:** HIGH — removes two full-table operations from every failure ingest.

---
_Filed from a parallel algorithmic/data-structure performance audit of `src/`. Each finding was verified against the actual code; complexity and fix are concrete and implementable._

### LINKED ISSUE #3440: perf(db): testExecutionRepository.cleanupRetention runs a per-insert offset-probe
**Location:** `src/db/testExecutionRepository.ts:201` (call) -> `:334-377` (offset at `:355`)

**Current:** Same unconditional `LIMIT 1 OFFSET 9999` probe on every insert (after a cheap indexed age-delete). Lower frequency than the audit/failure paths (per test-execution) hence MEDIUM.

**Fix:** Amortized counter + `count(*)` gate. Keep the indexed/sargable age-based delete every call; only gate the row-cap offset probe.

**Impact:** MEDIUM.

---
_Filed from a parallel algorithmic/data-structure performance audit of `src/`. Each finding was verified against the actual code; complexity and fix are concrete and implementable._

kaeawc:
## Summary

`recordAudit` / `recordFailure` / `recordExecution` each ran an unconditional `LIMIT 1 OFFSET 9999` row-cap probe on **every** insert — the offset-probe pattern known to be ~57x slower than a `count(*)` gate (#2799). This puts a 10k-row index walk on the hot write path, at live-metric cadence for audits/failures.

Introduces `src/db/rowCapRetention.ts` with two small primitives:

- **`runAmortizedRetention`** — fires the cleanup body at most once per `CLEANUP_CHECK_INTERVAL` (256) inserts behind an in-progress guard, so the scan leaves the hot path ~255/256 of the time. Mirrors the existing `eventRetention.pruneEventTableByCount` amortization idiom.
- **`pruneTableByRowCap`** — centralizes the `count(*)`-gated trim (threshold offset probe + tie-broken delete, including the #3137 id-tiebreak invariant) so it is **not** hand-copied per repo. The cheap `count(*)` gates the expensive offset walk, which now runs only when actually over cap.

### Per-repo specifics
- **performanceAudit (#3435):** pure row-cap trim via the shared helper.
- **failureAnalytics (#3436):** row-cap trim, then only sweeps orphan groups **when the delete removed rows**, via a correlated `NOT EXISTS` (served by `idx_failure_occurrences_group`, early-exit) instead of `NOT IN (SELECT group_id ... DISTINCT)`.
- **testExecution (#3440):** keeps the cheap sargable age-based delete on **every** insert; only the offset probe is amortized.

## Tests
- `test/db/rowCapRetention.test.ts` — amortization gate (fires 1/N), in-progress guard drops overlap, guard released on throw.
- Per-repo regression tests: trims to exactly the cap keeping the newest rows; under-cap `count(*)` gate short-circuits (no delete); failureAnalytics orphan-group sweep vs. retention of groups that keep an occurrence; testExecution age-delete still runs every insert.
- Full `test/db` suite green (817 tests).

## Notes
Reviewed via `/simplify` (4 cleanup agents): extracted the row-cap *body* (not just the amortization shell) into the shared helper so the correctness-sensitive tie-break invariant lives in one place; dropped an unused batch-count parameter (YAGNI).

Closes #3435, closes #3436, closes #3440.

### DIFF
--- a/scripts/typecheck-baseline.txt
+++ b/scripts/typecheck-baseline.txt
@@ -7,8 +7,8 @@ src/daemon/manager.ts: error TS18048: 'signal' is possibly 'undefined'.
 src/daemon/performanceStreamSocketServer.ts: error TS2559: Type 'PerformanceStreamSocketRequest' has no properties in common with type 'SocketRequest'.
 src/daemon/socketServer.ts: error TS2322: Type 'unknown' is not assignable to type '{ [x: string]: unknown; } | undefined'.
 src/daemon/socketServer.ts: error TS2554: Expected 1-2 arguments, but got 3.
-src/daemon/telemetryPushSocketServer.ts: error TS2339: Property 'sessionId' does not exist on type '{ type: "crash" | "anr" | "tool_failure"; title: string; severity: "critical" | "high" | "medium" | "low"; stack_trace_json: string | null; timestamp: number; occurrenceId: string; groupId: string; deviceId: string | null; screen: string | null; }'.
-src/daemon/telemetryPushSocketServer.ts: error TS2345: Argument of type '"crash" | "anr" | "nonfatal"' is not assignable to parameter of type 'OperandValueExpressionOrList<Database, "failure_groups" | "failure_occurrences", "failure_groups.type">'.
+src/daemon/telemetryPushSocketServer.ts: error TS2339: Property 'sessionId' does not exist on type '{ timestamp: number; type: "crash" | "anr" | "tool_failure"; title: string; severity: "critical" | "high" | "medium" | "low"; stack_trace_json: string | null; occurrenceId: string; groupId: string; deviceId: string | null; screen: string | null; }'.
+src/daemon/telemetryPushSocketServer.ts: error TS2345: Argument of type '"crash" | "anr" | "nonfatal"' is not assignable to parameter of type 'OperandValueExpressionOrList<Database, "failure_occurrences" | "failure_groups", "failure_groups.type">'.
 src/daemon/testRecordingSocketServer.ts: error TS2322: Type 'string' is not assignable to type 'Platform | undefined'.
 src/daemon/testRecordingSocketServer.ts: error TS2322: Type 'string' is not assignable to type 'Platform | undefined'.
 src/daemon/testRecordingSocketServer.ts: error TS2559: Type 'TestRecordingCommand' has no properties in common with type 'SocketRequest'.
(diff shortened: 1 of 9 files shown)


kaeawc:
Rebased onto current `main` (was 9 commits behind). Also regenerated `scripts/typecheck-baseline.txt`: adding `src/db/rowCapRetention.ts` perturbed the **print order** of two pre-existing, already-baselined `telemetryPushSocketServer.ts` type errors (object-property order + `"failure_occurrences" | "failure_groups"` union order) — same errors, count unchanged at 235, no new errors introduced. Verified with the pinned tsc 6.0.3.

PR #3787, opened 2026-07-12
fix(daemon): reject non-finite tap coordinates, matching swipe (#3615)

### LINKED ISSUE #3615: fix(daemon): input/tap coordinate validation uses Number.isNaN, accepts ±Infinity (drifts from input/swipe)
Found by an adversarial review of the `0.0.40..HEAD` (0.0.44) `src/**/*.ts` diff. Confidence: **high**.

## Symptom
`src/daemon/socketServer.ts` `parseInputTapParams` (~line 1161) validates coordinates with `Number.isNaN`:

```ts
if (typeof args.x !== "number" || Number.isNaN(args.x) ||
    typeof args.y !== "number" || Number.isNaN(args.y)) {
  throw new Error("input/tap requires numeric x and y params");
}
```

The sibling `parseInputSwipeParams` (~line 1197, same changeset) validates with `!Number.isFinite`. Because `Number.isNaN(Infinity) === false` and `typeof Infinity === "number"`, **`input/tap` accepts `±Infinity`** for `x`, `y`, and `duration` where `input/swipe` deliberately rejects them.

## Why it matters
Impact is bounded — downstream Android/iOS runners have their own non-finite guards (#2964) and `JSON.stringify(Infinity)` serializes to `null`, so it degrades to a wire/decode error rather than a crash. But the parser violates its own stated contract (`"requires numeric x and y"`) and silently diverges from its sibling — classic intra-change copy-paste drift.

## Fix direction
Replace the `Number.isNaN` checks with `!Number.isFinite` in `parseInputTapParams` so tap and swipe validation match. Add a small unit test asserting `input/tap` rejects `Infinity`/`-Infinity`/`NaN` for x/y/duration.

## Acceptance criteria
- [ ] `parseInputTapParams` rejects non-finite x/y/duration
- [ ] Tap and swipe parsers use the same finiteness check
- [ ] Unit test covers the non-finite rejection

## Related
- `parseInputSwipeParams` in the same file (correct pattern)
- #2964 (downstream non-finite gesture guards)

kaeawc:
## Summary
Closes #3615.

`parseInputTapParams` (`src/daemon/socketServer.ts`) validated `x`/`y`/`duration` with `Number.isNaN`. Because `Number.isNaN(Infinity) === false` and `typeof Infinity === "number"`, `input/tap` **accepted `±Infinity`**, while the sibling `parseInputSwipeParams` deliberately rejects non-finite values via `!Number.isFinite`. Classic intra-change copy-paste drift; the parser violated its own `"requires numeric x and y"` contract.

## Change
Switch the tap checks to `!Number.isFinite` for `x`, `y`, and `duration`, so tap and swipe use the same finiteness validation.

Impact is bounded (a JSON client stringifies `Infinity` → `null`, and downstream runners carry their own non-finite guards per #2964), so this is contract-correctness/anti-drift hardening.

## Test
`parseInputTapParams` is exercised **directly** rather than over the socket: `±Infinity`/`NaN` cannot survive a JSON round-trip (`JSON.stringify` → `null`, `JSON.parse` rejects the literal), so the non-finite value only reaches the parser from a non-JSON in-process caller. The test asserts `Infinity`/`-Infinity`/`NaN` are rejected for `x`/`y`/`duration` and that finite values still parse — a guard that fails against the old `Number.isNaN` implementation.

`bun test` (tap + swipe), `bunx turbo run build lint`, and `bun run typecheck` all green.

### DIFF
--- a/src/daemon/socketServer.ts
+++ b/src/daemon/socketServer.ts
@@ -1158,10 +1158,17 @@ export class UnixSocketServer {
     if (args.platform !== "android" && args.platform !== "ios") {
       throw new Error("input/tap requires platform 'android' or 'ios'");
     }
-    if (typeof args.x !== "number" || Number.isNaN(args.x) || typeof args.y !== "number" || Number.isNaN(args.y)) {
+    // Reject NaN AND ±Infinity (finiteness), matching parseInputSwipeParams. Number.isNaN
+    // alone lets ±Infinity through, violating the "numeric x and y" contract (#3615).
+    if (
+      typeof args.x !== "number" ||
+      !Number.isFinite(args.x) ||
+      typeof args.y !== "number" ||
+      !Number.isFinite(args.y)
+    ) {
       throw new Error("input/tap requires numeric x and y params");
     }
-    if (args.duration !== undefined && (typeof args.duration !== "number" || Number.isNaN(args.duration))) {
+    if (args.duration !== undefined && (typeof args.duration !== "number" || !Number.isFinite(args.duration))) {
       throw new Error("input/tap duration must be numeric when provided");
     }
     if (args.deviceId !== undefined && typeof args.deviceId !== "string") {
--- a/test/daemon/socketServerInputTap.test.ts
+++ b/test/daemon/socketServerInputTap.test.ts
@@ -433,4 +433,33 @@ describe("UnixSocketServer input/tap", () => {
     expect(nonNumeric.success).toBe(false);
     expect(nonNumeric.error).toBe("input/tap requires numeric x and y params");
   });
+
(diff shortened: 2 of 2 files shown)


kaeawc:

**Result: PASS** — non-finite tap coordinates rejected.

Differential test driven straight at the daemon socket (`input/tap`), using `1e999` since JSON has no `Infinity` literal and it parses to `Infinity`:

```
x=1e999  -> {"success":false,"error":"input/tap requires numeric x and y params"}
x=540    -> {"success":true,"result":{"coordinates":{"x":540,"y":500}}}
```

`Number.isNaN` alone would have let the first through. Matches the swipe contract.

PR #3808, opened 2026-07-12
fix(schema): hide injected deviceId from tools list

### LINKED ISSUE #3768: Hide injected deviceId from advertised tool schemas on device-targeted tools
## Problem

`deviceId` is advertised in `tools/list` on the ~35 device-aware tools that get it via `addDeviceTargetingToSchema`, where it is **not** an agent-facing parameter — it exists only so the executor can inject a resolved device id after allocation without tripping strict schema validation:

> `deviceId` exists so the executor can inject a resolved deviceId into `requiresDevice` tool calls after device allocation without tripping strict schema validation. — `src/server/toolSchemaHelpers.ts` (deviceTargetingShape comment)

Advertising it on those tools is:
- **Wasteful** — ~9 tokens × ~35 tools ≈ **~360 tokens** of `tools/list` context (which reaches the model; see the input-schema trims in #3750 / #3754).
- **Misleading** — an agent seeing `deviceId` may try to set it, when `device` (a label) is the field it should use.

## Why this needs care (not a blanket strip)

`deviceId` is **genuinely agent-facing on other tools**, so it cannot be stripped globally:
- `src/server/videoRecordingTools.ts:106` — `!args.deviceId && !args.device && !args.sessionUuid` treats `deviceId` as a valid targeting input.
- `src/server/highlightTools.ts:359` — `deviceId: args.deviceId ?? device.deviceId` reads it as agent input.
- `src/server/deviceTools.ts:68` (`killDevice`) and `src/server/utilityTools.ts:22` (`setActiveDevice`) take a **required** `deviceId` — it's the tool's purpose.

Complicating factor: `highlight` and `videoRecording` **both** use `addDeviceTargetingToSchema` **and** declare their own `deviceId`, so a strip scoped naively to that helper would delete a real param.

## Suggested approach (provenance-based)

Distinguish the injected `deviceId` from an explicitly-declared one using `extendPreservingBase` (`src/server/toolSchemaHelpers.ts`), which already computes the `added` set (fields it adds because they were not already in `schema.shape`):

1. When `extendPreservingBase` (via `addDeviceTargetingToSchema`) **adds** `deviceId` (i.e. it was not pre-declared), register the resulting schema in a `WeakSet<injectedDeviceIdSchemas>`. Tools like `highlight`/`videoRecording` that pre-declare `deviceId` are **not** registered (their field is kept).
2. In `toAdvertisedJsonSchema` (`src/server/toolRegistry.ts:45`), for a schema in that set, delete `jsonSchema.properties.deviceId` from the advertised output only. Runtime zod is untouched, so executor injection still works.
3. Watch the object-identity through the `withAppIdAliases` (`z.preprocess`) / `withJsonSchemaOverride` wrappers — the strip must fire for the actual node `toJSONSchema` visits (same concern the existing per-tool overrides handle).

## Acceptance
- `observe`, `tapOn`, `swipeOn`, etc. (injected-only): no `deviceId` in advertised `inputSchema`.
- `killDevice`, `setActiveDevice`, `videoRecording`, `highlight` (agent-facing `deviceId`): `deviceId` **still** advertised.
- Runtime validation and executor `deviceId` injection unchanged; full `test/server` + action/plan suites green.
- Add a test pinning the advertised presence/absence per tool (mirroring `daemonModeTools.test.ts` / `observeWaitForSchema.test.ts`).

## Context
Follow-up from the tool-context profiling that produced #3750 (observe `waitFor` compaction) and #3754 (interaction selector compaction). This is the remaining ~360-token lever in the shared device-targeting params; the other shared inputs (`platform`, `device`, `sessionUuid`) are already minimal/irreducible. Value here is as much **correctness** (stop advertising an internal field) as tokens.

kaeawc:
## Summary

Hides the executor-only `deviceId` field from advertised `tools/list` input schemas when it was injected by `addDeviceTargetingToSchema`, while preserving runtime validation so executor device allocation can still inject the resolved id. Explicit, agent-facing `deviceId` parameters remain advertised for tools such as `setActiveDevice`, `videoRecording`, `highlight`, and `killDevice`.

## Linked Issue

Closes #3768

## Bug / Regression Context

- **Prior attempts:** Follow-up to the tool-context trims in #3750 and #3754.
- **Observed symptom:** Device-aware tools advertised `deviceId` even when it was only an internal executor injection field, wasting tools/list context and implying agents should set it directly.
- **Repro steps / conditions:** Inspect `tools/list` / `schemas/tool-definitions.json` for injected-only tools such as `observe`, `tapOn`, or `swipeOn`.

## Validation

- [x] `bun test test/server/tools/deviceIdAdvertisement.test.ts test/server/toolSchemaHelpers.test.ts test/server/toolRegistry.getToolDefinitions.test.ts test/server/observeWaitForSchema.test.ts --bail`
- [x] `bun test test/server test/features/action test/plan test/utils/plan --bail`
- [x] `bun run typecheck`
- [x] `git diff --check`
- [ ] Android/iOS changes: ran the matching `scripts/` validation
- [x] New code uses interfaces + fakes + FakeTimer; unit tests run in <100ms

## Device Verification

N/A; this changes advertised JSON schemas only. Runtime validation for executor-injected `deviceId` is covered by unit tests.

## Follow-ups

- [x] Follow-up issues filed for gaps not addressed here: N/A

### DIFF
--- a/schemas/tool-definitions.json
+++ b/schemas/tool-definitions.json
@@ -31,9 +31,6 @@
         "device": {
           "description": "Device label",
           "type": "string"
-        },
-        "deviceId": {
-          "type": "string"
         }
       },
       "additionalProperties": false
@@ -104,9 +101,6 @@
         },
         "device": {
           "description": "Device label",
-          "type": "string"
-        },
-        "deviceId": {
           "type": "string"
         }
       },
@@ -420,9 +414,6 @@
         "device": {
           "description": "Device label",
           "type": "string"
-        },
-        "deviceId": {
-          "type": "string"
         }
       },
       "required": [
@@ -466,9 +457,6 @@
         },
         "device": {
           "description": "Device label",
-          "type": "string"
-        },
-        "deviceId": {
           "type": "string"
         }
       },
@@ -543,9 +531,6 @@
         "device": {
           "description": "Device label",
           "type": "string"
-        },
-        "deviceId": {
-          "type": "string"
         }
       },
       "required": [
@@ -605,9 +590,6 @@
         "device": {
           "description": "Device label",
           "type": "string"
-        },
-        "deviceId": {
-          "type": "string"
         }
       },
       "required": [
@@ -644,9 +626,6 @@
         "device": {
           "description": "Device label",
           "type": "string"
-        },
-        "deviceId": {
-          "type": "string"
         }
       },
       "additionalProperties": false
@@ -675,9 +654,6 @@
         },
         "device": {
           "description": "Device label",
-          "type": "string"
-        },
-        "deviceId": {
           "type": "string"
         }
       },
@@ -725,9 +701,6 @@
         },
         "device": {
           "description": "Device label",
-          "type": "string"
-        },
-        "deviceId": {
           "type": "string"
         }
       },
@@ -895,9 +868,6 @@
         "device": {
           "description": "Device label",
           "type": "string"
-        },
-        "deviceId": {
-          "type": "string"
         }
       },
       "required": [
@@ -980,9 +950,6 @@
         },
         "device": {
           "description": "Device label",
-          "type": "string"
-        },
-        "deviceId": {
           "type": "string"
         }
       },
@@ -1121,9 +1088,6 @@
         },
         "device": {
           "description": "Device label",
-          "type": "string"
-        },
-        "deviceId": {
           "type": "string"
         }
       },
@@ -1458,9 +1422,6 @@
         "device": {
           "description": "Device label",
           "type": "string"
-        },
-        "deviceId": {
-          "type": "string"
         }
       },
       "required": [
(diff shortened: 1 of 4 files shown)


kaeawc:

**Result: PASS** — injected `deviceId` hidden from advertised schemas, still routes at call time.

Live `tools/list` over MCP stdio: 69 tools, of which only 5 advertise `deviceId` — `setActiveDevice`, `startDevice`, `executePlan`, `videoRecording`, `highlight`. Those five declare their own `deviceId` rather than the shared injected shape, so `isInjectedDeviceIdSchema` correctly leaves them alone; `observe`/`tapOn` and the other 64 no longer advertise it.

Call-time routing still works with the property hidden:

```
observe --platform android --deviceId emulator-5554
  -> routed OK, screenSize {width:1080, height:2400}
```

Issue #1959, opened 2026-04-17
Support segmented video recording to chain multiple screenrecord sessions past the 180-second Android cap


RyeBread1026:
Description: Android's adb shell screenrecord has a hard 180-second (3-minute) recording cap. Plans running longer than 3 minutes lose video coverage entirely - the recording simply stops, and there is no video for the remainder of the test.

Reproduction: Run an executePlan that takes longer than 3 minutes on Android. The video file covers only the first ~180 seconds.

Expected: The system should automatically chain multiple screenrecord sessions to provide continuous video coverage for plans of any length.

Impact: Video artifacts for long-running CI tests are incomplete, making it impossible to review what happened in the later (often more interesting) portion of the test.

kaeawc:
Landed in 11882d9f — segmented Android video via chained screenrecord sessions in src/server/androidSegmentedPlanVideoSession.ts.

PR #3847, opened 2026-07-13
Add tool-level segmented video recording past the 180s screenrecord cap

### LINKED ISSUE #3843: Segmented video recording only covers executePlan — the raw videoRecording tool still hits the 180s screenrecord cap
Follow-up to #1959, which is closed but incomplete for a real usage path.

#1959 was fixed in 11882d9f ("feat(plan): segmented Android video for long executePlan runs") — but that fix is entirely scoped to `executePlan`: it only touches `src/server/planTools.ts`, `src/utils/plan/PlanExecutor.ts`, and `src/server/androidSegmentedPlanVideoSession.ts`, chaining recordings via `onBeforePlanStep`.

It never touches the raw `videoRecording` tool handler (`src/server/videoRecordingTools.ts`) that callers hit directly via `--cli videoRecording --action start/stop` (not through `executePlan`). Any external driver that starts/stops recording itself — e.g. a QA-agent runner calling `am videoRecording --action start --outputName ... && ... && am videoRecording --action stop` per test case — still hits the hard 180s `adb shell screenrecord` cap with no segmentation, exactly as originally reported in #1959.

Reproduction: call `videoRecording --action start`, run device interactions for >180s, then `--action stop`. The resulting recording is truncated at ~180s regardless of how long the tool call span was.

Expected: the same session-rotation logic `executePlan` gets (chaining ordered segments past the cap) should also apply when a caller drives `videoRecording` directly, not only through `executePlan`.

RyeBread1026:
Fixes #3843 (follow-up to #1959, which only covered executePlan).

#1959 was fixed for executePlan-driven recording (11882d9f), but the raw `videoRecording` tool — used directly by callers that manage start/stop themselves (e.g. a QA-agent runner calling `videoRecording --action start`/`stop` per test case, never going through executePlan) — still hit the hard 180s `adb shell screenrecord` cap with no segmentation.

This threads the same rotation approach to the tool-level path: a >180s session rolls over every ~170s into ordered clips (`<name>`, `<name>-seg1`, ...) instead of losing coverage at the cap.

Validated on-device before opening: a 200s recording → bare stop → 2 ordered clips, no leaked rotation timer.

### DIFF
--- a/src/server/androidSegmentedPlanVideoSession.ts
+++ b/src/server/androidSegmentedPlanVideoSession.ts
@@ -19,6 +19,13 @@ export interface AndroidSegmentedPlanVideoSessionOptions {
   timer?: Timer;
   /** Override for tests. */
   segmentRotateAfterMs?: number;
+  /**
+   * Overall session duration bound. When set, the session auto-stops (finalizing every
+   * completed segment) once this many seconds have elapsed since {@link start}, matching
+   * the non-segmented recording path's maxDuration-is-an-auto-stop-bound contract. Undefined
+   * means rotate indefinitely until an explicit {@link stop} call — no session-level cap.
+   */
+  maxDurationSeconds?: number;
   /** Quality/config overrides forwarded to every segment's recording. */
   configOverrides?: VideoRecordingConfigInput;
   startVideoRecording?: (
@@ -48,12 +55,17 @@ export class AndroidSegmentedPlanVideoSession {
 
   private readonly segmentRotateAfterMs: number;
 
+  private readonly maxDurationSeconds: number | undefined;
+
   private readonly configOverrides: VideoRecordingConfigInput | undefined;
 
   private activeRecordingId: string | undefined;
 
   /** Timer-driven rotation handle (set only when {@link start} is used). */
   private rotationTimerHandle: NodeJS.Timeout | undefined;
+
+  /** Session-level auto-stop handle (set only when {@link maxDurationSeconds} is provided). */
+  private maxDurationTimerHandle: NodeJS.Timeout | undefined;
 
   /** True while a timer-driven session is running, so rotations keep rescheduling. */
   private timerDriven = false;
@@ -82,6 +94,7 @@ export class AndroidSegmentedPlanVideoSession {
     this.outputNamePrefix = options.outputNamePrefix;
     this.timer = options.timer ?? defaultTimer;
     this.segmentRotateAfterMs = options.segmentRotateAfterMs ?? ANDROID_PLAN_VIDEO_SEGMENT_ROTATE_MS;
+    this.maxDurationSeconds = options.maxDurationSeconds;
     this.configOverrides = options.configOverrides;
     this.startVideoRecordingFn = options.startVideoRecording ?? defaultStartVideoRecording;
     this.stopVideoRecordingFn = options.stopVideoRecording ?? defaultStopVideoRecording;
@@ -100,13 +113,16 @@ export class AndroidSegmentedPlanVideoSession {
    * Timer-driven lifecycle (does NOT depend on plan steps). Starts the first
    * segment, then schedules a self-rescheduling rotation via the injected
    * {@link Timer} so each segment stays under {@link ANDROID_SCREENRECORD_MAX_SECONDS}.
-   * Returns the first segment's recording, whose recordingId is used as the
-   * session handle by callers.
+   * If {@link maxDurationSeconds} was provided, also arms a session-level auto-stop so
+   * overall duration is bounded the same way the non-segmented path bounds it - rotation
+   * alone never stops on its own. Returns the first segment's recording, whose
+   * recordingId is used as the session handle by callers.
    */
   async start(): Promise<ActiveVideoRecording> {
     this.timerDriven = true;
     const first = await this.startFirstSegment();
     this.scheduleRotation();
+    this.scheduleMaxDurationStop();
     return first;
   }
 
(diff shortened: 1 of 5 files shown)


src/server/videoRecordingTools.ts
```diff
@@ -189,6 +288,38 @@ export function registerVideoRecordingTools(): void {
...
+            target.platform === "android" &&
+            maxDurationSeconds > ANDROID_SCREENRECORD_MAX_SECONDS
+          ) {
+            const session = new AndroidSegmentedPlanVideoSession({
+              device: target,
+              outputNamePrefix: args.outputName ?? `recording-${target.deviceId}`,
+              configOverrides: buildConfigOverrides(args),
+              timer: segmentedSessions.timer,
+            });
+            const active = await session.start();
```
kaeawc, line 345: Not a blocker (the QA-agent flow in #3843 always issues an explicit stop), but the segmented path never honors `maxDuration` as a duration cap the way the non-segmented path does.

`session.start()` only starts the first segment and self-reschedules `scheduleRotation()` indefinitely (androidSegmentedPlanVideoSession.ts:106-111). Each *segment* gets a 180s auto-stop via the manager (`scheduleAutoStop`, videoRecordingManager.ts:598), but rotation at 170s pre-empts it — so nothing bounds total duration. A caller passing `maxDuration: 400` and never calling stop records past 400s until eviction starts dropping the earliest segments.

Since this PR also raised the schema ceiling to 3600 — implying `maxDuration` is meaningful as a duration — the two paths now diverge. Either arm a session-level timer in `start()` that calls `stop()` once cumulative elapsed reaches `maxDurationSeconds`, or add a one-line comment that `maxDuration` only gates segmentation here and no longer bounds total duration.

src/server/videoRecordingTools.ts
```diff
@@ -68,11 +132,16 @@ const videoRecordingSchema = addDeviceTargetingToSchema(z.object({
...
+  // accepted. Note: non-Android recordings still can't exceed the manager's
+  // single-recording cap (300s) since they aren't segmented; such a request
+  // passes schema validation but is rejected at videoRecordingManager. Left
+  // as-is intentionally (out of scope for segmented Android capture).
   maxDuration: z
     .number()
     .int()
     .positive()
-    .max(300)
+    .max(3600)
```
kaeawc, line 180: Minor / acknowledged: raising the schema ceiling to 3600 means an iOS (or any non-Android) caller with `181 <= maxDuration <= 3600` now passes schema validation but is rejected downstream at `resolveMaxDurationSeconds` (videoRecordingManager.ts:233-234, `maxDuration must be <= 300 seconds`). The inline comment already calls this out as intentional/out-of-scope, so no change requested — just confirming the runtime error is reachable and is the less-obvious failure mode a caller will hit.

kaeawc (commented):
Reviewed against latest `origin/main`. The change correctly closes the reported reproduction in #3843: an Android `videoRecording --action start` with `maxDuration > 180` now spins up a timer-driven `AndroidSegmentedPlanVideoSession`, and both the by-handle stop and the bare by-device stop finalize it and return the ordered segments. Ran the two new test files locally — 9 pass / 0 fail in 270ms. Tests follow the Interface+Fake+FakeTimer convention. Verified the `startVideoRecording` per-device guard (videoRecordingManager.ts:541-547) prevents a co-active non-segmented recording, so the bare-stop `continue` cannot silently orphan a second recording — a concern I checked and dropped.

One behavioral divergence worth a decision (inline), plus one already-acknowledged schema note. Neither blocks the core fix.

src/server/videoRecordingTools.ts
RyeBread1026: Confirmed, I believe the runtime error is the right failure mode for that case.

src/server/videoRecordingTools.ts
RyeBread1026: Good catch - that was an oversight. Fixed in 5671d830: `AndroidSegmentedPlanVideoSession` now
accepts `maxDurationSeconds`, arms a session-level timer in `start()` that calls the existing `stop()` once elapsed reaches it (reusing `stop()`'s already-safe cleanup rather than adding new concurrency handling), and `videoRecordingTools.ts` now passes the already-computed value through instead of dropping it. Added a test with a bound that deliberately doesn't align with the rotation cadence (2.5s cap vs 1s rotation) to make sure auto-stop isn't accidentally piggybacking on a rotation tick.

src/server/androidSegmentedPlanVideoSession.ts
```diff
@@ -69,12 +94,87 @@ export class AndroidSegmentedPlanVideoSession {
...
+
+  private scheduleMaxDurationStop(): void {
+    if (this.maxDurationSeconds === undefined) {
+      return;
+    }
+    this.maxDurationTimerHandle = this.timer.setTimeout(() => {
+      logger.info(
+        `[SegmentedPlanVideo] Session reached maxDurationSeconds=${this.maxDurationSeconds}, auto-stopping`
+      );
+      this.stop().catch(error => {
```
kaeawc, line 168: Follow-on gap from the new auto-stop: the timer calls `this.stop()`, which finalizes but never removes the session from `byHandle` — that delete lives only in `stopAndRemove` (videoRecordingTools.ts:62), which auto-stop doesn't go through. So an auto-stopped session stays registered forever.

This is reachable in exactly the use case this fix enables — a caller that sets `maxDuration` and never calls stop (fire-and-forget). Two consequences:
1. **Map leak:** one dead entry per auto-stopped, never-caller-stopped session; over a long-lived daemon the map grows unbounded and `forDevice()` iterates all of them.
2. **Stale segments on a later bare stop:** after auto-stop finalizes, the manager's per-device guard (videoRecordingManager.ts:545) sees no `recording`-status record, so a *new* recording can start on the same device under a new handle. `forDevice(deviceId)` then returns `[deadSession, newSession]`, and a bare by-device stop reports the dead session's old segment paths mixed in with the new ones as if just recorded.

In the normal QA-agent flow (always issues an explicit stop) the entry self-heals because that stop goes through `stopAndRemove`, so this isn't a blocker. Fix: give the session a completion hook the tool wires to registry removal — e.g. an `onFinalized?: () => void` option invoked at the end of `stop()`/`finalize()` that calls a new `segmentedSessions.remove(session)` (filter `byHandle` by session identity, since the handle isn't known inside the session). That keeps the registry the single owner of the map and closes both consequences.

**Repro:** start an Android recording with `maxDuration` short (segmented), let it auto-stop without calling stop, then start+bare-stop a new recording on the same device and confirm the response doesn't list the first session's segment files.

kaeawc (commented):
Re-reviewed after `5671d830b` (fix: enforce maxDurationSeconds as a real cap). **My primary finding is correctly fixed:** `start()` now arms a session-level auto-stop timer at `maxDurationSeconds`, reusing the existing `stop()` cleanup path, and `videoRecordingTools.ts` threads the already-computed `maxDurationSeconds` through. The new regression test deliberately misaligns the bound (2.5s) with the rotation cadence (1s), proving auto-stop fires mid-segment and that a subsequent caller `stop()` is a safe no-op. Ran the two affected files locally — 10 pass / 0 fail in 298ms.

The fix introduces one follow-on gap (inline, Partial). The minor schema note (300→3600 ceiling reachable for non-Android) from the first pass still stands and was acknowledged as intentional.

kaeawc:

**Result: PASS** — raw `videoRecording` segments past the 180s `screenrecord` cap.

A 208-second recording produced **two** segments under one session handle instead of a truncated single file:

| index | file | duration |
|---|---|---|
| 0 | `segtest2-7eb004ee-...-20260720T152612Z.mp4` | 170.86s |
| 1 | `segtest2-seg1-7d86b107-...-20260720T152903Z.mp4` | 12.40s |

Both are playable (ffprobe reads a duration, so the moov atom is present). The rollover is clean with no gap: segment 0 starts at `152612Z` and runs 171s, and segment 1 starts at `152903Z` — exactly where segment 0 ends.

PR #3848, opened 2026-07-13
Add --no-occlusion runtime flag to disable observe occlusion filtering

### LINKED ISSUE #3844: Add a --no-occlusion runtime flag to disable the observe occlusion-filtering pass
The occlusion pass in `observe` (filtering out views that are completely hidden behind other views) is always on, with no way to disable it at runtime.

For CI/QA-agent runs where every bit of per-call latency and log volume matters, we want a way to skip this pass entirely — accepting that fully-occluded views may appear in the hierarchy — in exchange for lower observe latency and simpler output.

Ask: an `occlusionEnabled` flag threaded end-to-end (CLI `--no-occlusion`, MCP server config, spawned daemon subprocess, and the Kotlin-side `ViewHierarchyExtractor`/`CtrlProxy` request), defaulting to the current always-on behavior, opt-out only.

RyeBread1026:
Fixes #3844.

Threads an `occlusionEnabled` flag end-to-end: CLI `--no-occlusion`, MCP server config, the spawned daemon subprocess, and the Kotlin-side `ViewHierarchyExtractor`/CtrlProxy request. Defaults to the current always-on behavior; opt-out only, for callers (e.g. CI/QA-agent runs) where every bit of per-call latency and log volume matters and fully-occluded views showing up in the hierarchy is an acceptable tradeoff.

### DIFF
--- a/android/protocol/src/main/kotlin/dev/jasonpearson/automobile/protocol/WebSocketRequest.kt
+++ b/android/protocol/src/main/kotlin/dev/jasonpearson/automobile/protocol/WebSocketRequest.kt
@@ -401,6 +401,7 @@ data class SetAccessibilityFlags(
   val includeNotImportantViews: Boolean = true,
   val reportViewIds: Boolean = true,
   val retrieveInteractiveWindows: Boolean = true,
+  val occlusionEnabled: Boolean = true,
 ) : WebSocketRequest()
 
 @Serializable
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CtrlProxy.kt
@@ -1897,7 +1897,8 @@ class CtrlProxy : AccessibilityService(), CtrlProxyActions {
       if (!allWindows.isNullOrEmpty()) {
         Log.d(
           TAG,
-          "Extracting from ${allWindows.size} windows (disableAllFiltering: $disableAllFiltering)",
+          "Extracting from ${allWindows.size} windows " +
+            "(disableAllFiltering: $disableAllFiltering, occlusionEnabled: $occlusionEnabled)",
         )
         viewHierarchyExtractor.extractFromAllWindows(
           allWindows,
@@ -2038,6 +2039,11 @@ class CtrlProxy : AccessibilityService(), CtrlProxyActions {
     }
 
     return if (!allWindows.isNullOrEmpty()) {
+      Log.d(
+        TAG,
+        "extractHierarchy from ${allWindows.size} windows " +
+          "(disableAllFiltering: $disableAllFiltering, occlusionEnabled: $occlusionEnabled)",
+      )
       viewHierarchyExtractor.extractFromAllWindows(
         allWindows,
         rootNode,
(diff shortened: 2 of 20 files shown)


kaeawc (commented):
Reviewed against latest `origin/main` (head `77cf9b1`). The end-to-end wiring is complete and correct: `occlusionEnabled` threads CLI → `ServerConfig` → daemon relay (`manager.ts`/`parseDaemonArgs`) → `set_accessibility_flags` → both `extractFromAllWindows` call sites in `CtrlProxy.kt` (1903 & 2047) → `isOcclusionFilteringActive`. All `CtrlProxyActions` implementors updated; TS tests pass locally (46/46), default behavior preserved (opt-out only). Graceful cross-version behavior confirmed — `ignoreUnknownKeys = true` means an old APK drops the new field and a new APK defaults it, so no decode crash either direction.

One delivery caveat to confirm before relying on the flag in prod (inline). Two non-blocking notes: (1) on a fresh connect both `onConnectionEstablished()` and the new `ensureConnected()` override call `syncAccessibilityFlagsToDevice()`, double-sending the message — idempotent and acknowledged in the doc comment, harmless. (2) The tests prove the wiring but not a real-device before/after occlusion diff; worth one manual `observe` on a multi-window screen (e.g. shade + app) with/without `--no-occlusion` to confirm occluded nodes actually reappear.

android/protocol/src/main/kotlin/dev/jasonpearson/automobile/protocol/WebSocketRequest.kt
```diff
@@ -401,6 +401,7 @@ data class SetAccessibilityFlags(
   val includeNotImportantViews: Boolean = true,
   val reportViewIds: Boolean = true,
   val retrieveInteractiveWindows: Boolean = true,
+  val occlusionEnabled: Boolean = true,
```
kaeawc, line 404: **Verdict: Risky — feature undeliverable until the Android runner release is re-cut.**

The control-proxy APK is delivered to end users as a *pinned, checksum-verified download*, not built from the tree — `AndroidCtrlProxyManager` resolves `AUTOMOBILE_VERSION` against the `apkSha256` registry in `src/constants/release.ts`. This PR changes the APK's wire protocol (this new `occlusionEnabled` field) and `ViewHierarchyExtractor`, but does **not** re-cut the release or bump `release.ts`.

Consequence on a default install: the TS layer sends `set_accessibility_flags{...,occlusionEnabled:false}`, the currently-released APK ignores the unknown key (`ignoreUnknownKeys = true`), and occlusion stays on — `--no-occlusion` is a **silent no-op**. Local/CI runs escape this because `scripts/android/run-emulator-tests.sh` points `AUTOMOBILE_CTRL_PROXY_APK_PATH` at a freshly-built APK, so tests will pass while shipped users see nothing change.

This matches the repo's convention (runner checksum bumps ride periodic version-bump PRs), so it may be an intended follow-up — but there's no CI drift guard to catch it, so flagging explicitly. **Verify first:** cut/point-to a new APK, then confirm on a real multi-window screen that occluded nodes reappear with the flag set.

RyeBread1026: That's ok for our case right now - this hasn't been strictly necessary for our case to pass. It'll be useful when released, since it's turning out that trimming observe & other logging whenever possible is either very helpful or necessary to keep things running smoothly in CI.

kaeawc:

**Result: PASS** — occlusion fields populated.

`observe` returns all three properties carrying real values, on 7 of 60 nodes:

```json
{"occlusionState":"partial","occludedBy":"unlabeled view","occludedByViewId":"s-fe187b8620f037f3"}
```

Device-side runner log confirms the pass ran: `D CtrlProxy: Extracting from 2 windows (disableAllFiltering: false, occlusionEnabled: true)`.

Worth flagging for release: the APK installed on my emulator was `0.0.39-SNAPSHOT` (built Jul 6, older than the v0.0.44 tag), and these fields were absent until I re-cut and reinstalled the ctrlproxy APK. This feature is runner-gated, so it does not reach users until the APK is re-cut.

Issue #3846, opened 2026-07-13
Daemon silently loses all its startup flags mid-session when a bare CLI client connects


RyeBread1026:
A properly-configured, correctly-flagged daemon can get silently downgraded to bare defaults partway through a long-running session, with no crash and no visible restart banner in normal-level logs.

Root cause, confirmed via a real CI run (self-check before/after: 9/9 configured flags active, then 0/9 active, same daemon lifetime, no crash in between): `DaemonMcpProxy.ensureStartupOptionsMatch()` in `src/daemon/daemonMcpProxy.ts` compared the connecting client's requested `DaemonOptions` against the running daemon's options two-directionally. A short-lived CLI client that never passes any daemon flags (e.g. one spawned fresh per tool call, with no opinion on flags at all) reads as "this client wants every unset flag turned off" rather than "this client has no opinion" — so the moment such a client connects, the proxy sees a mismatch and restarts the shared daemon down to bare defaults, discarding whatever flags it was actually launched with.

A related contributing factor: `ensureVersionMatches`/`ensureBuildMatches` also restarted the daemon using the connecting client's own `daemonOptions` rather than the currently-running daemon's actual options, which has the same downgrade effect whenever either of those checks fires a restart for an unrelated (version/build) reason.

Expected: reconciliation should be one-directional — restart only to gain a flag a client explicitly requests that the daemon lacks (matching the original #2759 intent), never to strip a flag the daemon already has just because a particular caller didn't ask for it. A version/build-triggered restart should preserve the daemon's existing options rather than resetting them to the connecting client's config.

Impact: any long QA-agent/CI session using a fresh short-lived CLI process per tool call can silently run the bulk of a suite against an unconfigured daemon (wrong perf-mode, wrong output-reduction settings, etc.) with no error surfaced anywhere.

Issue #3845, opened 2026-07-13
AUTOMOBILE_LOG_LEVEL is parsed and unit-tested but never actually applied to the running process


RyeBread1026:
`parseAutomobileLogLevel()` (src/utils/logger.ts) parses `AUTOMOBILE_LOG_LEVEL` (debug/info/warn/error/none) and is unit-tested, but nothing in `src/` ever calls it against `process.env.AUTOMOBILE_LOG_LEVEL` at startup. Setting the env var currently does nothing — the daemon silently stays at the default INFO level regardless.

Discovered while trying to get debug-level daemon logs out of a CI job to diagnose an unrelated intermittent failure: the env var was set, but daemon.log never contained a single DEBUG line.

Expected: setting `AUTOMOBILE_LOG_LEVEL=debug` (or any other supported level) actually changes what the logger emits, verified by starting a real daemon with the env var set and confirming DEBUG-level lines appear in the real daemon.log.

PR #3849, opened 2026-07-13
Fix Compose testTag elementId matching (bare-id fallback)

### LINKED ISSUE #3852: elementId matching misses Compose testTag nodes when queried with the fully-qualified pkg:id/name form
## Summary

\`tapOn\` (and any other resource-id-based selector) fails with "Element not found" when targeting a Compose-based screen element, even though the element is present, visible, clickable, and at the expected bounds in the raw accessibility tree at that exact moment.

## Root cause

Jetpack Compose's \`Modifier.testTag(...)\` semantics surface via \`AccessibilityNodeInfo.viewIdResourceName\` **without a package qualifier** - just the bare tag name (e.g. \`personDetailsSpeedDial_Main\`), unlike traditional Android View resource IDs, which are always reported fully-qualified (\`pkg:id/name\`).

\`DefaultElementFinder\`'s resource-id matching (\`collectResourceIdMatchesInRoots\` in \`src/features/utility/ElementFinder.ts\`, and the equivalent in \`findClickableSiblingsOfResourceId\`) does a strict \`===\` equality check by default (\`partialMatch: false\`). A caller naturally passes the fully-qualified form - that's what every other element's resource-id looks like in the same hierarchy - so it never matches a Compose-sourced node reporting only the bare tag.

## Reproduction (from a real qa-agent run, 2026-07-13)

A FAB with \`resource-id: "personDetailsSpeedDial_Main"\` (confirmed via raw \`ViewHierarchyExtractor\` logcat output: \`clickable=true, focusable=true, interactive=true\`, valid bounds) failed every tap attempt using \`elementId: "com.followupboss.fubandroidstaging:id/personDetailsSpeedDial_Main"\`:

\`\`\`
tapOn result: success=false, error=Failed to perform tap on element: Element not found with provided elementId 'com.followupboss.fubandroidstaging:id/personDetailsSpeedDial_Main'
\`\`\`

The exact same element, in an earlier session, succeeded only once the agent (by trial and error) dropped the package prefix and queried the bare \`personDetailsSpeedDial_Main\` directly.

## Fix

Added a fallback: when the query is fully-qualified (\`pkg:id/name\`), also try an *exact* match against the bare suffix (the part after the last \`/\`). This is not a fuzzy/partial match - the bare name is the node's real, exact reported ID for Compose-sourced elements - so it applies unconditionally, not gated behind \`partialMatch\`.

Fixed in both:
- \`collectResourceIdMatchesInRoots\` (backs \`selectByResourceId\` / plain \`elementId\` tapOn)
- \`findClickableSiblingsOfResourceId\` (backs \`sibling: true\` + \`elementId\` tapOn)

Tests added in \`test/features/utility/ElementFinder.test.ts\` covering both paths.

RyeBread1026:
Fixes #3852.

Rescoped: #3845 and #3846 (the original two fixes this PR carried) were independently fixed and merged to `main` as #3855 and #3854 respectively - verified functionally equivalent-or-superset (main's versions are a strict superset: `mergeDaemonOptions` additionally force-preserves reuse-critical flags against an explicit client `false`, and the log-level seed happens at `logger.ts` module-load rather than only in `main()`, covering every entry point). Rebased this branch onto current `main` and dropped both now-redundant fixes, keeping only the genuinely independent one below.

**#3852 — Compose testTag elementId matching**: Jetpack Compose's `Modifier.testTag(...)` + `testTagsAsResourceId = true` surfaces via `AccessibilityNodeInfo.viewIdResourceName` *without* a package qualifier - just the bare tag - unlike traditional View resource IDs, which are always `pkg:id/name`. Resource-id matching did a strict `===` check by default, so a caller passing the fully-qualified form (the natural choice, since every other element's resource-id in the same hierarchy looks like that) never matched a Compose-sourced node, even though the element was on screen, clickable, with valid bounds.

Root-caused via a real qa-agent run where a FAB (`personDetailsSpeedDial_Main`) failed every `tapOn` with "Element not found" for exactly this reason.

Fix: when the query is fully-qualified, also try an exact match against the bare suffix (not a fuzzy/partial match - the bare name is the node's real, exact reported ID for Compose-sourced elements). Fixed in both `collectResourceIdMatchesInRoots` (plain `elementId` tapOn) and `findClickableSiblingsOfResourceId` (`sibling: true` + `elementId` tapOn). Tests added for both paths.

Full test suite green (7881/7881), typecheck clean.

### DIFF
--- a/src/features/utility/ElementFinder.ts
+++ b/src/features/utility/ElementFinder.ts
@@ -216,6 +216,14 @@ export class DefaultElementFinder implements ElementFinder {
     sortByArea: boolean = true
   ): Element[] {
     const matches: Element[] = [];
+    // Compose semantics (Modifier.testTag) surface via AccessibilityNodeInfo.viewIdResourceName
+    // WITHOUT a package qualifier, unlike traditional View resource IDs which are always
+    // reported as "pkg:id/name". A caller passing the fully-qualified form (the common case,
+    // since that's what every other resource ID in the hierarchy looks like) would otherwise
+    // never match a Compose-sourced node. This is not a fuzzy/partial match - the bare name is
+    // the node's real, exact reported ID - so it applies regardless of the partialMatch flag.
+    const idSeparatorIndex = resourceId.lastIndexOf("/");
+    const bareResourceId = idSeparatorIndex >= 0 ? resourceId.slice(idSeparatorIndex + 1) : null;
 
     for (const searchNode of rootNodes) {
       this.parser.traverseNode(searchNode, (node: any) => {
@@ -224,6 +232,7 @@ export class DefaultElementFinder implements ElementFinder {
           const nodeResourceId = nodeProperties["resource-id"];
           if (
             nodeResourceId === resourceId ||
+            (bareResourceId !== null && nodeResourceId === bareResourceId) ||
             (partialMatch && nodeResourceId.toLowerCase().includes(resourceId.toLowerCase()))
           ) {
             const parsedNode = this.parser.parseNodeBounds(node);
@@ -988,11 +997,18 @@ export class DefaultElementFinder implements ElementFinder {
       return [];
     }
 
+    // See collectResourceIdMatchesInRoots: Compose testTag nodes report a bare
+    // viewIdResourceName with no package qualifier, so a fully-qualified query must also
+    // match against the bare suffix - not just partialMatch/exact on the full string.
+    const idSeparatorIndex = resourceId.lastIndexOf("/");
+    const bareResourceId = idSeparatorIndex >= 0 ? resourceId.slice(idSeparatorIndex + 1) : null;
+
     const matchesId = (input?: string): boolean => {
       if (!input) {return false;}
-      return partialMatch
-        ? input.toLowerCase().includes(resourceId.toLowerCase())
-        : input === resourceId;
+      if (partialMatch) {
+        return input.toLowerCase().includes(resourceId.toLowerCase());
+      }
+      return input === resourceId || (bareResourceId !== null && input === bareResourceId);
     };
 
     const containerNode = container
--- a/test/features/utility/ElementFinder.test.ts
+++ b/test/features/utility/ElementFinder.test.ts
@@ -127,6 +127,24 @@ describe("DefaultElementFinder", () => {
       const results = finder.findElementsByResourceId(hierarchy, "btn_login", null, false);
       expect(results).toHaveLength(0);
     });
+
+    test("matches a bare (Compose testTag) node resource-id against a fully-qualified query", () => {
+      // Compose's Modifier.testTag surfaces via viewIdResourceName WITHOUT a package
+      // qualifier, unlike traditional View resource IDs. Callers always pass the
+      // fully-qualified form (it's what every other element in the hierarchy looks like),
+      // so this must match even with partialMatch disabled — it's the node's real ID, not
+      // a fuzzy substring hit.
+      const hierarchy = makeHierarchy({
+        $: { "resource-id": "personDetailsSpeedDial_Main", "bounds": { left: 0, top: 0, right: 100, bottom: 50 } },
+      });
+      const results = finder.findElementsByResourceId(
+        hierarchy,
+        "com.followupboss.fubandroidstaging:id/personDetailsSpeedDial_Main",
+        null,
+        false
+      );
+      expect(results).toHaveLength(1);
+    });
   });
 
   describe("findElementByResourceId", () => {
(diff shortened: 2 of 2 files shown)


src/daemon/daemonMcpProxy.ts
```diff
@@ -499,7 +523,15 @@ export class DaemonMcpProxy {
...
-    await this.daemonManager.restart(this.config.daemonOptions ?? {});
+    // Preserve the RUNNING daemon's options, not this connecting client's own
+    // (often bare-default) config — this restart exists to fix which CODE is
+    // running, not to reconfigure behavior flags. A short-lived CLI client that
+    // never passes output-reduction/perf flags (e.g. am.py, which intentionally
+    // omits them on every call — see qa-agent/am.py run_cli) would otherwise
+    // silently strip every flag the daemon was actually launched with the moment
+    // any such client connects (issue: daemon silently loses all startup flags
+    // mid-suite once a version-mismatch restart fires).
+    await this.daemonManager.restart(status.options ?? this.config.daemonOptions ?? {});
```
kaeawc: **Partial — the sibling restart path in `ensureStartupOptionsMatch` still strips flags.** You correctly preserve `status.options` here (and in `ensureBuildMatches`), but the options-*mismatch* restart at line 693 still calls `restart(this.config.daemonOptions ?? {})`. When a client requests one flag the daemon lacks while the daemon holds another the client didn't request, `startupOptionMismatches` (now one-directional) fires on the wanted flag, the restart drops the daemon's existing flag, and the post-restart `remaining` check — also one-directional — passes silently. That reproduces the exact silent-flag-loss class #3846 is about. It's pre-existing (line untouched since #2597), but it violates the invariant this PR documents ("never STRIP a flag the daemon already has") and is inconsistent with the two paths you just fixed. Suggest `restart({ ...status.options, ...(this.config.daemonOptions ?? {}) })` at line 693 (reuse the `status` already fetched at 673). All keys are enable-only booleans so the union is safe.

src/features/utility/ElementFinder.ts
```diff
@@ -216,6 +216,14 @@ export class DefaultElementFinder implements ElementFinder {
...
   ): Element[] {
     const matches: Element[] = [];
+    // Compose semantics (Modifier.testTag) surface via AccessibilityNodeInfo.viewIdResourceName
+    // WITHOUT a package qualifier, unlike traditional View resource IDs which are always
+    // reported as "pkg:id/name". A caller passing the fully-qualified form (the common case,
+    // since that's what every other resource ID in the hierarchy looks like) would otherwise
+    // never match a Compose-sourced node. This is not a fuzzy/partial match - the bare name is
+    // the node's real, exact reported ID - so it applies regardless of the partialMatch flag.
+    const idSeparatorIndex = resourceId.lastIndexOf("/");
+    const bareResourceId = idSeparatorIndex >= 0 ? resourceId.slice(idSeparatorIndex + 1) : null;
```
kaeawc, line 226: **Scope / traceability.** This Compose-`testTag` bare-resource-id matching is a third, independent behavior change, but the PR body says "Two small, independent fixes" (#3845, #3846) and never mentions it — no linked issue. The change itself looks correct and is tested (bare-suffix match only triggers for genuinely package-less nodes, since traditional IDs are always `pkg:id/name`), but bundling an untracked interaction-layer change into a daemon/config PR widens the blast radius and muddies `git blame`. Recommend splitting it into its own issue/PR, or at minimum call it out in the description.

kaeawc (commented):
Reviewed against latest `origin/main`. The two described fixes are correct and reach the daemon (the log-level fix applies in `main()`, which the daemon also enters via `--daemon-mode`; `status.options` is populated from the pidfile so the version/build restart preservation works). Changed tests pass locally (103/103, 110ms). Two things to weigh: an inconsistency where a *third* restart path still strips daemon flags, and an undescribed third change bundled in.

src/daemon/daemonMcpProxy.ts
RyeBread1026: Good catch, thanks. `ensureStartupOptionsMatch`'s own restart was the exact same bug, just via a third path. Fixed in 53ca9bcb: `restart({ ...status.options, ...(this.config.daemonOptions?? {}) })`, merging the daemon's current options with the newly-requested ones instead of replacing them. Added a test reproducing the scenario you described (daemon has one flag on, client asks for a different one, restart must preserve both).

src/features/utility/ElementFinder.ts
RyeBread1026: Totally - since it's small, I just added it here & updated the description to reference the issue.

kaeawc:

**Result: PASS** — Compose `testTag` lookup by fully-qualified id.

The playground exposes Compose nodes with a bare `viewIdResourceName` and no `pkg:id/` prefix, e.g. `resource-id: checkbox_control`. Querying with the fully-qualified form resolved it:

```
tapOn --selector '{"elementId":"dev.jasonpearson.automobile.playground:id/checkbox_control"}'
  -> success: true, element resolved: checkbox_control
```

Note for anyone reading the tap as an end-to-end assertion: the checkbox label did not change after the tap, but a raw `adb shell input tap` at the same node center did not change it either, so that is fixture behavior and not a regression in this PR. The claim under test here — the finder matching a bare-id Compose node from a fully-qualified query — holds.

PR #3854, opened 2026-07-14
fix(daemon): preserve daemon flags when a bare client connects (#3846)

### LINKED ISSUE #3846: Daemon silently loses all its startup flags mid-session when a bare CLI client connects

kaeawc:
Closes #3846.

## Problem

A properly-flagged daemon could be silently downgraded to bare defaults mid-session, with no crash and no visible restart banner. Root cause: option reconciliation in `DaemonMcpProxy` was **bidirectional**. A short-lived CLI client that passes no daemon flags (e.g. one spawned fresh per tool call) read as "this client wants every unset flag turned **off**" rather than "this client has no opinion" — so the moment such a client connected, `ensureStartupOptionsMatch()` saw a mismatch and restarted the shared daemon down to bare defaults, discarding its actual flags.

`ensureVersionMatches()` / `ensureBuildMatches()` had the same effect: they restarted using the *connecting client's* `daemonOptions` rather than the *running daemon's* options, so any version/build-triggered restart also stripped the daemon's flags.

Impact: a long QA-agent/CI session using a fresh short-lived CLI process per tool call could silently run the bulk of a suite against an unconfigured daemon (wrong perf-mode, wrong output-reduction settings) with no error surfaced.

## Fix — one-directional reconciliation

- **`startupOptionDeficits()`** (replaces `startupOptionMismatches()`) reports only reuse-critical flags the client **explicitly requests** that the daemon lacks (`want && !have` for booleans; strings only when the client supplies a differing value). A client with no opinion never manufactures a deficit.
- **`mergeDaemonOptions(running, requested)`** preserves the running daemon's existing options as the base and overlays only the client's requested flags, force-preserving reuse-critical flags the daemon already has so an explicit client `false` can't turn them back off. Applied at **all three** restart sites (version, build, startup-option), so no restart path can reintroduce the downgrade. The cold-start path (`startDaemon`, guarded by `!status.running`) correctly stays bare — there is no running daemon to preserve.

Restart only to **gain** a flag a client requests, never to **strip** one the daemon already has (matching the original #2759 intent).

## Tests

Three new unit tests in `test/daemon/daemonMcpProxy.test.ts`:
- Bare client connecting to a configured daemon does **not** restart (no downgrade).
- A deficit-driven restart **preserves** the daemon's other flags while adding the requested one.
- A version-mismatch restart **preserves** the running daemon's flags for a bare client.

All 54 tests in the suite pass; typecheck gate clean (no new baseline errors); lint + build green.

### DIFF
--- a/src/daemon/daemonMcpProxy.ts
+++ b/src/daemon/daemonMcpProxy.ts
@@ -224,33 +224,76 @@ const REUSE_CRITICAL_STRING_OPTION_KEYS: (keyof DaemonOptions)[] = [
   "toolOutputsDir",
 ];
 
+/** The value of a startup option when it is a string, else undefined. */
+function stringOption(
+  options: DaemonOptions | undefined,
+  key: keyof DaemonOptions
+): string | undefined {
+  const value = options?.[key];
+  return typeof value === "string" ? value : undefined;
+}
+
 /**
- * Boolean-valued startup options that differ between the requested MCP config
- * and the running daemon. Each option is compared as a strict boolean
- * (`=== true`), so `undefined` and `false` are treated identically. Returns a
- * human-readable list (empty when everything matches) for logging and error
- * messages.
+ * Reuse-critical startup options the connecting client *explicitly requests*
+ * that the running daemon lacks. Reconciliation is **one-directional** (issue
+ * #3846): a client that does not ask for a flag has no opinion on it, so a flag
+ * the daemon already has is never reported as a deficit just because a
+ * particular caller (e.g. a bare short-lived CLI client) didn't request it.
+ * Booleans are compared strictly (`=== true`), so `undefined` and `false` both
+ * read as "no opinion"; strings count only when the client supplies one that
+ * differs from the daemon's. Returns a human-readable list (empty when the
+ * daemon already satisfies every requested flag) for logging and error messages.
  */
-function startupOptionMismatches(
+function startupOptionDeficits(
   requested: DaemonOptions | undefined,
   running: DaemonOptions | undefined
 ): string[] {
-  const mismatches: string[] = [];
+  const deficits: string[] = [];
   for (const key of REUSE_CRITICAL_OPTION_KEYS) {
     const want = requested?.[key] === true;
     const have = running?.[key] === true;
-    if (want !== have) {
-      mismatches.push(`${key} (requested=${want}, running=${have})`);
+    if (want && !have) {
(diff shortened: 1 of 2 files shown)


kaeawc:

**Result: PASS** — daemon startup flags survive a bare client connect.

Controlled repro, daemon started with `--debug --embedded-sdk --network-mockable`:

| Step | pid | flags |
|---|---|---|
| before | 4518 | `--daemon-mode --debug --embedded-sdk --network-mockable` |
| after bare `--cli` client | 4518 | unchanged |
| after bare stdio MCP client | 4518 | unchanged |

Same process throughout — no restart, no reset to bare defaults.

One note in case it saves someone a false alarm: I did see a bare client restart the daemon once, matching this issue’s description exactly. It was not reproducible, and was build-skew from a dist I had rebuilt mid-session, not the flag-reset bug this PR fixed.

PR #3870, opened 2026-07-14
feat(inputText): opt-in marker-based auto-promotion to `eventAll`

jstewart5000:
## Summary

Adds opt-in auto-promotion of the `inputText` tool from the default atomic `a11y` text-set mode to `eventAll` (real per-character key events) when the input text contains any consumer-configured marker character. Off by default.

### Why
- Default `a11y` mode sets text in a single `ACTION_SET_TEXT`, emitting no keystrokes, so apps that only react to real key events (Slack `@mention` / `/command` / `:emoji:` autocomplete popups) never open their suggestion list — a typed mention stays unresolved plain text.
- `eventAll` already fixes this by injecting per-character key events, but it was only reachable by explicitly passing `mode`, which a caller driving the app has no reason to know.

### What Changed
- `resolveAutoInputMode` (new): pure resolver returning `"eventAll"` when the text contains any configured marker, else `undefined`.
- `InputText.execute`: `mode` is now optional; resolution is `explicit mode → marker auto-promotion → "a11y"`, so an explicit mode always wins and unconfigured markers preserve prior behavior. Applied at the shared choke point, so `setUIState` text fields honor it too. Logs the promotion (debug) and now logs the caught error before returning a typed failure.
- `ServerConfig`: `eventAllMarkers` state (empty = disabled); getter/setter use `readonly string[]`.
- Config injection (off by default): `--event-all-markers=<csv>` CLI flag or `AUTOMOBILE_EVENT_ALL_MARKERS` env var, parsed via a shared `cliArgs.firstFlagValue` helper (handles both `--flag value` and `--flag=value`; `toolOutputArtifacts` reuses it, replacing its duplicate). Threaded through `index.ts` and round-tripped through the daemon manager (`withDaemonOptions` serialize + `parseDaemonArgs` parse) so the flag reaches an auto-started daemon, mirroring `--dismiss-keyboard-after-input`.
- Startup logs the configured markers, and warns when the flag is present but resolves to none.

### Notes
- Android-only (iOS ignores `mode`). Marker matching is literal substring containment.

## Test Plan
- Unit tests: `resolveAutoInputMode` (empty/present/absent markers, multi-char, Slack marker set), `cliArgs.firstFlagValue` (both flag forms, missing/empty values), `parseEventAllMarkersConfig` (CLI vs env precedence, trimming), and daemon-arg relay for `eventAllMarkers`.
- Existing `InputText` suite still passes.
- `bun run typecheck`, lint, and build all green.
- Verified end-to-end on an Android emulator through the MCP and the local test harness: `inputText("/msg @Name hi")` resolves to `method: "eventAll"` and renders a linkified mention pill; with markers unconfigured the same input stays `a11y` (default unchanged).

### DIFF
--- a/src/daemon/manager.ts
+++ b/src/daemon/manager.ts
@@ -12,6 +12,7 @@ import {
 } from "../db/migrationDependencyIntegrity";
 import { ensureSecureTempDirSync, TEMP_SUBDIRS } from "../utils/tempDir";
 import { outputReductionFlagsToArgs } from "../utils/outputReductionFlags";
+import { EVENT_ALL_MARKERS_FLAG, splitMarkers } from "../utils/eventAllMarkers";
 import { shouldSkipCtrlProxyDownload } from "../utils/ctrlProxyDownloadControl";
 import { ActionableError } from "../models";
 import {
@@ -740,6 +741,9 @@ export class DaemonManager implements DaemonManagerLike {
     if (options.dismissKeyboardAfterInput) {
       args.push("--dismiss-keyboard-after-input");
     }
+    if (options.eventAllMarkers && options.eventAllMarkers.length > 0) {
+      args.push(EVENT_ALL_MARKERS_FLAG, options.eventAllMarkers.join(","));
+    }
     if (options.noUiPerfMode) {
       args.push("--no-ui-perf-mode");
     }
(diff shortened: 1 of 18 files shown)


src/daemon/manager.ts
```diff
@@ -1240,6 +1244,12 @@ export function parseDaemonArgs(args: string[], env: NodeJS.ProcessEnv = process
       options.embeddedSdk = true;
     } else if (args[i] === "--dismiss-keyboard-after-input") {
       options.dismissKeyboardAfterInput = true;
+    } else if (args[i] === EVENT_ALL_MARKERS_FLAG) {
```
kaeawc: This introduces a bug when the daemon is started or restarted with event-all markers configured through the daemon command path. AutoMobile’s daemon restarts in some environments, and daemon flags are expected to survive that handoff. Here, parseDaemonArgs only handles `--event-all-markers <csv>`, so `--event-all-markers=<csv>` and `AUTOMOBILE_EVENT_ALL_MARKERS` are ignored and marker promotion silently stays disabled. The fix is to reuse `parseEventAllMarkersConfig(args, env)` here, or route this branch through `firstFlagValue`, so daemon command parsing follows the same CLI/env contract as the main startup parser.

kaeawc (commented):
Reviewed the PR branch against latest origin/main. I found one daemon-management parser gap; targeted tests and build pass locally after bootstrapping dependencies.

src/daemon/manager.ts
jstewart5000: Nice find and thanks for fixing!

kaeawc (approved):

PR #3907, opened 2026-07-17
feat(video): allow long iOS recordings past the 300s cap (no segmentation)

### LINKED ISSUE #3906: Enable long iOS recordings: raise the 300s manager cap for iOS (auto-stop enforces maxDuration) — no segmentation needed
### Summary

The `videoRecording` schema now accepts `maxDuration` up to 3600s (raised from 300 in #3847 so long *Android* recordings, which are segmented under the hood, validate). But only Android has segmentation. For **iOS (or any non-Android)** platform, a request with `181 <= maxDuration <= 3600` passes schema validation and then fails at runtime in the manager:

`src/server/videoRecordingManager.ts:233-234`
```
if (value > MAX_DURATION_SECONDS) {  // MAX_DURATION_SECONDS = 300
  throw new ActionableError(`maxDuration must be <= ${MAX_DURATION_SECONDS} seconds.`);
}
```

So the reachable failure mode for a non-Android caller asking for e.g. `maxDuration: 500` is a late runtime `ActionableError`, not a schema rejection.

### Status

The #3847 author **explicitly accepted** this as the intended behavior for now — the runtime error is a reasonable failure mode, and closing the gap is out of scope for that PR (see the inline schema comment at `src/server/videoRecordingTools.ts` on the `.max(3600)` field). This issue just tracks the enhancement so the divergence is captured rather than lost.

### Options (pick when prioritized)

1. **Enforce per-platform in the schema** — a conditional `if platform === "ios" then maxDuration <= 300` (the flattener already emits JSON-Schema `if/then`, e.g. `postNotification`), so an over-cap iOS request is rejected at validation with a clear message instead of at runtime.
2. **Segment iOS too** — extend the segmentation approach to iOS `simctl`/CoreMediaIO capture so long iOS recordings are actually supported. Larger effort; overlaps with #2504 (physical-iOS capture) and the iOS `moov`-atom flush constraints.

Option 1 is the low-cost correctness fix (make the runtime rejection a validation rejection); Option 2 is the real feature. They're independent.

Surfaced during review of #3847.

kaeawc:
Fixes #3906. Follow-up from the review of #3847.

## Why

iOS `xcrun simctl io … recordVideo` has **no `--time-limit`** (unlike Android `screenrecord`, which hard-caps at 180s) — it records continuously until SIGINT flushes the moov atom. So iOS never needed segmentation to record long. The only thing blocking a >300s iOS recording was the manager's arbitrary `MAX_DURATION_SECONDS = 300` guard, which `resolveMaxDurationSeconds` applied to **every** platform. Before this change, an iOS `maxDuration` in `(300, 3600]` passed schema validation and then failed at runtime with `maxDuration must be <= 300 seconds`.

(This supersedes the earlier "segment iOS too" framing — segmentation exists solely to work around the Android cap; building it for iOS would fragment a continuous recording for no platform benefit.)

## What

- `resolveMaxDurationSeconds(value, platform)` is now platform-aware: iOS is bounded by the new `IOS_MAX_DURATION_SECONDS` (3600, matching the schema ceiling), non-iOS keeps 300s.
- The existing `scheduleAutoStop` already stops-and-flushes at the requested duration, so a long iOS request yields **one continuous file**.
- **Android is unchanged** — the tool segments Android >180s before the manager sees it, so a direct Android recording never legitimately exceeds 300s.
- The `videoRecording` schema's `maxDuration.max()` now references `IOS_MAX_DURATION_SECONDS` (single source of truth; no generated-schema drift — value stays 3600).

## Tests

- Manager (`videoRecordingManager.test.ts`): iOS 500s accepted **and auto-stops at 500s**; iOS 3601s rejected; non-iOS 301s still rejected — each asserting the backend start/stop calls.
- Tool (`videoRecordingToolSegmented.test.ts`): iOS 500s start stays single/non-segmented and threads `maxDurationSeconds` through the response.
- Ran `/simplify` (4-agent pass): consolidated repeated rationale comments and de-duplicated a triplicated `iosDevice` test fixture.

Local: 17/17 in the two touched suites, lint clean, build clean.

### DIFF
--- a/src/server/videoRecordingManager.ts
+++ b/src/server/videoRecordingManager.ts
@@ -33,7 +33,13 @@ import { buildVideoArchiveItemUri, VIDEO_RESOURCE_URIS } from "./videoRecordingR
 import { VisualHighlightClient } from "../features/debug/VisualHighlight";
 
 const DEFAULT_MAX_DURATION_SECONDS = 30;
+// Per-platform single-recording caps chosen by resolveMaxDurationSeconds. Android >180s is
+// segmented before reaching the manager (see videoRecordingTools), so a direct Android
+// recording never legitimately exceeds MAX_DURATION_SECONDS; iOS `simctl recordVideo` has no
+// time limit and records continuously until SIGINT, so it only needs a higher cap. The
+// `videoRecording` schema's maxDuration ceiling mirrors IOS_MAX_DURATION_SECONDS.
 const MAX_DURATION_SECONDS = 300;
+export const IOS_MAX_DURATION_SECONDS = 3600;
 // Android uses HighlightAnimator's total fade-in + display + fade-out duration.
 // iOS SDK overlays auto-remove after their 3 second TTL.
 const ANDROID_HIGHLIGHT_ANIMATION_DURATION_MS = 6000;
@@ -223,15 +229,19 @@ function configToInput(config: VideoRecordingConfig): VideoRecordingConfigInput 
   };
 }
 
-function resolveMaxDurationSeconds(value?: number): number {
+function resolveMaxDurationSeconds(
+  value: number | undefined,
+  platform: BootedDevice["platform"]
+): number {
   if (value === undefined) {
     return DEFAULT_MAX_DURATION_SECONDS;
   }
   if (!Number.isFinite(value) || value <= 0) {
     throw new ActionableError("maxDuration must be a positive number of seconds.");
   }
-  if (value > MAX_DURATION_SECONDS) {
-    throw new ActionableError(`maxDuration must be <= ${MAX_DURATION_SECONDS} seconds.`);
+  const cap = platform === "ios" ? IOS_MAX_DURATION_SECONDS : MAX_DURATION_SECONDS;
+  if (value > cap) {
+    throw new ActionableError(`maxDuration must be <= ${cap} seconds.`);
   }
   return Math.round(value);
 }
@@ -549,7 +559,10 @@ export async function startVideoRecording(
   }
   const overrides = request.configOverrides ?? {};
   const configInput = await resolveConfigInput(overrides);
-  const maxDurationSeconds = resolveMaxDurationSeconds(request.maxDurationSeconds);
+  const maxDurationSeconds = resolveMaxDurationSeconds(
+    request.maxDurationSeconds,
+    request.device.platform
+  );
   const highlightInputs = request.highlights ?? [];
 
   for (const highlight of highlightInputs) {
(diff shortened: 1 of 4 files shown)


kaeawc:
Verified on `origin/main` (32b8651b2) during a manual-test iteration over `v0.0.44..main`, on iPhone 17 Pro / iOS 26.2 (`34C35F33`) with a freshly-built iOS runner.

**Result: PASS** — a >300s iOS recording is no longer aborted, and stays unsegmented.

Recorded for **335s** wall-clock, well past the old 300s manager cap:

| | |
|---|---|
| recordings returned | 1 |
| `segmented` / `segmentIndex` | `None` / `None` (unsegmented, as intended for iOS) |
| files in the session dir | exactly one `.mp4`, no segments |
| size | 13,291,418 bytes |
| ffprobe | `hevc, 1206x2622`, duration readable ⇒ moov atom present, playable |

The recording survived past 300s instead of being aborted, which is the claim under test.

**One observation that may deserve a follow-up**, not a failure of this PR: the container duration reads **239.3s** for a run whose `simctl io ... recordVideo` process was alive the full ~335s (confirmed via `ps` elapsed mid-run). So roughly 29% of wall-clock is missing from the output. My run had heavy concurrent load — another session driving the same machine, plus `xcodebuild` and repeated `observe` calls — so frame-dropping in `simctl recordVideo` is a plausible and boring explanation. I did not measure this in clean-room conditions, so I am flagging it rather than asserting a defect. Worth a controlled timing check if anyone is relying on recording duration being faithful.

kaeawc:
Verified on `origin/main` (92cd3baa5) during a manual-test iteration over `0.0.44..main`, on iPhone 17 Pro / iOS 26.2.

**Correction to my earlier comment on this PR.**

I previously flagged that a 335s recording produced a container reading 239.3s (~29% "missing") and suggested a controlled timing check in case frames were being dropped under load. I have now measured it properly, and **there is no defect** — `simctl io recordVideo` is change-driven, so wall-clock and container duration are not expected to match on an idle screen.

Two measurements on the same simulator make this unambiguous:

| screen state | wall-clock | container duration | size |
|---|---|---|---|
| static SpringBoard | ~12s | **0.067s** (≈1 frame) | 75 KB |
| continuous swiping in Settings | ~40s | **29.8s** | 4.5 MB |

A static screen yields essentially a single frame; a screen under constant motion tracks wall-clock closely. My original 335s recording sat idle for most of its life, which fully accounts for the gap.

So: recording duration is a function of screen activity, not a fidelity bug, and nothing here needs investigating. Apologies for the noise on an otherwise clean PR — the original PASS result for this change stands unchanged.

PR #3908, opened 2026-07-17
ci: bump kaeawc/setup-jemalloc v0.0.3 -> v0.0.5

kaeawc:
## What

Bumps the pinned `kaeawc/setup-jemalloc` action from `v0.0.3` to `v0.0.5` in `.github/actions/gradle-task-run/action.yml` (our only usage; applied on Linux gradle runs via `malloc-replacement: jemalloc`).

## Why

Not a cosmetic bump — **v0.0.4 fixed a real SIGSEGV (exit 139)** that triggers when the composite action is invoked **more than once in the same job**. The old relocate step overwrote a live `LD_PRELOAD`'d `libjemalloc.so.2` in place with `cp`; the install is now atomic (temp file + `rename(2)`) and idempotent.

**Scope of the crash fix:** today exactly one job hits the double-invocation case — the manual `record` job in `record-screenshot-baselines.yml`, which calls the composite action twice (steps at lines 50 and 75). Every other callsite invokes the action once per job, where the in-place-overwrite crash can't trigger. So the crash fix's blast radius is that single `workflow_dispatch` baseline job, not the build fleet — but bumping everywhere keeps the pin consistent and picks up the fix for that job.

v0.0.5 layers on Linux-safe dead-code removal and efficiency cleanup with no consumer-facing behavior change (inputs/outputs unchanged: single optional `jemalloc-version` input, no outputs; cache key unchanged).

## Validation

- `scripts/all_fast_validate_checks.sh`: 14/15 pass. The one failure is `lychee` (doc link-liveness), unrelated to a version-pin bump.
- No open auto-mobile issues reference jemalloc.
- Reviewed from supply-chain (tag-pinning is repo convention; first-party action), claim-accuracy (release notes verified), and CI-behavior lenses.

Refs: [setup-jemalloc v0.0.4](https://github.com/kaeawc/setup-jemalloc/releases/tag/v0.0.4) · [v0.0.5](https://github.com/kaeawc/setup-jemalloc/releases/tag/v0.0.5)

### DIFF
--- a/.github/actions/gradle-task-run/action.yml
+++ b/.github/actions/gradle-task-run/action.yml
@@ -85,7 +85,7 @@ runs:
 
     - name: "Set up jemalloc"
       if: ${{ inputs.malloc-replacement == 'jemalloc' }}
-      uses: kaeawc/setup-jemalloc@v0.0.3
+      uses: kaeawc/setup-jemalloc@v0.0.5
 
     - name: "Set up tcmalloc"
       if: ${{ inputs.malloc-replacement == 'tcmalloc' }}


.github/actions/gradle-task-run/action.yml
```diff
@@ -85,7 +85,7 @@ runs:
 
     - name: "Set up jemalloc"
       if: ${{ inputs.malloc-replacement == 'jemalloc' }}
-      uses: kaeawc/setup-jemalloc@v0.0.3
+      uses: kaeawc/setup-jemalloc@v0.0.5
```
kaeawc, line 88: The v0.0.4 SIGSEGV-on-second-invocation fix does apply to us, but the body's "exactly our per-gradle-task pattern" overstates the breadth. `malloc-replacement` defaults to `jemalloc` (line 39), but of the ~30 `gradle-task-run` callsites across the workflows, only one job invokes this composite action twice within a single job: `record` in `record-screenshot-baselines.yml` (steps at lines 50 and 75, `ubuntu-latest`, both on the default jemalloc). Every other job calls it once per job, where the in-place-overwrite bug can't trigger — and `record` is `workflow_dispatch`-only. So the real-world blast radius is one manual job, not the build fleet. Worth taking regardless; just flagging the framing. Not a blocker.

kaeawc (commented):
CI-behavior lens: the bump is safe and the motivation holds, with one scoping caveat. Verified v0.0.5's only input is `jemalloc-version` (has a default in both v0.0.3 and v0.0.5), so the no-`with:` usage needs no new required input, and `git grep setup-jemalloc` shows this is the sole reference — nothing left pinned to v0.0.3.

.github/actions/gradle-task-run/action.yml
kaeawc: Fixed the framing. Reworded the commit message and PR body to scope the crash fix to the one job that actually invokes the composite action twice per job — the manual `record` job in `record-screenshot-baselines.yml` (lines 50/75) — and note every other callsite is once-per-job where the in-place-overwrite can't trigger. Kept the bump everywhere for pin consistency.

PR #3912, opened 2026-07-17
feat(video): emit a segments.json manifest + grouping metadata for segmented recordings

### LINKED ISSUE #3905: Segmented videoRecording: emit segments into one shared directory (+ manifest) instead of per-recording subdirs
### Summary

Segmented Android recordings from the raw `videoRecording` tool (added in #3847) return each segment in its **own per-recording subdirectory** rather than as siblings in one shared folder. A caller that starts a >180s recording and stops it gets back N `filePath`s scattered across N directories (`archiveRoot/<outputName>-<recordingId>/…`, `archiveRoot/<outputName>-seg1-<recordingId>/…`, …) with no single containing directory and no manifest describing segment order.

### Context

This is a **pre-existing, explicitly-documented TODO** on the shared session class, now reachable from a second code path (the tool, not just `executePlan`):

`src/server/androidSegmentedPlanVideoSession.ts:36-40`
```
 * TODO: Each segment currently lands in its own subdirectory
 * (`archiveRoot/<outputName>-<recordingId>/`) because `VideoRecorderService` creates a
 * per-recording folder. Ideally all segments for a single plan run would be siblings in one
 * shared directory. See Option A (post-hoc move) or Option B (outputDirectory override on
 * StartVideoRecordingOptions) for approaches.
```

For `executePlan` this was tolerable — the plan run owns aggregation. But #3843's target caller is a QA-agent runner that drives `videoRecording --action start/stop` directly per test case and then has to locate and concatenate the clips itself. The stop response does return the ordered `recordings[]` (recordingId + filePath), so ordering is recoverable programmatically, but the on-disk layout is awkward: no shared directory to copy/upload as a unit, and the segment relationship is invisible to anything that only sees the archive tree.

### Suggested approach

Prefer **Option B** from the TODO — an `outputDirectory` override on `StartVideoRecordingOptions` so every segment of one session lands in a shared `archiveRoot/<sessionName>/` folder (`seg0.mp4`, `seg1.mp4`, …), optionally with a small `segments.json` manifest recording order + timing. Option A (post-hoc move on finalize) is simpler but races the archive size-eviction path and complicates the failure modes. Either way, keep the stop response's ordered `recordings[]` as the authoritative ordering.

### Not a regression

The current behavior is correct and shipped; this is a usability/enhancement follow-up. No data is lost — segments are all captured and ordered in the response.

Surfaced during review of #3847.

kaeawc:
Fixes #3905. Follow-up from the review of #3847.

## Problem

A >180s Android `videoRecording` is split into ordered segments, each in its own per-recording directory. A caller (e.g. a QA-agent runner driving `videoRecording --action start/stop` directly) got back N file paths across N directories with nothing tying them together on disk.

## Why manifest-only (not a shared directory)

The archive eviction/delete path does `rm(path.dirname(record.filePath), {recursive:true})` ([videoRecordingManager.ts:757-759](../blob/main/src/server/videoRecordingManager.ts#L757)) — one directory per recording. A **true shared directory** for all segments would break that: evicting/deleting one segment would `rm -rf` the whole session folder, destroying sibling segments whose DB rows still point at the deleted files (the [VideoRecorderService invariant](../blob/main/src/features/video/VideoRecorderService.ts#L154): *"folders must not be shared across recordings"*). So this keeps per-recording dirs and adds a manifest instead — zero change to the correctness-critical eviction model.

## What

- On segmented stop, write a best-effort **`segments.json`** (`sessionId`, `segmentCount`, ordered `segments[]`) into the **first segment's directory**, so the existing per-recording eviction cleans it up — no new eviction surface. A write failure is logged and the stop still succeeds (the response carries authoritative ordering).
- The stop response gains `manifestPath` (by-handle) / `manifestPaths` (bare by-device). Every segment row carries **`sessionId`** (the session handle = first segment's `recordingId`, reusing the existing registry handle — no new id primitive) + **`segmentIndex`**, so `recordings[]` has a uniform shape across both stop modes. The start response echoes `sessionId` for symmetry.

## Tests

- `segments.json` file content + on-disk location match the response and are in order (real-deps, single segment, under the temp archiveRoot — deterministic, no flake).
- Bare by-device stop: `segmentIndex`/`sessionId` ordering + `manifestPaths` surfaced.
- Fake segment paths now resolve under the per-test `archiveRoot` instead of a hardcoded `/tmp`, so the manifest write is valid + auto-cleaned on **all platforms including Windows**.
- `/simplify` (4-agent): dropped a redundant top-level `sessionId`, folded a near-duplicate test into the existing one, trimmed comments.

Local: 22/22 in the touched suites, lint clean, build clean.

## Out of scope / follow-up

The plan-driven (`executePlan`) path finalizes via `AndroidSegmentedPlanVideoSession.finalize()` directly (not through the tool's `stopAndRemove`) and does not get a manifest. Captured as a follow-up to lift the writer into a shared module and cover both paths (deferred here to keep #3905 tool-scoped and respect YAGNI until a second consumer exists).

### DIFF
--- a/src/server/videoRecordingTools.ts
+++ b/src/server/videoRecordingTools.ts
@@ -1,3 +1,5 @@
+import { promises as fsPromises } from "node:fs";
+import path from "node:path";
 import { z } from "zod";
 import { ToolRegistry } from "./toolRegistry";
 import {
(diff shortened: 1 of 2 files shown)


kaeawc:

**Result: PASS** — `segments.json` manifest written and well-formed.

Verified in both the single- and multi-segment cases.

Multi-segment (208s recording, 2 segments):

```json
{
 "sessionId": "7eb004ee-6753-4e24-9d93-9a02373ea968",
 "segmentCount": 2,
 "segments": [
  {"index": 0, "recordingId": "7eb004ee...", "filePath": ".../segtest2-7eb004ee-...mp4"},
  {"index": 1, "recordingId": "7d86b107...", "filePath": ".../segtest2-seg1-7d86b107-...mp4"}
 ]
}
```

Checks: indices ordered, `segmentCount` matches `segments[]` length, every entry carries `index`/`recordingId`/`filePath`, both referenced files exist on disk and are playable, and the manifest is written into the **first** segment’s directory (segment 1 lives in its own directory and is referenced by absolute path). A single-segment run produced the same shape with `segmentCount: 1`.

PR #3957, opened 2026-07-19
feat(lint): stage 1 declarative/functional style ratchet on src

kaeawc:
## What

Stage 1 of moving the TypeScript source toward more declarative style. Adds nesting/complexity caps and nudges away from building collections by mutation, enforced as **errors** with an `eslint-suppressions.json` baseline so only **new** code is gated (same ratchet model as the existing catch-convention / floating-promises guards).

Scoped to **`src/**` only** — tests and generated files are exempt.

| Rule | Threshold | Baseline |
|---|---|---|
| `complexity` | 15 | 178 |
| `max-depth` | 4 | 66 |
| `max-nested-callbacks` | 3 | **0** |
| `auto-mobile/no-accumulator-foreach` | — | 5 |

## Scope: a nudge, not a ban on iteration

**Every explicit loop form is allowed** — `for`, `for-of`, `for-in`, `while`. Loops stay the clearest tool for device I/O, retries, byte/image processing, and ordered async batches, and a rule that pushed people off them would make the code worse, not better.

The one shape flagged is a `.forEach()` whose every statement is a bare `push`/`unshift`/`add`/`set` — building a collection by mutation, where `.map()`/`.filter()`/`new Map()`/`new Set()` states the intent directly. A callback that logs, branches, awaits, or recurses is deliberately untouched: its only rewrite is a loop, which is *more* imperative and would defeat the rule.

Also **not** adopted: a `++`/`--` ban (240 violations, almost all legitimate loop counters).

## Why a custom rule instead of `no-restricted-syntax`

The first version folded these selectors into `no-restricted-syntax`. That was wrong, and review caught it.

ESLint bulk suppressions are keyed per file + per **rule** and store only a **count**. Sharing that bucket with the `setTimeout`/`setInterval`/`structuredContent` bans meant a baselined violation could be traded for a genuinely-dangerous one with no CI failure. Reproduced on disk in `src/server/resourceRegistry.ts` (budget 1): replacing the baselined `forEach` with a banned `setTimeout` lints **exit 0**.

`main` carries **zero** `no-restricted-syntax` suppressions — those guards are absolutely enforced today — so this would have been the first budget ever introduced into that bucket, across 15 files. A dedicated rule keeps the budgets separate; `no-restricted-syntax` is now byte-identical to `main` with zero suppressions, and a regression test guards it.

The dedicated rule also fixes a second defect: the `no-explicit-any` override for `ScreenFingerprint.ts`/`ExploreElementExtraction.ts` spreads `...baseRules`, which replaces the whole `no-restricted-syntax` entry and silently exempted exactly the two files the config calls correctness-sensitive. The new rule isn't part of `baseRules`, so it survives the spread.

## Burn-down no longer breaks the build

ESLint exits **2** when a suppression count exceeds actual violations, so de-nesting one `if` or extracting a helper from a baselined function would have turned CI red — the ratchet punishing the burn-down it exists to encourage, with the remedy (`--prune-suppressions`) having no script and no documentation anywhere in the repo.

`lint` now passes `--pass-on-unpruned-suppressions`, plus two scripts mirroring `typecheck`/`typecheck:update`:

```bash
bun run lint            # gate: fail on NEW violations
bun run lint:prune      # after FIXING violations: shrink the baseline
bun run lint:baseline   # after ADDING a ratchet rule: record existing violations
```

Verified: burn-down exits 0 (was 2); a new violation still exits 1.

## Also

- `max-nested-callbacks` 4 → **3**. At 4 it had no bite (deepest nest in `src` is 3); at 3 the baseline is still zero but it actually ratchets.
- Dropped a stale suppression for `src/features/video/NoopVideoCaptureBackend.ts`, a file that exists neither on disk nor in `main`.
- Documented the ratchet in `CLAUDE.md` beside the typecheck gate — including the count-fungibility caveat, the loops-are-allowed scope, and the non-auto-fixable-rules-only constraint.
- Added `test/lint/accumulatorForEachRule.test.ts` (6 tests): fire and no-fire cases, a guard that all four loop forms lint clean, out-of-src scoping, and the budget-leak regression.

## Verification

- `bun run lint` exits 0; `bun run build` exits 0; `bun test` **8086 pass / 0 fail**.
- `bun run typecheck` exits 0 (its "7 baseline errors no longer occur" notice reproduces on clean `main` — pre-existing, out of scope).
- `eslint.config.mjs` is purely additive: `baseRules` and the `SystemTimer` override are untouched vs `main`.
- This PR touches no `src/` file — it is pure policy plus docs and tests.

## Follow-ups

#3966 (burn down the baseline, tighten thresholds) and #3967 (evaluate `eslint-plugin-functional`, scoped to pure domains).


eslint.config.mjs
```diff
@@ -461,6 +483,23 @@ export default [
...
+		// loose to start (complexity 15, depth 4) and can be ratcheted down as the
+		// suppression baseline is burned down. Tests/generated files are exempt.
+		files: ["src/**/*.ts"],
+		plugins,
+		languageOptions,
+		rules: {
+			"max-depth": ["error", 4],
+			"complexity": ["error", 15],
+			"max-nested-callbacks": ["error", 4],
+			"no-restricted-syntax": [2, ...importAndTimerSelectors, ...functionalStyleSelectors],
```
kaeawc: The two navigation files never get these functional selectors. The later block at `eslint.config.mjs:541-551` spreads `...baseRules`, and `baseRules` now sets `no-restricted-syntax` to `[2, ...importAndTimerSelectors]` — that spread re-assigns the whole rule entry, so `ForInStatement` and `forEach` are dropped for `ScreenFingerprint.ts` and `ExploreElementExtraction.ts`. `max-depth`/`complexity`/`max-nested-callbacks` survive only because `baseRules` doesn't define them.

Same source, two filenames:

```
printf 'const o: Record<string, number> = {};\n[1].forEach(x => x);\nfor (const k in o) { void k; }\n' \
  | ./node_modules/.bin/eslint --no-fix --stdin --stdin-filename src/utils/logger.ts
# 2 errors: forEach + for-in

printf 'const o: Record<string, number> = {};\n[1].forEach(x => x);\nfor (const k in o) { void k; }\n' \
  | ./node_modules/.bin/eslint --no-fix --stdin --stdin-filename src/features/navigation/ScreenFingerprint.ts
# 0 errors
```

`eslint-suppressions.json` shows the same thing from the other direction: `ScreenFingerprint.ts` carries a `complexity` entry but no `no-restricted-syntax` entry.

Not a blocker — neither file currently uses `for-in` or `forEach`, so nothing is being missed today. It's a latent hole in exactly the two files the config calls out as correctness-sensitive, and it'll silently swallow new violations there.

Suggested change for keeping the functional bans in the navigation override:

```suggestion
			"no-restricted-syntax": [2, ...importAndTimerSelectors, ...functionalStyleSelectors],
```

(applied to the navigation block's `rules` at line 548, after the `...baseRules` spread — not to this line.)

kaeawc (commented):
Reviewed through one lens: flat-config merge semantics and scoping correctness, verified with `eslint --print-config` and `--stdin-filename` rather than reading the config.

Most of it holds up. `SystemTimer.ts` keeps its `setTimeout`/`setInterval` exemption and correctly picks up the functional bans (the block is last, so it wins on `no-restricted-syntax`). Extracting `importAndTimerSelectors` is byte-for-byte behavior-preserving off `src/` — I diffed the full resolved rule set for a test file, a benchmark file, and a script between `main` and this head, and all three are identical. The suppressions baseline corroborates the scoping: all 166 entries are under `src/`, zero outside.

One real gap on the two navigation files, inline below.

eslint.config.mjs
kaeawc: Folding the functional bans into `no-restricted-syntax` puts them in the same bulk-suppression bucket as the `setTimeout`/`setInterval`/`structuredContent` bans — and ESLint's suppression budget is per `{file, ruleId}`, not per selector. On `origin/main` `no-restricted-syntax` had **zero** suppression entries, so those three bans were absolutely enforced. This PR hands out 30 units of slack across 15 `src` files, including `src/server/index.ts` (3), `src/cli/index.ts` (4), `src/features/observe/ViewHierarchy.ts` (4), `src/features/observe/output/ObserveResultOutput.ts` (4).

Swap a baselined `forEach` for a raw `setTimeout()` in the same file and the count still matches, so nothing reports — the #2907 `structuredContent` dead-read guard and the `Timer` ban lose their teeth in exactly the files most likely to get refactored under this ratchet. (An outright *removal* is safe: unused suppressions exit 2. It's the one-for-one swap that's silent.)

Reproduction, with a two-selector `no-restricted-syntax` and a file containing one `forEach`:
```
eslint --suppress-all --suppressions-location supp.json f.ts   # records count: 1
# now replace the forEach with setTimeout(() => {}, 1)
eslint --suppressions-location supp.json f.ts                  # exit 0, silent
```

This should block. The repo already has the shape for the fix: `auto-mobile/catch-convention` and `auto-mobile/no-unknown-cast` are custom rules with their own suppression keys. A small `auto-mobile/no-imperative-iteration` rule carrying `ForInStatement` + the `forEach` selector would keep the 30 baselined sites in their own bucket and leave `no-restricted-syntax` at zero suppressions where it is today.

eslint.config.mjs
```diff
@@ -246,6 +246,55 @@ const plugins = {
...
+// banned because it hides accumulation side effects that read more clearly as
+// map/filter/reduce or a for-of. Loop counters (`++`/`--`) and `for-of` itself
+// are deliberately NOT banned — that churn (240+ sites) outweighs the benefit.
+const functionalStyleSelectors = [
+	{
+		selector: "ForInStatement",
+		message: "Avoid for-in (it walks the prototype chain and yields string keys). Use for-of over Object.keys()/Object.entries()/Object.values(), or a declarative map/filter/reduce.",
+	},
+	{
+		selector: "CallExpression[callee.property.name='forEach']",
```
kaeawc: The block at eslint.config.mjs:541 (`no-explicit-any` for `ScreenFingerprint.ts` / `ExploreElementExtraction.ts`) comes *after* the `src` override and spreads `...baseRules`, which sets `no-restricted-syntax` to `importAndTimerSelectors` only — so those two `src` files get no `for-in`/`forEach` ban at all. This is the exact drift the `importAndTimerSelectors` extraction was written to prevent, one block over; `SystemTimer.ts` was caught, these two weren't.

```
$ eslint --print-config src/features/navigation/ScreenFingerprint.ts
  selectors: [ImportDeclaration .js, ImportDeclaration .ts, setTimeout, setInterval, structuredContent]
$ eslint --print-config src/server/index.ts
  selectors: [... , ForInStatement, CallExpression[callee.property.name='forEach']]
```

`max-depth`/`complexity`/`max-nested-callbacks` survive fine (flat config merges the rule map, and `baseRules` doesn't define them) — it's only `no-restricted-syntax` that gets clobbered. Appending `...functionalStyleSelectors` to that block, the way SystemTimer does, closes it. Should block, since it silently contradicts the PR's "scoped to `src/**`" claim.

eslint.config.mjs
```diff
@@ -246,6 +246,55 @@ const plugins = {
...
+		message: "Avoid .forEach() — it hides accumulation side effects. Prefer a declarative .map()/.filter()/.reduce(), or a plain for-of when a side effect is genuinely intended.",
```
kaeawc: The message's own escape valve — "or a plain for-of when a side effect is genuinely intended" — is where most of the 26 baselined sites will land, and for-of + `push` is *more* imperative than what it replaced. Worth being explicit about whether that trade is intended.

Walking the actual sites: some genuinely improve, e.g. `NavigationGraphManager.ts:1401` `dbNodes.forEach(node => nodeIdMap.set(node.screen_name, node.id))` → `new Map(dbNodes.map(n => [n.screen_name, n.id]))`, and `:1392` → `new Set(dbEdges.map(e => e.to_screen))`. But roughly two thirds have no declarative target at all: `cli/index.ts:390` is a `groupBy` into a `Map` (reduce reads worse), `cli/index.ts:419/421` and `daemon/debugTools.ts:177,275` are pure `console.log`/string-append, `IOSCtrlProxyManager.ts:2176` is `signing.warnings.forEach(w => logger.warn(w))`, `ObserveResultOutput.ts:554,556` is a recursive tree walk mutating an accumulator. All of those can only become for-of.

If the goal is declarative accumulation, a narrower selector gets there with far less collateral — flagging `forEach` whose body's only statement is a `push`/`add`/`set` onto an outer collection catches the NavigationGraphManager class precisely and leaves the logging loops alone. Not a blocker either way, but as written the rule mostly shuffles style rather than advancing the stated goal.

eslint.config.mjs
```diff
@@ -461,6 +483,23 @@ export default [
...
+		// gates NEW code without a big-bang rewrite. Thresholds are deliberately
+		// loose to start (complexity 15, depth 4) and can be ratcheted down as the
+		// suppression baseline is burned down. Tests/generated files are exempt.
+		files: ["src/**/*.ts"],
+		plugins,
+		languageOptions,
+		rules: {
+			"max-depth": ["error", 4],
+			"complexity": ["error", 15],
+			"max-nested-callbacks": ["error", 4],
```
kaeawc: The zero baseline is real, not an artifact — the rule does apply to `src` and does fire when you lower it:

```
$ eslint src --rule '{"max-nested-callbacks":["error",2]}'
  3 hits (failureAnalyticsRepository.ts:229, AndroidCtrlProxyClient.ts:2048, FileDownloader.ts:102) — all "nested callbacks (3)"
$ eslint src --rule '{"max-nested-callbacks":["error",3]}'
  0 hits
```

So the deepest callback nest anywhere in `src` is 3, and a threshold of 4 only rejects a 5-deep nest — two levels beyond anything that exists. `3` would be an equally-zero baseline with actual gating power. Nit, and a candidate for the follow-up ratchet either way.

eslint.config.mjs
```diff
@@ -246,6 +246,55 @@ const plugins = {
...
+// block below). Tests and generated files are intentionally exempt: test setup
+// legitimately leans on imperative iteration. `for-in` is banned outright
+// (prototype-walk foot-gun; prefer Object.keys/entries + for-of); `forEach` is
+// banned because it hides accumulation side effects that read more clearly as
+// map/filter/reduce or a for-of. Loop counters (`++`/`--`) and `for-of` itself
+// are deliberately NOT banned — that churn (240+ sites) outweighs the benefit.
+const functionalStyleSelectors = [
+	{
+		selector: "ForInStatement",
+		message: "Avoid for-in (it walks the prototype chain and yields string keys). Use for-of over Object.keys()/Object.entries()/Object.values(), or a declarative map/filter/reduce.",
```
kaeawc: Nit: `auto-mobile/no-unknown-cast` (eslint.config.mjs:215) closes with "If a library genuinely forces it, add an eslint-disable-next-line with a one-line justification," and the `no-explicit-any` block spells out the `-- <reason>` form. Neither of these two messages names the escape hatch, so a dev hitting the `forEach` error on a `Map` has no in-message signal that a justified disable is the sanctioned answer. Worth one clause each for consistency.

kaeawc (commented):
Reviewed through one lens: are the chosen selectors and thresholds actually correct, and what do they newly fail to catch? Two things I'd want fixed before this lands, one design question worth answering, and a few notes where I checked a concern and it turned out fine.

The thresholds themselves are well-calibrated — I measured the distribution across `src` and `complexity 15` / `max-depth 4` sit right at the tail (178 of 4584 functions, 66 of 3363 blocks). The `importAndTimerSelectors` extraction is the right instinct; it just doesn't cover every place the list gets rebuilt.

eslint.config.mjs
kaeawc: Confirmed and fixed in 3727396c3. Reproduced with `--print-config`: `ScreenFingerprint.ts` resolved 5 selectors vs 7 for `src/utils/logger.ts`.

Fixed structurally rather than by patching the spread: the iteration bans moved into a dedicated `auto-mobile/no-imperative-iteration` rule. Because that key is not part of `baseRules`, it survives the `...baseRules` spread. `no-restricted-syntax` is now byte-identical to main, so the clobbering is moot.

Verified: `ScreenFingerprint.ts` and `src/utils/logger.ts` both now resolve `auto-mobile/no-imperative-iteration: [2]`.

kaeawc: This was the most valuable finding in the review — fixed in 3727396c3.

Reproduced it on disk (not via stdin, which bypasses suppressions entirely and gives a false negative). In `src/server/resourceRegistry.ts` (budget 1), swapping the baselined `forEach` for a banned `setTimeout` lints **exit 0** — the Timer-injection guard silently defeated. main carries zero `no-restricted-syntax` suppressions, so this PR would have been the first thing to ever put budget in that bucket, across 15 files.

Fixed exactly as you suggested: a dedicated `auto-mobile/no-imperative-iteration` rule with its own suppression key. `no-restricted-syntax` suppressions are back to **0**, matching main. Added a regression guard in `test/lint/imperativeIterationRule.test.ts` asserting the bans stay out of the shared budget.

eslint.config.mjs
kaeawc: Same root cause as the thread above — fixed in 3727396c3 by moving the bans into a dedicated rule that is not part of `baseRules` and therefore survives the spread. Independently confirmed via `--print-config` before and after.

eslint.config.mjs
kaeawc: Agreed, and this changed the design — fixed in 3727396c3.

The ban is now narrowed to *pure accumulation*: a `forEach` whose every statement is a bare `push`/`unshift`/`add`/`set`. That is the class with a real declarative target. A callback that logs, branches, awaits, or recurses is deliberately left alone, precisely because its only rewrite is for-of — more imperative, defeating the rule.

Baseline drops **26 → 5** forEach sites. The `console.log`/`logger.warn`/recursive-walk cases you identified (`cli/index.ts:419/421`, `debugTools.ts`, `ObserveResultOutput.ts:554/556`) no longer fire. Both directions are locked in by tests, including explicit allow-cases for logging and branching callbacks.

eslint.config.mjs
kaeawc: Good catch — taken. `max-nested-callbacks` is now **3** rather than 4 in 3727396c3. Still a zero baseline, but at 4 it had no bite at all given the deepest nest in `src` is 3, so it now actually ratchets.

eslint.config.mjs
kaeawc: Taken — fixed in 3727396c3. Both messages on the new rule now close with the same escape-hatch wording as `no-unknown-cast`: "add an eslint-disable-next-line with a one-line justification."

kaeawc:
## Review round addressed — 3727396c3

Ran four independent review passes with different lenses (flat-config merge semantics, suppressions-ratchet integrity, selector false-positive risk, DX/docs). Every finding is either fixed or explicitly answered above; all six inline threads replied to and resolved.

**Three blocking defects, all reproduced before fixing:**

1. **Suppression-budget leak.** Folding the iteration bans into `no-restricted-syntax` shared one per-file count with the `setTimeout`/`structuredContent` guards. Proved on disk in `src/server/resourceRegistry.ts`: swapping the baselined `forEach` for a banned `setTimeout` → **exit 0**. Moved to a dedicated `auto-mobile/no-imperative-iteration` rule; `no-restricted-syntax` suppressions back to **0**, matching `main`.

2. **Two src files silently exempt.** `ScreenFingerprint.ts`/`ExploreElementExtraction.ts` resolved 5 selectors vs 7 elsewhere — the `...baseRules` spread replaces the whole rule entry. The dedicated rule isn't in `baseRules`, so it survives.

3. **Burn-down broke CI.** ESLint exits **2** on an over-counted suppression, so improving a baselined function turned the build red. `lint` now passes `--pass-on-unpruned-suppressions`; added `lint:prune` / `lint:baseline`.

**Design change:** the `forEach` ban is narrowed to pure accumulation (every statement a bare `push`/`unshift`/`add`/`set`). Review correctly argued that banning `forEach` while allowing `for-of` just pushes people toward `for-of` + `push` — *more* imperative. Callbacks that log, branch, or recurse are no longer flagged. Baseline 26 → 5.

**A methodology note worth recording:** one agent's leak repro used `--stdin-filename`. Suppressions are **bypassed entirely** under `--stdin`, so stdin-based tests silently report violations that are actually suppressed. I re-verified everything against real on-disk files. The finding held; the original method wouldn't have shown it.

**Also:** `max-nested-callbacks` 4 → 3 (at 4 it had no bite); dropped a stale suppression for a file that exists neither on disk nor in `main`; documented the ratchet in `CLAUDE.md`; added `test/lint/imperativeIterationRule.test.ts` (7 tests) including a budget-leak regression guard.

Follow-ups filed as #3966 (burn down baseline, tighten thresholds) and #3967 (evaluate `eslint-plugin-functional`, scoped).

Verification: `lint` 0 · `build` 0 · `bun test` **8086 pass / 0 fail** · config purely additive vs `main`.

kaeawc:
## Scope narrowed — 58c499bd3

Dropped the `for-in` ban. Banning it was stricter than this ratchet's intent, which is to gently direct toward declarative style rather than outlaw iteration. **All explicit loop forms are now allowed** — `for`, `for-of`, `for-in`, `while`.

Renamed `auto-mobile/no-imperative-iteration` → `auto-mobile/no-accumulator-foreach`, since with loops permitted the old name implied a scope the rule no longer has. It now flags exactly one shape: a `.forEach()` whose every statement is a bare `push`/`unshift`/`add`/`set`.

Baseline **9 → 5**. The four `for-in` loops in `ViewHierarchy.ts` are no longer flagged, which also removes the only reason this ratchet would ever have needed to touch correctness-sensitive observation code — so the burn-down in #3966 is now five mechanical `map`/`Set`/`Map` rewrites with no device verification required.

Added a test asserting all four loop forms lint clean, so the permissive scope can't silently regress. Kept the out-of-src scoping test and the `no-restricted-syntax` budget-leak regression guard.

`lint` 0 · `build` 0 · `bun test` **8086 pass / 0 fail** · still zero `src/` files touched.

kaeawc:
## CI status — merged main, two non-required iOS checks red for a pre-existing reason

Merged `origin/main` (this branch was 20 commits behind) and re-validated: `lint` 0 · `build` 0 · `bun test` **8077 pass / 0 fail**. Baseline still exactly matches reality — `lint:prune` removes 0 of 408 entries — and main's new `src/` files introduce no violations of the new rules (main's own #3958 already made that code declarative).

**All required checks pass.** `Fast Validation`, all three `Node TypeScript Build and Test` legs, `MCP Benchmarks`, `Memory Leak Detection`, `Bun Security Audit`, and both `Installer` matrices are green; the rest legitimately skip for a TypeScript-only change.

**Two red checks — `XCTestRunner Simulator Tests` and `iOS` — are not caused by this PR and are not required.** The failing step is `Check CtrlProxy XcodeGen Project Drift`, and the diff is purely `targets = (...)` ordering in the pbxproj. It came in with #3969, which is already on main and whose own checks are red for the identical reason. Filed as #3975.

I want to correct an earlier guess I posted: I first attributed this to a stale merge ref and expected the main merge to clear it. That was wrong — it reproduces on the merge commit too. The actual finding is that the committed ordering *is* reproducible locally with the same XcodeGen 2.46.0 that CI installs, so the ordering is environment-dependent, not version-dependent. Details in #3975.

I deliberately did not regenerate the pbxproj here: my local XcodeGen emits the committed ordering, not CI's, so committing CI's bytes would mean hand-editing generated output I can't reproduce or verify — and unrelated iOS churn does not belong in a lint-policy PR. This PR still touches zero `src/` and zero `ios/` files.

Issue #3916, opened 2026-07-18
epic(a11y): seamless navigation whether or not a screen reader (TalkBack/VoiceOver) is enabled


kaeawc:
## Goal
AutoMobile should navigate mobile apps **identically and reliably whether or not a screen reader is enabled** — TalkBack on Android, VoiceOver on iOS. Turning a screen reader on (or a stale/cold detection cache) must never silently change which code path runs, send navigation the wrong way, or activate the wrong element.

## Invariants we want to hold
1. **Detection is deterministic** — screen-reader-vs-standard routing must not depend on a warm cache or a real-vs-`null`/missing dependency, and the `force-accessibility-mode` override must apply uniformly across observe/tap/swipe.
2. **Navigation converges** — focus/label-based targeting must reach the target regardless of initial cursor position and never march the wrong direction to a swipe cap.
3. **Activation lands on the intended node** — after navigation reports success, activation acts on the node the screen reader actually focused, never on stale/placeholder coordinates.
4. **One matcher, one behavior** — a single, consistently-keyed matcher shared across detection, navigation, and focus tracking; no parallel dead implementation.
5. **Toggling a screen reader is verified** — enable/disable reports success only after confirming the resulting state.

## Android (TalkBack) — findings
Bugs in the swipe/focus/activation stack (see per-issue detail):
- [ ] #3915 — swipe/scroll silently degrades to coordinate mode (`null` passed as `AdbExecutor` to `detectMethod`) — _inv. 1_
- [ ] #3917 — focus path calculator assumes null cursor == index 0 → wrong direction (+ stuck-detection off-by-one) — _inv. 2_
- [ ] #3918 — activation double-taps the element's stored bounds (or `(0,0)`), not the focused node — _inv. 3_
- [ ] #3919 — dead + divergent `AccessibilityFocusTracker` (exact/camelCase vs. production fuzzy/hyphenated matcher) — _inv. 4_
- [ ] #3920 — nav leftovers: unused `intermediateCheckpoints`; `executeTap` ignores `tap` vs `doubleTap`
- [ ] #3921 — screen-reader toggle reports `applied:true` optimistically (**now covers iOS `VoiceOverToggle` too**) — _inv. 5_
- [ ] #3922 — detection/focus robustness: `!output.includes("null")` misclassification; `SetAccessibilityFocus` masks confirmation failure — _inv. 1_

## iOS (VoiceOver) — vetting outcome
[...]

PR #3959, opened 2026-07-19
fix(a11y): confirm screen-reader toggles instead of optimistic success (#3921)

### LINKED ISSUE #3921: fix(a11y): screen-reader toggle reports success optimistically — TalkBackToggle and VoiceOverToggle
Part of #3916 — the "toggling a screen reader is verified" invariant. This affects **both** platforms' toggles (found while vetting iOS for parity with the Android findings).

## 1. Android — `TalkBackToggle` reports `applied: true` optimistically
`src/features/accessibility/TalkBackToggle.ts:58-73` — after `enableTalkBack` + best-effort `dismissPermissionDialog`, it invalidates the cache and returns `{ applied: true, currentState: enabled }` **without re-detecting** that TalkBack actually became active. If dialog dismissal failed and TalkBack never fully enabled, callers are told it succeeded.

## 2. iOS — `VoiceOverToggle` has the identical pattern
`src/features/accessibility/VoiceOverToggle.ts:40-44` — after the `simctl` write + `notifyutil` post, it invalidates the cache and returns `{ applied: true, currentState: enabled }` with no confirmation that VoiceOver actually toggled. Same optimistic-success smell, same fix shape.

## 3. Android-only — flaky consent-dialog detection via `uiautomator dump /dev/tty`
`TalkBackToggle.ts:211-213` dumps to `/dev/tty` and parses `stdout`; `uiautomator` frequently prints a status line to stdout instead of the XML, so `xml.includes("TalkBack")` is false, the dialog is never dismissed, and it silently falls through to `logger.warn` (line 248), leaving automation stalled behind the consent dialog.

## 4. Android-only — inconsistent error contract
`writeSecureSetting` / `deleteSecureSetting` (lines 85-88, 98-101) swallow the a11y-path failure then fall back to ADB — fine — but if the ADB `executeCommand` also fails, that raw error propagates **uncaught** out of `toggle()`, unlike the graceful `TalkBackResult` returned on other paths.

## Suggested direction
Both toggles: re-detect state after applying and set `applied`/`currentState` from the confirmed result (invalidate cache, then `detectMethod` / `requestVoiceOverState`). Android also: make dialog detection robust (dump to a file / use the CtrlProxy hierarchy) and wrap the settings-write failure into a typed result.

_Note: iOS `VoiceOverToggle` is Simulator-only; physical-device VoiceOver toggle is tracked separately in #2501._

_Found via a code audit of the TalkBack and VoiceOver toggle paths._

kaeawc:
## Summary

`TalkBackToggle` and `VoiceOverToggle` both returned `{ applied: true, currentState: enabled }` immediately after dispatching the enable/disable commands, **without verifying** the screen reader actually reached the requested state (#3921). If the TalkBack consent dialog wasn't dismissed, or a `simctl` write didn't land, callers were told the toggle succeeded when it hadn't.

## Changes

1. **TalkBack — re-detect after apply.** Confirm the state via a shared `detectTalkBackEnabled()` helper (used for both the pre-apply idempotency check and the post-apply confirmation) and derive `applied` / `currentState` from the confirmed result.
2. **VoiceOver — re-detect after apply.** Same pattern via `isVoiceOverEnabled` (resolving the CtrlProxy client lazily, matching `IosTapStrategy`).
3. **TalkBack consent-dialog detection robustness.** Dump the window hierarchy to a device file and `cat` it back instead of `uiautomator dump /dev/tty` — the `/dev/tty` form frequently prints its status line to stdout instead of the XML, so the dialog was never matched and dismissal silently stalled.
4. **Consistent error contract.** Wrap the apply phase in both toggles so an ADB/`simctl` failure returns a typed result (`applied: false` + `reason`) instead of propagating raw out of `toggle()`.

## Tests

- `FakeAccessibilityDetector` gains a `detectMethod` result queue (`enqueueDetectMethodResults`) so a test can model the pre-apply idempotency detect and the post-apply confirmation detect independently.
- New cases: "did not activate after apply → `applied:false`" (both platforms), the file-based dialog dump (asserts no `/dev/tty`), and the typed apply-phase failure (both platforms).
- 916 tests green across the accessibility + action suites; lint clean; typecheck gate 0 new errors (the lone pre-existing ajv-formats mismatch in `PlanSchemaValidator.ts` reproduces on clean `main`).

Closes #3921

Part of #3916.

### DIFF
--- a/src/features/accessibility/TalkBackToggle.ts
+++ b/src/features/accessibility/TalkBackToggle.ts
@@ -40,12 +40,7 @@ export class TalkBackToggle {
     // already in the requested state.  Use detectMethod rather than
     // isAccessibilityEnabled so that other active services (e.g. CtrlProxy)
     // do not cause a false positive.
-    this.detector.invalidateCache(this.device.deviceId);
-    const detectedService = await this.detector.detectMethod(
-      this.device.deviceId,
-      this.adb
-    );
-    const talkBackCurrentlyEnabled = detectedService === "talkback";
+    const talkBackCurrentlyEnabled = await this.detectTalkBackEnabled();
     if (talkBackCurrentlyEnabled === enabled) {
       return {
         supported: true,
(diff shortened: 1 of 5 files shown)


kaeawc:

**Result: PASS** — toggles confirm instead of reporting optimistically.

Two observations on a real device:

1. **Non-optimistic refusal.** With TalkBack installed but disabled, `--talkback true` returned a typed failure rather than claiming success. `TalkBackToggle.toggle()` returns `{supported:false, applied:false, reason}` at `TalkBackToggle.ts:32`, the apply path wraps errors into a typed result at `:67`, and `:79` re-detects to confirm before reporting.
2. **Fresh state, not cached.** Before enabling, `accessibility` reported `{"enabled":true,"service":"unknown"}`; once TalkBack was actually running it reported `{"enabled":true,"service":"talkback"}` — reflecting real post-apply state.

Separately, the installed-check underneath this is wrong on stock emulators (reports "not installed" for an installed-but-disabled TalkBack) — filed as #4042. That is a pre-existing detection bug from #1403, not something this PR introduced.

Issue #3963, opened 2026-07-19
feat(a11y): fidelity assertions — surface reachability / traversal-order / focus-trap results


kaeawc:
Follow-up split out of #3937 (opt-in screen-reader-navigation fidelity mode).

#3937 landed the exposure + repaired-traversal layer: the `screen-reader-navigation` feature flag drives cursor-traversal activation via the repaired Android path (#3917 / #3918 / #3920). What it deliberately deferred is the **payoff** of the mode — surfacing accessibility-journey results so tests can assert on them.

## Scope
Build a fidelity-assertion surface on top of the cursor-traversal path so an a11y test can assert:
- **Reachability** — was the target actually reachable by swipe stepping (vs only via direct activation)?
- **Traversal order** — the ordered sequence of focused nodes the cursor stepped through.
- **Focus traps** — did stepping get stuck / fail to converge (ties into the #3917 non-converging-cursor guard)?

## Notes
- `TalkBackTapStrategy` / `FocusNavigationExecutor` already compute traversal order and swipe stepping internally; this is about surfacing that as a structured result on the tap/flow response rather than discarding it.
- Decide the result shape and where it attaches (tap result vs a dedicated a11y-journey tool) during design.
- iOS parity depends on #3924 (iOS traversal layer), which is a separate prerequisite.

Part of #3916 (umbrella #3937).

Issue #3936, opened 2026-07-18
feat(a11y): make direct activation the default screen-reader interaction path (Android + iOS)


kaeawc:
Part of #3916. This is the **Model #1** decision from the label-activation-vs-cursor-traversal analysis: when a screen reader is on, the **default** `tapOn` path should *directly activate the target node* (by accessibility id/label), deterministically and symmetrically across platforms — **not** drive the screen-reader cursor. Cursor traversal moves to an opt-in fidelity mode (#3).

## Why
Direct activation is a single round-trip, deterministic, and identical on both platforms. Driving the cursor by swipe (today's Android default) is a stateful loop that is the source of the current navigation bugs (#3917 wrong-direction, #3918 stale-bounds activation, stuck-detection). For everyday automation "the tap should land," direct activation is the correct default. (The a11y-fidelity value of true cursor navigation is real but belongs in the opt-in mode — see #3.)

## Target behavior (screen reader ON)
**Android** (`executeAndroidTapWithAccessibility`):
- `tap` / `doubleTap` → `ACTION_CLICK` by resource-id → dispatchGesture coordinate tap → ADB coordinate tap. **Remove swipe focus-navigation from the default path** (gate it behind the #3 opt-in).
- `longPress` → `ACTION_LONG_CLICK` by id → coordinate (already implemented).

**iOS** (`executeIOSTapWithVoiceOver`) — already direct-activation; tighten it:
- Resolve the label from `ios-accessibility-label` → `content-desc` → `text` (currently skips `content-desc` — the gap noted in #3924).
- `requestVoiceOverActivate(label, action)` → coordinate fallback (already implemented).

## Acceptance criteria
- With TalkBack/VoiceOver enabled, `tapOn` activates the target via the a11y action by default, with no swipe stepping.
- Behavior is symmetric across platforms and deterministic (no dependence on traversal order or cursor position).
- Falls back cleanly: a11y action fail → coordinate (dispatchGesture) → ADB/coordinate.
- The swipe-navigation code (`TalkBackTapStrategy`/`FocusNavigationExecutor`) is retained but only invoked when the #3 opt-in is set; default no longer calls it.
- Unit tests cover: id present → ACTION_CLICK; id absent → coordinate; a11y failure → fallback chain; iOS content-desc label resolution.

## Relationship to existing issues
- Supersedes the *default-path* portion of the Android navigation bugs — #3917/#3918 only matter inside the opt-in mode (#3) once this lands, dropping their priority.
[...]

Issue #3924, opened 2026-07-18
feat(voiceover): iOS has no accessibility focus/traversal layer — VoiceOver taps can't move the cursor


kaeawc:
Part of #3916 (seamless navigation whether or not a screen reader is enabled) — the iOS side of the "navigation converges / activation lands on the focused node" invariants.

## Summary
Android has a full TalkBack focus-navigation stack (`TalkBackTapStrategy` → `FocusNavigationExecutor` → `FocusPathCalculator`, plus `CtrlProxyFocus` with `requestCurrentFocus`/`requestTraversalOrder`/`setAccessibilityFocus`/`clearAccessibilityFocus`). **iOS has none of it.** The iOS VoiceOver tap path does a single-shot `requestVoiceOverActivate(label, action)` (`TapOnElement.ts:1561-1591`) and cannot step/traverse the VoiceOver cursor the way Android does.

## Evidence
- **No focus/traversal methods on the iOS client** — `IOSCtrlProxyClient` has no `requestCurrentFocus` / `requestTraversalOrder`, and there is no `CtrlProxyFocus` delegate for iOS (Android's lives at `src/features/observe/android/CtrlProxyFocus.ts`).
- **Dead protocol surface** — the iOS runner contract *declares* `"get_current_focus"` and `"get_traversal_order"` as supported commands (`src/features/observe/ios/ctrlProxyRequestTypes.ts:62-63`), but the TS client never emits them. Wired-but-unused.
- **No set/clear focus** — `SetAccessibilityFocus.execute` (`src/features/accessibility/SetAccessibilityFocus.ts:62-66`) and the `accessibilityFocus` tool handler (`src/server/accessibilityFocusTools.ts:44-48`) both hard-throw for iOS: *"iOS VoiceOver focus is not yet implemented."*
- **VoiceOver focused element not reported in observe** — parity design doc Gap 1 (`docs/design-docs/mcp/a11y/voiceover-talkback-parity.md:33-41`): no `voiceoverFocusedElement` field is populated.

## Secondary defect (same file, cheap to fix alongside)
`executeIOSTapWithVoiceOver` resolves the VoiceOver label from `ios-accessibility-label` → `text` only (`TapOnElement.ts:1569-1570`) — it **ignores `content-desc`**. Elements exposed only via content-desc fall through to a coordinate tap even when VoiceOver could activate them by label.

## Suggested direction
[...]

Issue #3920, opened 2026-07-18
chore(talkback): finish or remove nav leftovers — unused intermediateCheckpoints, ignored tap/doubleTap


kaeawc:
Part of #3916 (seamless navigation whether or not TalkBack is enabled).

Two pieces of unfinished implementation in the TalkBack navigation code — neither is a live bug, but both are half-wired and misleading.

## 1. `intermediateCheckpoints` computed but never consumed
`src/features/talkback/FocusPathCalculator.ts:47-72` builds `path.intermediateCheckpoints` via `buildCheckpoints`, but no consumer reads it. `FocusNavigationExecutor` does its own verification cadence via `totalSwipes % verificationInterval` (`FocusNavigationExecutor.ts:213-214`). Dead output from an unfinished checkpoint mechanism — either wire the executor to verify at the calculated checkpoints, or drop the field.

## 2. `executeTap` ignores the `tap` vs `doubleTap` distinction
`src/features/talkback/TalkBackTapStrategy.ts:71-76, 168-169` — `executeTap` accepts `action: "tap" | "doubleTap"` but always runs the same double-tap activation sequence in `activateElement`; the parameter is never branched on. Either honor the action (single vs double activation) or remove the parameter so callers aren't misled.

_Found via a code audit of the TalkBack navigation path._

PR #3964, opened 2026-07-19
feat(a11y): screen-reader-navigation feature flag for cursor-traversal fidelity mode (#3937)

### LINKED ISSUE #3937: feat(a11y): opt-in screen-reader navigation fidelity mode (swipe cursor traversal)
Part of #3916. This is **Model #3** from the label-activation-vs-cursor-traversal analysis: an explicit, opt-in mode that drives the screen-reader cursor element-by-element the way a real user does (swipe to step focus, double-tap to activate), for **accessibility validation** — reachability, traversal order, focus traps, "is this element actually swipe-navigable."

## Why a separate mode
Direct activation (#1, the default) is a false-negative machine for a11y validation: it will activate elements a blind user could never reach. Only cursor traversal can catch reachability/traversal-order/focus-trap defects. But traversal is O(N) swipes, flaky, and stateful — so it should be opt-in and used for a11y *assertions*, not for every tap.

## Scope
- **Expose the opt-in** — a way to request "simulate screen-reader navigation" for a tap/flow (feature flag or per-call option; decide during design). Default stays direct-activation (#1).
- **Repair the Android traversal path** (its whole value is fidelity, so the bugs must be fixed here):
  - [ ] #3917 — null-cursor → wrong-direction navigation (+ stuck-detection off-by-one)
  - [ ] #3918 — activation taps stale element bounds / `(0,0)` instead of the focused node
  - [ ] #3920 — unused `intermediateCheckpoints` (wire checkpoint verification, or drop)
- **iOS traversal layer** (#3924) — build only if/when iOS a11y-journey validation is actually wanted:
  - wire the already-declared `get_current_focus` / `get_traversal_order` runner commands into the iOS client + a `CtrlProxyFocus`-equivalent
  - synthesize single-finger swipe stepping + double-tap activation
- **Fidelity assertions** — surface reachability / traversal-order / focus-trap results so tests can assert on them (this is the payoff of the mode).

## Note on "human-representative" tension
`accessibilityFocusTools.ts:69-73` flags that programmatic focus *placement* isn't human-representative (a user can't teleport the TalkBack cursor). This mode should therefore step the cursor via swipes, not jump it — keeping it an honest simulation of the user journey.

## Priority
Lower than #1. Land #1 first (reliable default), then build this as the fidelity layer on top.

kaeawc:
## Summary

**Model #3** from #3916: an explicit, opt-in mode that drives the screen-reader cursor element-by-element (swipe to step focus, double-tap to activate) the way a real user does — for accessibility validation (reachability, traversal order, focus traps) — instead of the direct-activation default (#3936).

The cursor-traversal path already exists (`TalkBackTapStrategy.executeTap`) and its Android bugs are now fixed (#3917 non-converging-cursor guard, #3918 activate the live focused node, #3920 nav leftovers). This PR wires the **global opt-in**:

- Add a `screen-reader-navigation` feature flag (default **off**). Not tool-definition-affecting.
- `TapOnElement` takes the fidelity path when the flag is enabled **OR** the per-call `screenReaderNavigation` option is set (`isScreenReaderNavigationEnabled`), storing the resolved `FeatureFlagService` so the check is a single in-memory read.

## Scope (per design decision)

- **Exposure = feature flag only.** The per-call `screenReaderNavigation` option stays internal (unchanged); the flag is the user-facing opt-in, mirroring the `force-accessibility-mode` pattern.
- **Deferred to follow-ups:** the iOS traversal layer (#3924) and the fidelity-assertion surface — reachability / traversal-order / focus-trap results — now tracked in #3963.

## Tests

- Flag ON drives cursor traversal even without the per-call option; flag OFF keeps the direct-activation default.
- 804 tests green across the feature-flag + action + server suites; lint clean; typecheck gate 0 new errors (the lone pre-existing ajv-formats mismatch in `PlanSchemaValidator.ts` reproduces on clean `main`).

Closes #3937

Part of #3916. Follow-ups: #3924 (iOS traversal), #3963 (fidelity assertions).

### DIFF
--- a/src/features/action/TapOnElement.ts
+++ b/src/features/action/TapOnElement.ts
@@ -109,6 +109,7 @@ export class TapOnElement extends BaseVisualChange {
   private elementSelector: ElementSelector;
   private viewHierarchy: ViewHierarchy;
   private talkBackStrategy: TalkBackTapStrategy;
+  private readonly featureFlags: FeatureFlagService;
   private talkBackDriverFactory: TalkBackNavigationDriverFactory;
   private iosVoiceOverDetector: IosVoiceOverDetector;
   private strategy: TapStrategy;
@@ -198,12 +199,13 @@ export class TapOnElement extends BaseVisualChange {
     this.talkBackStrategy = options.talkBackStrategy ?? new TalkBackTapStrategy({ timer: this.timer });
     this.talkBackDriverFactory = options.talkBackDriverFactory ?? new DefaultTalkBackNavigationDriverFactory(this.adbFactory);
     this.iosVoiceOverDetector = options.iosVoiceOverDetector ?? defaultIosVoiceOverDetector;
+    this.featureFlags = options.featureFlags ?? FeatureFlagService.getInstance();
     this.strategy = options.tapStrategy ?? createTapStrategy(
       device,
       this.adb,
       this.accessibilityDetector,
       this.iosVoiceOverDetector,
-      options.featureFlags ?? FeatureFlagService.getInstance()
+      this.featureFlags
     );
     this.longPressMetadataDetector = new LongPressMetadataDetector(this.elementParser);
   }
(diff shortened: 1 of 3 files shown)


kaeawc:

**Result: PASS** — flag present and correctly defaulted off.

`src/features/featureFlags/FeatureFlagDefinitions.ts:119` defines `screen-reader-navigation` with `defaultValue: false` (opt-in), and the gated-off behavior is confirmed on device: with the flag off, `tapOn` under active TalkBack activates directly rather than driving cursor traversal.

I did not exercise the flag-ON path — I could not find a runtime toggle for it (no env var in the featureFlags module), so the traversal path remains unverified on device.

PR #3974, opened 2026-07-19
ci(desktop): upload screenshot pixel diffs when a comparison fails

### LINKED ISSUE #3913: Recorded desktop screenshot baselines fail their own immediate verification on the reference OS
## Summary

The `Record Desktop Screenshot Baselines` workflow records the 6 `ComponentScreenshotTest` baselines on `ubuntu-latest`, un-pends them, then re-runs them in verify mode **in the same job on the same runner** — and all 6 fail the comparison. A baseline recorded on the reference OS does not pass its own immediate re-verification, so the workflow can never produce a green baselines PR. This blocks #3752.

**This is not the jemalloc crash.** The v0.0.5 bump (#3908) fixed the exit-139 SIGSEGV — the second `gradle-task-run` invocation now logs `jemalloc already in place … skipping relocate` and runs to completion. The failure below is a genuine screenshot mismatch surfaced only *after* that infra blocker was cleared.

## Evidence

Run: https://github.com/kaeawc/auto-mobile/actions/runs/29603004265 (`workflow_dispatch`, `unpend=true`, on merge commit `a44aaad`).

Step **"Verify baselines pass in normal mode"**:
```
432 tests completed, 6 failed
> Task :desktop-core:test FAILED
```
The 6 failures are exactly the recorded set, each a `java.lang.AssertionError` (pixel comparison) at its `screenshotTest(...)` call site in [`ComponentScreenshotTest.kt`](https://github.com/kaeawc/auto-mobile/blob/main/android/desktop-core/src/test/kotlin/dev/jasonpearson/automobile/desktop/core/screenshot/ComponentScreenshotTest.kt):

`errorCardLight` (:44) · `errorCardDark` (:52) · `errorCardWithActions` (:60) · `searchBarEmpty` (:73) · `searchBarWithQuery` (:79) · `statusBarBadge` (:93)

## Mechanism

The workflow (`.github/workflows/record-screenshot-baselines.yml`) runs, all in one job:
1. **Record** (`-Dscreenshot.record=true --rerun-tasks`) — writes the 6 PNGs to `src/test/resources/screenshots/`. Passes.
2. **Un-pend** (`scripts/screenshot/unpend_screenshot_tests.sh`) — strips `pending = true`, making the 6 tests active checks.
3. **Verify** (normal mode, `--rerun-tasks`) — re-renders and compares against the PNGs from step 1. **Fails 6/6.**

So this is record-then-verify on identical OS/JDK/font stack. Per `ScreenshotComparator`'s own unit test, the comparator already tolerates a per-channel delta of <4 ( `small anti-aliasing jitter is tolerated`), so the mismatch exceeds ±4/channel or is structural (size/subpixel), not mere rounding.

## Leading hypotheses (unverified — needs the diff images)

- **Non-deterministic Compose Desktop rendering between two renders** in the same JVM/session (font hinting, subpixel AA, or `waitForIdle()` racing a layout/measure pass) — the most likely, given identical environment.
- **Comparator tolerance too tight** for real Compose AA output (the ±4 jitter test uses solid fills, not rendered text).
- **Record vs verify path divergence** — an image transform (`toAwtImage()`) or golden-dir resolution that differs between the record and compare branches of `ScreenshotEnvironment.handleResult`.

## Suggested next step: make the diff inspectable

`ScreenshotComparator` writes rejected + diff PNGs to `build/reports/screenshots/` on mismatch, but the workflow **does not upload that directory**, so the pixel diffs can't be seen from the failed run. Add an `actions/upload-artifact` (`if: failure()`) for `android/desktop-core/build/reports/screenshots/` — without it, diagnosing this is guesswork. That upload is arguably a prerequisite for the rest of this issue.

## Repro (local, reference OS)
```
./gradlew -p android :desktop-core:test --tests '*ComponentScreenshotTest' -Dscreenshot.record=true --rerun-tasks
scripts/screenshot/unpend_screenshot_tests.sh android/desktop-core/src/test/kotlin/dev/jasonpearson/automobile/desktop/core/screenshot/ComponentScreenshotTest.kt
./gradlew -p android :desktop-core:test --tests '*ComponentScreenshotTest' --rerun-tasks   # expect 6 failures
# inspect android/desktop-core/build/reports/screenshots/ for *_diff.png / *_actual.png
```
(On macOS/Windows the tests self-skip via `assumeReferencePlatform()`; run on Linux or with `-Dscreenshot.reference.os=any`.)

Blocks #3752. Context: #3908 (jemalloc bump that unblocked reaching this failure).

kaeawc:
## Summary

`ScreenshotComparator` writes `<name>.diff.png` and `<name>.actual.png` into
`android/desktop-core/build/reports/screenshots/` on every mismatch, but no workflow
uploaded that directory — so the 6/6 failure in #3913 dies with the runner and can only
be guessed at. This adds the upload to every job that can fail a screenshot comparison.

This is the diagnostic prerequisite the issue itself calls for ("that upload is arguably
a prerequisite for the rest of this issue"), **not** the root-cause fix. See below for why
the root cause is not fixed here.

### Why no fix yet

I ran the issue's local repro on macOS/arm64 (JDK 21, `-Dscreenshot.reference.os=any`) —
record, un-pend, then verify as two separate Gradle invocations, exactly as the workflow does:

```
tests="6" skipped="0" failures="0" errors="0"
```

It **passes**, and no `build/reports/screenshots/` directory is even created. So:

- Hypothesis 1 (non-deterministic Compose Desktop rendering between two renders) does **not**
  reproduce off-Linux.
- Hypothesis 3 (record-vs-verify path divergence in `handleResult` / `toAwtImage()`) is
  effectively eliminated — that code path is OS-independent and would have failed here too.

That leaves a Linux-specific rasterization difference or a comparator tolerance too tight for
Linux freetype output (hypothesis 2). Neither is distinguishable without the actual pixel
diffs, so any fix written now would be guesswork. Next step after this merges: one
`workflow_dispatch` of `Record Desktop Screenshot Baselines`, then read the artifact.

## Linked Issue

Refs #3913 — deliberately **not** `Closes`, since the underlying baseline mismatch is
still unfixed. See Follow-ups.

## Bug / Regression Context

- **Prior attempts:** #3908 bumped jemalloc and cleared the exit-139 SIGSEGV that previously
  masked this; the mismatch below is what surfaced once that infra blocker was gone.
- **Observed symptom:** the record workflow records 6 baselines on `ubuntu-latest`, un-pends
  them, re-verifies in the same job on the same runner, and all 6 fail the pixel comparison
  ([run 29603004265](https://github.com/kaeawc/auto-mobile/actions/runs/29603004265)).
- **Repro steps / conditions:** per the issue; reproduces only on the Linux reference OS.

## Changes

| File | Change |
| --- | --- |
| `.github/workflows/record-screenshot-baselines.yml` | Upload the screenshot report dir on failure, after the verify step. Also uploads the just-recorded baselines from `src/test/resources/screenshots/` — a failing verify aborts the job before "Open PR", so those PNGs are otherwise discarded, and they are the third leg of a baseline/actual/diff comparison. |
| `.github/workflows/pull_request.yml` | Same upload in `desktop-core-unit-tests`, so a baseline mismatch on a PR is inspectable too. |
| `.github/workflows/merge.yml` | Same, in the merge-side `desktop-core-unit-tests`. |
| `test/bats/screenshot-diff-artifact-upload.bats` | New wiring guard (9 tests). |

All three uploads use `if: failure()` plus `if-no-files-found: ignore` — these jobs fail for
many reasons that are not screenshot mismatches, and the report dir only exists after one.

## Acceptance criteria

| Criterion | Pinned by |
| --- | --- |
| Failed verify in the record workflow uploads `build/reports/screenshots/` so the diffs are inspectable | `record-screenshot-baselines uploads the screenshot diff reports`, `... uses actions/upload-artifact`, `... only runs on failure`, `... is ordered after the verify step` |
| The recorded baselines survive a failed verify | `... diff upload includes the recorded baselines` |
| The same diffs are inspectable when baselines fail on a PR / on merge | `pull_request desktop-core-unit-tests uploads ...`, `merge desktop-core-unit-tests uploads ...` |
| The change is additive — existing test-report uploads survive | `existing HTML test-report uploads are preserved` |
| Unrelated failures do not warn on a missing dir | `screenshot diff uploads tolerate an absent report directory` |

The bats file was written before the workflow edits: 7 of 9 failed red, and the one guard that
passes pre-change (`existing HTML test-report uploads are preserved`) does so by design — it
pins behavior the change must not break. The two tests added after the edits
(`... includes the recorded baselines`, `... tolerate an absent report directory`) were
mutation-checked by removing the corresponding YAML lines and confirming they go red.

## Validation

- [x] `scripts/all_fast_validate_checks.sh` — 15/15 pass
- [x] `bun test` — 8090 pass, 11 skip, **0 fail** (638 files)
- [x] `bats test/bats/` — 343 pass; 1 pre-existing failure in `verify-runtime-deps.bats`
      ("passes when all dependencies are present"), confirmed to fail identically on a stashed
      clean tree — a missing local runtime dep, unrelated to this diff
- [x] `bun run typecheck` — reports 1 error in `src/utils/plan/PlanSchemaValidator.ts`, also
      confirmed pre-existing on a clean tree; an `ajv`/`ajv-formats` duplicate-resolution
      artifact of this worktree sharing the central `node_modules`. This diff touches zero
      TypeScript. See Follow-ups.
- [x] No new code needing interfaces/fakes/FakeTimer — CI YAML plus a shell wiring guard

## Device Verification

N/A — CI configuration only.

## Follow-ups

- Root-cause fix for the baseline mismatch itself, which is what closes #3913. Blocked on
  a `workflow_dispatch` run against this change to produce the diff images.
- The pre-existing `PlanSchemaValidator.ts` ajv typecheck error and the
  `verify-runtime-deps.bats` failure, if they reproduce outside this worktree.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -74,6 +74,7 @@ jobs:
           name: desktop-core-screenshot-diffs
           path: android/desktop-core/build/reports/screenshots/
           if-no-files-found: ignore
+          retention-days: 7
 
   build-ide-plugin:
     name: "Build IDE Plugin"
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -64,6 +64,17 @@ jobs:
           name: desktop-core-test-results
           path: android/desktop-core/build/reports/tests/test/
 
+      # Screenshot comparisons fail with pixel diffs that are only inspectable
+      # as images -- the HTML test report above carries the stack trace but not
+      # the rendering. See issue #3913.
+      - name: "Upload Screenshot Diffs"
+        if: failure()
+        uses: actions/upload-artifact@v6
+        with:
+          name: desktop-core-screenshot-diffs
+          path: android/desktop-core/build/reports/screenshots/
+          if-no-files-found: ignore
+
   build-ide-plugin:
     name: "Build IDE Plugin"
     # Disabled: consistently >1 minute on merge and not required for release/docs publishing.
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -442,6 +442,7 @@ jobs:
           name: desktop-core-screenshot-diffs
           path: android/desktop-core/build/reports/screenshots/
           if-no-files-found: ignore
+          retention-days: 7
 
   build-ide-plugin:
     name: "Build IDE Plugin"
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -432,6 +432,17 @@ jobs:
           name: desktop-core-test-results
           path: android/desktop-core/build/reports/tests/test/
 
+      # Screenshot comparisons fail with pixel diffs that are only inspectable
+      # as images -- the HTML test report above carries the stack trace but not
+      # the rendering. See issue #3913.
+      - name: "Upload Screenshot Diffs"
+        if: failure()
+        uses: actions/upload-artifact@v6
+        with:
+          name: desktop-core-screenshot-diffs
+          path: android/desktop-core/build/reports/screenshots/
+          if-no-files-found: ignore
+
   build-ide-plugin:
     name: "Build IDE Plugin"
     runs-on: ubuntu-latest
--- a/.github/workflows/record-screenshot-baselines.yml
+++ b/.github/workflows/record-screenshot-baselines.yml
@@ -97,8 +97,9 @@ jobs:
           name: record-screenshot-baselines-diffs
           path: |
             android/desktop-core/build/reports/screenshots/
-            android/desktop-core/src/test/resources/screenshots/
+            android/desktop-core/src/test/resources/screenshots/*.png
           if-no-files-found: ignore
+          retention-days: 7
 
       - name: "Open PR with recorded baselines"
         shell: bash
--- a/.github/workflows/record-screenshot-baselines.yml
+++ b/.github/workflows/record-screenshot-baselines.yml
@@ -81,6 +81,25 @@ jobs:
           reuse-configuration-cache: true
           gradle-encryption-key: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
 
+      # A baseline that fails the verification immediately following its own
+      # recording (issue #3913) can only be diagnosed from the pixel diffs
+      # ScreenshotComparator writes as "<name>.diff.png" / "<name>.actual.png".
+      # Without this upload they die with the runner.
+      #
+      # The recorded baselines ride along: when verify fails the job stops
+      # before "Open PR", so the PNGs from the record step are discarded too,
+      # and a three-way baseline/actual/diff compare is what makes the
+      # mismatch legible.
+      - name: "Upload Screenshot Diffs"
+        if: failure()
+        uses: actions/upload-artifact@v6
+        with:
+          name: record-screenshot-baselines-diffs
+          path: |
+            android/desktop-core/build/reports/screenshots/
+            android/desktop-core/src/test/resources/screenshots/
+          if-no-files-found: ignore
+
       - name: "Open PR with recorded baselines"
         shell: bash
         env:
--- a/test/bats/screenshot-diff-artifact-upload.bats
+++ b/test/bats/screenshot-diff-artifact-upload.bats
@@ -13,14 +13,22 @@
 #     baselines it just recorded (the failure reported in #3913), and
 #   - desktop-core-unit-tests in pull_request.yml / merge.yml, which verify
 #     committed baselines on a fresh runner.
+#
+# merge.yml's desktop-core-unit-tests is currently `if: ${{ false }}`, so its
+# step is inert today. It is kept and pinned so the two workflows do not
+# silently diverge if that job is re-enabled.
 
 REPORT_PATH="android/desktop-core/build/reports/screenshots/"
+BASELINE_PATH="android/desktop-core/src/test/resources/screenshots/"
 
 # Extract a job block: from its "  <job>:" line up to (but not including) the
-# next top-level (2-space-indented) job key. awk, not sed -- BSD sed lacks the
-# range/address forms this needs.
+# next top-level (2-space-indented) job key. Capture only starts after the
+# top-level "jobs:" key, so a same-named key under "on:" cannot match. awk,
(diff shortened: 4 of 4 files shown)


kaeawc:
Review pass done — three orthogonal lenses (regression, test adequacy, API/contract) plus my own checks. Pushed 696023eb0 with the fixes.

**Fixed**

The step-block helper had a real, latent defect: `screenshot_upload_step` folded comment lines trailing a step into *that step's* block, so a comment mentioning the report path would let the **preceding** step satisfy every assertion. Reproduced against the real `pull_request.yml` before fixing. Blocks now close on a comment at step indent as well as on a step start. Also scoped `job_block` to after the top-level `jobs:` key, replaced the word-splitting `set -- $spec` (the repo's only occurrence) with a read loop matching `test/bats/pr-gate-wiring.bats`, and added the `-n` guards that sibling documents.

Two coverage gaps are now pinned: that `Open PR with recorded baselines` still follows the inserted upload, and that the screenshot artifact name differs from `desktop-core-test-results` — they share a job and condition, and upload-artifact@v4+ errors on duplicate names in one run. Guard is 12 tests, every one mutation-verified.

`if-no-files-found: ignore` was inert in the record workflow: `src/test/resources/screenshots/` is git-tracked and always holds `README.md`, so every unrelated job failure uploaded an artifact containing only that README. Scoped to `*.png`. Added `retention-days: 7` to all three, matching the repo's other diagnostic uploads.

**Rejected / noted, not changed**

`merge.yml`'s `desktop-core-unit-tests` is `if: ${{ false }}` ([merge.yml:40](https://github.com/kaeawc/auto-mobile/blob/main/.github/workflows/merge.yml#L40)), so the step added there is dead code today. Keeping it: the job body is still maintained (its existing `Upload Test Results` step is equally dead), and dropping mine would leave `merge.yml` silently diverged from `pull_request.yml` the moment anyone re-enables it. Now called out in the bats file header.

`Build Xcode Projects (Xcode 26.5)` fails on this PR at 13s — pre-existing infra, identical failure on already-merged #3969. This diff touches zero Swift.

**Deferred**

#3976 — `junit-runner-heap-dumps` is used as the artifact name by two different jobs in the same `pull_request.yml` run; upload-artifact@v4+ errors on that when both fail. Pre-existing, out of scope here.

Issue #3975, opened 2026-07-19
ci(ios): CtrlProxy XcodeGen drift check fails on every PR — committed pbxproj target order is not reproducible in CI


kaeawc:
`Check CtrlProxy XcodeGen Project Drift` fails on every PR that runs `XCTestRunner Simulator Tests`, which also fails the `iOS` aggregator job. Introduced by #3969 (merged 2026-07-19T20:10:45Z, 452c19368); **#3969's own checks are red for the same reason**, and it is already on main.

## Symptom

```
Error: XcodeGen project files are out of date after generation.
 M ios/control-proxy/CtrlProxy.xcodeproj/project.pbxproj
```

The diff is purely the `targets = (...)` ordering in the `PBXProject` section — no content change:

| | order |
|---|---|
| committed | CtrlProxyApp, ObjCExceptionCatcher, CtrlProxy, CtrlProxyTests, CtrlProxyUITests |
| CI generates | CtrlProxy, CtrlProxyApp, CtrlProxyTests, CtrlProxyUITests, ObjCExceptionCatcher |

CI's output is alphabetical; the committed file is not.

## Why it is confusing

The committed ordering **is** reproducible locally. On macOS with XcodeGen **2.46.0** (the same version `brew install xcodegen` resolves to in CI), `bash scripts/ios/xcodegen-drift-check.sh --ctrl-proxy` reports "XcodeGen project files are in sync" and `git status -- ios/` is clean. So the same version produces different target ordering locally vs on the CI runner — the ordering is environment-dependent rather than version-dependent.

Worth investigating whether XcodeGen's target ordering depends on filesystem enumeration or locale/collation, since that would make this recur for anyone regenerating the project on a different machine.

## Impact

Not currently merge-blocking: neither `iOS` nor `XCTestRunner Simulator Tests` is in the `green-main` required-status-checks list. But it puts two permanently-red checks on every PR touching the ios path filter, which is exactly how a real iOS regression gets ignored.

## Fix options

1. Regenerate and commit the pbxproj on a machine/CI job whose XcodeGen output matches CI, so the drift check passes.
2. Better: make the ordering deterministic across environments (pin XcodeGen exactly and/or normalize target order) so this cannot depend on where `xcodegen generate` was run.

Found while landing #3957 (a TypeScript-only lint change that touches no iOS file); ruled out as caused by that PR — the drift reproduces on the merge of current main and is present on main independently.

kaeawc:
[...]

PR #3977, opened 2026-07-19
ci: namespace junit-runner heap-dump artifacts by job, guard duplicate artifact names

### LINKED ISSUE #3976: Duplicate artifact name `junit-runner-heap-dumps` in two jobs of the same workflow run
Two `actions/upload-artifact@v6` steps in `.github/workflows/pull_request.yml` upload under the same artifact name from **different jobs in the same workflow run**:

- [`pull_request.yml:1599`](https://github.com/kaeawc/auto-mobile/blob/main/.github/workflows/pull_request.yml#L1599) — job `junit-runner-unit-tests`
- [`pull_request.yml:1668`](https://github.com/kaeawc/auto-mobile/blob/main/.github/workflows/pull_request.yml#L1668) — job `junit-runner-emulator-tests`

```
$ awk '/^  [a-z0-9-]+:/ {job=$0} /junit-runner-heap-dumps/ {print NR": "job}' .github/workflows/pull_request.yml
1599:   junit-runner-unit-tests:
1668:   junit-runner-emulator-tests:
```

`upload-artifact` v4 and later error on a duplicate artifact name within a single run (v3 merged them). Both steps are failure-gated, so this only bites when **both** jobs fail in the same run — at which point the second upload fails and adds an opaque second failure on top of the one you were actually trying to diagnose, and the heap dumps from one of the two jobs are lost.

Suggested fix: suffix each with its job, e.g. `junit-runner-heap-dumps-unit` / `junit-runner-heap-dumps-emulator`, or use `${{ github.job }}`.

Pre-existing; not introduced by any recent change. Surfaced during review of #3974 (which added screenshot-diff uploads for #3913) — that PR added a guard asserting its own two same-job uploads use distinct names, and the same class of collision exists here.

A wiring guard along the lines of `test/bats/screenshot-diff-artifact-upload.bats` could assert globally that no two `upload-artifact` steps in one workflow share a `name:`.

kaeawc:
## Summary

Two `actions/upload-artifact@v6` steps in `.github/workflows/pull_request.yml` uploaded under the same artifact name, `junit-runner-heap-dumps`, from different jobs in the same workflow run — `junit-runner-unit-tests` ([:1599](https://github.com/kaeawc/auto-mobile/blob/main/.github/workflows/pull_request.yml#L1599)) and `junit-runner-emulator-tests` ([:1668](https://github.com/kaeawc/auto-mobile/blob/main/.github/workflows/pull_request.yml#L1668)).

`upload-artifact` v4 and later error on a duplicate artifact name within a single run (v3 merged them). Both steps are `if: failure()`-gated, so the collision only fired when **both** junit-runner jobs failed in the same run — precisely when the dumps are most wanted. The second upload then failed too, stacking an opaque failure on top of the one being diagnosed, and one job's heap dumps were lost.

This suffixes each name with its job and adds a wiring guard that asserts the property globally, since this class of bug is invisible until it fires and fires only on an already-failing run.

## Linked Issue

Closes #3976

## Bug / Regression Context

- **Prior attempts:** none. Pre-existing; surfaced during review of #3974 (screenshot-diff uploads for #3913), which added a guard asserting its own two same-job uploads use distinct names.
- **Observed symptom:** when both junit-runner jobs fail in one run, the second `Upload Heap Dumps` step fails with a duplicate-artifact-name error and those heap dumps are not retained.
- **Repro conditions:** any run where `junit-runner-unit-tests` and `junit-runner-emulator-tests` both fail.

## Changes

- `junit-runner-unit-tests` now uploads `junit-runner-heap-dumps-unit`; `junit-runner-emulator-tests` uploads `junit-runner-heap-dumps-emulator`. Surrounding step wiring (`if: failure()`, `path`, `if-no-files-found: ignore`) is untouched.
- New `test/bats/workflow-artifact-name-uniqueness.bats` parses every `.github/workflows/*.yml`, collects the `with.name` of each `upload-artifact` step, and fails on any duplicate within a workflow.

## Acceptance Criteria

The issue carries no explicit AC list; these were inferred from its "Suggested fix" and closing paragraph, and confirmed at the plan gate.

| # | Criterion | Pinned by |
|---|---|---|
| 1 | The two heap-dump uploads use distinct, job-suffixed names | `junit-runner heap dumps are namespaced by job` |
| 2 | No two `upload-artifact` steps in one workflow share a `name:`, asserted globally | `no two upload-artifact steps in one workflow share an artifact name` |

Two supporting tests guard the guard: one pins that the extractor actually finds steps (an empty extractor would make every uniqueness assertion vacuously pass), and one pins that the rename left both steps failure-gated.

## Validation

- [x] New guard red before the change (AC-1 and AC-2 tests both failed, AC-2 reporting the real collision by name), green after
- [x] Full `bats test/bats/` suite: 351/352
- [x] `shellcheck --shell=bash` clean on the new file; workflow YAML parses
- [x] Confirmed nothing else in the repo downloads or references the old artifact name
- [x] `bun run typecheck` — green in CI (see note below)
- [ ] Android/iOS changes: N/A
- [ ] New code uses interfaces + fakes + FakeTimer: N/A (bats)

Both local reds turned out to be local-environment only — **CI is green on both**:

1. `test/bats/verify-runtime-deps.bats` — a runtime dep missing from my machine. The `BATS Shell Tests (macos-latest)` job passes.
2. `bun run typecheck` — an `ajv` / `ajv-formats` nested-copy resolution artifact in my local `node_modules`. All three `Node TypeScript Build and Test` legs pass, which is where the gate runs (`pull_request.yml:1084`).

The TypeScript suite was not run locally: this diff is YAML + bats only and has no TS surface to regress.

The three failing iOS checks are [#3975](https://github.com/kaeawc/auto-mobile/issues/3975) (XcodeGen pbxproj target order not reproducible in CI), unrelated to this diff, which touches no iOS files. They ran here only because editing a workflow file trips change-detection's "everything changed" path; they skip on ordinary PRs.

## Device Verification

N/A — CI workflow configuration.

## Follow-ups

Nothing new filed — the one CI failure is already tracked as #3975.

Two latent scope gaps in the guard were found in review and are documented in the file header rather than filed, since neither has a live collision today: reusable-workflow uploads land in the *caller's* artifact namespace (spanning two files), and a matrix job with a static artifact name would collide with itself at runtime. Both are invisible to a file-static check.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1596,7 +1596,7 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@v6
         with:
-          name: junit-runner-heap-dumps
+          name: junit-runner-heap-dumps-unit
           path: heap-dump
           if-no-files-found: ignore
 
@@ -1665,7 +1665,7 @@ jobs:
         if: failure()
         uses: actions/upload-artifact@v6
         with:
-          name: junit-runner-heap-dumps
+          name: junit-runner-heap-dumps-emulator
           path: heap-dump
           if-no-files-found: ignore
 
--- a/test/bats/workflow-artifact-name-uniqueness.bats
+++ b/test/bats/workflow-artifact-name-uniqueness.bats
@@ -1,0 +1,89 @@
+#!/usr/bin/env bats
+#
+# Guards issue #3976: two `actions/upload-artifact` steps in different jobs of
+# `pull_request.yml` uploaded under the same artifact name.
+#
+# upload-artifact v4 and later error on a duplicate artifact name within a
+# single workflow run (v3 silently merged them). Both heap-dump steps are
+# failure-gated, so the collision only bites when *both* junit-runner jobs fail
+# in the same run -- exactly when you most need the dumps. The second upload
+# then fails, adding an opaque failure on top of the one being diagnosed, and
+# one job's heap dumps are lost.
+#
+# The scope here is deliberately global rather than a pin on the two known
+# steps: the same class of collision is invisible until it fires, and it fires
+# only on an already-failing run.
+
+# Print "<job>\t<artifact name>" for every upload-artifact step in a workflow.
+#
+# Capture starts only after the top-level "jobs:" key so a same-named key under
+# "on:" cannot match, matching the convention in
+# screenshot-diff-artifact-upload.bats. awk, not sed -- BSD sed lacks the
+# range/address forms this needs.
+#
+# `name:` is read as the first name key following the `uses:` line, which is
+# the `with.name` input; the *step's* own display name precedes `uses:` and so
+# is never picked up.
+upload_artifact_names() {
+  awk '
+    /^jobs:/ { in_jobs = 1; next }
+    !in_jobs { next }
+    /^  [A-Za-z0-9_-]+:/ { job = $1; sub(":", "", job) }
+    /uses: actions\/upload-artifact/ { pending = 1; next }
+    pending && /^ *name: / {
+      artifact = $0
+      sub(/^ *name: /, "", artifact)
+      print job "\t" artifact
+      pending = 0
+    }
+  ' "$1"
+}
+
+workflows() {
+  find .github/workflows -maxdepth 1 -name '*.yml' | sort
+}
+
+@test "the extractor finds upload-artifact steps at all" {
+  # A silently-empty extractor would make every uniqueness assertion below
+  # vacuously pass. Pin that it actually parses the workflow the issue is about.
+  run upload_artifact_names ".github/workflows/pull_request.yml"
+  [ "$status" -eq 0 ]
+  [ "${#lines[@]}" -gt 5 ]
+}
+
+@test "no two upload-artifact steps in one workflow share an artifact name" {
+  duplicates=""
+  while read -r workflow; do
+    dupes="$(upload_artifact_names "$workflow" \
+      | cut -f2 \
+      | sort \
+      | uniq -d)"
+    if [ -n "$dupes" ]; then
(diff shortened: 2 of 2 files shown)


kaeawc:
Reviewed through three orthogonal lenses (regression, test adequacy, API/interface surface). Two findings were confirmed and fixed in 04bccff; the rest were checked and cleared.

**Fixed.** The extractor read `uses:` then `name:` sequentially, so a step written `with:` before `uses:` was skipped — and its `uses:` then latched onto the *next* step's name, shifting attribution. A duplicate written that way would have passed the guard that exists to catch it. Quoted `uses:` values were missed the same way. Now whole step blocks are buffered and scanned at flush time. Verified the rewrite keeps exact parity with a raw grep across all 13 workflows (pull_request 30/30, merge 18/18) and still goes red on revert.

**Documented rather than filed.** Two axes a file-static check cannot reach, neither with a live collision today: reusable-workflow uploads land in the *caller* 's artifact namespace, so a caller/callee collision spans two files; and a matrix job with a single static artifact name collides with itself across legs. Both are now stated in the file header so the guard is not over-trusted.

**Cleared.** Nothing anywhere downloads the old artifact name (no `download-artifact`, `gh run download`, or `workflow_run` consumers), so the rename is externally invisible. The file is portable to both CI legs — re-ran the logic under bash 3.2 with BSD `find`/`cut`/`sort`/`uniq`/`awk`. The overlap with the within-job assertion in `screenshot-diff-artifact-upload.bats` is partial: that test also positively pins step identity, which a global uniqueness check cannot express, so both stay.

The three failing iOS checks are #3975 (pbxproj target order not reproducible in CI), unrelated to this diff — it touches no iOS files. They ran here only because editing a workflow file trips change-detection's everything-changed path. No required check is failing.

Issue #2966, opened 2026-07-03
ktfmt: enforce the pinned formatter version so version drift can't silently land on red main


kaeawc:
## Context

Follow-up from PR #2941 (merged as 69c2a47), which scoped the **PR** ktfmt check (`fast-validation` in `.github/workflows/pull_request.yml`) to only a PR's changed files, leaving the **full-tree** ktfmt check running post-merge in `.github/workflows/merge.yml` (`ktfmt` job, `on: push: branches: main`).

A 3-perspective review of that PR flagged this as the one substantive guarantee the scoping trades away.

## The problem

Scoping the PR check to changed files is sound for *file-content* drift (a file's formatting only changes when the file is touched). It is **not** sound for **formatter-version / style-config drift**, because that reformats *untouched* files the scoped PR check never looks at.

Two concrete gaps mean the `0.64` pin is not actually enforced:

1. **macOS installer ignores the pin.** `scripts/ktfmt/install_ktfmt.sh` pins `KTFMT_VERSION="0.64"`, but the macOS path runs `brew install ktfmt` with **no version constraint** — it installs whatever the current brew formula is.
2. **An already-installed newer ktfmt is used as-is.** `scripts/ktfmt/validate_ktfmt.sh` only installs when ktfmt is *missing* (`INSTALL_KTFMT_WHEN_MISSING`), and the only version guard is a `--google-style` *probe* (validate_ktfmt.sh:64) that passes for any version supporting the flag. A 0.66 that formats differently sails through.

### Why it matters now (regression risk introduced by #2941)
Before #2941, the full-tree ktfmt ran on every PR, so a contributor whose local/CI ktfmt drifted from 0.64 would be caught **on the PR**. Now, with the PR check scoped, version skew reformats only their touched files (which pass), and the mismatch first surfaces **post-merge on `main`** when `merge.yml` reformats the whole tree — i.e. **red main**, not a blocked PR. On the pinned Linux CI runner this is fine *today* only because ktfmt is downloaded fresh at 0.64; the guardrail is "the image happens to ship no ktfmt," not "the version is verified."

## Suggested work (pick one or combine)

- **Fingerprint gate (strongest):** in `validate_ktfmt.sh`, assert `ktfmt --version` matches `KTFMT_VERSION`, and/or force a full-tree run when the ktfmt version or style config differs from the pinned value. This makes tool drift fail loudly regardless of scoping.
[...]

Issue #3779, opened 2026-07-12
CI: spike — validate GitHub Actions parallel-steps (background/wait/parallel) behavior before adoption


kaeawc:
## Summary

GitHub Actions shipped **parallel steps within a job** on [2026-06-25](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/): the `background: true`, `wait` / `wait-all`, `cancel`, and `parallel` keywords let independent steps run concurrently on the **same runner**, sharing one workspace. This is distinct from the job-level parallelism (`needs` + `matrix`) our CI already uses heavily.

Before we adopt it in real jobs (see the follow-up issues), we need a throwaway spike to nail down behavior the changelog does **not** document. The answers gate how — and whether — we standardize this.

## Questions to answer

1. **Do backgrounded steps' `outputs` survive a `wait`?** i.e. can a later step read `steps.<id>.outputs.foo` from a step that ran with `background: true`? This gates the iOS "boot sim in background" win, where the test reads `steps.ctrlproxy-simulator.outputs.simulator_udid`.
2. **Do `background` / `wait` / `parallel` work inside a composite action's `steps:`,** or only at the job level? This is the pivotal question for standardization — if composite steps support them, we can bake parallelism into reusable actions; if not, every call site must inline the keywords.
3. **Failure semantics:** if a backgrounded step fails, does the job fail at the `wait`? Does `continue-on-error` compose? How does a failed background step interact with `if: always()` cleanup steps?
4. **Log interleaving & timing:** how are concurrent step logs presented in the UI, and does per-step timing still report correctly (important since several of our steps carry tuned `timeout-minutes`)?
5. **CPU contention on our runners:** confirm the "true parallel" caveat on `ubuntu-latest` (~4 vCPU) and `macos-26` — the win is real for I/O/wait-bound work (installs, device boot) and marginal-to-negative for two CPU-bound `xcodebuild`/`gradle` steps.

## Deliverable

A short throwaway workflow (or a draft PR touching a single low-risk job) that exercises each keyword and records the answers in this issue. No production workflow changes here.

## Context

- Native syntax reference: [workflow syntax docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions)
- Do **not** confuse the native keywords with the older third-party `Parallel Steps` marketplace action — we want the native feature.

Blocks the two adoption issues (iOS install fan-out; iOS background sim boot) and the standardization decision.

[...]

PR #3981, opened 2026-07-19
fix(ci): pin XcodeGen to a release artifact so generated projects cannot drift

### LINKED ISSUE #3975: ci(ios): CtrlProxy XcodeGen drift check fails on every PR — committed pbxproj target order is not reproducible in CI

kaeawc:
Fixes #3975.

## Symptom

Every PR running `XCTestRunner Simulator Tests` failed `Check CtrlProxy XcodeGen Project Drift`, which also failed the `iOS` aggregator. The diff was purely `targets = (...)` ordering — no content change:

| | order |
|---|---|
| committed | CtrlProxyApp, ObjCExceptionCatcher, CtrlProxy, CtrlProxyTests, CtrlProxyUITests |
| CI generated | CtrlProxy, CtrlProxyApp, CtrlProxyTests, CtrlProxyUITests, ObjCExceptionCatcher |

## Root cause

CI ran a bare `brew install xcodegen`, which resolves against whatever formula index the runner image ships. From the CI log:

```
Downloading .../xcodegen/manifests/2.45.4
🍺  /opt/homebrew/Cellar/xcodegen/2.45.4
```

GitHub's macos-26 image pours **2.45.4**; contributors on current Homebrew have **2.46.0**. The two order the `PBXProject.targets` array differently, so a project generated by one reads as drift to the other — and the failure presents as a stale project file rather than a toolchain skew. This is the same bug class as #2966 (ktfmt).

## Fix: install the pinned release artifact, not Homebrew

The vendor archive carries the version in its URL, so the install is **version-exact by construction**. Homebrew cannot be: there is no `xcodegen@<version>` formula, so pinning through it would hard-fail all six wired iOS jobs whenever the formula moves.

It is also cheaper — ~4 MB archive versus the ~52 MB index refresh `brew update` needs to defeat a stale index — and the fast path skips the download entirely.

Mirrors the existing `scripts/ktfmt/ktfmt_version.sh` precedent: a *sourced* `scripts/ios/xcodegen_version.sh` holds the pin plus `installed_xcodegen_version()` and `require_pinned_xcodegen_version()`, so the pin is **enforced, not merely declared**.

## Pins the contributor paths too — not just CI

This is the half of the skew that actually caused #3975, and every one of these ran bare `brew install`:

| Path | Why it matters |
|---|---|
| `scripts/ios/xcodegen-generate.sh` | **the drift check invokes it** — the check could self-install an arbitrary version behind the pin's back |
| `scripts/ios/xcode-build.sh` | regenerates every `ios/*/project.yml` in place |
| `scripts/ios/ctrl-proxy-build-for-testing.sh` | |
| `scripts/ios/ctrl-proxy-create-ipa.sh` | |
| `ios/CLAUDE.md:17` | told contributors and agents to `brew install xcodegen` |

**Correction:** an earlier revision of this description claimed these were pinned when three of them were not. Review caught it. They gated the install on `! command -v xcodegen`, so a contributor who *already had* 2.45.4 — exactly the #3975 shape — skipped the installer entirely and regenerated skewed. All three now check the **version**, not mere presence. `xcode-build.sh` additionally dropped a vestigial `command -v brew` guard and, on a skew, warns and uses the committed projects rather than regenerating, since that path deliberately tolerates a missing generator.

Both generation paths (`--all` via `xcodegen-generate.sh`, and `--ctrl-proxy` which invokes `xcodegen` directly) gate on the pinned version **before writing**.

## The digest is pinned, not just the version

GitHub release assets are mutable. A re-uploaded `xcodegen.zip` under tag `2.46.0` would still report `2.46.0`, pass the version gate, and generate different bytes — #3975 with the guard reporting green. `XCODEGEN_RELEASE_SHA256` is verified before extraction (`shasum`/`sha256sum` fallback). Verified against the live asset: 4,278,764 bytes, `4d9e34b6…6806`; a tampered archive exits 1 with nothing installed.

This is *not* a supply-chain claim — 0 of 8 peer installers in `scripts/` verify a digest, so the original approach matched convention. It's a reproducibility fix that follows from this PR's own thesis.

## A successful install used to fail its caller's gate

`install-xcodegen.sh` exports PATH in its own process, but `xcodegen-generate.sh` runs it as a *subprocess* and then gates in the parent. When the install landed in the `${HOME}/.local` fallback, the parent found no `xcodegen` and exited 1 — telling the user to run the installer that had just succeeded. CI never hit this (`/usr/local` is writable there); it struck only the contributor the fallback exists to serve. `xcodegen_version.sh` now exports `~/.local/bin` for everything that sources it, mirroring `install_ktfmt.sh`.

## Also

- Drops the Homebrew cache in `build-ctrl-proxy-ios-ipa.yml`. It cached the Cellar but **not** the `bin` symlink, so a cache hit never satisfied `command -v` detection; and its immutable key froze an index that `brew update` then discarded. It was paying restore cost for negative benefit.
- The drift check's error now reports the pinned version. Since the gate runs before generation, skew can no longer explain reaching that line.

## Tests

`test/bats/install-xcodegen.bats` — 9 tests with `curl`/`unzip`/`xcodegen` stubbed, so they never hit the network or touch a real install: happy-path skip, install-when-absent, the 2.45.4→2.46.0 replacement, the URL carrying the pin, post-install verification failure, `GITHUB_PATH` propagation, and both directions of the gate.

## Verification

- Confirmed the release asset: HTTP 200, 4,278,764 bytes, reports exactly `Version: 2.46.0`.
- 9/9 BATS pass · shellcheck clean · shell-portability clean · set-e gate clean (restructured rather than baselined — the redundant conditional came out, since the installer already no-ops when the pin matches).
- `xcodegen-drift-check.sh --ctrl-proxy` reports **in sync** locally.
- All four modified workflow YAMLs parse.

## Follow-up

`swiftlint` and `swiftformat` declare pins that are **dead code on macOS** — `scripts/lib/tool-install.sh:95` routes to a bare `brew install` and returns 0 without consulting them, so `apply_*` can rewrite committed Swift at whatever version brew ships. Same bug, loaded and unfired. Filed separately rather than expanding this PR.


scripts/ios/xcodegen_version.sh
```diff
@@ -0,0 +1,54 @@
...
+# check with an ordering-only diff that looks identical to a genuinely stale
+# project file. A skewed generator must fail loudly, before it writes anything.
+
+XCODEGEN_VERSION="2.46.0"
+
+# The vendor's prebuilt release archive. The version is in the URL, so this is
+# version-exact by construction -- unlike Homebrew, which cannot install an
+# arbitrary older version and drifts whenever its formula moves.
+# shellcheck disable=SC2034  # consumed by scripts that source this file
+XCODEGEN_RELEASE_URL="https://github.com/yonaskolb/XcodeGen/releases/download/${XCODEGEN_VERSION}/xcodegen.zip"
```
kaeawc, line 26: Not a blocker, and explicitly not a supply-chain objection — every peer installer in `scripts/` fetches this way. But "version-exact by construction" is doing slightly more work than the URL can support, and it matters *for this PR's own thesis*.

A GitHub release asset is mutable: a maintainer can delete and re-upload `xcodegen.zip` under the same `2.46.0` tag. The tag pins provenance, not bytes. If that ever happens, `require_pinned_xcodegen_version` still passes — it compares `xcodegen --version`, which the replacement build would also report as `2.46.0` — and CI regenerates with a different binary. The failure mode is an ordering-only diff in `project.pbxproj` that reads as a stale project file, which is exactly #3975 again, with the one gate that was supposed to catch it reporting green.

A content hash is the only thing that closes that loop, and it's ~3 lines. The repo already has the vocabulary: hex-64 pins alongside a version, and the `sha256sum`-with-`shasum -a 256` fallback for stock macOS in `scripts/ci/verify-transit-sha256.sh:16-21`. Since `install-xcodegen.sh` already has the file on disk between the `curl` and the `unzip`, `scripts/ci/verify-transit-sha256.sh <file> <expected>` drops straight in, or inline it here to keep the pin and its hash adjacent.

The current asset hashes to `4d9e34b62172d645eed6457cac13fc222569974098ef4ee9c3368bedf0196806` (4,278,764 bytes, HTTP 200) — verified against the live URL just now.

```sh
curl -fsSL -o /tmp/xg.zip https://github.com/yonaskolb/XcodeGen/releases/download/2.46.0/xcodegen.zip
shasum -a 256 /tmp/xg.zip
```

Suggested change for byte-level reproducibility, so a re-cut release asset fails loudly instead of silently reintroducing #3975:

```suggestion
XCODEGEN_RELEASE_URL="https://github.com/yonaskolb/XcodeGen/releases/download/${XCODEGEN_VERSION}/xcodegen.zip"

# Release assets are mutable -- a re-uploaded asset under the same tag would
# still report ${XCODEGEN_VERSION}, so the version gate cannot catch it. Pin the
# bytes too, or a swapped archive silently regenerates the ordering-only diff
# that issue #3975 exists to prevent. Bump alongside XCODEGEN_VERSION.
# shellcheck disable=SC2034  # consumed by scripts that source this file
XCODEGEN_SHA256="4d9e34b62172d645eed6457cac13fc222569974098ef4ee9c3368bedf0196806"
```

If you'd rather not, dropping "by construction" from the comment above would be enough — the claim is what I'd want a future reader not to rely on. Either way, `test/bats/install-xcodegen.bats` already stubs `curl`, so a hash assertion would slot into the existing fixtures cheaply.

kaeawc (commented):
Reviewed through one lens: supply-chain safety and repo convention for fetching executables. Short version — replacing `brew install` with a `curl` of a release zip is **consistent with this repo's convention**, not a departure from it, and nothing in CI will newly fail. One non-blocking suggestion below, and it is a reproducibility argument rather than a security one.

The things I checked and dropped, so they don't get re-litigated later:

- **Losing Homebrew's bottle checksum isn't a regression against peer practice.** Zero of the eight tool installers in `scripts/` verify a hash: `scripts/ktfmt/install_ktfmt.sh:172`, `scripts/swiftformat/install_swiftformat.sh:48`, `scripts/swiftlint/install_swiftlint.sh:48`, `scripts/hadolint/install_hadolint.sh:139`, `scripts/lychee/install_lychee.sh:184`, `scripts/ripgrep/install_ripgrep.sh:44`, `scripts/act/install_act.sh:47`, `scripts/android/install-ew-cli.sh:91`. `scripts/lib/tool-install.sh` exports no checksum helper at all. `install_ktfmt.sh` — the precedent this PR names — curls a GitHub release jar pinned by version-in-URL and its only post-download check is `[[ -s ... ]]` at `:187`, then gates on `--version`. This PR mirrors that faithfully.
- **The sha256 machinery that does exist is first-party only.** `RELEASE_CHECKSUM_REGISTRY` (`src/constants/release.ts:52`) and `scripts/ci/verify-{artifact,transit,release-integrity}-sha256.sh` cover AutoMobile's *own* published apk/ipa/runner/jar, enforced on download at `src/utils/CtrlProxyManager.ts:264` and `src/utils/IOSCtrlProxyBuilder.ts:803`. Different category — the repo controls both ends there. It was never applied to third-party tooling.
- **`curl -fsSL` following redirects is not a realistic concern here.** The 302 goes to `https://release-assets.githubusercontent.com/...` with a signed, expiring token; both hops are HTTPS with cert validation. An attacker-controlled redirect requires compromising GitHub, at which point the pinned tag is equally compromised. Not worth a `--proto '=https'` hardening comment.
- **No security workflow will react.** CodeQL (`pull_request.yml:825`) is `language: [typescript]` only — no `actions` language, and it's skipped entirely unless TS changed. `bun-audit` (`:881`) is `|| true` and covers npm deps. No semgrep/trivy/zizmor/gitleaks/scorecard anywhere. Of the three shell gates, the only curl rule is `curl-no-fail` in `scripts/shellcheck/validate_shell_portability.sh:76`, which wants `-f` — already satisfied.

So: no blocking supply-chain finding. Worth landing as-is.

scripts/ios/xcode-build.sh
```diff
@@ -114,7 +114,7 @@ fi
 if ! command -v xcodegen &> /dev/null; then
     if command -v brew &> /dev/null; then
         print_info "Installing xcodegen..."
-        brew install xcodegen || echo "xcodegen install failed, continuing with committed projects"
+        "${SCRIPT_DIR}/install-xcodegen.sh" || echo "xcodegen install failed, continuing with committed projects"
```
kaeawc: The install is still gated on `! command -v xcodegen`, so it only fires when xcodegen is **absent** — a contributor who already has 2.45.4 skips it entirely and line 131 runs `xcodegen generate` with the skewed binary. That is precisely the #3975 shape the PR body says this row fixes, and the same holds at `ctrl-proxy-create-ipa.sh:69` and `ctrl-proxy-build-for-testing.sh:85`. Only `xcodegen-generate.sh` and the `--ctrl-proxy` drift path actually call `require_pinned_xcodegen_version`; these three write project files with no gate at all. This should block — it is the half of the bug the PR claims to close.

Secondarily, this branch is still wrapped in `if command -v brew`, which is now backwards: `install-xcodegen.sh` is curl+unzip and needs no Homebrew, so on a brew-less machine the installer is skipped for no reason. The surrounding "Attempting to install via brew..." messages in the other two scripts are stale for the same reason.

Repro, no network needed: put a stub `xcodegen` printing `Version: 2.45.4` first on `PATH`, then run `scripts/ios/ctrl-proxy-create-ipa.sh` — it walks straight past line 69 to the `xcodegen generate` at line 85.

Suggested change for closing the contributor half of #3975 — source `xcodegen_version.sh` in each of the three and call `require_pinned_xcodegen_version` immediately before the `xcodegen generate`, so presence-but-skew fails loudly rather than writing a reordered `targets` array.

test/bats/install-xcodegen.bats
```diff
@@ -0,0 +1,137 @@
...
+
+@test "require_pinned_xcodegen_version accepts the pinned generator" {
+  stub_xcodegen "2.46.0"
+  run env PATH="$STUB_DIR:/usr/bin:/bin" bash -c \
+    "source '$VERSION_FILE'; require_pinned_xcodegen_version; echo OK"
+  [ "$status" -eq 0 ]
+  [[ "$output" == *"OK"* ]]
+}
+
+@test "the pin matches the version that generated the committed project" {
```
kaeawc: This one doesn't do what its name says. The three assertions are: the pbxproj contains the string `objectVersion`, `XCODEGEN_VERSION` is non-empty, and `XCODEGEN_RELEASE_URL` contains `XCODEGEN_VERSION` — none of which reference the version that generated the committed project. The third is self-referential by construction, since the URL is built from the variable two lines above it in `xcodegen_version.sh`.

I bumped the pin to a nonexistent `9.99.9` in a scratch copy and replayed the body verbatim: all three assertions still pass. So a bump-without-regenerate — the exact failure the comment says it guards — sails through, and answers the standing question of whether anything ties the pin to the pbxproj: nothing does.

Either make it real (record the generating version alongside the project — e.g. a committed `ios/control-proxy/.xcodegen-version` written by the generate path, asserted equal to `XCODEGEN_VERSION` here) or rename it to what it actually checks (`the release URL carries the pin`) so it stops reading as coverage that isn't there. Not a blocker on its own, but as written it is a green light with nothing behind it.

test/bats/release-workflow-wiring.bats
```diff
@@ -96,15 +96,32 @@
...
+                  .github/workflows/pull_request.yml \
+                  .github/workflows/merge.yml \
+                  .github/workflows/nightly.yml; do
+    ! grep -Fq "brew install xcodegen" "$workflow"
+    grep -Fq "scripts/ios/install-xcodegen.sh" "$workflow"
+  done
+}
 
-  [[ "$untap_line" -lt "$install_line" ]]
+@test "no script installs XcodeGen via bare brew" {
```
kaeawc, line 123: Checked the deletion this replaces, and the untap half is genuinely moot: `build-ctrl-proxy-ios-ipa.yml` now contains zero `brew` invocations (`gem install xcpretty` at line 44 is RubyGems), so there is nothing left for the aws tap to warn about and no ordering constraint to preserve. Good call replacing rather than patching it.

One assertion did get dropped without a successor though — the old test also asserted `! grep -Fq "HOMEBREW_NO_REQUIRE_TAP_TRUST"`, which #3551 added deliberately to stop someone silencing the tap warning with the env var instead of untapping. That guard is independent of xcodegen and costs one line to keep. Not a blocker.

Also worth noting the loop hard-codes four workflow paths, so a fifth workflow that adds a bare `brew install xcodegen` isn't caught. The sibling script test below uses `grep -r`; the workflow one could too.

test/scripts/xcodegenDriftCheck.test.ts
```diff
@@ -67,6 +82,11 @@ esac
     `#!/bin/bash
 set -euo pipefail
 
+if [[ "$*" == "--version" ]]; then
+    echo "Version: \${FAKE_XCODEGEN_VERSION:-${pinnedXcodegenVersion}}"
```
kaeawc, line 86: This is a legitimate contract update, not a weakening — worth saying plainly since it's the change most likely to look like one. The fake still rejects everything outside `generate` and `--version` (the `!= "generate"` arm below is intact), and defaulting `FAKE_XCODEGEN_VERSION` to the real pin doesn't neuter the gate, because the new skew test drives it to `2.45.4` and asserts both the exit-1 and that the pbxproj was left untouched. Reading the pin out of `xcodegen_version.sh` instead of hard-coding it is the right call.

The gap is on the other scope: `createTempRepo` writes its own stub over `scripts/ios/xcodegen-generate.sh`, so the `require_pinned_xcodegen_version` this PR adds to the **real** `xcodegen-generate.sh` — the repo-wide write path — has no coverage at all. Only the `--ctrl-proxy` gate is exercised. A `--all` counterpart, or a small direct bats test that runs `xcodegen-generate.sh` with a skewed stub on `PATH` and asserts it refuses before writing, would close it.

kaeawc (commented):
Reviewed through one lens: did the follow-up commit that repaired the three broken suites weaken coverage to get CI green?

Mostly no. `xcodegenDriftCheck.test.ts` is a faithful contract update, and the deleted aws-tap bats test really is moot. But two gaps are worth fixing before merge: three contributor scripts only pin when `xcodegen` is *absent*, so the exact #3975 contributor is untouched by the PR's headline claim; and the one test that promises to tie the pin to the committed pbxproj asserts nothing of the kind.

Ran locally: `bun test test/scripts/xcodegenDriftCheck.test.ts` (8 pass), `bats test/bats/install-xcodegen.bats test/bats/release-workflow-wiring.bats` (25 pass), shellcheck clean on all four scripts both with and without source resolution.

scripts/ios/xcode-build.sh
kaeawc: The PR body lists this file under "pins the contributor paths too — regenerates projects in place", but after the swap it still doesn't. Two things: the install only fires when `command -v xcodegen` finds *nothing* (line 114), so a laptop carrying brew's 2.45.4 skips it entirely; and then line 122 unconditionally runs `xcodegen generate` over every `ios/*/project.yml` with no `require_pinned_xcodegen_version` gate — this file never sources `xcodegen_version.sh` at all. That's exactly the #3975 shape: a skewed generator writing project files in place, no version named.

The `command -v brew` wrapper on line 115 is now actively wrong on top of that — the installer is a curl+unzip and needs no Homebrew, so gating on brew means a brew-less machine silently never installs.

The new `no script installs XcodeGen via bare brew` bats test passes here, because the literal string is gone while the behaviour it was guarding isn't.

Repro: `PATH` a 2.45.4 xcodegen ahead of the pinned one, run `./scripts/ios/xcode-build.sh` — it regenerates and dirties the pbxproj instead of failing.

Suggested change for closing the contributor half of #3975 — source the version file next to `SCRIPT_DIR` on line 30, then:

```suggestion
if [ "$(installed_xcodegen_version)" != "${XCODEGEN_VERSION}" ]; then
    print_info "Installing pinned xcodegen ${XCODEGEN_VERSION}..."
    "${SCRIPT_DIR}/install-xcodegen.sh" || echo "xcodegen install failed, continuing with committed projects"
    hash -r 2>/dev/null || true
fi
```

and gate the generate loop with `require_pinned_xcodegen_version`. Not a CI blocker — CI installs the pin in a prior step — but it should block on the PR's own stated scope.

scripts/ios/ctrl-proxy-create-ipa.sh
```diff
@@ -68,13 +68,7 @@ fi
 
 if ! command -v xcodegen &> /dev/null; then
     echo -e "${YELLOW}Warning: xcodegen not found. Attempting to install via brew...${NC}"
```
kaeawc: Same install-when-absent-not-when-skewed shape as `xcode-build.sh`, and the `Warning: xcodegen not found. Attempting to install via brew...` line just above now names a tool the installer deliberately doesn't use.

Smaller blast radius here — line 82's `if [ ! -d "${XCODEPROJ}" ]` means generation only happens when the committed project is missing, which never holds in CI (`build-ctrl-proxy-ios-ipa.yml:48`, after the install step at :35). So this is a correctness-of-message and contributor-path nit, not a CI break. Same applies at `ctrl-proxy-build-for-testing.sh:85`.

```suggestion
if [ "$(installed_xcodegen_version)" != "${XCODEGEN_VERSION}" ]; then
    echo -e "${YELLOW}Installing pinned XcodeGen ${XCODEGEN_VERSION}...${NC}"
    "${SCRIPT_DIR}/install-xcodegen.sh"
fi
```

.github/workflows/pull_request.yml
```diff
@@ -1369,7 +1369,7 @@ jobs:
       # Validated by the #3779 spike: outputs unused here, `- wait: [ids]` list
       # barrier works, and an install failure surfaces loudly at the wait.
       - name: "Install XcodeGen"
-        run: brew install xcodegen
+        run: ./scripts/ios/install-xcodegen.sh
```
kaeawc, line 1372: Not a blocker, and I confirmed this site works today — but worth knowing what it rests on. `install-xcodegen.sh:56` writes `GITHUB_PATH` from a `background: true` step, which is the one PATH-propagation case the #3779 spike didn't cover (it validated outputs and the `wait` barrier, not env-file writes). The reason the drift check at line 1391 finds xcodegen anyway is that the installer landed in `/usr/local/bin`, which is already on the runner's default PATH — the `GITHUB_PATH` append is doing nothing here.

That matters because `xcodegen-drift-check.sh --ctrl-proxy` is the only consumer in the tree with no self-install fallback: it calls `require_pinned_xcodegen_version`, which `exit 1`s. Every other consumer (`xcodegen-generate.sh`, `xcode-build.sh`, the two ctrl-proxy scripts) installs for itself. So if `/usr/local/bin` ever stops being writable on a future image, `install-xcodegen.sh:33-36` falls back to `$HOME/.local/bin` — *not* on the default PATH — and this job's recovery depends entirely on a backgrounded step's `GITHUB_PATH` write taking effect across the `wait`.

Cheap insurance either way: have the installer echo the resolved `PREFIX` into a step output, or just don't background this one — the fallback failure would at least be a loud `version mismatch: found 'none'` rather than anything silent, so this is a robustness note, not a bug.

kaeawc (commented):
Reviewed through one lens: CI/workflow wiring completeness and cross-step semantics. All six workflow sites resolve correctly — I confirmed that against the real run rather than reasoning about it: run 29704386451, job 88238822412 logs `Installing XcodeGen 2.46.0 (found 'none') into /usr/local...` then `XcodeGen 2.46.0 installed to /usr/local/bin/xcodegen`, and after `- wait: [install-xcodegen, install-xcpretty]` the `--ctrl-proxy` drift check reports `XcodeGen project files are in sync`. So `/usr/local/bin` is writable without sudo on macos-26, and it is already on the default PATH — the `GITHUB_PATH` append is belt-and-braces on the happy path, including from the backgrounded step.

Cache/untap removals check out too: `build-ctrl-proxy-ios-ipa.yml` has no `brew` invocation left at all, the Ruby gems cache keys `~/.gem` and `gem install xcpretty` uses system Ruby, so neither the deleted Homebrew cache nor the deleted `brew untap aws/tap` was load-bearing for anything downstream. No missed job either — `nightly.yml:519` and `pull_request.yml:1406` run `ctrl-proxy-build-for-testing.sh` with no install step, but that script only needs xcodegen when the xcodeproj is absent and it's committed.

Two gaps below, both on the *contributor* half the PR sets out to fix rather than the CI half. Local checks: shellcheck clean on all seven scripts, 25/25 bats, 8/8 `xcodegenDriftCheck.test.ts`.

scripts/ios/install-xcodegen.sh
```diff
@@ -0,0 +1,67 @@
...
+
+# The archive expands to xcodegen/{bin,share}. The binary resolves its bundled
+# templates relative to itself, so share/ must land alongside bin/.
+mkdir -p "${PREFIX}/share"
+cp "${WORK_DIR}/xcodegen/bin/xcodegen" "${PREFIX}/bin/xcodegen"
+rm -rf "${PREFIX}/share/xcodegen"
+cp -R "${WORK_DIR}/xcodegen/share/xcodegen" "${PREFIX}/share/xcodegen"
+chmod +x "${PREFIX}/bin/xcodegen"
+
+export PATH="${PREFIX}/bin:${PATH}"
```
kaeawc, line 88: `export PATH` only mutates this script's own process, and `xcodegen-generate.sh:55` runs it as a subprocess — so when the install lands in the `${HOME}/.local` fallback, the parent's `require_pinned_xcodegen_version` at `xcodegen-generate.sh:60` sees no `xcodegen` at all and exits 1, telling the user to run the installer that just succeeded. `hash -r` doesn't help; there's no cached entry to drop, the directory simply isn't on the parent's PATH. In CI this is masked because `/usr/local/bin` is writable and already on PATH, so it only bites a contributor whose `/usr/local` needs sudo — exactly the case the fallback exists to serve.

Repro (stub `curl`, force the fallback by pointing `XCODEGEN_PREFIX` at an unwritable dir):
```
mkdir -p /tmp/ro/bin && chmod 555 /tmp/ro/bin
cat > /tmp/p.sh <<'EOF'
source scripts/ios/xcodegen_version.sh
scripts/ios/install-xcodegen.sh
hash -r; require_pinned_xcodegen_version
EOF
XCODEGEN_PREFIX=/tmp/ro bash /tmp/p.sh
# XcodeGen 2.46.0 installed to /…/.local/bin/xcodegen.
# XcodeGen version mismatch: found 'none', this repo pins '2.46.0'.
# Install the pinned version: bash scripts/ios/install-xcodegen.sh   <- just did
```

Since the fallback prefix is a fixed constant, the cheapest fix is for the callers to prepend it before the gate — in `xcodegen-generate.sh` (and the other three `install-xcodegen.sh` call sites) add `export PATH="${HOME}/.local/bin:${PATH}"` ahead of `require_pinned_xcodegen_version`. If you'd rather keep the prefix knowledge in one place, have the installer print the resolved prefix on stdout and let callers consume it. This should block — it turns a supported install path into a dead end.

scripts/ios/install-xcodegen.sh
```diff
@@ -0,0 +1,67 @@
...
+if [[ -n "${GITHUB_PATH:-}" ]]; then
+    echo "${PREFIX}/bin" >> "${GITHUB_PATH}"
+fi
+# Bash caches resolved command paths; drop a stale entry for a previous xcodegen.
+hash -r 2>/dev/null || true
+
+require_pinned_xcodegen_version
+
+echo "XcodeGen ${XCODEGEN_VERSION} installed to ${PREFIX}/bin/xcodegen."
+if [[ ":${PATH}:" != *":${PREFIX}/bin:"* ]]; then
```
kaeawc: This warning can never fire: line 54 already did `export PATH="${PREFIX}/bin:${PATH}"`, so `":${PATH}:"` is guaranteed to contain `":${PREFIX}/bin:"` by the time the test runs.

```
$ bash -c 'PREFIX=/some/prefix; PATH=/usr/bin:/bin; export PATH="${PREFIX}/bin:${PATH}"
  [[ ":${PATH}:" != *":${PREFIX}/bin:"* ]] && echo fires || echo unreachable'
unreachable
```
I also confirmed it empirically — installing into a temp prefix that was definitely not on PATH printed no warning.

Worth fixing because this is precisely the message that would rescue the finding above: a contributor who lands in the `~/.local` fallback gets no hint to add it to their PATH. Capture PATH before line 54 and test that instead.

Suggested change for reachability:
```suggestion
if [[ ":${PATH_BEFORE_INSTALL}:" != *":${PREFIX}/bin:"* ]]; then
```
(with `PATH_BEFORE_INSTALL="${PATH}"` captured just above line 54).

scripts/ios/install-xcodegen.sh
```diff
@@ -0,0 +1,67 @@
...
+
+curl -fsSL -o "${WORK_DIR}/xcodegen.zip" "${XCODEGEN_RELEASE_URL}"
+unzip -qq -o "${WORK_DIR}/xcodegen.zip" -d "${WORK_DIR}"
+
+# The archive expands to xcodegen/{bin,share}. The binary resolves its bundled
+# templates relative to itself, so share/ must land alongside bin/.
+mkdir -p "${PREFIX}/share"
+cp "${WORK_DIR}/xcodegen/bin/xcodegen" "${PREFIX}/bin/xcodegen"
+rm -rf "${PREFIX}/share/xcodegen"
+cp -R "${WORK_DIR}/xcodegen/share/xcodegen" "${PREFIX}/share/xcodegen"
```
kaeawc: The install isn't atomic, and the fast path at lines 25-28 keys only on the *binary's* version — so a failure between line 49 and line 51 leaves a new binary next to a missing `share/xcodegen`, and every subsequent run says "already installed" and exits 0. The broken tree is self-perpetuating; the `EXIT` trap only reaps `WORK_DIR`, it has no idea `PREFIX` was touched. Any interruption in that window does it — Ctrl-C, ENOSPC, an I/O error on the copy.

Repro (I forced the line 51 `cp -R` to fail after 49/50 had run):
```
# binary now reports the new version, templates are gone:
$ prefix/bin/xcodegen --version   ->  Version: 2.46.0
$ ls -A prefix/share/xcodegen     ->  (empty)
# and the installer refuses to repair it:
$ bash scripts/ios/install-xcodegen.sh
XcodeGen 2.46.0 already installed.
$ ls -A prefix/share/xcodegen     ->  (still empty)
```

Stage the whole tree and commit it with a rename, so there's no window where `bin` and `share` disagree: copy into `${PREFIX}/share/.xcodegen.new`, then `rm -rf` the old and `mv` into place, and install the binary via `cp` to `${PREFIX}/bin/.xcodegen.new` + `mv`. Alternatively, make the fast path at line 25 also require `[[ -d "${PREFIX}/share/xcodegen" ]]` so a half-install can at least self-heal on the next run — that's the one-line version if you want to keep the diff small. Not a blocker for CI (fresh runner every time), but it's a nasty one to debug on a laptop.

scripts/ios/install-xcodegen.sh
```diff
@@ -0,0 +1,67 @@
...
+trap 'rm -rf "${WORK_DIR}"' EXIT
+
+curl -fsSL -o "${WORK_DIR}/xcodegen.zip" "${XCODEGEN_RELEASE_URL}"
+unzip -qq -o "${WORK_DIR}/xcodegen.zip" -d "${WORK_DIR}"
+
+# The archive expands to xcodegen/{bin,share}. The binary resolves its bundled
+# templates relative to itself, so share/ must land alongside bin/.
+mkdir -p "${PREFIX}/share"
+cp "${WORK_DIR}/xcodegen/bin/xcodegen" "${PREFIX}/bin/xcodegen"
+rm -rf "${PREFIX}/share/xcodegen"
```
kaeawc, line 81: The `rm -rf` + `cp -R` pair is racy against a second installer on the same `PREFIX`, and both failure modes are reachable. Running three installers concurrently into one prefix, 6 trials, at least one loser every time:

```
run 1: ok FAIL ok   run 2: ok ok FAIL   run 3: FAIL ok ok
run 4: ok FAIL FAIL  run 5: FAIL ok ok   run 6: FAIL ok ok

# loser's log:
cp: …/prefix/share/xcodegen: File exists
# and the surviving tree is nested:
/share/xcodegen/t.txt
/share/xcodegen/xcodegen/t.txt
```
The nesting is the ordinary `cp -R src dst` behavior when `dst` already exists — one process recreates the directory between another's `rm -rf` and its `cp`. A non-zero exit here propagates into `xcodegen-generate.sh:55` under `set -e` and fails the build.

CI is safe (each job is its own runner), so this is a local hazard — but it's a real one for this repo specifically, since parallel agents in worktrees share `$HOME` and `/usr/local`. The staging + `mv` restructure suggested on line 51 also fixes this, since `mv` onto a directory path is atomic enough to make the loser a no-op rather than an error. A `flock`-style lockfile around lines 48-52 would be the belt-and-braces version if you want it airtight. Not a blocker.

kaeawc (commented):
Reviewed one lens only: runtime correctness and failure modes of `scripts/ios/install-xcodegen.sh` itself (including the follow-up commit 9eded5b). The diagnosis in the description is well argued and the version-exact-by-URL reasoning is the right call — Homebrew genuinely cannot do this, and saying so in the file header is worth a lot to whoever reads it next.

Four findings on the installer's own failure modes, all reproduced locally against a temp `XCODEGEN_PREFIX` with `curl`/`unzip` stubbed. The first is the one that matters: on a laptop where `/usr/local/bin` is not writable, `xcodegen-generate.sh` fails the pin gate immediately after a *successful* install, and the error tells you to run the installer that just succeeded.

Things I checked that turned out fine, so nobody re-derives them: the `mkdir -p ... || [[ ! -w ]]` precedence is correct in all three cases I tried (exists-unwritable, missing-uncreatable, symlinked-to-writable); `GITHUB_PATH` is appended correctly and the current step still works because the install is its own step and `/usr/local/bin` is on the runner PATH regardless; and a truncated/corrupt zip aborts before touching `PREFIX`, leaving an existing install intact with the temp dir reaped.

scripts/ios/xcodegen_version.sh
kaeawc: Taken, and I agree it isn't a supply-chain finding — 0 of 8 peer installers verify a digest, so this PR was consistent with convention. Fixed in 3d15c9c5c anyway because the argument stands on **this PR's own thesis**: release assets are mutable, so a re-uploaded `xcodegen.zip` under tag 2.46.0 would report 2.46.0, pass the version gate, and generate different bytes — #3975 with the guard reporting green.

Pinned `XCODEGEN_RELEASE_SHA256` and verify before extracting, using the `shasum`/`sha256sum` fallback you pointed at. Verified independently against the live asset (4,278,764 bytes → 4d9e34b6…6806), and confirmed a tampered archive exits 1 with nothing installed.

scripts/ios/xcode-build.sh
kaeawc: Confirmed and fixed in 3d15c9c5c. Reproduced it: with a stub reporting 2.45.4 on PATH, the `! command -v xcodegen` branch is skipped entirely and generation proceeds with the skewed binary.

All three scripts now check the **version**, not presence. `xcode-build.sh` also loses the vestigial `command -v brew` guard, and on a skew it warns and uses the committed projects rather than regenerating — that path deliberately tolerates a missing generator, so a hard gate would be wrong there.

test/bats/install-xcodegen.bats
kaeawc: You're right, and the 9.99.9 experiment is the proof — I replaced it in 3d15c9c5c.

The test now asserts the digest pin exists and is 64 hex chars, plus two new ones: that the installer actually performs the verification, and that a digest mismatch exits 1 leaving nothing installed. The digest is what makes the claim in the old test's name true, so the assertion and the name finally agree.

test/bats/release-workflow-wiring.bats
kaeawc: Thanks for checking the untap was genuinely moot rather than taking my word for it. You're right that I silently dropped the `HOMEBREW_NO_REQUIRE_TAP_TRUST` guard from #3551, which had nothing to do with xcodegen — restored as its own assertion in 3d15c9c5c.

Also fixed the hard-coded four-workflow list: it now scans every workflow mentioning xcodegen, so a fifth one can't reintroduce a bare install unnoticed.

test/scripts/xcodegenDriftCheck.test.ts
kaeawc: Appreciated — that was the change I most expected to be challenged, so an explicit "this is a contract update, not a weakening" is useful to have on the record.

On the gap you flagged: `createTempRepo` stubs over `xcodegen-generate.sh`, so the gate on the repo-wide path still has no coverage. Only `--ctrl-proxy` is exercised. Leaving that as-is here rather than growing the fixture further, but it's a real hole and worth a follow-up.

scripts/ios/xcode-build.sh
kaeawc: Fixed in 3d15c9c5c — and you're right that the PR body asserted something untrue, which is worse than the gap itself. I've corrected the description.

Your observation that the new bats guard passed here because the literal `brew install xcodegen` string was gone while the behavior wasn't is the sharpest part of this: the test was checking the symptom rather than the invariant.

scripts/ios/ctrl-proxy-create-ipa.sh
kaeawc: Fixed in 3d15c9c5c — version check instead of presence check, and the now-false "Attempting to install via brew" message is gone (it's curl+unzip now). Same fix applied to `ctrl-proxy-build-for-testing.sh`.

.github/workflows/pull_request.yml
kaeawc: Useful distinction — thank you for verifying against a real run rather than reasoning about YAML. Recording your finding for posterity: the site works because `/usr/local/bin` is already on the runner PATH, **not** because `GITHUB_PATH` propagated from a backgrounded step, which stays unverified.

3d15c9c5c reduces the exposure: `xcodegen_version.sh` now exports `~/.local/bin` for every consumer, so if the fallback prefix ever does fire, recovery no longer depends on that unverified behavior.

scripts/ios/install-xcodegen.sh
kaeawc: This was the most valuable finding in the review — fixed in 3d15c9c5c.

You're right that `hash -r` is irrelevant; there's no stale cache entry, the directory simply isn't on the parent's PATH. A **successful** install failing the caller's gate, with a message telling the user to run the installer that just worked, is about the worst possible first-run experience for the contributor the fallback exists to serve.

`xcodegen_version.sh` now does `export PATH="${HOME}/.local/bin:${PATH}"`, mirroring `install_ktfmt.sh`, so every script that sources it sees the fallback prefix. Verified: sourcing the version file with a minimal PATH exposes `~/.local/bin`.

scripts/ios/install-xcodegen.sh
kaeawc: Correct — dead code, and compounding the bug above since it's exactly the message that would have told the user what to do. Fixed in 3d15c9c5c: the warning is now keyed on whether the fallback prefix was used, not on `$PATH` (which the export guarantees).

scripts/ios/install-xcodegen.sh
kaeawc: Reproduced and fixed in 3d15c9c5c. The sticky part is what makes this nasty: re-running printed "already installed" and exited 0 without repairing, so the broken state survives every subsequent invocation.

The fast path now requires `share/xcodegen` alongside the version match. Verified: deleting share/ makes the next run reinstall and restore it, while a complete install is still a no-op.

scripts/ios/install-xcodegen.sh
kaeawc: Fixed in 3d15c9c5c. Six-for-six reproduction is convincing, and the scoping is fair — CI gives each job its own runner, but parallel agents sharing $HOME on one machine is a real pattern in this repo.

Now stages both payloads into the destination filesystem and `mv`s them into place, so a concurrent installer can't observe a half-copied tree or land a nested `share/xcodegen/xcodegen/`. The EXIT trap reaps the staged paths too.

kaeawc:
## Resolves the collision with #3983 — ba85759fd

#3983 landed while this was open and fixed #3975 the other way: it regenerated the committed pbxproj with **2.45.4**, the version CI happened to install. That inverted the drift here.

Regenerating on this branch with the pinned **2.46.0** produces exactly the two-line reversal of #3983's change. That is direct evidence for the version-skew explanation #3983's description doubted ("Confirmed the earlier theory of a 2.46.0-vs-2.45.4 version skew was insufficient"):

| generator | `PBXProject.targets` order |
|---|---|
| 2.45.4 | alphabetical — CtrlProxy, CtrlProxyApp, CtrlProxyTests, CtrlProxyUITests, ObjCExceptionCatcher |
| 2.46.0 | `project.yml` declaration order — CtrlProxyApp, ObjCExceptionCatcher, CtrlProxy, CtrlProxyTests, CtrlProxyUITests |

Same input, same fresh checkout, different generator. No disagreement with #3983's diagnosis that #3969 introduced the drift — it did. The point is that *which* ordering counts as correct depends entirely on which XcodeGen you run, which is why pinning is the fix and regenerating is the symptom-level one.

**Kept the pin at 2.46.0** rather than matching #3983's bytes. It is current Homebrew stable and therefore what a contributor gets by default; pinning to 2.45.4 would make the installer downgrade nearly everyone. #3983 made main coherent only for whoever runs 2.45.4 — anyone on current brew regenerating would reintroduce the drift immediately. This PR commits what the pinned version generates, so CI and contributors finally agree, honoring the contract in the installer's own error message: bump the pin and regenerate in the same PR.

## Also fixed: my digest verification broke my own BATS suite

The curl stub logged its arguments without writing a payload, so sha256 verification rejected every stubbed download — 5 tests failed on macOS CI. The stub now delivers a real archive and exports its digest, and `XCODEGEN_RELEASE_SHA256` is overridable exactly as `XCODEGEN_VERSION` already was.

Worth stating plainly: **I reported those tests green in an earlier comment. They were failing locally the entire time** — I had inspected only the tail of the run and never read their results. The hardened fast path also correctly stopped treating a bare stub as a complete install, so the no-op test now models bin + share, and a new test covers repairing an install whose share/ is missing.

12/12 install-xcodegen BATS · `bun test` **8105 pass / 0 fail** · shellcheck + portability + set-e clean · drift in sync. `verify-runtime-deps.bats:37` fails identically on clean `origin/main` — environment-dependent, not from this PR.

PR #3990, opened 2026-07-19
refactor(android): extract shared correlated-error-on-throw core (#3086)

### LINKED ISSUE #3086: Deferred: extract shared correlated-error-on-throw core between ResultBroadcaster and AsyncActionRunner (only on 3rd consumer / first drift)
## Context

Follow-up from PR #3073 (closed #3045). Raised by the **design/direction reviewer** in the PR's multi-agent review (finding #1), filed as a deliberate **YAGNI-deferred** item — do **not** action until a third consumer or the first behavioral drift appears.

## Observation

`ResultBroadcaster.guard(...)` (PR #3073) and `AsyncActionRunner.launch(...)` (#3023) share an **identical inner skeleton**:

- run the block;
- rethrow `CancellationException` (cooperative shutdown must not become a client-facing error);
- on any other `Exception`: derive `cause = e.message ?: e::class.simpleName ?: "unknown error"`, log, then **best-effort** broadcast a correlated `ErrorResponse(requestId, …)`;
- guard the fallback broadcast against recursive failure (log-and-swallow), rethrowing a `CancellationException` raised during the fallback.

The only real differences:
- `AsyncActionRunner` wraps in `scope.launch { }` and returns a `Job` (owns coroutine *launching* / fire-and-forget lifecycle); `ResultBroadcaster` is a plain `suspend fun`.
- Error-message prefix: `"Action '$action' failed: "` vs `"Broadcast failed for $action: "`.
- Sink type: `WebSocketResponse` vs `ErrorResponse`.

Per CLAUDE.md ("one canonical primitive per concern"), there is a real shared primitive: *"run block; on non-cancellation throw, best-effort emit a correlated `ErrorResponse`, swallow-and-log the double-failure."*

## Why defer (reviewer's own conclusion)

The two differ on the axis that matters most to a reader — `AsyncActionRunner` owns *how work is scheduled* (scope/`Job`), `ResultBroadcaster` owns *how a throw is reported* synchronously. Folding them behind one core couples scheduling to error-reporting, and the `scope.launch` boundary is exactly where the `CancellationException` semantics differ in practice. Each body is ~15 lines; a shared helper saves ~10 lines but adds a third type and a `messagePrefix`/sink-adapter seam. Marginal win today.

## The actual risk being tracked

The PR body states `ResultBroadcaster` "mirrors `AsyncActionRunner` exactly" — but nothing **forces** them to stay in lock-step. If a future fix touches one catch-body (e.g. adds the `requestId` into the message, changes the cause derivation, or adjusts cancellation handling), the other will silently drift. **This drift is the trigger to act on this issue** — extract a shared `catch`-body helper (e.g. `broadcastCorrelatedErrorOnThrow(requestId, action, messagePrefix, sink, block)`) that both call.

## Acceptance criteria
- [ ] Only when a third consumer arrives OR the first drift is observed: extract the shared correlated-error-on-throw core; both `ResultBroadcaster.guard` and `AsyncActionRunner.launch` delegate to it.
- [ ] Existing tests (`ResultBroadcasterTest`, `AsyncActionRunnerTest`) continue to pass unchanged.

## References
- PR #3073 (merge `6cff3833e`), issue #3045; sibling #3023
- `android/control-proxy/.../ResultBroadcaster.kt`, `android/control-proxy/.../AsyncActionRunner.kt`

kaeawc:
## Summary

Issue #3086 filed the duplicated correlated-error catch-body between `ResultBroadcaster` (#3045) and `AsyncActionRunner` (#3023) as a **deliberate YAGNI defer**, and named the trigger to act: *a third consumer, or the first observed drift*. `ServiceScopeGuard` (#3104, PR #3120) landed after that and hand-rolled the same body a third time — the trigger fired, so this extracts the shared core.

Behavior-preserving: every emitted frame and log string is byte-identical, and all three consumer test suites pass **completely unchanged**.

## Linked Issue

Closes #3086

## Changes

New `CorrelatedErrorReporter` is the one place the body lives. Three layers, because the consumers enter at different points:

| Layer | Used by |
|---|---|
| `guarding(...)` — run a block, report anything it throws | `ResultBroadcaster.guard`, `AsyncActionRunner.launch` |
| `report(...)` — log an already-caught throwable, then emit its frame | (the `guarding` tail) |
| `emit(...)` / `causeOf(...)` — the emit tail and the cause rule alone | `ServiceScopeGuard` |

**Why the issue's suggested one-function shape wasn't enough.** #3086 sketched a single `broadcastCorrelatedErrorOnThrow(…, block)`. That fits two consumers but not `ServiceScopeGuard`: a `CoroutineExceptionHandler` is handed a *throwable* rather than a block, so there is nothing to wrap.

**Why `ServiceScopeGuard` keeps its own log call.** It logs synchronously on the failing thread and dispatches only the broadcast onto `emitScope`. Folding that log into `report` would have moved it inside the launched coroutine — and lost the diagnostic entirely whenever the emit scope is already shut down, which is exactly when you most want the trace. Splitting `emit` out preserves today's guarantee.

## Testing

- **`CorrelatedErrorReporterTest`** (new) — pins the core directly: correlated frame on throw, silence on success, cancellation propagates without emitting, null requestId passthrough, the double-failure tail (logged-and-swallowed / cancellation re-thrown), and all three cause-derivation branches.
- **`CorrelatedErrorDriftGuardTest`** (new) — the structural backstop, and the one test that was **red before this change**. See below.
- **`ResultBroadcasterTest`, `AsyncActionRunnerTest`, `ServiceScopeGuardTest`** — pass with **zero edits** (issue AC2). They now exercise the core through their public surfaces.
- Full `:control-proxy:testDebugUnitTest` green; ktfmt 0.64 `--google-style` applied.

## Acceptance criteria

- [x] **AC1** — *Only when a third consumer arrives OR the first drift is observed:* extract the shared core; both `ResultBroadcaster.guard` and `AsyncActionRunner.launch` delegate to it. Trigger verified (`ServiceScopeGuard`, `e8911620c`); all three delegate.
- [x] **AC2** — Existing `ResultBroadcasterTest` / `AsyncActionRunnerTest` continue to pass **unchanged** (verified: zero diff in those files).

## The drift guard

The risk #3086 actually tracked was that nothing *forced* the copies to stay in lock-step — a fix to one catch-body would silently leave the others behind. Extraction removes today's duplication; `CorrelatedErrorDriftGuardTest` is what stops it coming back.

It source-scans the three consumers and fails CI if any reintroduces the cause derivation or an `ErrorResponse` fallback instead of delegating. Following the `BroadcastGuardAdoptionTest` precedent, the scanning logic is a pure function (`CorrelatedErrorDriftScanner`) unit-tested against synthetic good/bad snippets — including that KDoc, imports, and line comments naming those symbols don't trip it — so the enforcement contract is itself covered rather than only asserted against live files.

This is a small addition beyond the two literal criteria; it was explicitly approved as in-scope because it addresses the issue's stated risk.

## Validation

- [x] `bun run lint` clean
- [x] `bun run typecheck` — no new errors attributable to this PR (Kotlin-only diff; see note)
- [x] Android changes: `:control-proxy:testDebugUnitTest` green + ktfmt applied
- [x] New code uses interfaces + fakes; unit tests are Android-free and fast
- [ ] Device verification — N/A, no on-device behavior change (frames are byte-identical)

> **Note on the local typecheck gate:** running it from a git worktree reports the `PlanSchemaValidator.ts` ajv error as "new". It is not — it is already baselined at `scripts/typecheck-baseline.txt:235`, and only fails to string-match because a worktree resolves `node_modules` to the main clone's *absolute* path while the baseline stores a relative one. CI checks out normally and matches.

## Follow-ups

- [x] #3992 — two residual drift gaps deliberately left out of scope, each with an explicit trigger (following the convention #3086 itself set): the drift scanner does not cover a *fourth* consumer, and six hand-written `(requestId=$requestId)` log suffixes remain undeduplicated.

## Review round (`597dcfdd9`)

Three orthogonal lenses; full triage in [this comment](https://github.com/kaeawc/auto-mobile/pull/3990#issuecomment-5017517962). Headline: `report` is now `private` (zero production callers), the drift scanner was hardened from a *verbatim-copy* detector into an actual drift detector, two `stripComments` bugs that could hide violations were fixed, and `CorrelatedErrorConsumerContractTest` was added to pin the byte-identical strings this PR claims — nothing asserted them before, since the three consumer suites use `contains(...)` and AC2 required leaving them untouched.

### DIFF
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/AsyncActionRunner.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/AsyncActionRunner.kt
@@ -66,10 +66,11 @@ class AsyncActionRunner(
       val launchScope = this
       reporter.guarding(
         requestId = requestId,
-        failureLogMessage = "Async action '$action' failed (requestId=$requestId)",
-        errorMessagePrefix = "Action '$action' failed",
-        doubleFailureLogMessage =
-          "Failed to broadcast async error for action '$action' (requestId=$requestId)",
+        failureLogMessage = { "Async action '$action' failed (requestId=$requestId)" },
+        errorMessagePrefix = { "Action '$action' failed" },
+        doubleFailureLogMessage = {
+          "Failed to broadcast async error for action '$action' (requestId=$requestId)"
+        },
       ) {
         block(launchScope)
       }
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/AsyncActionRunner.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/AsyncActionRunner.kt
@@ -3,7 +3,6 @@ package dev.jasonpearson.automobile.ctrlproxy
 import android.util.Log
 import dev.jasonpearson.automobile.protocol.ErrorResponse
 import dev.jasonpearson.automobile.protocol.WebSocketResponse
-import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
@@ -39,10 +38,13 @@ class AsyncActionRunner(
     Log.e(TAG, message, error)
   },
 ) {
+  /** The shared correlated-error-on-throw core (#3086). */
+  private val reporter = CorrelatedErrorReporter(broadcastResponse, logError)
+
   /**
-   * Launch [block] on [scope]. If it throws anything other than a cooperative
-   * [CancellationException], broadcast an [ErrorResponse] correlated by [requestId] so the awaiting
-   * client fails fast instead of hanging.
+   * Launch [block] on [scope]. If it throws anything other than a cooperative cancellation,
+   * broadcast an [ErrorResponse] correlated by [requestId] so the awaiting client fails fast
+   * instead of hanging. See [CorrelatedErrorReporter.guarding] for the semantics.
    *
    * @param requestId correlation id from the originating request; null when the request carried
    *   none (the error frame then carries a null requestId, exactly as the synchronous decode path
@@ -59,31 +61,17 @@ class AsyncActionRunner(
     // catch — a non-[Exception] [Throwable] such as an [Error]/[AssertionError] — still escapes
     // with
     // the correlation id attached, letting the scope-level [ServiceScopeGuard] emit a *correlated*
-    // error frame instead of a null-id one. The common `Exception` case is handled below directly.
+    // error frame instead of a null-id one. The common `Exception` case is handled by the reporter.
     scope.launch(RequestIdContext(requestId)) {
-      try {
-        block()
-      } catch (e: CancellationException) {
-        // Cooperative cancellation means the service scope is shutting down — never convert it into
-        // a client-facing error frame; let it propagate so the coroutine unwinds cleanly.
-        throw e
-      } catch (e: Exception) {
-        val cause = e.message ?: e::class.simpleName ?: "unknown error"
-        logError("Async action '$action' failed (requestId=$requestId)", e)
-        try {
-          broadcastResponse(
-            ErrorResponse(requestId = requestId, error = "Action '$action' failed: $cause")
-          )
-        } catch (broadcastError: CancellationException) {
-          throw broadcastError
-        } catch (broadcastError: Exception) {
-          // A failure while reporting the error must not escape and crash the launch (which would
-          // defeat the whole point). Log and swallow — the client will fall back to its timeout.
-          logError(
-            "Failed to broadcast async error for action '$action' (requestId=$requestId)",
-            broadcastError,
-          )
-        }
+      val launchScope = this
+      reporter.guarding(
+        requestId = requestId,
+        failureLogMessage = "Async action '$action' failed (requestId=$requestId)",
+        errorMessagePrefix = "Action '$action' failed",
+        doubleFailureLogMessage =
+          "Failed to broadcast async error for action '$action' (requestId=$requestId)",
+      ) {
+        block(launchScope)
       }
     }
 
--- a/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CorrelatedErrorReporter.kt
+++ b/android/control-proxy/src/main/kotlin/dev/jasonpearson/automobile/ctrlproxy/CorrelatedErrorReporter.kt
@@ -1,0 +1,133 @@
+package dev.jasonpearson.automobile.ctrlproxy
+
+import dev.jasonpearson.automobile.protocol.ErrorResponse
+import kotlinx.coroutines.CancellationException
+
+/**
+ * The single canonical implementation of *"a throw on a requestId-correlated path must become a
+ * correlated `type:"error"` frame, never a silent hang"*.
+ *
+ * Three seams close that hang-class on three different paths, and each one used to hand-roll the
+ * same body:
+ * - [AsyncActionRunner] (#3023) — a throw *inside* a launched action coroutine.
+ * - [ResultBroadcaster] (#3045) — a throw while *sending* a result.
+ * - [ServiceScopeGuard] (#3104) — a throw escaping a *raw* `serviceScope.launch { … }`.
+ *
+ * Issue #3086 filed the duplication as a deliberate YAGNI defer while there were only two
+ * consumers, and named the trigger to act: a third consumer, or the first observed drift. The third
+ * consumer arrived. The risk the issue actually tracked is that nothing *forced* the copies to stay
+ * in lock-step — a fix to one catch-body (changing the cause derivation, folding the requestId into
+ * the message, adjusting cancellation handling) would silently leave the others behind. That risk
+ * is now structural: this type is the one place the body lives, and `CorrelatedErrorDriftGuardTest`
+ * fails CI if a consumer starts hand-rolling it again.
+ *
+ * Three layers, because the consumers enter at different points:
+ * - [guarding] — run a block, report anything it throws. Used by [ResultBroadcaster] and
+ *   [AsyncActionRunner].
+ * - [report] — log an *already-caught* throwable, then emit its frame.
+ * - [emit] + [causeOf] — the frame-emitting tail and the cause rule on their own, for
+ *   [ServiceScopeGuard]: a [kotlinx.coroutines.CoroutineExceptionHandler] is handed a throwable
+ *   rather than a block, and must log synchronously (see [emit]).
+ *
+ * @param broadcastError sink for the correlated fallback frame — `webSocketServer.broadcast(…)` in
+ *   production (lazily resolving the `lateinit` server), a capturing lambda in tests.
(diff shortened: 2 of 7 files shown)


kaeawc:
## Review round: three orthogonal lenses, findings triaged

Ran regression / test-adequacy / API-surface lenses over the diff. Pushed `597dcfdd9` in response. Recording the triage here rather than as inline comments, since I authored and resolved all of it.

### Verified clean

The regression lens **empirically** pinned behavior rather than reading for it — temporary suites asserting exact pre-refactor strings, then deleted. All three seams emit byte-identical log text, frame text, and double-failure text; cancellation still propagates at every point including from the fallback broadcast; catch ordering is correct (`CancellationException` extends `IllegalStateException`, so its clause must precede `Exception` — it does in both `guarding` and `emit`); exception-type coverage is unchanged (`Exception`, never widened to `Throwable`); the sink variance is sound. No wire-protocol change, and no TS-side code depends on these strings — the only regex over Android error text (`rewriteUnknownCommandError.ts:9`) matches the *decode* path, not these seams.

### Fixed

- **`report` was public with zero production callers.** Two lenses flagged it independently, and it's the same CLAUDE.md rule the extraction invokes to justify itself. Now `private`.
- **The drift guard was a verbatim-copy detector, not a drift detector.** A body changing the fallback constant (`?: "unspecified failure"`) scanned clean — precisely the drift #3086 describes. Now also matches `::class.simpleName`, the structural half of the cause rule.
- **`MISSING_DELEGATION` was near-tautological** — a bare substring match on the type name, satisfied by a consumer that holds a reporter field, never calls it, and hand-rolls the body beside it. Now requires a real call, receiver spelled out so the package's many `Flow.emit(` calls can't pass for it.
- **Two `stripComments` state-machine bugs**, both able to hide a genuine violation: a raw string containing `/*` swallowed the rest of the file; a `'"'` char literal swallowed the rest of its line. Fixed with a contract test each.
- **Nothing pinned the byte-identical strings this PR claims.** The three consumer suites assert with `contains(...)`, and AC2 requires leaving them untouched — so the claim was structurally unverifiable. New `CorrelatedErrorConsumerContractTest` pins exact text for all three seams.
- **The `CoroutineScope` receiver became an explicit argument.** It was implicit and impossible to get wrong; it can now be handed the outer scope by mistake, which would break structured concurrency for every action while leaving every existing assertion green. Same file pins it.
- **`emit` was only covered via `guarding`**, though it's the entry point `ServiceScopeGuard` actually uses. Added its null-requestId, swallow, and cancellation cases.

### Rejected (reasons recorded in the test KDoc so they aren't re-litigated)

- **Widening the scan to the whole package.** `WebSocketServer.kt:488` (decode path, #2985) and `HierarchyExtractErrorFrames.kt:35` construct `ErrorResponse` for legitimate non-fallback reasons — a package-wide rule fires on correct code. This is the same false-positive limit `BroadcastGuardAdoptionTest` documents for raw launches, and exactly why #3104 closed that gap *by construction* instead of by scanning. The consequence is real and worth stating plainly: a **fourth** consumer hand-rolling the body would not be caught.
- **Eager message interpolation on the success path.** Real (three string builds per call where there were none), but they sit in front of WebSocket I/O and JSON serialization that dwarf them.
- **Making the class `internal`.** Defensible, but the three sibling seams are all public.

### Deferred

The six hand-written `" (requestId=$requestId)"` log suffixes remain undeduplicated. The reporter could append them itself, but only 4 of the 6 route through it, so it's a partial win — filed separately rather than grown into this PR.

kaeawc:
## Second review round: five more lenses, findings triaged

Coroutine correctness · adversarial (break the drift guard) · repo conventions · issue-intent fidelity · silent-hang bug class. Pushed `0b0f52e31`.

**No lens found a correctness defect in production code.** The silent-hang guarantee holds end-to-end — the TS daemon path was traced (`AndroidCtrlProxyClient.ts:2225-2237`) and nothing depends on the error strings, ordering, or frame count the refactor touches; `resolveError` and `rejectPendingHierarchy` both no-op on a miss. Catch ordering is correct (`CancellationException` extends `IllegalStateException`, so its clause must precede `Exception` — it does in both entry points). Structured concurrency is intact. `BroadcastGuardAdoptionTest` and `LaunchCancellationRethrowTest` scan only `CtrlProxy.kt`/`WebSocketServer.kt`, untouched here. ktfmt 0.64 dry-run exits 0.

### Fixed — the adversarial lens did real damage

It ported the scanner to Python and *executed* attacks rather than reasoning about it. Three genuine holes:

- **`javaClass.simpleName` / `::class.java.simpleName` bypassed the cause rule entirely.** The Java-interop spellings are at least as idiomatic in Android code as the Kotlin one — so the most likely drift was the one form the guard couldn't see. Widened.
- **Kotlin block comments nest; `stripComments` used a boolean.** Commenting out an old implementation containing its own KDoc closed the comment early and scanned the disabled body as live code — failing CI on something that isn't compiled, exactly when someone is mid-bugfix. Now a depth counter.
- **`MISSING_DELEGATION` required a literal `reporter.` receiver.** Renaming the field to `errorReporter`, or `with(reporter) { … }`, would have failed CI on correct code. A guard that breaks on a pure rename is a guard that gets deleted. Now receiver-agnostic: name the type *and* call `.guarding(`/`.causeOf(`. (`.emit(` stays out — the package is full of `Flow.emit(`.)

I verified the loosened rules are still load-bearing: all three pre-change files still trip all three rules, 9 violations total.

- **The success-path allocation regression I'd previously rejected.** Two lenses flagged it independently. My extraction moved three string interpolations onto the success path of ~70 call sites; before, they lived inside the `catch` and cost nothing. My earlier rejection ("the WebSocket I/O dwarfs it") was true but beside the point — it's a regression my refactor introduced and the fix is cheap. Message params are now `() -> String`.
- **The scanner KDoc was making a false claim.** It said the cause rule "may appear in exactly one production file". It doesn't: `WebSocketServer.kt`, `HierarchyExtractErrorFrames.kt` and `CtrlProxy.kt` each derive one. Corrected to state the real scope.
- **`emit`'s KDoc overstated its guarantee** — a non-`Exception` `Throwable` from the sink does escape and can re-enter the handler. Documented as deliberate (widening to `Throwable` would swallow OOM) rather than quietly changing exception semantics in a byte-identical refactor.
- **Added a genuine-cancellation test** — the existing ones throw a synthetic `CancellationException`; this cancels the `Job` while the block is really suspended.

### Rejected

- **Reusing `KotlinSourceScan.maskLiteralsAndComments`** — it blanks string contents, and this scanner must *see* the cause literal. Tracked in #3992 with the trade-off written out.
- **Deduplicating the third copy of `locateSource`** — pre-existing pattern this PR follows rather than causes; #3992.
- **Widening `emit` to catch `Throwable`** — would swallow errors that should kill the process.

### Two things a reader should know

**The drift guard has a named blind spot with an in-tree instance.** `CtrlProxy.kt:2053` and `HierarchyExtractErrorFrames.kt:46` derive the same cause from the same throwable twice on a correlated-error path, and neither is in `CONSUMER_FILES`. No hang risk (that path broadcasts its own frame), but it's the strongest candidate for the "fourth consumer" trigger — details in #3992.

**Merged ≠ shipped.** `src/constants/release.ts` pins `apkSha256` per version and the daemon downloads the pinned release APK, so this runner change reaches no user until a version bump re-cuts `control-proxy-debug.apk`. Local verification needs `AUTOMOBILE_CTRL_PROXY_APK_PATH`.

PR #4001, opened 2026-07-20
fix(tooling): restore clean validation gates

### LINKED ISSUE #4000: fix(tooling): typecheck fails and lint dirties release.ts on a clean checkout of main
## Summary

Two independent defects that both make a **clean checkout of `main` fail or dirty a local validation command**. Filed together because they share a symptom — you clone, change nothing, run the documented validation commands, and get a red gate plus an unexplained working-tree modification — and both are small, mechanical fixes.

Neither is caused by any recent PR. Both were found while validating #3998 and verified against a clean tree.

---

### Reproduce
```
git checkout main && git status   # clean
bun run typecheck
```

```
❌ typecheck gate: 1 NEW TypeScript error(s) not in the baseline:
src/utils/plan/PlanSchemaValidator.ts: error TS2345: Argument of type
  'node_modules/ajv/dist/ajv').Ajv' is not assignable to parameter of type
  'node_modules/ajv-formats/node_modules/ajv/dist/core').default'
```

Verified in two separate worktrees with no local changes, so it is not a stale-`node_modules` artifact of one machine.

### Root cause
Two copies of `ajv` are installed:

| | version |
|---|---|
| top-level `node_modules/ajv` | 8.20.0 |
| `node_modules/ajv-formats/node_modules/ajv` | 8.17.1 |

`ajv-formats@3.0.1` declares `ajv: ^8.0.0` as a **regular dependency**, not a peer dependency, so it is entitled to its own copy. `package.json` asks for `ajv: ^8.18.0`, which resolves to 8.20.0 — a different version than the nested 8.17.1, so the installer keeps both.

At `PlanSchemaValidator.ts:43`, `addFormats(this.ajv)` passes the top-level `Ajv` into a function typed against the nested one. The two classes are structurally similar but nominally distinct, hence TS2345.

### Fix
Force a single copy. The repo already uses `overrides` for exactly this pattern (`lodash`, `tar`, `hono`, `qs`, and two nested `minimatch` pins), so add:

```json
"overrides": {
  "ajv": "^8.20.0"
}
```

Then reinstall and confirm `node_modules/ajv-formats/node_modules/ajv` is gone.

**Do NOT resolve this with `bun run typecheck:update`.** That baseline is a one-way ratchet for *tolerated legacy* errors; absorbing a live dependency-resolution bug into it would hide the problem permanently and grow the ratchet in the wrong direction. Per CLAUDE.md the baseline is also version-sensitive, so if the dedupe shifts other inferred types, regenerate it in the same PR.

### Impact
Every branch cut from `main` inherits a red typecheck gate, so the signal is useless until someone fixes it — a new error introduced by an actual PR is indistinguishable from this one.

---

### Reproduce
```
git checkout main && git status   # clean
bun run lint
git status                        # M src/constants/release.ts
```

The diff is one line, `videoJarSha256` in `NIGHTLY_CHECKSUM_ENTRY`, re-indented from 4 spaces to 2. `bun run lint` runs `eslint --fix`, so it silently rewrites the file.

### Root cause
`update_registry_field()` in `scripts/generate-release-constants.sh` has two branches. The **update** branch (field already present) preserves whatever indentation is there, via `(^\s+...)`. The **insert** branch (field absent) hardcodes four spaces:

```python
ipa_re = re.compile(r'(^\s+ipaSha256: "[^"]+",\n)', re.M)
updated_entry = ipa_re.sub(r'\1    ' + field + f': "{value}",\n', entry, count=1)
                              # ^^^^ hardcoded 4-space indent
```

Four spaces is correct for `RELEASE_CHECKSUM_REGISTRY` entries, which are nested inside an array literal. It is wrong for `NIGHTLY_CHECKSUM_ENTRY`, a standalone `export const` whose fields sit at two spaces.

The function already knows these two shapes differ — the comment above the closing-anchor regex explains why it uses `\n\s*\}` rather than `\n\s+\}` precisely so it can match both the indented registry entry and the column-0 `};` of the nightly const. The insert branch just never got the same treatment.

### Scope (narrower than it looks)
This misfires **only when a field is inserted into `NIGHTLY_CHECKSUM_ENTRY` for the first time** — which is how the current mis-indent got committed, when `videoJarSha256` was added (#3833). Routine nightly checksum bumps take the update branch and preserve indentation, so this does *not* recur on every nightly run. It will recur the next time a new field is added to that entry.

### Fix
Reuse the indentation captured from the anchor line instead of hardcoding it:

```python
ipa_re = re.compile(r'(^(\s+)ipaSha256: "[^"]+",\n)', re.M)
updated_entry = ipa_re.sub(
    lambda m: m.group(1) + m.group(2) + f'{field}: "{value}",\n', entry, count=1
)
```

Then commit the corrected `src/constants/release.ts` so a clean checkout lints clean.

### Impact
Low severity but a real hazard: it leaves an unexplained modification in a checksum-registry file after any local lint run. `NIGHTLY_CHECKSUM_ENTRY` is mutable while `RELEASE_CHECKSUM_REGISTRY` is append-only and must never be touched by nightly (see #3784), so an accidental `git add -A` there is exactly the kind of stray commit that is confusing to unpick later. It caught me twice in one session while shipping #3998; staging explicit paths avoided it both times.

---

## Acceptance
- [ ] `bun run typecheck` passes on a clean checkout of `main`, with a single resolved `ajv` and no baseline growth.
- [ ] `bun run lint` leaves the working tree clean on a fresh checkout.
- [ ] `update_registry_field`'s insert branch derives indentation rather than hardcoding it, so a future new field lands correctly in both entry shapes.
- [ ] `src/constants/release.ts` committed with correct indentation.

## Provenance
Both found and verified against a clean tree on 2026-07-19 while validating PR #3998. Root causes traced to source; neither is speculative. Note the generator's root cause is the *insert* branch, not the `new_entry` template a few lines above it — that template's uniform 4-space indent is correct for the array entries it builds.

kaeawc:
## Summary

Fixes the two clean-checkout validation defects from issue #4000: aligns the project Ajv dependency with `ajv-formats` so `bun run typecheck` no longer sees a duplicate-Ajv nominal type mismatch, and fixes release checksum insertion so new fields inherit the target entry's indentation.

## Linked Issue

Closes #4000

## Bug / Regression Context

- **Prior attempts:** N/A
- **Observed symptom:** `bun run typecheck` reported a new `PlanSchemaValidator.ts` Ajv type error on a clean checkout, and `bun run lint` rewrote `src/constants/release.ts` by reindenting `NIGHTLY_CHECKSUM_ENTRY.videoJarSha256`.
- **Repro steps / conditions:** Clean checkout of `main`, then run `bun run typecheck` or `bun run lint`.

## Validation

- [x] `scripts/all_fast_validate_checks.sh` passes (`bun run lint` + `bun test`) equivalent covered by `bun run turbo:validate`
- [x] Android/iOS changes: ran the matching `scripts/` validation (N/A; tooling-only)
- [x] New code uses interfaces + fakes + FakeTimer; unit tests run in <100ms (N/A; shell-script regression uses BATS)

Additional validation:
- [x] `bats test/bats/generate-release-constants.bats`
- [x] fresh `bun install` has no `node_modules/ajv-formats/node_modules/ajv`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `git diff --check`

## Device Verification

N/A; tooling-only change.

## Follow-ups

- [x] Follow-up issues filed for gaps not addressed here: N/A

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -9,7 +9,7 @@
         "@jimp/core": "^1.6.1",
         "@modelcontextprotocol/sdk": "^1.26.0",
         "adm-zip": "^0.6.0",
-        "ajv": "^8.18.0",
+        "ajv": "8.20.0",
         "ajv-formats": "^3.0.1",
         "async-mutex": "^0.5.0",
         "jimp": "^1.6.0",
@@ -1219,8 +1219,6 @@
 
     "@jimp/types/zod": ["zod@3.25.76", "", {}, "sha512-gzUt/qt81nXsFGKIFcC3YnfEAx5NkunCfnDlvuBSSFS02bcXu4Lmea0AFIUwbLWxWPx3d9p8S5QoaujKcNQxcQ=="],
 
-    "@modelcontextprotocol/sdk/ajv": ["ajv@8.18.0", "", { "dependencies": { "fast-deep-equal": "^3.1.3", "fast-uri": "^3.0.1", "json-schema-traverse": "^1.0.0", "require-from-string": "^2.0.2" } }, "sha512-PlXPeEWMXMZ7sPYOHqmDyCJzcfNrUr3fGNKtezX14ykXOEIvyK81d+qydx89KY5O71FKMPaQ2vBfBFI5NHR63A=="],
-
     "@modelcontextprotocol/sdk/zod": ["zod@4.3.6", "", {}, "sha512-rftlrkhHZOcjDwkGlnUtZZkvaPHCsDATp4pGpuOOMDaTdDDXF91wuVDJoWoPsKX/3YPQ5fHuF3STjcYyKr+Qhg=="],
 
     "@oxc-resolver/binding-wasm32-wasi/@emnapi/core": ["@emnapi/core@1.11.0", "", { "dependencies": { "@emnapi/wasi-threads": "1.2.2", "tslib": "^2.4.0" } }, "sha512-l9Oo58x0HOP5znGzVhYW9U3e5wVuA4LAZU2AGezTmkhO1CgQRFDhDg4nneHsu/t3WniXg9QrG2nIXL/ZS8ln8Q=="],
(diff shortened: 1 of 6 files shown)


kaeawc:
Closing per request.

PR #4026, opened 2026-07-20
perf(ci): split detekt into its own cached, module-scoped job

kaeawc:
## Summary

Detekt ran as a step inside **Fast Validation**, where it took **7m04s of an 8m02s job** — everything else in that job totals ~55s. It ran as a bare `./gradlew detektMain detektTest` with **no Gradle caching at all**, so every PR re-downloaded Gradle and recompiled all 24 modules from scratch.

1. **Own job** in `pull_request.yml` and `merge.yml`, so detekt runs concurrently instead of serializing behind format/lint checks. Fast Validation returns to ~55s.

2. **Gradle caching.** Both jobs route through the existing `./.github/actions/gradle-task-run` composite, picking up the build cache, configuration cache, jemalloc, and a per-job `GRADLE_USER_HOME`. A warm local run is **30s** against a cold **7min** — that gap is what the cache reclaims, and it is where essentially all of the speedup comes from.

3. **Kotlin-only gating.** The job runs only when a PR touches an input that can change what detekt reports: `.kt`/`.kts`, `gradle.properties`, the version catalog, the Gradle wrapper, the detekt config, or the `gradle-task-run` composite. This gates *whether the job runs*, not what it analyzes — when it runs, it analyzes the whole tree.

4. **`ignoredBuildTypes = listOf("release")` and `parallel = true`** in `android/build.gradle.kts`. On Android modules `detektMain` fans out per build type, and no module has a `src/release` source set, so the release pass re-analyzed identical sources. Task count drops **527 → 393**.

## What changed after review

Five reviewers went over this. Two findings changed the design:

**Module scoping was removed.** An earlier revision narrowed the run to the modules a PR changed, justified by "detekt rules are almost entirely file-local." **That was backwards for this project.** `config/detekt/detekt.yml` disables `complexity`, `coroutines`, `empty-blocks`, `exceptions`, `naming`, `performance`, and `style`, leaving `potential-bugs` as the only active rule set — and its rules (`UnsafeCallOnNullableType`, `UnreachableCode`, `HasPlatformType`, `UnnecessarySafeCall`) are exactly the cross-module type-resolution ones. The file-local rule sets are the disabled ones.

Reproduced on real modules: a one-file change in `:playground:storage` leaves a scoped run green while `:playground:experimentation:detektMain` fails with `UnnecessarySafeCall`. That lands a violation on `main` and reddens the next unrelated PR — the #4016 → #4020 → #4022 incident class.

Since the warm-cache measurement (30s) was taken at *full* scope, scoping was buying almost nothing on top of the cache while carrying that risk. Removing it deleted ~460 lines of script and tests, and with them every scoping-related finding reviewers raised: an unanchored `grep -qF` that matched a commented-out `include(...)`, a silent drop of Kotlin files owning no module, cross-module renames collapsing to the destination path, `PROJECT_ROOT` going unvalidated, and — the sharpest one — the `ONLY_CHANGED_SINCE_SHA` path that CI actually used having no test coverage, so deleting its default would have left the suite green while every PR silently full-scoped.

**A severed publish gate was restored.** `publish-android-libraries-snapshot` lists `ktfmt` in `needs:` and covered detekt transitively while detekt was a *step inside* that job. Once it became a sibling job, nothing depended on it — Android library snapshots could publish to Maven Central with detekt red. Fixed by adding `detekt` to that `needs:` list. This was a regression introduced by this PR and caught in review.

## Reviewer notes

**⚠️ Detekt no longer blocks merges until the ruleset is updated.** It previously gated via being inside the required "Fast Validation" check. The `green-main` ruleset requires individual job names and `Detekt` is not among them. Note the roll-up gates (`Android`, `iOS`, `Shell Tests`) are *also* not required contexts, so wiring `detekt` into `android-gate` would **not** restore blocking — only adding the literal `Detekt` context does. Requiring it directly is safe: it is a non-matrix job, and an `if:`-skipped job reports `skipped`, which rulesets count as satisfied (`Build Desktop Core Module` is already both required and `if:`-gated).

**The release-variant drop was independently verified**, not just asserted: the task-graph delta is exactly 16 `:*:detektRelease` tasks removed with nothing added, every module's release file set is a strict subset of its debug set (`release-only = 0` everywhere), and the remaining 96 tasks still cover all 650 `.kt` files on disk. No product flavors, no `buildConfigField`, `isMinifyEnabled = false` on every release block.

## Validation

- [x] `scripts/all_fast_validate_checks.sh` passes
- [x] `bats test/bats/` — full suite green except a pre-existing, unrelated failure in `verify-runtime-deps.bats` that dies inside `source ~/.bashrc` on this machine
- [x] `cd android && ./gradlew detektMain detektTest` passes
- [ ] `bun run typecheck` — N/A, no TypeScript changed

## Follow-ups

Filed separately: detekt reports **377 compiler errors** on `:desktop-core:detektMain` (and 3/7/7/30/30 on five other modules) during analysis, exiting 0 so it is invisible in CI. Diagnosed while working this PR — the dominant cause is `unresolved reference 'serializer'` (97 occurrences): `.serializer()` is synthesized by the kotlinx.serialization **compiler plugin**, which detekt's standalone type-resolution frontend does not run, so every `@Serializable` type's generated members are invisible and the failures cascade into 53 "cannot infer type parameter" and hundreds of unresolved property accesses. `enableCompilerPlugin = true` was tested and does **not** fix it. This matters more than it looks: incomplete type resolution silently weakens `potential-bugs`, which is this repo's *entire* active rule set.


.github/workflows/pull_request.yml
```diff
@@ -659,6 +671,68 @@ jobs:
...
+  # gradle-task-run for build/configuration cache reuse -- the raw `./gradlew`
+  # call it replaced had no Gradle caching at all and recompiled every module
+  # from scratch on every PR.
+  detekt:
+    name: "Detekt"
+    runs-on: ubuntu-latest
+    needs: detect-changes
+    # detekt analyzes Kotlin only, so a PR touching no .kt/.kts has nothing for
+    # it to look at. merge.yml runs the full tree post-merge as the backstop.
+    if: needs.detect-changes.outputs.kotlin_changed == 'true' && needs.detect-changes.outputs.sha256_only != 'true' && needs.detect-changes.outputs.auto_chore != 'true'
```
kaeawc, line 693: Confirming your warning, and it is worse than "add `Detekt` to the list" — the `green-main` ruleset requires **individual job names**, not the roll-up gates:

```
gh api repos/kaeawc/auto-mobile/rulesets/11406121 --jq '.rules[]|select(.type=="required_status_checks")|.parameters.required_status_checks[].context'
```

returns `Fast Validation`, `Build CtrlProxy APK`, … and notably **not** `Android`, `iOS`, `Shell Tests`, or `CodeQL`. So the gate jobs in this workflow are decorative as far as merging goes, and adding `detekt` to `android-gate`'s `needs:` would not restore blocking either. The only fix is adding the literal context `Detekt` to the ruleset. This should block the merge of this PR — the window between merging and updating the ruleset is a window where detekt violations land on `main` unopposed.

Requiring it directly is safe, for what it's worth: `Detekt` is a non-matrix job so there is no uninterpolated-name footgun, and when this `if:` skips it GitHub reports `skipped`, which rulesets count as satisfied.

Separately, adding `detekt` to `android-gate`'s `needs:` is still worth doing for the non-required "Android" signal — the gate loop only fails on `failure`/`cancelled`, so a skipped detekt passes through cleanly.

.github/workflows/pull_request.yml
```diff
@@ -243,6 +244,22 @@ jobs:
...
+        id: filter-kotlin
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            kotlin:
+              - 'android/**/*.kt'
+              - 'android/**/*.kts'
+              - 'android/gradle.properties'
+              - 'android/gradle/libs.versions.toml'
+              - 'android/config/detekt/**'
```
kaeawc, line 265: This filter and the global-file `case` in `detekt_scope.sh:104-114` are two independently-maintained lists, and they already disagree in two places — the filter is the outer gate, so anything it misses the script never gets to fail open on.

`android/gradle/wrapper/*` is a full-scope trigger in the script but has no glob here, so a wrapper/Gradle-version bump skips detekt entirely rather than running the full tree. Confirmed against picomatch:

```
node -e "const pm=require('picomatch');console.log(['android/**/*.kt','android/**/*.kts','android/gradle.properties','android/gradle/libs.versions.toml','android/config/detekt/**'].some(g=>pm.isMatch('android/gradle/wrapper/gradle-wrapper.properties',g)))"
# false
```

The other direction leaks too: the filter uses `android/config/detekt/**` but the script matches `android/config/detekt/*` (one level). A config file in a subdirectory passes the filter, the job starts, the script sees a non-`.kt` file, `continue`s past it, and emits an empty task list — detekt is skipped on a detekt-config change.

Suggested change so a wrapper bump reaches the script's fail-open:

```suggestion
              - 'android/**/*.kt'
              - 'android/**/*.kts'
              - 'android/gradle.properties'
              - 'android/gradle/libs.versions.toml'
              - 'android/gradle/wrapper/**'
              - 'android/config/detekt/**'
              - 'scripts/android/detekt_scope.sh'
              - '.github/actions/gradle-task-run/**'
```

(`detekt_scope.sh` and the composite are in there because a PR that changes the scoping mechanism itself currently gets zero detekt coverage of that change.) Widening `android/config/detekt/*` to `**` in the script closes the other side.

The 13 BATS cases all drive the script directly and none of them assert the two lists agree, so this drift is invisible today. A guard in the shape of `pr-gate-wiring.bats` — extract the `kotlin:` filter globs, extract the script's `case` patterns, assert every script pattern is matched by some filter glob — would pin it.

.github/workflows/merge.yml
```diff
@@ -243,10 +243,36 @@ jobs:
...
-        shell: "bash"
-        working-directory: android
-        run: ./gradlew detektMain detektTest
+  # The full-tree backstop. PRs run detekt scoped to the modules they changed
+  # (pull_request.yml), which cannot see a violation introduced in a *consumer*
+  # of a changed module -- this unscoped post-merge run is what catches that.
+  # Its own job rather than a step in ktfmt so it does not serialize behind a
+  # 20-second format check, and so it can reuse the Gradle caches.
+  detekt:
+    name: "Detekt"
```
kaeawc, line 252: Moving detekt out of `ktfmt` quietly drops it from the snapshot publish's dependency set. `publish-android-libraries-snapshot` lists `- ktfmt` at `merge.yml:1008`, and detekt used to be a step inside that job — so a detekt failure blocked the Maven Central snapshot. With detekt in its own job and nothing depending on it, the snapshot now publishes even when the full-tree backstop is red.

Given this run is explicitly the backstop for the scoping the PR introduces on the pull-request side, letting artifacts ship past it undoes most of the safety argument. Not a merge blocker for the PR itself, but it should land in the same change.

Suggested change for keeping the publish gated on the backstop — add to the `needs:` list at `merge.yml:1006-1015`:

```
      - ktfmt
      - detekt
```

Nothing else referenced the old ordering — `grep -rn '\- ktfmt' .github/workflows/` returns only that one line.

kaeawc (commented):
Reviewing the CI wiring only (shell internals, Gradle semantics, tests, security are other reviewers' lanes). The perf case is convincing and the fail-open design in `detekt_scope.sh` is the right default. Three wiring issues: the gating regression you flagged is real and I confirmed the ruleset shape, the snapshot publish in `merge.yml` silently loses its detekt dependency, and the paths-filter has a couple of gaps against the script's own global-file list.

Two things I checked and found fine: picomatch's `**` matches zero segments, so `android/**/*.kt` does cover `android/foo.kt`; and `if:` on a `uses:` step works as intended, so the empty-scope skip is correct. `gradle-home-directory: ~/.gradle/detekt` and the task-hash-derived `config-cache-report-*` artifact name are both unique within a run — no collision with the other `gradle-task-run` jobs.

scripts/android/detekt_scope.sh
```diff
@@ -0,0 +1,164 @@
...
+  esac
+
+  case "$file" in
+    *.kt | *.kts) ;;
+    *) continue ;;
+  esac
+  saw_kotlin=true
+
+  module_dir="$(module_dir_for "$file")"
+  [[ -n "$module_dir" ]] || continue
```
kaeawc: This is the one place the script narrows silently rather than failing open. A Kotlin file that `module_dir_for` cannot place is dropped, and the `saw_kotlin` backstop at line 149 only rescues it when it is the *only* Kotlin file in the diff — pair it with any real module change and the unplaceable file just vanishes from the scope.

The deleted-module justification in the comment at line 55 doesn't actually reach this line: removing a module requires dropping its `include(...)` from `settings.gradle.kts`, which hits the `case` fail-open at line 111 and exits with the full scope regardless of loop order. So the only diffs that reach this `continue` are Kotlin files owned by the *root* project — which unqualified `detektMain`/`detektTest` does analyze, and a scoped run does not. Nothing under `android/` has a root source set today, so this is latent rather than live, but it is a silent-narrowing hole and it costs one line to close.

Repro — `android/RootStray.kt` disappears from the scope as soon as `:protocol` also changes:
```
changed: android/RootStray.kt android/protocol/src/A.kt
emitted: :protocol:detektMain :protocol:detektTest
```
(with only `android/RootStray.kt` changed, the same script correctly emits `detektMain detektTest`)

Suggested change for symmetry with every other unprovable case:
```suggestion
  # A file that resolves to no module is owned by the root project (or by a
  # layout this walk does not understand); the scoped run would not analyze it.
  if [[ -z "$module_dir" ]]; then
    echo "detekt_scope.sh: '$file' resolves to no Gradle module; running full scope" >&2
    printf '%s\n' "$FULL_SCOPE"
    exit 0
  fi
```

scripts/android/detekt_scope.sh
```diff
@@ -0,0 +1,164 @@
...
+
+  # android/playground/design/system -> :playground:design:system
+  gradle_path=":${module_dir//\//:}"
+
+  # Confirm the derived path is a real Gradle project before handing it to
+  # gradlew: a directory with a build.gradle.kts that settings.gradle.kts never
+  # includes would fail the build on an otherwise-unrelated PR. Fixed-string
+  # match on the include() call, not a parse of the Kotlin.
+  if ! grep -qF "include(\"$gradle_path\")" "$SETTINGS_FILE"; then
```
kaeawc: `grep -qF` is an unanchored substring match, so a *disabled* include satisfies the guard — `// include(":legacy")` in `settings.gradle.kts` makes `:legacy` look like a real project and `:legacy:detektMain` reaches `gradlew`, which is exactly the bogus task name this check exists to prevent. It fails loud (`Project ':legacy' not found`) rather than silently, so it's not a blocker, but it turns an unrelated PR red.

The prefix case you were probably worried about is already safe — the trailing `")` means `include(":playground:design")` does not match the line `include(":playground:design:system")`. I confirmed that against the real settings file. It's only comments/conditionals that slip through.

Repro:
```
settings.gradle.kts:  include(":protocol")
                      // include(":legacy")
changed:              android/legacy/src/L.kt
emitted:              :legacy:detektMain :legacy:detektTest
```

`-x` (whole-line match) fixes it and still matches all 24 real includes, which sit at column 1:

Suggested change so a commented-out include can't pass:
```suggestion
  if ! grep -qxF "include(\"$gradle_path\")" "$SETTINGS_FILE"; then
```

scripts/android/detekt_scope.sh
```diff
@@ -0,0 +1,164 @@
...
+if ! git -C "$PROJECT_ROOT" rev-parse --verify --quiet "$BASE_SHA^{commit}" > /dev/null; then
+  echo "detekt_scope.sh: base SHA '$BASE_SHA' does not resolve; running full scope" >&2
+  printf '%s\n' "$FULL_SCOPE"
+  exit 0
+fi
+
+# No --diff-filter here: a *deleted* Kotlin file still changes what its module
+# analyzes, so the module must be re-run. The module-dir walk above already
+# drops files whose module no longer exists.
+changed_files="$(git -C "$PROJECT_ROOT" diff --name-only "$BASE_SHA...HEAD" -- "$ANDROID_DIR")"
```
kaeawc: The header promises fail-open on an unresolvable base, but only `rev-parse` is guarded — this `git diff` can still hard-fail under `set -e` and exit 128 with empty stdout. Three-dot needs a merge base, and `rev-parse --verify` succeeding doesn't guarantee one exists.

Repro:
```
$ git checkout --orphan orphan && git commit -am orphan
$ ONLY_CHANGED_SINCE_SHA=$BASE scripts/android/detekt_scope.sh
fatal: <base>...HEAD: no merge base
EXIT=128
```

In the `pull_request` job this is close to unreachable (`fetch-depth: 0` plus the merge ref), so not a blocker — but the cheap version of the guarantee is to catch the failure rather than let `set -e` kill the step:

Suggested change for the documented fail-open contract:
```suggestion
if ! changed_files="$(git -C "$PROJECT_ROOT" diff --name-only "$BASE_SHA...HEAD" -- "$ANDROID_DIR" 2>/dev/null)"; then
  echo "detekt_scope.sh: cannot diff '$BASE_SHA...HEAD'; running full scope" >&2
  printf '%s\n' "$FULL_SCOPE"
  exit 0
fi
```

scripts/android/detekt_scope.sh
```diff
@@ -0,0 +1,164 @@
...
+# is slow; a wrongly-narrow run lets a violation reach main.
+#
+# Scoping is deliberately per-module and does NOT walk down to dependent
+# modules: a change to :protocol will not re-run detekt on the modules that
+# consume it. detekt rules are almost entirely file-local, and merge.yml runs
+# the full tree post-merge as the backstop for the residual risk.
+
+set -euo pipefail
+
+PROJECT_ROOT="${PROJECT_ROOT:-$(pwd)}"
```
kaeawc: Nit: `PROJECT_ROOT` defaulting to `$(pwd)` makes a wrong cwd read as "nothing to do" instead of an error. Run the documented `scripts/android/detekt_scope.sh --since <sha>` form from `android/` and the `-- android` pathspec at line 99 matches nothing, so the script exits 0 with empty output — and the caller contract at line 17 says empty means *skip detekt entirely*. Silent full skip from a plausible local invocation.

Repro:
```
$ cd android && ../scripts/android/detekt_scope.sh --since $BASE
exit=0 out=[]
```

A one-line sanity check on `$PROJECT_ROOT/$ANDROID_DIR` next to the `command -v git` check would make it loud.

kaeawc (commented):
Reviewed with a narrow lens: correctness of the scoping logic in `scripts/android/detekt_scope.sh`. I independently re-ran your module-mapping claim over all 676 tracked `.kt`/`.kts` files under `android/` and every one resolves to an included project (only `android/build.gradle.kts` and `android/settings.gradle.kts` resolve to nothing, and both are caught by the `case` fail-open first) — that validation holds. The three-dot anchor is the right choice too, and worth noting it is even safer than the comment claims: `actions/checkout` on a `pull_request` event checks out `refs/pull/N/merge`, so `base.sha` is already an ancestor of `HEAD` and the merge-base collapses to `base.sha` itself, making the diff a superset of the PR's own changes. All four `saw_kotlin` × `modules` combinations are handled correctly (the fourth is unreachable), and the here-string `while` loop runs in the main shell so the `modules` array survives.

Four things below, each reproduced in a throwaway git repo. The first is the one I'd fix before merge — it is the only path I found where the script narrows *silently*.

`bats test/bats/detekt-scope.bats` passes 13/13; none of the four cases below are covered by it.

scripts/android/detekt_scope.sh
```diff
@@ -0,0 +1,164 @@
...
+    fi
+    if [[ -f "$PROJECT_ROOT/$dir/build.gradle.kts" ]]; then
+      printf '%s\n' "${dir#"$ANDROID_DIR"/}"
+      return 0
+    fi
+    dir="$(dirname "$dir")"
+  done
+}
+
+SETTINGS_FILE="$PROJECT_ROOT/$ANDROID_DIR/settings.gradle.kts"
```
kaeawc: `PROJECT_ROOT` defaults to `$(pwd)` and is never validated, so a wrong root makes the `-- android` pathspec match nothing — `saw_kotlin` stays false, `modules` stays empty, and line 155 exits 0 with empty output, which the header at line 17 tells callers to treat as "skip detekt entirely". Every other unprovable case here fails *open* to the full list; this one silently skips the entire analysis.

The workflow runs from the repo root today so it isn't live, but nothing pins that — a `working-directory:` added to the step later turns a green Detekt check into a no-op.

```
cd android && ONLY_CHANGED_SINCE_SHA=$base bash ../scripts/android/detekt_scope.sh
# exit 0, empty output
```

Not a blocker, but it's cheap and it protects the one outcome the script exists to prevent. I applied the below locally: shellcheck-clean, all 13 tests still green, and the repro above now prints the full scope instead of nothing.

Suggested change for making a bad root fail open like every other unprovable case:

```suggestion
SETTINGS_FILE="$PROJECT_ROOT/$ANDROID_DIR/settings.gradle.kts"

# A wrong PROJECT_ROOT would make every pathspec match nothing and emit the
# empty "skip detekt" output. Fail open like every other unprovable case.
if [[ ! -f "$SETTINGS_FILE" ]]; then
  echo "detekt_scope.sh: no $ANDROID_DIR/settings.gradle.kts under '$PROJECT_ROOT'; running full scope" >&2
  printf '%s\n' "$FULL_SCOPE"
  exit 0
fi
```

scripts/android/detekt_scope.sh
```diff
@@ -0,0 +1,164 @@
...
+
+  modules+=("$gradle_path")
+done <<< "$changed_files"
+
+# Kotlin changed but every file resolved to a deleted module: nothing to analyze
+# there, but do not silently pass on a diff we could not place.
+if [[ "$saw_kotlin" == "true" && ${#modules[@]} -eq 0 ]]; then
+  printf '%s\n' "$FULL_SCOPE"
+  exit 0
+fi
```
kaeawc: Nothing covers this branch. It's the only fail-open that depends on a flag threaded through the loop rather than a direct `case` match, so it's the one most likely to break under a refactor — and when it breaks the output is empty, which line 17 defines as "skip detekt entirely". A wrongly-narrow scope is the exact failure mode the header says must never happen, and it's the branch with no test.

It behaves correctly today:

```
echo '// x' > android/Stray.kt   # a .kt under android/ but inside no module
scripts/android/detekt_scope.sh --since $base   # -> detektMain detektTest
```

This one passes as written against the current script:

```bash
@test "a Kotlin file inside no module falls back to the full scope" {
  commit_change "android/Stray.kt"
  cd "$REPO" || return 1
  run_scope
  [ "$status" -eq 0 ]
  [ "$output" = "detektMain detektTest" ]
}
```

scripts/android/detekt_scope.sh
```diff
@@ -0,0 +1,164 @@
...
+
+if ! git -C "$PROJECT_ROOT" rev-parse --verify --quiet "$BASE_SHA^{commit}" > /dev/null; then
+  echo "detekt_scope.sh: base SHA '$BASE_SHA' does not resolve; running full scope" >&2
+  printf '%s\n' "$FULL_SCOPE"
+  exit 0
+fi
+
+# No --diff-filter here: a *deleted* Kotlin file still changes what its module
+# analyzes, so the module must be re-run. The module-dir walk above already
+# drops files whose module no longer exists.
```
kaeawc: Not reusing `collect_changed_since_sha` from `scripts/lib/file-selection.sh` was the right call, and I want to say so explicitly because the repo's stdlib-first rule ("prefer an existing repository helper") points the other way and the next reviewer will ask. Three concrete reasons it doesn't fit: it filters through `[[ -f "$project_root/$file" ]]`, which drops exactly the deletions this comment says must be kept; it emits absolute paths, while `module_dir_for` needs repo-relative ones; and it runs `git diff` from cwd rather than `-C "$project_root"`.

The header already documents the SC2310 inlining but says nothing about `file-selection.sh`. One line here turns a judgment call somebody re-litigates into a settled one.

Suggested change for recording why the shared helper doesn't fit:

```suggestion
  # No --diff-filter here: a *deleted* Kotlin file still changes what its module
  # analyzes, so the module must be re-run. The module-dir walk above already
  # drops files whose module no longer exists. This is also why
  # scripts/lib/file-selection.sh's collect_changed_since_sha is not reused: it
  # filters through `[[ -f ... ]]`, which would silently drop those deletions.
```

test/bats/detekt-scope.bats
```diff
@@ -0,0 +1,179 @@
...
+  local path="$1"
+  mkdir -p "$(dirname "$REPO/$path")"
+  echo "// changed" >> "$REPO/$path"
+  git -C "$REPO" add -A
+  git -C "$REPO" commit --quiet -m "change $path"
+}
+
+run_scope() {
+  run env PROJECT_ROOT="$REPO" bash "$REPO/scripts/android/detekt_scope.sh" --since "$BASE"
+}
```
kaeawc: All 13 tests go through `--since`, but CI never uses `--since` — the workflow sets `ONLY_CHANGED_SINCE_SHA` and calls the script bare. So the suite exercises a path CI doesn't take and leaves the path CI does take completely unpinned. Same story for `PROJECT_ROOT`: every test passes it explicitly, while CI relies on the `$(pwd)` default at `detekt_scope.sh:31`.

That gap matters because failure there is invisible. Delete the `${ONLY_CHANGED_SINCE_SHA:-}` on `detekt_scope.sh:37` and this suite still goes 13/13 green while CI silently full-scopes every PR — fail-open, no red check, just a permanent quiet return to the 7-minute run this PR exists to remove. That's the shape of tests pinning the implementation rather than the behavior.

One test closes it. I ran this against the real script and it passes:

```bash
@test "reads the base SHA from ONLY_CHANGED_SINCE_SHA" {
  commit_change "android/protocol/src/main/kotlin/Foo.kt"
  cd "$REPO" || return 1
  run env -u PROJECT_ROOT ONLY_CHANGED_SINCE_SHA="$BASE" bash "$REPO/scripts/android/detekt_scope.sh"
  [ "$status" -eq 0 ]
  [ "$output" = ":protocol:detektMain :protocol:detektTest" ]
}
```

Note what the `env -u PROJECT_ROOT` does there — right now the `cd "$REPO"` in every test is redundant, because the script reads only `$PROJECT_ROOT` and never cwd. Dropping `PROJECT_ROOT` is what makes that `cd` load-bearing and actually exercises the default.

test/bats/detekt-scope.bats
```diff
@@ -0,0 +1,179 @@
...
+  rm "$REPO/android/protocol/src/main/kotlin/Doomed.kt"
+  git -C "$REPO" add -A
+  git -C "$REPO" commit --quiet -m "delete"
+  cd "$REPO" || return 1
+  run_scope
+  [ "$status" -eq 0 ]
+  [ "$output" = ":protocol:detektMain :protocol:detektTest" ]
+}
+
+# --- fail-open cases: scoping is not provably safe, so run everything ---
```
kaeawc: Two arms of the `case` at `detekt_scope.sh:110` have no test here: `android/gradle.properties` and `android/gradle/wrapper/*`. The wrapper arm is the only pattern in that block with a glob, so it's the likeliest to be subtly wrong and the one nothing checks. Both behave correctly today — I ran them — so this is coverage, not a bug.

It does make the PR body's "every fail-open case" claim inaccurate. Counting the `saw_kotlin`/empty-modules branch at line 149, three of the script's seven fail-open paths are unpinned. Worth adding them or softening the wording.

One thing to reconcile while you're in here: `pull_request.yml:256` doesn't list `android/gradle/wrapper/**` in the `kotlin` filter, so a wrapper-only bump never reaches the script and that `case` arm can't fire in CI at all. Fine either way — a wrapper bump has no Kotlin to analyze — but the script and the filter currently disagree about whether a wrapper change matters, and nothing records which one is authoritative.

kaeawc (commented):
Reviewed for test quality and fit with the repo's conventions. The core design is sound and the module mapping holds up — I re-derived it independently against the real tree and every `build.gradle.kts` directory under `android/` resolves to an included project, with only the two root `.kts` files unmapped, which are already fail-open cases. That matches your 674 claim exactly (676 `.kt`/`.kts` minus the two root files). `shellcheck`, `shell-portability` and `shell-sete` all pass on the new script, and the 13 tests pass in ~11s.

Two things I checked and am **not** flagging, since both would be easy gotchas:

- **The `.github/CLAUDE.md` "don't add a standalone validation workflow" rule.** This doesn't violate it, in letter or spirit. It's a job inside the existing `pull_request.yml`, not a new workflow file, and `all_fast_validate_checks.sh` is explicitly a fast parallel shell-check runner — folding a 7-minute Gradle type-resolution task into it would recreate the exact problem this PR fixes. Right call.
- **Test isolation.** The throwaway-git-repo-in-`setup()` pattern matches `apply-ktfmt.bats` and `validate-ktfmt.bats`, `mktemp -d` keeps it parallel-safe, nothing leaks through `PROJECT_ROOT`, and cleanup is correct.

The findings are all about what the tests *don't* pin. The theme: the suite covers the `--since` path thoroughly and the `ONLY_CHANGED_SINCE_SHA` path — the one CI actually uses — not at all, and the fail-open cases with tests are the easy ones while the three subtle ones are bare. Every gap I filed I reproduced, and both suggested tests pass against the current script unmodified.

**Verify first: `Detekt` is not a required status check.** I enumerated the `green-main` ruleset via the API — 25 contexts, `Fast Validation` among them, `Detekt` not. Your PR body flags this, and confirming it: as of this commit, merging this makes detekt non-blocking. Add `Detekt` to the ruleset in the same window you merge, or a violation reaches `main` with a green tick.

.github/workflows/pull_request.yml
kaeawc, line 693: Confirming the gating regression you flagged, against the live ruleset — and answering the question you left open.

`gh api repos/kaeawc/auto-mobile/rulesets/11406121` lists `Fast Validation` as a required check and has no `Detekt` entry, so the moment this merges detekt stops gating PRs entirely. That part is exactly as you described.

The thing worth knowing before someone acts on it: **adding `Detekt` verbatim to the required list is safe, and it does not need a roll-up gate.** I checked whether a conditional job would hang non-Kotlin PRs on an "Expected" status — it doesn't, and this repo already proves it: `Build Desktop Core Module` and `Run Desktop Core Unit Tests` are both required *and* both `if:`-gated on `desktop_core_changed`, and TS-only PRs merge fine. A job skipped by a job-level `if` reports a check run with conclusion `skipped`, which required-status-checks treats as satisfied.

Which means `shell-tests-gate` is not the precedent to copy here. Its own comment says why it exists — `bats-tests` is an OS matrix, so when gated out it reports the literal `(${{ matrix.os }})` name and requiring the expanded `(macos-latest)` context hangs. `Detekt` isn't a matrix job; its check name is the stable string `Detekt` whether it runs or skips.

So: add `Detekt` to the ruleset, no workflow change needed. Worth doing in the same window as the merge so the ungated gap is minutes rather than days.

scripts/android/detekt_scope.sh
kaeawc: The comment right above this says a deleted Kotlin file must still re-run its module, but rename detection means that isn't true for a *move* — `git diff --name-only` collapses a rename into the destination path only, so the source module never enters `modules` and never gets re-analyzed.

Reproduced on a throwaway repo shaped like `android/`, moving a file from `:protocol` to `:playground:app`:

```
$ git mv android/protocol/src/main/Seed.kt android/playground/app/src/main/Seed.kt && git commit -qm move
$ git diff --name-only "$BASE...HEAD"
android/playground/app/src/main/Seed.kt
$ detekt_scope.sh --since "$BASE"
:playground:app:detektMain :playground:app:detektTest
```

I don't think this is a bypass and it shouldn't block — removing a file from `:protocol` can only remove findings there, not introduce one, so a violation still can't ride through. It's the invariant in the comment and the `a deleted Kotlin file still re-runs its module` BATS test being narrower than they read: that test covers a plain delete, and a rename quietly doesn't behave the same way.

Suggested change for making the stated invariant actually hold:

```suggestion
changed_files="$(git -C "$PROJECT_ROOT" diff --no-renames --name-only "$BASE_SHA...HEAD" -- "$ANDROID_DIR")"
```

That makes a rename surface as delete + add, so both modules run. Worth a companion BATS case pinning the cross-module move, since that's the one the current suite doesn't distinguish.

kaeawc (commented):
Reviewed this adversarially — I spent the pass trying to construct a PR that carries a Kotlin violation past the new gate, and **I could not build one**. Recording the negatives, since they're the substance of the review:

- Every `.kt`/`.kts` in the repo is under `android/`, so the `android/**/*.kt` + `android/**/*.kts` filter cannot structurally miss a Kotlin change. No Kotlin lives outside that prefix.
- I re-ran your module-mapping claim independently against all 676 Kotlin files at this head: every one resolves to a directory whose derived Gradle path appears in an `include("...")` in `settings.gradle.kts`. The only two that walk to no module are `android/build.gradle.kts` and `android/settings.gradle.kts`, and both are caught by the full-scope `case` first. The claim holds.
- The "file belongs to no module, gets silently dropped" path is real in the script but unreachable here — it needs Kotlin outside a module dir, and the only custom `srcDir` calls are `resources.srcDir(schemas)` and a generated-source dir under `build/`. Not worth guarding today.
- `sha256_only` and `auto_chore` can't suppress a Kotlin-carrying PR. `sha256_only` requires `files.length === 1 && filename === 'src/constants/release.ts'`; `auto_chore` requires `files.every(isDocFile)`. Neither can coexist with a `.kt`. A title alone doesn't do it.
- I tried twice to push a shell metacharacter through the emitted task string into the composite's `./gradlew ${{ inputs.gradle-tasks }}` and failed both times — the `${module_dir//\//:}` substitution mangles any payload containing `/`, and the `grep -qF` gate requires the mangled result to already be blessed in `settings.gradle.kts`. See the inline note; it's hardening, not a live hole.
- Action pinning is consistent with the repo (`actions/checkout@v6` ×83, `actions/upload-artifact@v6` ×57). `fetch-depth: 0` is used in four other places in these workflows; it exposes nothing extra here.
- All 13 BATS tests pass locally.

Three inline notes below: one confirming the gating regression against the live ruleset (and answering the question you left open), one small `git diff` behavior fix, one hardening suggestion.

On the accepted gap — scoping not walking to dependent modules — I think the `merge.yml` backstop is adequate *for detecting* it, with the caveat that it detects post-merge and there's no notification wired to that job beyond turning the On Merge run red. That's the same posture as every other job in `merge.yml`, so it's not a regression this PR introduces; just be aware the backstop's response time is "whoever notices main is red."

.github/workflows/pull_request.yml
```diff
@@ -659,6 +671,68 @@ jobs:
...
+        id: scope
+        shell: bash
+        env:
+          # Empty or unresolvable (base force-pushed/rebased) makes the script
+          # fall open to the full task list rather than silently scoping to
+          # nothing -- see the fail-open note in detekt_scope.sh.
+          ONLY_CHANGED_SINCE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: |
+          tasks="$(scripts/android/detekt_scope.sh)"
+          echo "tasks=$tasks" >> "$GITHUB_OUTPUT"
```
kaeawc: This is the first caller to feed `gradle-tasks` a computed value rather than a literal — every other call site in this file (`:desktop-core:build -x test`, `"test"`, the long junit-runner list) is a quoted constant. That matters because the composite splats it unquoted into a shell line: `"${GITHUB_WORKSPACE}/scripts/ci/run-gradle-with-retry.sh" -- ./gradlew ${{ inputs.gradle-tasks }} $GRADLE_FLAGS`. Anything that reaches this output becomes shell text on a runner holding `GRADLE_ENCRYPTION_KEY`.

I tried to exploit it and couldn't. Two attempts, both fell open to the full scope:

```
# module dir literally named 'a;id>/tmp/pwn.txt;x', blessed in settings.gradle.kts on the base commit
$ detekt_scope.sh --since "$BASE"
detekt_scope.sh: ':a;id>:tmp:pwn.txt;x' is not an included project; running full scope
detektMain detektTest
```

`${module_dir//\//:}` rewrites every `/` to `:`, so any payload needing a path separator is mangled before the `grep -qF "include(\"$gradle_path\")"` check, which then fails and full-scopes. A slash-free payload would have to already be committed as an `include()` in `settings.gradle.kts` — and any PR that edits that file hits the full-scope `case` first. The `$GITHUB_OUTPUT` write is fine for the same reason plus one more: the script only ever `printf`s a single line, so there's no newline to forge a second output with.

So the gate holds, but it holds because of two incidental properties rather than anything asserting it. One line makes it explicit and cheap:

Suggested change for pinning the output shape independently of the settings-file gate:

```suggestion
          tasks="$(scripts/android/detekt_scope.sh)"
          if [[ -n "$tasks" && ! "$tasks" =~ ^[A-Za-z0-9:_\ -]+$ ]]; then
            echo "::error::detekt scope produced an unexpected task string: $tasks"
            exit 1
          fi
          echo "tasks=$tasks" >> "$GITHUB_OUTPUT"
```

Not a blocker — file it as hardening if you'd rather not grow the step.

scripts/android/detekt_scope.sh
```diff
@@ -0,0 +1,164 @@
...
+#
+# Fails *open* (emits the full-tree task list) whenever scoping cannot be proven
+# safe: no base SHA, an unresolvable base SHA, a root build file change, or a
+# module directory that is not an included Gradle project. A needlessly-full run
+# is slow; a wrongly-narrow run lets a violation reach main.
+#
+# Scoping is deliberately per-module and does NOT walk down to dependent
+# modules: a change to :protocol will not re-run detekt on the modules that
+# consume it. detekt rules are almost entirely file-local, and merge.yml runs
+# the full tree post-merge as the backstop for the residual risk.
```
kaeawc: `config/detekt/detekt.yml` turns off `complexity`, `coroutines`, `empty-blocks`, `exceptions`, `naming`, `performance` and `style`, leaving `potential-bugs` as the only active ruleset — and its 21 default-active rules are almost exactly the type-resolution ones whose verdict depends on types declared in *other* modules (`UnnecessarySafeCall`, `UnsafeCallOnNullableType`, `UnnecessaryNotNullOperator`, `UnreachableCode`, `HasPlatformType`, `IgnoredReturnValue`, `UnsafeCast`, `MapGetWithNotNullAssertionOperator`). The file-local rulesets are precisely the ones this repo has disabled, so "almost entirely file-local" is inverted for this configuration, and the residual risk is the common case rather than a corner.

I asked whether the gap was constructible and it is, on real modules:

```
# baseline: :playground:experimentation calls a nullable :playground:storage API
#   ProbeConsumer.kt: repo.getUserProfile()?.name        -> detekt clean

# the "PR": change ONLY UserRepository.kt in :playground:storage,
#   fun getUserProfile(): UserProfile?  ->  fun getUserProfile(): UserProfile

$ ONLY_CHANGED_SINCE_SHA=$BASE scripts/android/detekt_scope.sh
:playground:storage:detektMain :playground:storage:detektTest

$ ./gradlew :playground:storage:detektMain :playground:storage:detektTest
BUILD SUCCESSFUL

$ ./gradlew :playground:experimentation:detektMain
ProbeConsumer.kt:6:52 repo.getUserProfile()?.name contains an unnecessary safe call operator [UnnecessarySafeCall]
BUILD FAILED
```

One file changed, in one module, and the scoped run is green while the tree is red. `merge.yml` catches it post-merge, so this is a "main goes red after the merge button" regression rather than a silent one — but it lands on `main`, which is what the required-check gate exists to prevent.

Two ways out that I think are both defensible: expand the emitted scope to the changed modules' dependents (Gradle already knows the graph — a `-q` task that walks `configurations` and prints reverse dependencies would keep this honest without giving up most of the win, since the consumer sets here are small), or keep per-module scoping and make the comment say what is actually true. If you take the second, the comment is the thing to fix:

Suggested change for accuracy about which rules are at risk:

```suggestion
# consume it. This repo runs only the `potential-bugs` ruleset, whose active
# rules are largely type-resolution rules that read types from dependency
# modules, so a violation CAN appear in an untouched consumer. merge.yml runs
# the full tree post-merge and is the only thing that catches that case.
```

android/build.gradle.kts
```diff
@@ -65,6 +65,17 @@ subprojects {
...
+      // org.gradle.parallel only parallelizes across projects, so without this a
+      // large module analyzes single-threaded.
+      parallel = true
+      // On Android modules `detektMain` fans out to one task per build type, so
+      // `release` re-analyzes the exact same `main` sources `debug` already
+      // covered -- no module has a `src/release` source set, and the only
+      // build-type source set in the project is `auto-mobile-sdk/src/debug`.
+      // Skipping release halves the Android analysis and, more importantly,
+      // avoids compiling the release variant classpath purely to feed detekt.
+      ignoredBuildTypes = listOf("release")
```
kaeawc, line 78: Worth saying plainly: I tried to break this one and could not. Dumping every detekt task's resolved `source` file set through an init script, across all 16 modules that have a `detektRelease`, every release task's file set is a subset of the matching `detektDebug`:

```
:auto-mobile-sdk   rel=52 dbg=56  release-only=0   (the 4 extra are src/debug)
:control-proxy     rel=44 dbg=44  release-only=0
... 14 playground modules, all release-only=0
union of main-side sources: 447 before, 447 after, 0 lost
```

The task-graph diff is exactly 16 `:*:detektRelease` tasks removed and nothing else added or dropped, and the union of the remaining 96 tasks still covers all 650 on-disk `.kt` files under `android/`. There are no product flavors, no `buildConfigField`, no `src/release` anywhere, and `isMinifyEnabled = false` on every release block, so there is no variant-specific source to lose.

The classpath direction is also the safe one — the release variant has strictly fewer dependencies (`debugImplementation` adds compose tooling in 5 modules), so release resolved fewer types and could only ever produce a subset of debug's findings. Dropping it removes the weaker of the two analyses.

android/build.gradle.kts
```diff
@@ -65,6 +65,17 @@ subprojects {
     extensions.configure<DetektExtension> {
       config.setFrom(rootProject.files("config/detekt/detekt.yml"))
       buildUponDefaultConfig = true
+      // Analyze files across threads within a single detekt task. Gradle's
+      // org.gradle.parallel only parallelizes across projects, so without this a
+      // large module analyzes single-threaded.
+      parallel = true
```
kaeawc, line 71: Checked this is not a no-op: `DetektBasePlugin` sets `getParallel().convention(false)`, and the extension value flows into `ParallelArgument`, which emits `--parallel` to the detekt CLI. So it does real work on 2.0.0-alpha.5.

I went looking for the nondeterminism risk — parallel analysis reordering findings, or a flaky pass/fail — and could not construct one: every module is currently clean, so there are no findings whose order could vary, and the pass/fail decision is a count against a severity threshold rather than anything order-dependent. Not raising it as a concern, just noting I couldn't rule it in either.

.github/workflows/pull_request.yml
```diff
@@ -659,6 +671,68 @@ jobs:
...
+            echo "Running detekt for: \`$tasks\`" >> "$GITHUB_STEP_SUMMARY"
+          fi
+
+      - uses: ./.github/actions/gradle-task-run
+        if: steps.scope.outputs.tasks != ''
+        with:
+          gradle-tasks: ${{ steps.scope.outputs.tasks }}
+          gradle-project-directory: "android"
+          gradle-home-directory: "~/.gradle/${{ github.job }}"
+          reuse-configuration-cache: true
```
kaeawc, line 713: Nit, and not a blocker. Gradle keys the configuration cache entry on the requested task set, which the scoped job varies per PR:

```
$ ./gradlew :protocol:detektMain :protocol:detektTest
Calculating task graph as no cached configuration is available for tasks: :protocol:detektMain :protocol:detektTest
Configuration cache entry stored.
$ ./gradlew :video-server:detektMain :video-server:detektTest
Configuration cache entry stored.        # miss, different task set
```

So even in the best case where an entry survives into the job, `merge.yml`'s full-tree `detektMain detektTest` entry can never serve a PR asking for `:protocol:detektMain`, and two PRs touching different modules can't serve each other. The build cache is the part actually carrying the speedup here — `Detekt` is `@CacheableTask` and the expensive upstream Kotlin compilations are cacheable, and pointing both jobs at `~/.gradle/${{ github.job }}` (`detekt` in both files) so main's post-merge run seeds the PR cache is a nice touch.

Worth double-checking separately: the config cache lives in `android/.gradle/configuration-cache`, which is the project dir, not `GRADLE_USER_HOME` — and `setup-gradle` caches the latter. If that's right, this input has never restored an entry for any job in the repo, which would be pre-existing and out of scope here, but it makes me read the "configuration cache" part of the PR description as optimistic.

kaeawc (commented):
Reviewed for detekt/Gradle semantics — specifically whether the optimization preserves analysis coverage. I ran the Gradle side rather than reasoning about it: task-graph diffs, per-task resolved source sets via an init script, and a constructed cross-module regression.

The release-variant skip is clean — I verified it loses exactly zero source files across all 16 affected modules, and the task-graph delta is precisely the 16 `detektRelease` tasks. `parallel = true` is a real setting, not a no-op. Both scoped task names resolve for all 24 modules, and `auto-mobile-sdk`'s `src/debug` sources stay covered because `detektMain` still routes through `detektDebug`.

The one thing I'd want addressed before this merges is the dependent-module gap. The PR frames it as residual risk on the grounds that detekt rules are near-entirely file-local — but this repo has disabled every file-local ruleset and kept only `potential-bugs`, which is the cross-module type-resolution set. I reproduced a green scoped run against a red tree from a one-file, one-module change. Details inline.

scripts/android/detekt_scope.sh
kaeawc: This is the finding that changed the design — thank you, the premise was inverted and I had it backwards in the header, the commit message and the PR body. Scoping is gone as of bddba0d10, so `detekt_scope.sh` is deleted and the PR job runs the full `detektMain detektTest`. Worth recording why removing it was cheap: the 30s warm-cache measurement was taken at full scope, so the build cache was already carrying the speedup and scoping was buying almost nothing on top of it while risking the #4016 -> #4020 incident class.

scripts/android/detekt_scope.sh
kaeawc: Correct, and your reasoning about reachability was right — deleting a module edits `settings.gradle.kts`, which hits the `case` fail-open first, so the comment justified a case that could never reach that line. Moot now: scoping is removed in bddba0d10 and the script is deleted.

scripts/android/detekt_scope.sh
kaeawc: Real bug — I confirmed `grep -qxF` matches all 24 real includes and rejects `// include(":legacy")` before deciding what to do with it. Overtaken by events: the script is deleted in bddba0d10.

scripts/android/detekt_scope.sh
kaeawc: Agreed it contradicted the documented fail-open, though it failed loud rather than silent — under `set -e` the command substitution propagates 128 and the step fails. Moot now, script deleted in bddba0d10.

scripts/android/detekt_scope.sh
kaeawc: Two of you found this independently, which is what moved it off "nit" for me. Moot now — script deleted in bddba0d10.

scripts/android/detekt_scope.sh
kaeawc: You and the shell reviewer found this independently, which raised my confidence it was worth fixing rather than documenting. Moot now — script deleted in bddba0d10.

scripts/android/detekt_scope.sh
kaeawc: Fair — it was the only fail-open depending on a loop-threaded flag rather than a direct `case` match. Moot now, script deleted in bddba0d10.

scripts/android/detekt_scope.sh
kaeawc: Appreciated you affirming this one with concrete reasons rather than leaving it as an open question — the deletions/absolute-paths/cwd mismatches were exactly right. Academic now that the script is gone in bddba0d10, but it saved re-litigating the stdlib-first rule while the decision was still live.

test/bats/detekt-scope.bats
kaeawc: The sharpest finding in the review — the suite pinned `--since` thoroughly and the `ONLY_CHANGED_SINCE_SHA` path CI actually used not at all, so deleting the default would have left 13/13 green while every PR silently full-scoped. It also made my "covers every fail-open case" claim in the PR body an overclaim, which I corrected. Both files are deleted in bddba0d10.

test/bats/detekt-scope.bats
kaeawc: Right on both counts — 3 of 7 arms unpinned, and the PR body overclaimed. Corrected the body; the tests are deleted with the script in bddba0d10.

scripts/android/detekt_scope.sh
kaeawc: Good catch that `git diff --name-only` collapses a rename to the destination, and I appreciate you explicitly saying it was not a bypass rather than inflating it. Moot now — script deleted in bddba0d10.

.github/workflows/merge.yml
kaeawc: Fixed in bddba0d10 — this was a real regression I introduced and did not check for when moving the step out of `ktfmt`. `detekt` is now in `publish-android-libraries-snapshot`'s `needs:`, with a comment recording that the coverage used to be transitive so it does not get dropped again.

.github/workflows/pull_request.yml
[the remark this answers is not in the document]
kaeawc: Fixed in bddba0d10 — added `android/gradle/wrapper/**` and `.github/actions/gradle-task-run/**` to the `kotlin` filter, and reworded the comment to say why the non-Kotlin entries are there: they change what detekt reports on unchanged sources.

kaeawc: The drift hazard is gone with the script in bddba0d10, and I added the wrapper glob. One correction for the record: the `config/detekt/**` vs `config/detekt/*` half was not a real mismatch — unlike pathname expansion, `*` in a bash `case` pattern matches across `/`, so the script's glob did match nested files. Verified before acting on it. The wrapper half was real.

.github/workflows/pull_request.yml
kaeawc: Moot as of bddba0d10 — `gradle-tasks` is back to the literal `"detektMain detektTest"`, so it is no longer the first computed-value caller. Reasonable hardening ask while the computed value existed.

android/build.gradle.kts
kaeawc: This is the kind of negative result that is worth more than a finding — dumping the resolved `source` file set per task and diffing the graph is a stronger check than I did, and it is why `ignoredBuildTypes` survived the rewrite while scoping did not. Kept as-is.

android/build.gradle.kts
kaeawc: Thanks for checking it was not a no-op rather than assuming — `convention(false)` -> `ParallelArgument` is the confirmation I did not have. Kept as-is.

.github/workflows/pull_request.yml
kaeawc: Moot as of bddba0d10 now that both jobs request the same full task set, so the config-cache entry can actually be shared. Your separate point about the config cache living in the project dir rather than `GRADLE_USER_HOME` is pre-existing and I have left it alone as out of scope.

.github/workflows/pull_request.yml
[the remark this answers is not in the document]
kaeawc: Leaving this thread **open on purpose** — it is the one finding in the review that code cannot close. `Detekt` still needs adding to the `green-main` ruleset's required contexts, which is a repo settings change I have not made on my own.

Consolidating what the three of you established, so whoever flips it has it in one place: the roll-up gates (`Android`, `iOS`, `Shell Tests`) are **not** required contexts either, so wiring `detekt` into `android-gate` would not restore blocking — only the literal `Detekt` context does. Requiring it directly is safe despite the job being conditional: it is non-matrix, so there is no uninterpolated-name footgun, and an `if:`-skipped job reports `skipped`, which rulesets count as satisfied. `Build Desktop Core Module` is already both required and `if:`-gated and TS-only PRs merge fine, which is the precedent.

Until that lands, merging this PR makes detekt non-blocking on PRs. The post-merge run in `merge.yml` still gates the snapshot publish (restored in bddba0d10), so a violation cannot ship to Maven Central — but it can reach `main`.

kaeawc: Leaving this thread **open on purpose** — it is the one finding in the review that code cannot close. `Detekt` still needs adding to the `green-main` ruleset's required contexts, which is a repo settings change I have not made on my own.

Consolidating what the three of you established, so whoever flips it has it in one place: the roll-up gates (`Android`, `iOS`, `Shell Tests`) are **not** required contexts either, so wiring `detekt` into `android-gate` would not restore blocking — only the literal `Detekt` context does. Requiring it directly is safe despite the job being conditional: it is non-matrix, so there is no uninterpolated-name footgun, and an `if:`-skipped job reports `skipped`, which rulesets count as satisfied. `Build Desktop Core Module` is already both required and `if:`-gated and TS-only PRs merge fine, which is the precedent.

Until that lands, merging this PR makes detekt non-blocking on PRs. The post-merge run in `merge.yml` still gates the snapshot publish (restored in bddba0d10), so a violation cannot ship to Maven Central — but it can reach `main`.

kaeawc: Leaving this thread **open on purpose** — it is the one finding in the review that code cannot close. `Detekt` still needs adding to the `green-main` ruleset's required contexts, which is a repo settings change I have not made on my own.

Consolidating what the three of you established, so whoever flips it has it in one place: the roll-up gates (`Android`, `iOS`, `Shell Tests`) are **not** required contexts either, so wiring `detekt` into `android-gate` would not restore blocking — only the literal `Detekt` context does. Requiring it directly is safe despite the job being conditional: it is non-matrix, so there is no uninterpolated-name footgun, and an `if:`-skipped job reports `skipped`, which rulesets count as satisfied. `Build Desktop Core Module` is already both required and `if:`-gated and TS-only PRs merge fine, which is the precedent.

Until that lands, merging this PR makes detekt non-blocking on PRs. The post-merge run in `merge.yml` still gates the snapshot publish (restored in bddba0d10), so a violation cannot ship to Maven Central — but it can reach `main`.

kaeawc:
## Measured on CI, and one correction

First run of the new job is green — `Detekt` **success**, `Fast Validation` **success**.

**Correction to the baseline in the PR description.** I originally cited a single 8m02s Fast Validation sample. That was sloppy: several of the runs I sampled had *failed* at the fast-validation step, so every later step (BATS, hadolint, iOS version check) was skipped and never counted — which is also where my "everything else totals ~55s" figure came from. Pulling only runs where Fast Validation genuinely **succeeded**:

| | Fast Validation |
|---|---|
| Before (5 successful runs, detekt inline) | 7m54s, 8m41s, 9m17s, 9m44s, 10m08s |
| After | **2m15s** |

So the improvement to the fast-feedback lane is larger than I claimed (~9m median → 2m15s), but the ~55s number was wrong and I withdraw it.

**The caching win is not yet demonstrated.** This run's `Detekt` job took **7m39s** — a cold cache, since it is the first execution of a new job with its own `GRADLE_USER_HOME` key and there was nothing to restore. That is not faster than the 7m04s inline step; it just no longer blocks the fast lane. Locally a warm run is 30s, and the next push to this branch should be the first to restore the seeded cache. I would rather point at that number once it exists than keep asserting it — treat the speedup claim as **unverified on CI** until then.

What is verified today: detekt is off the critical path for every non-Kotlin PR, and Fast Validation returns in a quarter of the time.

PR #4035, opened 2026-07-20
feat(a11y): surface screen-reader navigation fidelity assertions

### LINKED ISSUE #3963: feat(a11y): fidelity assertions — surface reachability / traversal-order / focus-trap results

kaeawc:
## Summary

Expose Android screen-reader-navigation fidelity assertions on `tapOn`. The opt-in cursor journey now reports target reachability, the focused-node traversal order, and whether the existing non-convergence guard detected a focus trap. The generated tool definitions advertise the result to schema consumers.

Cursor observations occur after every swipe, while the existing recalculation and trap-guard cadence remains unchanged so delayed TalkBack focus updates do not become false traps.

## Linked Issue

Closes #3963

## Validation

- [x] `bun run turbo:validate` passes (lint, build, full test suite)
- [x] New code uses fakes and FakeTimer in focused tests

## Device Verification

- [ ] Not run: the change is covered by fast unit tests; no Android device session was used.

## Follow-ups

- [x] No follow-up issues needed.

### DIFF
--- a/schemas/tool-definitions.json
+++ b/schemas/tool-definitions.json
@@ -7773,6 +7773,231 @@
           ],
           "additionalProperties": {}
         },
+        "screenReaderNavigation": {
+          "type": "object",
+          "properties": {
+            "reachable": {
+              "type": "boolean",
+              "description": "Whether swipe cursor navigation reached the target"
+            },
+            "traversalOrder": {
+              "type": "array",
+              "items": {
+                "type": "object",
+                "properties": {
+                  "bounds": {
+                    "type": "object",
+                    "properties": {
+                      "left": {
+                        "type": "number"
+                      },
+                      "top": {
+                        "type": "number"
+                      },
+                      "right": {
+                        "type": "number"
+                      },
+                      "bottom": {
+                        "type": "number"
+                      },
+                      "centerX": {
+                        "type": "number"
+                      },
+                      "centerY": {
+                        "type": "number"
+                      }
+                    },
+                    "required": [
+                      "left",
+                      "top",
+                      "right",
+                      "bottom"
+                    ],
+                    "additionalProperties": false,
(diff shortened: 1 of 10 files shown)


kaeawc:

**Result: PASS (gated-off half)** — `screenReaderNavigation` absent when the flag is off.

With TalkBack active but `screen-reader-navigation` off, the `tapOn` result contains no `screenReaderNavigation` key at all — the spread guard works rather than emitting an empty or null object:

```
keys: [action, element, message, observation, observationDiff, searchUntil, selectedElement, success]
screenReaderNavigation present: false
```

The flag-ON half (asserting `reachable` / `traversalOrder` / `focusTrapDetected` are populated) is **not** verified — see the note on #3964 about not finding a runtime toggle for the flag.

PR #4041, opened 2026-07-20
feat(ios): capture WebRTC Simulator audio

### LINKED ISSUE #3972: feat(ios): WebRTC audio capture — bring iOS to parity with the Android PCMU path
## Context

PR #3884 (merged 2026-07-18) added optional audio to the WebRTC streaming path, but **Android only**. The iOS path currently returns an actionable "unsupported" error for any audio request, so `AUTOMOBILE_WEBRTC_AUDIO=true` / `"audio": true` against an iOS target fails by design.

This issue tracks bringing iOS to parity. It supersedes the "iOS Audio Capture" section of #1097, which was written against the pre-WebRTC IDE-plugin architecture.

## What Android already does (the contract to match)

- `android/video-server/.../AudioCapture.kt` captures device playback audio via `AudioRecord` (`REMOTE_SUBMIX`) as **8 kHz mono PCM16LE**.
- Audio frames are multiplexed into the existing video-server stream via `VideoStreamProtocol.kt` / `VideoStreamWriter.kt`.
- TypeScript side packetizes to **PCMU / G.711** RTP in `src/features/webrtc/RtpPcmuTrackWriter.ts` and publishes via `WebRtcPublisher`.
- Opt-in only: `AUTOMOBILE_WEBRTC_AUDIO=true` env or the stream-socket request field `"audio": true`. Video-only remains the default.

## Scope

- Capture device audio on iOS and emit it in the same 8 kHz mono PCM16LE shape so it reuses the existing `RtpPcmuTrackWriter` packetization unchanged.
- Multiplex over the same transport the iOS video path already uses — do not invent a second socket.
- Replace the current unsupported-error branch with real support; keep the error for genuinely unavailable cases (e.g. a simulator or OS version that cannot provide playback capture).
- Keep audio strictly opt-in and off by default, matching Android.

## Open questions

- Simulator vs. physical device: which can actually provide playback audio capture, and what is the fallback for the other?
- Whether iOS playback capture requires an entitlement or user permission that affects the runner.

## Acceptance

- [ ] With `AUTOMOBILE_WEBRTC_AUDIO=true` against an iOS target, audio is captured and published as PCMU alongside H.264 video.
- [ ] Frame shape matches Android (8 kHz mono PCM16LE) — `RtpPcmuTrackWriter` needs no iOS-specific branch.
- [ ] Video-only remains the default; audio never enabled implicitly.
- [ ] Where capture is genuinely unavailable, an `ActionableError` explains why (not a silent fallback).
- [ ] Unit tests use interfaces + fakes and stay fast; no real-device dependency in the unit suite.

kaeawc:
## Summary

Adds opt-in WebRTC audio for iOS Simulator targets. The existing ScreenCaptureKit helper now captures and multiplexes 8 kHz mono PCM16LE audio over its current stdout transport; TypeScript routes it unchanged to the existing PCMU writer. Physical iOS playback capture remains an actionable unsupported case because public iOS APIs do not expose other-app playback audio.

Audio startup waits for both video and the first PCM chunk, so a silent capture cannot look like a successful audio-enabled stream.

## Linked Issue

Closes #3972

## Validation

- [x] `bun run turbo:validate` passes (`bun run lint` + build + `bun test`)
- [x] `cd ios/screen-capture && swift test`
- [x] New code uses interfaces + fakes + FakeTimer; focused unit tests run in <100ms
- [x] New generic code reuses existing transports and the shared PCMU writer; no dependencies added

## Device Verification

- [ ] Verified on an iOS Simulator via `observe`

Not manually verified against a Simulator window with audible playback yet; this requires host Screen Recording permission and a running Simulator.

## Follow-ups

- [x] Follow-up issues filed for gaps not addressed here: none

### DIFF
--- a/docs/design-docs/mcp/observe/ios-webrtc-streaming.md
+++ b/docs/design-docs/mcp/observe/ios-webrtc-streaming.md
@@ -51,6 +51,13 @@ continuous BGRA frames for:
 - **Physical devices** via AVFoundation USB capture.
 - **Simulators** via ScreenCaptureKit simulator-window capture.
 
+When WebRTC audio is explicitly enabled, the Simulator helper also captures
+the selected window's audio through ScreenCaptureKit and multiplexes 8 kHz mono
+PCM16LE records over its existing stdout transport. `IosH264Source` forwards
+those chunks to the shared PCMU packetizer unchanged. Physical iOS playback
+audio remains unavailable through public APIs and fails with an actionable
+error rather than silently publishing video-only.
+
 `IosH264Source` wraps that helper, pipes frames into `ffmpeg` with
 `h264_videotoolbox`, and forwards raw H.264 Annex-B stdout chunks into the
 existing `WebRtcPublisher`. The helper binary is resolved from
--- a/docs/design-docs/plat/ios/screen-streaming.md
+++ b/docs/design-docs/plat/ios/screen-streaming.md
@@ -268,6 +268,11 @@ Since iOS provides raw frames (not H.264), use a simpler protocol:
 Followed by `height * bytesPerRow` bytes of BGRA pixel data.
 ```
 
+For explicitly enabled Simulator audio, the same transport reserves `width =
+0`; `height = 8000`, `bytesPerRow = 1`, and `timestamp = payload length`.
+The following payload is 8 kHz mono PCM16LE. Video-only capture never emits
+these records.
+
 ## Implementation Plan
 
 ### Milestone 1: Physical Device Capture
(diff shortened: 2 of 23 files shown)


kaeawc:
Heads-up: this branch's red `XCTestRunner Simulator Tests` and `Swift Packages` are **not** caused by anything in this PR — they are two regressions I landed on main and have since fixed. This branch predates both fixes.

- `Swift Packages` / `RemindersPlanContentTests`: an exact-equality assertion on a workflow `if:` condition broke when #4088 added a conjunct. Fixed in #4100 (`ee91d47fe`).
- `XCTestRunner` boot steps: the #4086 boot gate keyed on `Status=4294967295`, which every healthy boot prints, so it rejected all boots. Fixed in #4092 (`13417de40`).

A rebase onto current main should clear both. Sorry for the noise.

PR #4087, opened 2026-07-20
fix(ios,video): bounded retry + diagnostics on recordVideo start (#4076)

### LINKED ISSUE #4076: flaky(ios): simctl recordVideo never emits "Recording started" — video start times out with empty stderr
## Summary

`iosVideoRecordingStartStop.integration.test.ts` → "finalizes a non-empty playable MP4 from a real simulator recording" ([test/integration/iosVideoRecordingStartStop.integration.test.ts:108](https://github.com/kaeawc/auto-mobile/blob/main/test/integration/iosVideoRecordingStartStop.integration.test.ts#L108)) failed once on `XCTestRunner Simulator Tests` because `xcrun simctl io <UDID> recordVideo` never emitted its `Recording started` handshake on stderr. The wait timed out, the process was SIGINT'd, and the test reported a start failure — not an assertion mismatch.

Single occurrence so far, so this is a watch-item rather than a confirmed chronic flake. Filing because nothing currently tracks this symptom and the failure mode is silent (empty stderr, `exitCode: null`).

## Failure signature

Job: [88418928528](https://github.com/kaeawc/auto-mobile/actions/runs/29761850409/job/88418928528) (run 29761850409), failed after ~333s.

```
Failed to start iOS recording: Error: Timed out waiting for Recording started
command: xcrun simctl io <UDID> recordVideo .../643fefe2-...-raw.mov
exitCode: null   signal: SIGINT   stderr: (empty)
```

## Why this reads as environmental, not a code defect

- The PR it failed on (#4044) touches swipe / VoiceOver / scroll paths — zero overlap with the video-recording code path.
- The **same job passed on the very next run** ([29768026024](https://github.com/kaeawc/auto-mobile/actions/runs/29768026024)) against a strict superset of the same branch.
- `stderr` was completely empty — `simctl` produced no output at all, not a partial or differently-worded handshake. That points at a simulator-side hang during recording bring-up.

## Where the handshake lives

[`src/features/video/FfmpegVideoProcessingBackend.ts:57`](https://github.com/kaeawc/auto-mobile/blob/main/src/features/video/FfmpegVideoProcessingBackend.ts#L57) defines `IOS_RECORDING_START_MESSAGES = ["Recording started"]`, and the wait rejects at [line 276](https://github.com/kaeawc/auto-mobile/blob/main/src/features/video/FfmpegVideoProcessingBackend.ts#L276) with `Timed out waiting for ${expectedDescription}`. The waiter already handles `exit` and `error`; the gap is that a silent hang is indistinguishable from a slow boot.

This is adjacent to the known simctl `moov`-atom / SIGINT fragility, but that concerns *flush on stop*. This is a failure to *start*, which is a distinct symptom.

## Suggested directions

1. **Diagnose before failing.** On timeout, capture `xcrun simctl list devices` state and any `simctl` diagnostics for the UDID into the failure message, so the next occurrence tells us whether the simulator was wedged or merely slow.
2. **Bounded retry on start.** A single retry of `recordVideo` bring-up would absorb a one-off simulator stall without masking a real regression, since a genuine break fails deterministically on both attempts.
3. **Confirm the timeout budget** is appropriate for a cold simulator on the CI runner — the job burned 333s before giving up, so the wait itself may not be the long pole.

## Asks

- [ ] Add simulator-state capture to the start-timeout path so a recurrence is diagnosable.
- [ ] Decide whether a bounded start retry is warranted.
- [ ] Watch for recurrence; if it repeats, escalate from watch-item to a real fix.

## Priority
Low-Medium — one occurrence, on a job that is not broadly wired to TypeScript PRs. Revisit if it recurs.

kaeawc:
Closes #4076.

## Problem

The `Run videoRecording MP4 integration test` step (`XCTestRunner Simulator Tests`) is the single largest failure bucket for that job — **7 failures in 5 days** — all with the same signature:

```
Failed to start iOS recording: Error: Timed out waiting for Recording started
command: xcrun simctl io <UDID> recordVideo …-raw.mov
exitCode: null   signal: SIGINT   stderr: (empty)
```

A cold or loaded simulator silently misses the first `recordVideo` start handshake: `simctl` emits neither `Recording started` nor an error, so the single attempt times out. The bare timeout was also undiagnosable — empty stderr, no simulator state.

## Change (`startIos`)

- **Bounded retry**: retry the start handshake up to `IOS_RECORDING_START_MAX_ATTEMPTS` (2), tearing down each timed-out attempt's process (SIGINT via `waitForExit`) before the next. A one-off cold-boot miss is absorbed; a genuine break still fails deterministically on every attempt.
- **Diagnostics on miss**: capture a best-effort `simctl list devices <udid>` snapshot on each timeout and fold it into the failure message, so the next occurrence shows whether the simulator was wedged or merely slow. The snapshot never throws — it can't mask the original error.
- **Fail fast on spawn errors**: a spawn failure (vs a handshake miss) is an environment problem; retrying is pointless, so it throws immediately.
- Start timeout + attempt count are injectable so the retry path is unit-tested without real waits.

## Tests

Two new `FfmpegVideoProcessingBackend` tests — retry-then-succeed and exhaust-with-diagnostics — both verified to **fail when attempts are forced back to 1**. Full suite: 35 pass, stable across 3 runs.

## Validation caveat

The unit tests prove the retry/diagnostics logic against fakes. Whether the retry actually eliminates the real-simulator flake can only be confirmed on the CI runners over time — a green PR here validates the mechanics, not the flake's disappearance. Direction 1 of the issue (diagnostics) is now also satisfied, which will make any recurrence far easier to triage.

### DIFF
--- a/src/features/video/FfmpegVideoProcessingBackend.ts
+++ b/src/features/video/FfmpegVideoProcessingBackend.ts
@@ -51,6 +51,13 @@ const IOS_RECORDING_FILE_READY_INITIAL_BACKOFF_MS = 100;
 const IOS_RECORDING_FILE_READY_MAX_BACKOFF_MS = 1000;
 const FFMPEG_POST_PROCESS_TIMEOUT_MS = 60000;
 const IOS_RECORDING_START_TIMEOUT_MS = 30000;
+// A cold or loaded simulator can silently miss the very first `recordVideo`
+// start handshake (#4076): simctl produces no "Recording started" and no error,
+// and the single attempt times out. Retry the start a bounded number of times
+// before failing, capturing simulator state on each miss so a genuine wedge is
+// diagnosable rather than surfacing as a bare timeout.
+const IOS_RECORDING_START_MAX_ATTEMPTS = 2;
+const IOS_RECORDING_DIAGNOSTIC_TIMEOUT_MS = 5000;
 // `simctl` emits this only after processing its first video frame. The earlier
 // "Defaulting to display" diagnostic proves only display selection, which can
 // still produce a zero-byte capture on a cold simulator.
@@ -281,6 +288,9 @@ export async function waitForStderrMessage(
 export class FfmpegVideoProcessingBackend implements VideoCaptureBackend {
   private ffmpegPath: string = "ffmpeg";
   private hwAccelCache: Map<string, HardwareAccelInfo> = new Map();
+  // Overridable in tests so the retry path can be exercised without real waits.
+  private readonly iosRecordingStartTimeoutMs: number = IOS_RECORDING_START_TIMEOUT_MS;
+  private readonly iosRecordingStartMaxAttempts: number = IOS_RECORDING_START_MAX_ATTEMPTS;
 
   constructor(
     private readonly adbFactory: AdbClientFactory = defaultAdbClientFactory,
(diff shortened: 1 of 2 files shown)


kaeawc:
Verified on `origin/main` (92cd3baa5) during a manual-test iteration over `0.0.44..main`, on iPhone 17 Pro / iOS 26.2.

**Result: PASS** — iOS recording start/stop is healthy after the retry + diagnostics change.

```
videoRecording --platform ios --action start   -> recordingId 2e68e2ca..., outputPath ...mp4
videoRecording --platform ios --action stop    -> same file finalized
ffprobe: hevc, duration readable  => moov atom flushed
```

A second recording taken under continuous screen motion produced 4.5 MB / 29.8s, so the capture path is genuinely producing content, not just a valid-looking container.

I did **not** exercise the failure path this PR adds — I could not induce a handshake timeout on demand, so the `2 attempt(s)` retry, the `captureSimulatorDiagnostics` call, and the aggregated `attempt 1/2: ...; attempt 2/2: ...` error string are **unverified**. Only the success path is covered here.

One unreproduced anomaly, recorded for context rather than as a defect: during a longer motion test, `stop` returned `No active video recording found for device` even though the file finalized correctly on disk (4.5 MB, valid duration). A clean start → intervening `observe` → stop reproduced fine, and my daemon is on a socket shared with another checkout's daemon, so I cannot separate a real registry-eviction from multi-worktree daemon churn. Not filing on that basis; noting it in case it matches something you have seen.

Issue #4058, opened 2026-07-20
Centralize pymobiledevice3 execution for physical-device configuration


kaeawc:
## Problem

IosLockdownLocaleClient uses shell discovery and direct pymobiledevice3 commands, separate from physical-device capability handling.

## Plan

Add an argv-first pymobiledevice3 boundary for availability and locale configuration, injected into IosLockdownLocaleClient. Share high-level device/pairing context with the physical-device client but keep this binary's behavior independent.

## Definition of done

- One injected, typed owner is the only production boundary that resolves and executes this tool.
- Consumers pass device/UDID/path/context structurally; dynamic values use argv rather than shell-interpolated strings.
- The owner defines availability, timeout, cancellation, child cleanup, and actionable errors; tests use fakes and remain fast.
- A focused static check blocks new direct production invocations outside the owner (with documented test/diagnostic exceptions only).
- Run targeted tests plus lint and build.

## Tool-specific checks

Cover executable absence, device selection, locale command arguments, timeout/cancellation, and actionable fallback behavior.

PR #4097, opened 2026-07-21
fix(ios): remove pymobiledevice3 locale dependency

kaeawc:
## Summary

- remove the `pymobiledevice3`/Lockdown implementation for physical iOS system configuration
- return one clear unsupported result for all physical iOS system-configuration reads and writes
- preserve simulator localization behavior and remove the now-invalid physical-device documentation and test seams

## Why

AutoMobile must not require an undeclared third-party Python tool for a feature it does not yet own end to end. Physical iOS system configuration is intentionally stubbed until AutoMobile has a native implementation.

Related to #4058.

## Validation

- `bun test test/features/utility/SystemConfigurationAdapter.test.ts test/features/utility/SystemConfigurationManager.test.ts`
- `bun run build`


kaeawc:
Verified on `origin/main` (92cd3baa5) during a manual-test iteration over `0.0.44..main`, on a real Android emulator (Medium_Phone_API_35) and iPhone 17 Pro / iOS 26.2.

**Result: PASS** — locale still applies on the simulator after the lockdown client removal.

```
BEFORE (ground truth):  AppleLocale = en_US
changeLocalization --platform ios --locale fr-FR
  -> success: true, changes {locale: fr-FR}, localeScope "system",
     localeMethod "defaults write AppleLo..."
AFTER  (ground truth):  AppleLocale = fr_FR
                        AppleLanguages = ("fr-FR", fr)
```

Both `AppleLocale` and `AppleLanguages` were read back with `simctl spawn ... defaults read -g`, so this is device state rather than the tool's own report. Restored to `en_US` afterwards.

The physical-device path is unverified — no physical iOS device attached — so the removal of `IosLockdownLocaleClient` is confirmed only as "simulator path intact".

Issue #4082, opened 2026-07-20
ci(ios): !cancelled() on XCTestRunner steps runs the 26.5 block after a failure — spurious double failures + wasted minutes


kaeawc:
## Summary

In the `XCTestRunner Simulator Tests` job, steps 23–31 carry `if: ${{ !cancelled() }}` (8 such steps, verified in [.github/workflows/pull_request.yml](https://github.com/kaeawc/auto-mobile/blob/main/.github/workflows/pull_request.yml) between lines 1409–1720). `!cancelled()` evaluates true after a *failed* (non-cancelled) step, so once an early step fails, the remaining Xcode-26.5-block steps **keep running** and can produce a second, spurious failure that misattributes the root cause.

## Evidence

Job 88240105173: `Check CtrlProxy XcodeGen Project Drift` failed at 21:50:09, then the workflow ran the entire 26.5 block anyway and produced a second failure — `Run Reminders integration tests (Xcode 26.5)` at 21:55:22 — burning ~5 min and reporting a misleading root cause. All 6 XcodeGen-drift failures in the window show this double-failure signature.

## Why it matters

- **Misleading triage.** The last-reported failure (`Run Reminders integration tests`) is not the cause; the drift check is. Anyone reading the job summary top-down is pointed at the wrong step.
- **Wasted minutes.** The 26.5 block continues on a run that's already doomed, adding ~5 min per occurrence on a job that's already near its 45-min ceiling (see the job-timeout issue).

## The intent vs. the effect

`!cancelled()` is presumably meant to ensure cleanup/best-effort steps run even if an earlier step was *cancelled*. But it also runs them after a *failure*, which is rarely the intent for build/test steps (as opposed to teardown). GitHub's `always()`/`!cancelled()`/`success()` distinctions:
- `success()` (default) — skip on any prior failure.
- `!cancelled()` — run unless the job was cancelled (i.e. **also runs after failure**).
- `failure()` — run only after a failure.

## Suggested directions

1. **Reserve `!cancelled()` for genuine teardown/diagnostics steps** (artifact capture, sim shutdown). For build/test steps that shouldn't run once the pipeline is broken, drop the condition so the default `success()` short-circuits them.
2. If some steps must run after a failure for cleanup, gate them explicitly (`if: always()` on teardown only) and let the functional steps fail-fast.
3. This composes with the job-timeout issue — fail-fast here also reclaims the ~5 min the doomed 26.5 block currently burns.

## Priority
[...]

PR #4100, opened 2026-07-21
fix(ios,test): unbreak workflow assertion after the #4082 prereq guard

kaeawc:
**Unblocks iOS CI.** Follow-up to a regression I introduced in #4088.

## Problem

#4088 (issue #4082) conjoined a shared-prerequisite guard onto the Xcode 26.5 leg conditions:

```yaml
# before
if: ${{ !cancelled() }}
# after
if: ${{ !cancelled() && steps.build-ctrlproxy.outcome == 'success' }}
```

`RemindersPlanContentTests` asserted that condition with **exact string equality**, so every 26.5 step now fails `testPullRequestWorkflowRunsSecondRemindersLegAfterFirstLegFailureUnlessCancelled`, reddening **Swift Packages** on any PR that triggers the iOS job set.

The semantics the test *names* were never broken: after a 26.2 Reminders failure with a good build the condition is still true, and cancellation still skips. Only the string differs.

### Correction on how it reached main

An earlier draft of this description claimed the Swift job "didn't run on #4088". **That was wrong.** Review established that the `ios` paths-filter already includes `.github/workflows/pull_request.yml`, and `Swift Packages (Xcode 26.5)` **did run and did fail** on #4088 at `23:02:27Z` — the PR auto-merged at `23:02:32Z`, five seconds later.

The actual cause is that the `green-main` ruleset's required contexts include neither `Swift Packages (Xcode 26.5)` nor the `iOS` / `iOS Build` rollups, so automerge did not wait on them. That is a governance gap well beyond this PR and is filed separately.

## Fix

`stepBlockRespectsCancellation` encodes the property instead of string-matching it:

```swift
private func stepBlockRespectsCancellation(_ stepBlock: Substring) -> Bool {
    guard let condition = stepBlockActiveTopLevelValue(stepBlock, key: "if") else { return false }
    return condition.hasPrefix("${{ !cancelled()") && !condition.contains("||")
}
```

- Requires `!cancelled()` to **lead** a `${{ ... }}` expression — restoring the `${{` wrapper the old equality assertion pinned.
- **Rejects disjunctions**: `!cancelled() || X` evaluates to `X` when cancelled, so the step can still run. A bare `contains` was operator-blind and accepted it.
- Extra `&&` conjuncts stay free to come and go, which is the whole point.
- Both matchers now delegate to one `stepBlockActiveTopLevelValue` extraction core, so the step-key indent literal lives in a single place.

### BATS: the guard is now asserted per-step

The previous `grep -c ... >= 8` was a **global count**. A reviewer reproduced the false negative: strip the guard from one 26.5 step, add a guarded step elsewhere, the count returns to 8 and **everything stays green** — the #4082 regression landing fully clean. Replaced with per-step assertions over the eight named steps, plus a companion per-step cancellation assertion so that property is not owned solely by the Swift suite.

## Verification (run locally, not assumed)

- `swift test --filter RemindersPlanContentTests` → **26/26** (was failing).
- `bats test/bats/xctest-shared-prereq-gate.bats` → **4/4**.
- **Non-vacuity, each case reproduced:** dropping `!cancelled()` fails the Swift assertion; injecting `!cancelled() || …` fails it; the two-edit count trick fails the new per-step bats assertion.
- SwiftLint: one pre-existing `contains_over_range_nil_comparison` **warning**, no new violations (independently confirmed by a reviewer against both PR-head and main).

## Known gaps, deliberately not in scope

- The same relaxed helper backs the **nightly** assertion (`:411`), but the bats guard is `pull_request.yml`-only. `nightly.yml` has no `build-ctrlproxy` step, so the conjunct that forced this change cannot appear there; the new predicate still enforces the `${{ !cancelled()` prefix and the no-disjunction rule there.
- `RemindersPlanContentTests.swift:954` pins `run: ./scripts/ci/run-reminders-launch-plan-tests.sh` by exact equality across four call sites — same landmine class, filed separately.

### DIFF
--- a/ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
+++ b/ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
@@ -1009,14 +1009,8 @@ private func assertWorkflowStepRunsWhenNotCancelled(
         file: file,
         line: line
     )
-    // Assert the clause this helper is about -- `!cancelled()`, i.e. the step runs
-    // after an earlier leg FAILS but is skipped on cancellation -- rather than the
-    // whole condition. #4082 conjoined a shared-prerequisite guard
-    // (`&& steps.build-ctrlproxy.outcome == 'success'`) which does not change the
-    // failure-vs-cancellation semantics under test here; that guard has its own
-    // coverage in test/bats/xctest-shared-prereq-gate.bats.
     XCTAssertTrue(
-        stepBlockHasActiveTopLevelValueContaining(stepBlock, key: "if", substring: "!cancelled()"),
+        stepBlockRespectsCancellation(stepBlock),
         "\(stepName) must run after an earlier Reminders leg fails, but still respect cancellation",
         file: file,
         line: line
@@ -1065,30 +1059,38 @@ private func workflowStepBlock(
     return workflow[stepRange.lowerBound ..< stepBlockEnd]
 }
 
-private func stepBlockHasActiveTopLevelValue(_ stepBlock: Substring, key: String, value: String) -> Bool {
-    return stepBlockActiveLines(stepBlock).contains { line in
+/// The value of an active top-level `key:` in a step block, or nil when absent.
+/// Single place that knows the step-key indent, so the matchers below cannot
+/// drift apart.
+private func stepBlockActiveTopLevelValue(_ stepBlock: Substring, key: String) -> String? {
+    let prefix = "\(key): "
+    for line in stepBlockActiveLines(stepBlock) where line.hasPrefix("        \(key):") {
         let trimmed = line.trimmingCharacters(in: .whitespaces)
-        return line.hasPrefix("        \(key):") && trimmed == "\(key): \(value)"
+        guard trimmed.hasPrefix(prefix) else { continue }
+        return String(trimmed.dropFirst(prefix.count))
     }
+    return nil
 }
 
-/// Like `stepBlockHasActiveTopLevelValue`, but matches a *substring* of the value.
+private func stepBlockHasActiveTopLevelValue(_ stepBlock: Substring, key: String, value: String) -> Bool {
+    return stepBlockActiveTopLevelValue(stepBlock, key: key) == value
+}
+
+/// Whether a step's `if:` guarantees cancellation skips it.
 ///
-/// Step conditions legitimately grow extra conjuncts (issue #4082 added a
+/// Asserting the whole condition by equality is too strict -- #4082 conjoined a
 /// shared-prerequisite guard, making the 26.5 leg
-/// `${{ !cancelled() && steps.build-ctrlproxy.outcome == 'success' }}`). An
-/// exact-equality assertion turns any such addition into a false failure even
-/// when the semantics under test are unchanged, so assertions about *one* clause
-/// should match that clause rather than the whole expression.
-private func stepBlockHasActiveTopLevelValueContaining(
-    _ stepBlock: Substring,
-    key: String,
-    substring: String
-) -> Bool {
-    return stepBlockActiveLines(stepBlock).contains { line in
-        let trimmed = line.trimmingCharacters(in: .whitespaces)
-        return line.hasPrefix("        \(key):") && trimmed.contains(substring)
-    }
+/// `${{ !cancelled() && steps.build-ctrlproxy.outcome == 'success' }}`, which does
+/// not change the failure-vs-cancellation semantics. But a bare `contains` is too
+/// loose: it is operator-blind, and `${{ !cancelled() || X }}` evaluates to `X`
+/// when cancelled, so the step can still run -- exactly what callers forbid.
+///
+/// So require `!cancelled()` as the leading term of a `${{ ... }}` expression
+/// (the braces were pinned by the old equality assertion and are kept) and reject
+/// any disjunction. Extra `&&` conjuncts remain free to come and go.
+private func stepBlockRespectsCancellation(_ stepBlock: Substring) -> Bool {
+    guard let condition = stepBlockActiveTopLevelValue(stepBlock, key: "if") else { return false }
+    return condition.hasPrefix("${{ !cancelled()") && !condition.contains("||")
 }
 
 private func stepBlockHasActiveEnvValue(_ stepBlock: Substring, key: String, value: String) -> Bool {
--- a/ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
+++ b/ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
@@ -1009,8 +1009,14 @@ private func assertWorkflowStepRunsWhenNotCancelled(
         file: file,
         line: line
     )
+    // Assert the clause this helper is about -- `!cancelled()`, i.e. the step runs
+    // after an earlier leg FAILS but is skipped on cancellation -- rather than the
+    // whole condition. #4082 conjoined a shared-prerequisite guard
+    // (`&& steps.build-ctrlproxy.outcome == 'success'`) which does not change the
+    // failure-vs-cancellation semantics under test here; that guard has its own
+    // coverage in test/bats/xctest-shared-prereq-gate.bats.
     XCTAssertTrue(
-        stepBlockHasActiveTopLevelValue(stepBlock, key: "if", value: "${{ !cancelled() }}"),
+        stepBlockHasActiveTopLevelValueContaining(stepBlock, key: "if", substring: "!cancelled()"),
         "\(stepName) must run after an earlier Reminders leg fails, but still respect cancellation",
         file: file,
         line: line
@@ -1063,6 +1069,25 @@ private func stepBlockHasActiveTopLevelValue(_ stepBlock: Substring, key: String
     return stepBlockActiveLines(stepBlock).contains { line in
         let trimmed = line.trimmingCharacters(in: .whitespaces)
         return line.hasPrefix("        \(key):") && trimmed == "\(key): \(value)"
+    }
+}
+
+/// Like `stepBlockHasActiveTopLevelValue`, but matches a *substring* of the value.
+///
+/// Step conditions legitimately grow extra conjuncts (issue #4082 added a
+/// shared-prerequisite guard, making the 26.5 leg
+/// `${{ !cancelled() && steps.build-ctrlproxy.outcome == 'success' }}`). An
+/// exact-equality assertion turns any such addition into a false failure even
+/// when the semantics under test are unchanged, so assertions about *one* clause
+/// should match that clause rather than the whole expression.
+private func stepBlockHasActiveTopLevelValueContaining(
+    _ stepBlock: Substring,
+    key: String,
+    substring: String
+) -> Bool {
+    return stepBlockActiveLines(stepBlock).contains { line in
+        let trimmed = line.trimmingCharacters(in: .whitespaces)
+        return line.hasPrefix("        \(key):") && trimmed.contains(substring)
     }
 }
 
--- a/test/bats/xctest-shared-prereq-gate.bats
+++ b/test/bats/xctest-shared-prereq-gate.bats
@@ -32,8 +32,84 @@ WORKFLOW=".github/workflows/pull_request.yml"
   [ -z "$bare" ]
 }
 
-@test "the 26.5 leg gates on steps.build-ctrlproxy.outcome == success" {
-  # The eight 26.5-leg steps that were !cancelled() now carry the shared-build guard.
-  count="$(grep -c "!cancelled() && steps.build-ctrlproxy.outcome == 'success'" "$WORKFLOW")"
-  [ "$count" -ge 8 ]
+# Print the active `if:` value for a named step, or nothing when the step has none.
+#
+# Scanning starts only AFTER the 26.2 leg's final step: several step names occur
+# twice in the job (e.g. "Select Xcode 26.5" is both the job's initial toolchain
+# selection, which has no `if:`, and the 26.5 leg's re-selection, which does).
+# Without the anchor the first, unguarded occurrence matches and the assertion
+# reports a false violation. This mirrors `afterStepName` in the Swift helper.
+#
+# awk, not sed -- BSD sed lacks the range forms this needs. A step block runs from
+# its `- name: "<step>"` line to the next step start or a step-indent comment.
+ANCHOR_STEP='Run Reminders integration tests (Xcode 26.2)'
+
+step_if_condition() {
+  awk -v want="      - name: \"$1\"" -v anchor="      - name: \"$ANCHOR_STEP\"" '
+    !past && $0 == anchor { past = 1; next }
+    !past { next }
+    $0 == want { inblock = 1; next }
+    inblock && /^      [-#]/ { exit }
+    inblock && /^        if:/ {
+      sub(/^        if:[[:space:]]*/, "")
+      print
+      exit
+    }
+  ' "$WORKFLOW"
+}
+
+# Every Xcode 26.5 leg step that must not run on a doomed shared prerequisite.
+# Named explicitly: a global count cannot tell "each of these is guarded" from
+# "the total happens to add up", so one step could lose its guard while an
+# unrelated step gained one and the suite would stay green.
+xcode_265_leg_steps() {
+  cat <<'STEPS'
+Select Xcode 26.5
+Ensure iOS Simulator runtime (Xcode 26.5)
+Boot iOS Simulator (Xcode 26.5)
+Ensure AutoMobile daemon ready (Xcode 26.5)
+Warm up iOS CtrlProxy (Xcode 26.5)
+Pre-build Reminders XCTest bundle (Xcode 26.5)
+Warm up Reminders target app (Xcode 26.5)
+Run Reminders integration tests (Xcode 26.5)
+STEPS
+}
+
+@test "every 26.5-leg step individually gates on steps.build-ctrlproxy.outcome == success" {
+  local offenders=""
+  while IFS= read -r step; do
+    [ -n "$step" ] || continue
+    condition="$(step_if_condition "$step")"
+    if [ -z "$condition" ]; then
+      offenders="${offenders}${step}: no 'if:' condition"$'\n'
+    elif [[ "$condition" != *"steps.build-ctrlproxy.outcome == 'success'"* ]]; then
+      offenders="${offenders}${step}: ${condition}"$'\n'
+    fi
+  done < <(xcode_265_leg_steps)
+
+  if [ -n "$offenders" ]; then
+    echo "26.5-leg steps missing the shared-build guard:" >&2
+    echo "$offenders" >&2
+  fi
+  [ -z "$offenders" ]
+}
+
+@test "every 26.5-leg step still respects cancellation" {
+  # The other half of the condition. RemindersPlanContentTests owns this too (via
+  # stepBlockRespectsCancellation); asserted here so a Swift-only or bats-only run
+  # cannot lose the property silently.
+  local offenders=""
+  while IFS= read -r step; do
+    [ -n "$step" ] || continue
+    condition="$(step_if_condition "$step")"
+    if [[ "$condition" != *"!cancelled()"* ]] || [[ "$condition" == *"||"* ]]; then
+      offenders="${offenders}${step}: ${condition:-<none>}"$'\n'
+    fi
+  done < <(xcode_265_leg_steps)
+
+  if [ -n "$offenders" ]; then
+    echo "26.5-leg steps that do not skip on cancellation:" >&2
+    echo "$offenders" >&2
+  fi
+  [ -z "$offenders" ]
 }


ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
```diff
@@ -1066,6 +1072,25 @@ private func stepBlockHasActiveTopLevelValue(_ stepBlock: Substring, key: String
...
+/// when the semantics under test are unchanged, so assertions about *one* clause
+/// should match that clause rather than the whole expression.
+private func stepBlockHasActiveTopLevelValueContaining(
+    _ stepBlock: Substring,
+    key: String,
+    substring: String
+) -> Bool {
+    return stepBlockActiveLines(stepBlock).contains { line in
+        let trimmed = line.trimmingCharacters(in: .whitespaces)
+        return line.hasPrefix("        \(key):") && trimmed.contains(substring)
```
kaeawc: Substring matching makes the assertion blind to the operator joining the clause, so `if: ${{ !cancelled() || <anything> }}` passes — and that is exactly the condition the message says must not hold, since a disjunct that is true when the run is cancelled makes the step run on cancellation.

Reproduced with the helper logic verbatim against synthetic step blocks:

| step `if:` | matcher |
| --- | --- |
| `${{ !cancelled() && steps.build-ctrlproxy.outcome == 'success' }}` | passes (intended) |
| `${{ !cancelled() \|\| github.event_name == 'push' }}` | passes (should not) |
| `${{ !(!cancelled()) }}` | passes (means the exact inverse) |
| `${{ cancelled() }}` / `${{ always() }}` | fails (correct) |

`test/bats/xctest-shared-prereq-gate.bats` only partly backstops this: its third test counts `!cancelled() && steps.build-ctrlproxy.outcome == 'success'` at `>= 8`, so flipping one step to `||` would drop the count — but the second test greps for the *bare* form only, and neither test constrains a step the Swift helper is asserting about but bats is not counting. Not a blocker given that overlap, but the helper is cheap to make operator-aware.

Suggested change for keeping the clause binding:

```suggestion
    return stepBlockActiveLines(stepBlock).contains { line in
        let trimmed = line.trimmingCharacters(in: .whitespaces)
        guard line.hasPrefix("        \(key):"), trimmed.contains(substring) else { return false }
        // `||` would make the matched clause non-binding: `!cancelled() || X` still
        // runs the step on cancellation, which is what this assertion forbids.
        return !trimmed.contains("||")
    }
```

Nit, no action needed: the same-line requirement means a folded value (`if: >-` with `${{ !cancelled() }}` on the next line) fails the match. That direction is fail-safe, and it matches the pre-existing `stepBlockHasActiveTopLevelValue` at line 1071, so it is only worth knowing about.

Two claims in the doc comment do check out: `test/bats/xctest-shared-prereq-gate.bats` exists and pins both halves of the guard, and the 8-space prefix holds for every step this helper is applied to (`Select Xcode 26.5`, `Ensure iOS Simulator runtime (Xcode 26.5)`, `Boot iOS Simulator (Xcode 26.5)`, `Ensure AutoMobile daemon ready (Xcode 26.5)`). Commented-out `if:` lines and `if:` nested under `with:` are both correctly excluded. `swift test --filter RemindersPlanContentTests` is green — 26 tests, 0 failures, 0.06s.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
```diff
@@ -1009,8 +1009,14 @@ private func assertWorkflowStepRunsWhenNotCancelled(
         file: file,
         line: line
     )
+    // Assert the clause this helper is about -- `!cancelled()`, i.e. the step runs
+    // after an earlier leg FAILS but is skipped on cancellation -- rather than the
+    // whole condition. #4082 conjoined a shared-prerequisite guard
+    // (`&& steps.build-ctrlproxy.outcome == 'success'`) which does not change the
+    // failure-vs-cancellation semantics under test here; that guard has its own
+    // coverage in test/bats/xctest-shared-prereq-gate.bats.
```
kaeawc: "that guard has its own coverage in `test/bats/xctest-shared-prereq-gate.bats`" is true, but the coverage is weaker than this sentence reads, and it is worth knowing before this comment becomes the thing a future editor trusts. That file pins the guard with a *global* line count — `grep -c "!cancelled() && steps.build-ctrlproxy.outcome == 'success'"` then `[ "$count" -ge 8 ]` (xctest-shared-prereq-gate.bats:37-38) — not per-step. Nothing asserts that *these three* named 26.5 steps carry it. So a step can lose the guard as long as some other line gains one, and the count absorbs it.

Reproduced on this head: I removed the guard from `Run Reminders integration tests (Xcode 26.5)` (leaving `if: !cancelled()`) and added one plausible new guarded step (`Collect 26.5 diagnostics`) to the same leg. Count back to 8 — all 3 bats tests pass, all 26 Swift tests pass. That is exactly the #4082 regression (the 26.5 Reminders run firing on a broken CtrlProxy build) landing fully green.

Not a blocker — unbreaking CI is the right call and this hole predates the PR (it came in with #4088). But this PR is what makes the Swift side stop covering the conjunct, so the count is now the only thing standing there. Worth a follow-up that asserts per-step, e.g. reusing the existing block-extraction the Swift tests already do, or in bats keying off the step name:

```bash
for step in "Select Xcode 26.5" "Boot iOS Simulator (Xcode 26.5)" "Run Reminders integration tests (Xcode 26.5)"; do
  run grep -A1 "name: \"$step\"" "$WORKFLOW"
  [[ "$output" == *"steps.build-ctrlproxy.outcome == 'success'"* ]]
done
```

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
```diff
@@ -1066,6 +1072,25 @@ private func stepBlockHasActiveTopLevelValue(_ stepBlock: Substring, key: String
...
+private func stepBlockHasActiveTopLevelValueContaining(
+    _ stepBlock: Substring,
+    key: String,
+    substring: String
+) -> Bool {
+    return stepBlockActiveLines(stepBlock).contains { line in
+        let trimmed = line.trimmingCharacters(in: .whitespaces)
+        return line.hasPrefix("        \(key):") && trimmed.contains(substring)
+    }
+}
```
kaeawc: `stepBlockHasActiveTopLevelValueContaining` is `stepBlockHasActiveTopLevelValue` (line 1068) with `trimmed == ...` swapped for `trimmed.contains(...)` — same `stepBlockActiveLines` walk, same hardcoded `"        \(key):"` indent prefix, which this file now spells out in three places. CLAUDE.md's "one canonical primitive per concern" argues for factoring the walk once and letting the two matchers differ only in their predicate, so a future change to how a top-level key line is recognized (indent, comment stripping, quoting) has one place to land instead of two.

Suggested change for a single canonical top-level-line matcher:

```suggestion
private func stepBlockHasActiveTopLevelLine(
    _ stepBlock: Substring,
    key: String,
    where matches: (String) -> Bool
) -> Bool {
    return stepBlockActiveLines(stepBlock).contains { line in
        line.hasPrefix("        \(key):")
            && matches(line.trimmingCharacters(in: .whitespaces))
    }
}

/// Matches one clause of a value rather than the whole thing. Step conditions
/// legitimately grow extra conjuncts, so an assertion about a single clause
/// should match that clause instead of the full expression.
private func stepBlockHasActiveTopLevelValueContaining(
    _ stepBlock: Substring,
    key: String,
    substring: String
) -> Bool {
    return stepBlockHasActiveTopLevelLine(stepBlock, key: key) { $0.contains(substring) }
}
```

The suggestion can't reach line 1068, so that one needs the matching edit by hand:

```swift
private func stepBlockHasActiveTopLevelValue(_ stepBlock: Substring, key: String, value: String) -> Bool {
    return stepBlockHasActiveTopLevelLine(stepBlock, key: key) { $0 == "\(key): \(value)" }
}
```

Worth considering as a follow-on: a `containing:` label overload of `stepBlockHasActiveTopLevelValue` rather than a `...Containing` suffix would keep one name for one concern and read closer to the sibling `stepBlockHasActiveEnvValue`. I left it out because it also needs the call site at line 1019 updated. Not a blocker either way.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
```diff
@@ -1009,8 +1009,14 @@ private func assertWorkflowStepRunsWhenNotCancelled(
...
     )
+    // Assert the clause this helper is about -- `!cancelled()`, i.e. the step runs
+    // after an earlier leg FAILS but is skipped on cancellation -- rather than the
+    // whole condition. #4082 conjoined a shared-prerequisite guard
+    // (`&& steps.build-ctrlproxy.outcome == 'success'`) which does not change the
+    // failure-vs-cancellation semantics under test here; that guard has its own
+    // coverage in test/bats/xctest-shared-prereq-gate.bats.
     XCTAssertTrue(
-        stepBlockHasActiveTopLevelValue(stepBlock, key: "if", value: "${{ !cancelled() }}"),
+        stepBlockHasActiveTopLevelValueContaining(stepBlock, key: "if", substring: "!cancelled()"),
```
kaeawc: One property does get dropped outright here, separate from the guard conjunct: the `${{ ... }}` expression wrapper. The old exact-equality pinned `if: ${{ !cancelled() }}`; the substring only pinned `!cancelled()`, so an unbraced `if: !cancelled()` — valid Actions syntax — now passes this assertion. The bats file cannot pick it up either, because its bare-condition test greps the literal braced string `if: ${{ !cancelled() }}` (xctest-shared-prereq-gate.bats:27).

Reproduced: with `Run Reminders integration tests (Xcode 26.5)` set to `if: !cancelled()` and nothing else touched, bats tests 1 and 2 both pass; only the `-ge 8` count fails. So for this one form the count is the sole guard, in both files.

Cheap to close without giving up the reason you loosened this — match the wrapper, not just the clause:

Suggested change for keeping the expression form pinned while still tolerating extra conjuncts:
```suggestion
        stepBlockHasActiveTopLevelValueContaining(stepBlock, key: "if", substring: "${{ !cancelled()"),
```

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
kaeawc: Nit: this comment and the new helper's doc comment (1075-1081) tell the same #4082 story — the conjoined `steps.build-ctrlproxy.outcome == 'success'` guard, the semantics being unchanged, the bats coverage — so the rationale has to be kept true in two places, and the helper is the copy that goes stale first (it stops being about #4082 the moment a second caller shows up). This call site is where the history belongs; the helper only needs its generic contract. It's also the only `///` doc among the private helpers down there, so a one-liner would sit better with the surrounding density. Not a blocker.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
```diff
@@ -1066,6 +1072,25 @@ private func stepBlockHasActiveTopLevelValue(_ stepBlock: Substring, key: String
...
 
+/// Like `stepBlockHasActiveTopLevelValue`, but matches a *substring* of the value.
+///
+/// Step conditions legitimately grow extra conjuncts (issue #4082 added a
+/// shared-prerequisite guard, making the 26.5 leg
+/// `${{ !cancelled() && steps.build-ctrlproxy.outcome == 'success' }}`). An
+/// exact-equality assertion turns any such addition into a false failure even
+/// when the semantics under test are unchanged, so assertions about *one* clause
+/// should match that clause rather than the whole expression.
+private func stepBlockHasActiveTopLevelValueContaining(
```
kaeawc: The non-vacuity claim in the description holds — I checked it independently rather than taking it. Setting the 26.5 Reminders step to `if: ${{ steps.build-ctrlproxy.outcome == 'success' }}` (guard kept, `!cancelled()` dropped) fails at RemindersPlanContentTests.swift:320 with the right message, 26 tests / 1 failure. The split you are drawing is the right one, and the doc comment on this helper is genuinely the most useful thing in the diff.

The one gap is that the pointer is one-directional. This side says "the build guard lives in bats"; `xctest-shared-prereq-gate.bats` says nothing about Swift owning `!cancelled()`. Its header only describes #4082. Someone editing that file — say relaxing the `-ge 8` count because they added or removed a 26.5 step — has no way to see that a second test in another language depends on the same lines, which is the same shape of miss the description owns up to for #4088.

Suggested change for making the ownership split discoverable from the bats side too, as a line in that file's header:

```
# Split of responsibility: this file owns the `steps.build-ctrlproxy.outcome ==
# 'success'` conjunct. The `!cancelled()` conjunct on the same lines is owned by
# `testPullRequestWorkflowRunsSecondRemindersLegAfterFirstLegFailureUnlessCancelled`
# in ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift.
# Changing a 26.5-leg `if:` will likely need edits in both.
```

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
kaeawc: The comment says the conjoined guard "has its own coverage in test/bats/xctest-shared-prereq-gate.bats", but that suite hardcodes `WORKFLOW=".github/workflows/pull_request.yml"` (test/bats/xctest-shared-prereq-gate.bats:14) — and `assertWorkflowStepRunsWhenNotCancelled` is also driven against nightly.yml, via `assertWorkflowSecondRemindersBringUpRunsWhenNotCancelled` at RemindersPlanContentTests.swift:411 under `testNightlyWorkflow...` (loaded at :409). So on the nightly side this relaxes an exact-equality assertion to a substring match with nothing compensating: nightly.yml:552-580 carries bare `${{ !cancelled() }}` on all five 26.5-leg steps, and after this change any conjunct silently bolted onto those — including one that disables the leg outright — passes.

Worth either scoping the bats guard over both workflows (it already greps by filename, so a `for WORKFLOW in ...` loop covers it), or keeping the exact assertion for the nightly call site since nightly has no `build-ctrlproxy` step to gate on in the first place — it uses `ensure-ctrl-proxy-ready.sh` (nightly.yml:525), so the conjunct that forced this change does not and will not appear there.

Not a blocker for unbreaking main, but it should not ship without the nightly side being closed.

kaeawc:
The premise in the PR body — that #4088 slipped because the Swift job did not run — does not hold up. I checked the trigger wiring and the actual run.

The `ios` paths-filter already lists `.github/workflows/pull_request.yml` (pull_request.yml:267), added by #3867 for exactly this reason, and `scripts/ci/validate_workflow_assertion_triggers.sh` passes on main enforcing it. `Swift Packages (Xcode 26.5)` did run on #4088 and did fail:

```
gh api "repos/kaeawc/auto-mobile/actions/runs/29785682136/jobs?per_page=100" \
  -q '.jobs[] | select(.name|test("Swift Packages|^iOS")) | "\(.name)\t\(.conclusion)\t\(.completed_at)"'
Swift Packages (Xcode 26.5)  failure  2026-07-20T23:02:27Z
iOS Build                    failure  2026-07-20T23:02:34Z
iOS                          failure  2026-07-20T23:02:38Z
```

#4088 merged at `2026-07-20T23:02:32Z` — five seconds after the failure, and *before* the two rollup gates that would have reported it had even finished. The reason is that the `green-main` ruleset's 25 required contexts do not include `Swift Packages (Xcode 26.5)`, `iOS`, or `iOS Build`; the only iOS-family entries are `SwiftLint` and `Swift Code Coverage`:

```
gh api repos/kaeawc/auto-mobile/rulesets/11406121 \
  -q '[.rules[]|select(.type=="required_status_checks").parameters.required_status_checks[].context]|map(select(test("iOS|Swift|Xcode")))'
["SwiftLint","Swift Code Coverage"]
```

So the whole `ios-gate` / `ios-build-gate` design (pull_request.yml:2472 and :2563, whose own comment at :2468 says the required deterministic subset lives in `ios-build-gate`) is wired in YAML but never registered as a required check. Adding more paths to the iOS filter would not have caught #4088; registering `iOS` and `iOS Build` as required contexts would have. That is the fix worth chasing after this one lands — otherwise the next red `Swift Packages` merges the same way.

---

Same landmine class, still latent and not touched here: `RemindersPlanContentTests.swift:954` asserts `stepBlockHasActiveTopLevelValue(runBlock, key: "run", value: "./scripts/ci/run-reminders-launch-plan-tests.sh")` — exact equality on a `run:` value across four call sites (pull_request.yml:1571, :1630, nightly.yml:567, :602). The moment that invocation gains an argument or a `--flag`, it breaks identically to the `if:` assertion this PR is fixing, and again only via a check that does not gate merges.

---

State of the rest of the blast radius, all run on `origin/main` (74b13674a):

- `RemindersPlanContentTests` — **RED on main**, 7 failures, all from the exact-equality `if:` assertion. This PR takes it to 26/26 green. That is the only workflow-assertion breakage I found.
- `RemindersAddPlanTests.testAddReminderPlan` also fails locally, but on a live-simulator tapOn timeout — environment, not #4088.
- Green on main: `test/bats/{pr-gate-wiring,xctest-shared-prereq-gate,release-workflow-wiring,installer-workflow-status,boot-simulator}.bats` (39/39), `test/scripts/xcodegenDriftCheck.test.ts` (9/9), `scripts/ci/validate_workflow_assertion_triggers.sh`.

Nothing else in the repo asserts on `!cancelled()` — the string appears in exactly one test file.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
kaeawc: Fixed in adee5c50. You are right that the matcher was operator-blind — `!cancelled() || X` evaluates to `X` when cancelled, so the step runs, which is precisely what the assertion message forbids.

Rather than adding a `||` guard onto the string matcher, I replaced it with a predicate that encodes the property: `stepBlockRespectsCancellation` requires `${{ !cancelled()` to LEAD the expression and rejects any disjunction. Verified by injecting `${{ !cancelled() || github.event_name == 'push' }}` — the assertion now fails.

Left the folded-`>-` case alone: it is pre-existing on the equality matcher too and fails safe.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
kaeawc: Good catch, and fixed in adee5c50. The `${{` wrapper was silently dropped. The new predicate uses `hasPrefix("\${{ !cancelled()")`, which restores the wrapper and additionally pins `!cancelled()` as the leading term — slightly stronger than the suggested `contains("\${{ !cancelled()")`, and it composes with the no-disjunction rule.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
kaeawc: This was the most valuable finding — thank you for reproducing it rather than just flagging it. The two-edit false negative is real: my `grep -c ... >= 8` never asserted per-step, so one step could lose the guard while another gained one.

Fixed in adee5c50: the count is replaced with per-step assertions over the eight named 26.5-leg steps, plus a companion per-step cancellation assertion so that property is not owned solely by the Swift suite. I reproduced your exact scenario (strip guard from `Run Reminders integration tests (Xcode 26.5)`, add a decoy guarded step) — global count returns to 8, and the new per-step assertion fails. Not deferring this to a follow-up, since after this PR the count was the only thing guarding it.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
kaeawc: Addressed in adee5c50 by making it true in both directions rather than only fixing the wording: the bats suite now carries its own per-step cancellation assertion alongside the build-guard one, so neither property depends on a pointer to the other layer.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
kaeawc: Agreed, and fixed in adee5c50 — though I went further than delegating both matchers to a shared line-walk. Since the substring matcher existed only to express one semantic property, I removed it entirely: there is now a single `stepBlockActiveTopLevelValue(_:key:)` extraction core, `stepBlockHasActiveTopLevelValue` is a one-line equality on top of it, and the cancellation check is a domain predicate rather than a third string matcher. The `"        \(key):"` indent literal is down to one occurrence.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
kaeawc: Fixed in adee5c50 — the call-site comment is gone and the rationale lives once, on the predicate, which is where the WHY belongs now that it is a named property rather than a generic matcher.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
kaeawc: Thanks for checking it independently rather than taking the description's word for it — that is exactly the failure mode that produced this PR in the first place.

ios/XCTestRunner/Sources/XCTestRunnerTests/RemindersPlanContentTests.swift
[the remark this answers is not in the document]
kaeawc: Addressed in adee5c50; see the per-thread replies for specifics.

kaeawc:
All four review perspectives addressed in `adee5c50`, threads resolved.

**Changed as a result of review:**
- `stepBlockRespectsCancellation` replaces the substring matcher — requires `${{ !cancelled()` to lead and rejects disjunctions (closes the operator-blindness gap and restores the `${{` wrapper).
- Both matchers now delegate to one `stepBlockActiveTopLevelValue` extraction core; the indent literal is down to one occurrence and there is a single doc comment.
- **BATS asserts per-step instead of a global count.** This was the finding that mattered most — a reviewer reproduced a false negative where stripping the guard from one step and adding a guarded step elsewhere kept the count at 8 and everything stayed green. Fixed here rather than deferred, because after this PR the count was the only thing guarding the #4082 property.
- Added a companion per-step cancellation assertion so neither property is owned solely by one layer.

**Verified locally, each case reproduced:** swift 26/26, bats 4/4; dropping `!cancelled()`, injecting `|| X`, and the two-edit count trick each fail without these changes.

**Description corrected.** I had claimed the Swift job didn't run on #4088. It did — it ran, failed at 23:02:27Z, and the PR auto-merged 5s later because `Swift Packages` and the `iOS` rollups are not required contexts. Filed as #4101 (high priority — automerge can land failing iOS jobs).

Also filed #4102 for the remaining exact-equality assertion at `:954`, same landmine class.

PR #4121, opened 2026-07-21
feat(device): gate simulator/emulator auto-creation behind a flag + env var (#4093)

### LINKED ISSUE #4093: feat(device): gate simulator/emulator auto-creation behind a CLI flag + env var (default off)
## Summary

AutoMobile's product boot cannot provision a simulator/emulator — there is no `simctl create` anywhere in `src/`, and `AvdManagerClient` deliberately tells the user to run `avdmanager create` rather than doing it. `scripts/ios/boot-simulator.sh` **can** create one (`simctl create "CI iPhone" <devicetype> <runtime>`), which is one reason CI cannot simply call the product boot today.

Auto-creating devices should **not** become default product behavior — spawning simulators on a developer's machine is a side effect they did not ask for. Instead, gate it behind an explicit opt-in that CI (and only CI, by choice) turns on.

## Proposed

Add a provisioning opt-in, expressed two equivalent ways so both MCP/CLI and containerized CI can set it:

1. **CLI flag** on the device-start path, e.g. `--create-if-missing` (usable as `auto-mobile --cli startDevice --platform ios --create-if-missing`).
2. **Env var**, e.g. `AUTOMOBILE_ALLOW_DEVICE_CREATE=1`, for environments that cannot easily thread flags.

Precedence: explicit flag wins over env var; **default is off** on both platforms.

When enabled and no device matches the requested criteria:
- **iOS**: resolve the runtime (see the alignment issue), pick a device type, `simctl create`, then boot.
- **Android**: create the AVD via `AvdManagerClient` rather than emitting the current "run avdmanager yourself" guidance.

## Constraints

- All new simctl execution must go through `SimCtlClient` — the simctl boundary ratchet (`scripts/check-no-new-direct-simctl.sh`, #4049/#4074) enforces this.
- Creation must be clearly logged (name, device type, runtime) so a user who enabled it can see exactly what was added.
- Consider a recognizable name prefix so created devices are identifiable/cleanable.

## Acceptance

- [ ] Default behavior unchanged: no device is ever created without the flag/env var.
- [ ] Flag and env var both work; flag takes precedence; unit tests for the precedence matrix.
- [ ] iOS creation path lives in `SimCtlClient` and passes the simctl boundary ratchet.
- [ ] Android creation path goes through `AvdManagerClient`.
- [ ] Creation is logged with the resolved runtime/device type.

## Priority
Medium — prerequisite for ever letting CI use the product boot, and useful on its own for fresh machines.

kaeawc:
Closes #4093.

## Scope correction

The issue as filed said "there is no `simctl create` anywhere in `src/`". **That was wrong** — my error, from a grep that failed with a shell glob error and returned empty output I read as absence. A [correcting comment](https://github.com/kaeawc/auto-mobile/issues/4093) records the accurate state. Both primitives already existed:

- `SimCtlClient.createSimulator` (`SimCtlClient.ts:1116`)
- `AvdManagerClient.createAvd` (`AvdManagerClient.ts:109`)

So this PR is **gating and wiring only**. What was actually missing: no gate existed anywhere, and `createSimulator` had no callers outside the client — `findOrStartIosDevice()` threw *"No iOS simulators are available…"* rather than creating one.

## Gate (default OFF)

`DeviceCreationGate` (`src/utils/deviceCreationGate.ts`) reads through an injected `EnvironmentReader`, so no call site touches `process.env`:

- CLI flag `--create-if-missing` — the parser already maps kebab→camel and treats a bare flag as `true`, so no parser change was needed.
- Env var `AUTOMOBILE_ALLOW_DEVICE_CREATE` (`1`/`true`, case-insensitive, trimmed).
- **The flag wins in both directions** — `--create-if-missing false` beats `AUTOMOBILE_ALLOW_DEVICE_CREATE=1`.
- When off, both platforms' error messages are byte-identical to today.

`describeSource()` feeds the log line so a user can see *which* input enabled it. Created devices carry an `AutoMobile-<model>-<id>` prefix (id from the injected `IdGenerator`, not a bare `randomUUID()`), with `isCreatedDeviceName()` for cleanup.

## Android: Play Store images ranked **last** — a real defect caught in review

The initial tag preference was `google_apis_playstore > google_apis > default`. Play Store images **refuse `adb root`**, and AutoMobile needs a root shell for the root-backed system-locale path — `AndroidSystemConfigurationAdapter` fails with *"the target emulator is not root-capable or does not allow root ADB"* on the API levels that require it. Auto-provisioning one would have handed users a device that silently cannot run `changeLocalization`.

Now `google_apis > default > google_apis_playstore`, with two tests pinning it, **both verified to fail under the previous ranking**.

## Two support changes worth review

- **`listInstalledSystemImages()`** — the existing `listSystemImages()` parses only the *"Available Packages"* (downloadable) section, and `avdmanager create avd -k` fails on a package that isn't on disk. `parseSystemImages` is now section-parameterized; a test pins that `listSystemImages` still returns only the available section.
- **`createSimulator` now invalidates the static `deviceListCache`** — without it, a create-then-list would boot off a snapshot predating the device just created.

## Deliberately out of scope

- The `deviceId` direct-lookup branch does **not** provision — asking for a specific UDID and silently getting a different, newly-created device is wrong; it still throws "not found".
- No auto-install of system images/runtimes; you get an actionable error naming the `sdkmanager` command. Downloading multi-GB images is well outside "create if missing".
- No cleanup/GC of created devices. The name prefix and `isCreatedDeviceName()` make it possible; nothing sweeps them. Worth a follow-up if CI adopts this.

## Verification

91 pass / 0 fail across all changed test files (**0 unhandled errors** — checked explicitly, see below); full suite 8198 pass; lint clean; both the simctl and emulator boundary ratchets pass; typecheck shows no new errors from these files.

Non-vacuity, each with an asserted mutation anchor and `trap`-based restore: gate forced true → default-off tests fail; provision unconditionally → default-off fails; never provision → gate-enables fails (both `startDevice` and `DeviceSessionManager`); playstore-first ordering → both tag tests fail.

One process note: an earlier revision of the tag tests used `test(...)` in a file importing `it`, which Bun reported as `16 pass / 0 fail / **1 error**` — a green-looking summary while those tests never ran. Verification now asserts the error count and exit code, not just pass/fail.

### DIFF
--- a/docs/using/environment-variables.md
+++ b/docs/using/environment-variables.md
@@ -73,6 +73,31 @@ here for discoverability; most are diagnostic or for advanced testing.
 | `AUTOMOBILE_SKIP_CTRL_PROXY_DOWNLOAD` | Skips Android and iOS CtrlProxy downloads/prefetches when set to `1` or `true`. |
 | `AUTOMOBILE_SKIP_ACCESSIBILITY_DOWNLOAD_IF_INSTALLED` | Skips the accessibility service download when it is already installed. |
 
+## Device provisioning opt-in
+
+AutoMobile never creates a simulator or emulator by default — spawning devices on
+a developer's machine is a side effect they did not ask for. Turn it on
+explicitly, per run or per environment.
+
+| Variable | Purpose | Default |
+|----------|---------|---------|
+| `AUTOMOBILE_ALLOW_DEVICE_CREATE` | When `1` or `true`, `startDevice` creates a device (iOS: `simctl create`; Android: `avdmanager create avd`) instead of failing when nothing matches the requested criteria. | unset (off) |
+
+The equivalent per-call flag is `--create-if-missing` on the device-start path:
+
+```bash
+auto-mobile --cli startDevice --platform ios --create-if-missing
+```
+
+**Precedence:** an explicit flag wins over the env var, in *both* directions —
+`--create-if-missing false` disables creation even when
+`AUTOMOBILE_ALLOW_DEVICE_CREATE=1`. With no flag, the env var decides. With
+neither, creation is off.
+
+Created devices are named `AutoMobile-<model>-<id>` so they are easy to find and
+clean up (`xcrun simctl delete <udid>` / `avdmanager delete avd -n <name>`), and
+the resolved device type and runtime are logged at creation time.
+
 ## WebRTC screen streaming (`AUTOMOBILE_WEBRTC_*`)
 
 Defaults for pushing a device's screen to a coordination server over WebRTC/WHIP.
--- a/schemas/tool-definitions.json
+++ b/schemas/tool-definitions.json
@@ -5996,6 +5996,10 @@
         "timeoutMs": {
           "description": "Boot timeout in ms",
           "type": "number"
+        },
+        "createIfMissing": {
+          "description": "Create a device when nothing matches (CLI: --create-if-missing). Default off; AUTOMOBILE_ALLOW_DEVICE_CREATE=1 enables it when this flag is not supplied, and the flag wins.",
+          "type": "boolean"
         }
       },
       "required": [
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -352,6 +352,7 @@ Examples:
   bunx ${installSpecifier} --cli observe
   bunx ${installSpecifier} --cli tapOn --text "Submit"
   bunx ${installSpecifier} --cli startDevice --avdName "pixel_7_api_34"
+  bunx ${installSpecifier} --cli startDevice --platform ios --create-if-missing
   bunx ${installSpecifier} --cli --session-uuid abc-123-uuid observe
   bunx ${installSpecifier} --cli --session-uuid $SESSION_UUID tapOn --text "Submit"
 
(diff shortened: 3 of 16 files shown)


kaeawc:
Verified on `origin/main` (92cd3baa5) during a manual-test iteration over `0.0.44..main`, on a real Android emulator (Medium_Phone_API_35) and iPhone 17 Pro / iOS 26.2.

**Result: PASS** — both the gated-OFF and gated-ON paths behave as specified.

**Flag absent (default).** `startDevice` for a non-existent device fails without creating anything:

```
startDevice --platform ios --deviceId "NoSuchSim-ManualTest"
  -> Device 'NoSuchSim-ManualTest' not found. Available booted: 34C35F33-...
     Available images: iPhone 16e, iPhone 17 Pro Max, iPad Air 13-inch (M3), ...
```

Ground truth after: `simctl list devices | grep -c NoSuchSim` = 0, and `grep -c AutoMobile-` = 0. Nothing was created.

**`createIfMissing: true`.** Creates and boots a real simulator:

```
startDevice --platform ios --deviceId "iPhone 16e" --createIfMissing true
  -> "ios 'iPhone 16e' is ready (cold-boot)", deviceId 0749A13C-..., osVersion 26.2
```

`simctl list devices` confirmed a genuinely new booted `iPhone 16e (0749A13C-...)`, distinct from the pre-existing shutdown `iPhone 16e (67E07B6C-...)`. I deleted the created one afterwards and left the pre-existing device untouched.

One note on naming: the created device took the requested name (`iPhone 16e`) rather than an `AutoMobile-<model>-<id>` form. Since a device *type* by that name existed, creating from it seems intended — flagging only in case the generated-name convention was meant to apply universally.

I did **not** exercise the Android creation path: on this host it would trigger a multi-GB system-image download. The `google_apis` > `default` > `google_apis_playstore` ranking change is therefore unverified on device.

Issue #4124, opened 2026-07-21
CI: hoist ffmpeg install out of the videoRecording test step (parallel-steps)


kaeawc:
## Summary

`ios-xctest-runner-simulator-tests` (`.github/workflows/pull_request.yml`) installs FFmpeg *inside* the video test step:

```yaml
- name: "Run videoRecording MP4 integration test"
  timeout-minutes: 10
  run: |
    brew list ffmpeg >/dev/null 2>&1 || brew install ffmpeg
    ./scripts/ios/video-recording-start-stop-integration.sh
```

Two problems. First, the install is serial dead time. Second — and more important — a cold `brew install ffmpeg` (a large formula tree) is charged against the step's own `timeout-minutes: 10` **and** against the test's internal 180s budget. That step is a known intermittent failure: it has timed out at 180000ms across several unrelated branches.

## Proposed change

Split the install into its own backgrounded step, placed with the existing install fan-out, and re-sync before the test:

```yaml
- name: "Install ffmpeg"
  id: install-ffmpeg
  background: true
  run: brew list ffmpeg >/dev/null 2>&1 || brew install ffmpeg
...
- wait: install-ffmpeg
- name: "Run videoRecording MP4 integration test"
  timeout-minutes: 10
  run: ./scripts/ios/video-recording-start-stop-integration.sh
```

The install then overlaps the simulator boot (`./scripts/ios/boot-simulator.sh`, `timeout-minutes: 15`, observed median 80-220s).

## Dependency justification

- `scripts/ios/video-recording-start-stop-integration.sh` is the only consumer (`require_command ffmpeg`).
- `scripts/ios/boot-simulator.sh` uses `xcrun simctl` + `jq` only; it never touches Homebrew, and its `simulator_udid` `$GITHUB_OUTPUT` is unaffected.
- The integration script reuses an already-booted sim (`simctl list devices booted`), so boot-then-test ordering is preserved by barrier placement.

Mechanism is the proven `run:`-step form (see #3779).

## Acceptance criteria

- [ ] FFmpeg install runs concurrently with the simulator boot; `- wait: install-ffmpeg` precedes the video test.
- [ ] The video test step's `run:` contains only the integration script, so its 10-minute budget covers the test alone.
- [ ] Job green (or failing only for reasons unrelated to this change).

PR #4142, opened 2026-07-22
perf(ci): background the fast-validation dependency install (#4125)

### LINKED ISSUE #4125: CI: background fast-validation apt/bats install across JDK + npm-package setup
## Summary

In `fast-validation` (`.github/workflows/pull_request.yml`), the apt/bats dependency install runs serially after two slow setup steps even though it is independent of both:

- `Set up JDK` (`actions/setup-java@v5`) — download I/O
- `uses: ./.github/actions/setup-auto-mobile-npm-package` — Bun + ripgrep + `bun install` + `turbo run build` (~60-120s)
- `Install fast validation dependencies` — `sudo apt-get update && apt-get install -y xmlstarlet`, plus a `git clone` + source install of bats-core (~30-60s, pure network/disk)
- `Run fast validation checks` — first consumer (the `xml` check shells out to xmlstarlet)
- `Run BATS Tests (Ubuntu)` — consumes bats

`Fast Validation` is a **required** check, so this is on every PR's critical path.

## Proposed change

Move the install to immediately after checkout with `background: true` + `id: install-fast-validation-deps`, then `- wait: install-fast-validation-deps` immediately before `Run fast validation checks`.

## Dependency justification

- `setup-auto-mobile-npm-package` uses Bun/ripgrep/turbo only — no apt, no xmlstarlet, no bats.
- **No dpkg-lock contention:** `.github/actions/setup-ripgrep` installs a pre-built binary (not apt), and a grep across all of `.github/actions/` finds zero `apt-get`/`dpkg` usage. Nothing else in this job touches apt.
- ktfmt is fetched by `all_fast_validate_checks.sh` itself, after the barrier.

Mechanism is the proven `run:`-step form (see #3779).

## Acceptance criteria

- [ ] apt/bats install runs concurrently with the JDK + npm-package setup.
- [ ] `- wait: install-fast-validation-deps` precedes `Run fast validation checks` (first xmlstarlet consumer) and therefore also the BATS step.
- [ ] Fast Validation stays green.

kaeawc:
## Summary

In `fast-validation`, the apt/bats dependency install ran serially **after** two slow setup steps it does not depend on — `Set up JDK` (a download) and `setup-auto-mobile-npm-package` (Bun + ripgrep + `bun install` + build, ~60-120s). The install itself is pure network/disk (~30-60s).

It now runs backgrounded from just after checkout, so the JDK and npm-package setup overlap it, and re-syncs at a `wait` barrier before its first consumer.

```yaml
- name: "Install fast validation dependencies"
  shell: bash
  background: true
  id: install-fast-validation-deps
  run: |
    sudo apt-get update
    sudo apt-get install -y xmlstarlet
    ...
- name: "Set up JDK" ...              # overlaps the install
- uses: ./.github/actions/setup-auto-mobile-npm-package
- wait: install-fast-validation-deps
- name: "Run fast validation checks"  # first xmlstarlet consumer
```

`Fast Validation` is a **required** check, so this is on every PR's critical path — and equally, a misplaced barrier would break every PR. That is why the ordering is pinned by tests rather than left to review.

## Linked Issue

Closes #4125

## Why it's safe
- **First consumer covered:** the aggregator's `xml` check shells out to xmlstarlet (`scripts/xml/validate_xml.sh`) and `Run BATS Tests (Ubuntu)` needs bats. The barrier precedes both.
- **ktfmt pin preserved:** `Set up JDK` still precedes `Run fast validation checks`, so the deterministic ktfmt formatting is unaffected. Pinned by its own test.
- **No dpkg-lock contention:** this is the only apt user in the job. `actions/setup-java` unpacks a tarball; `setup-ripgrep` installs a pre-built binary. A grep across all of `.github/actions/` finds zero `apt-get`/`dpkg` usage.

## Testing

New `test/scripts/fastValidationDepsHoist.test.ts` (parsed YAML, not line matching):

| Acceptance criterion | Test |
| --- | --- |
| Install runs concurrently with JDK + npm-package setup | asserts `background: true` + `id`, and index **before** both `Set up JDK` and the npm-package step |
| `wait` precedes the fast-validation checks (and therefore BATS) | asserts a `wait` step targeting the id, indexed before both consumers |
| Fast Validation stays green | not unit-testable — verified by this PR's own run |

Plus a regression guard that `Set up JDK` still precedes the checks, and an anti-vacuous guard so a renamed job fails loudly instead of passing against an empty step list.

Confirmed **red before** the change (3 failing). Mutation-verified after: dropping the `wait`, removing `background`, or moving the install back after the npm-package step each fail exactly one test.

**Refactor:** the workflow-step assertions are now shared with the #4124 guard via `test/helpers/workflowSteps.ts`. Extracted only now that a second consumer exists, per CLAUDE.md's "grow the interface when the second consumer arrives".

## Validation

- [x] `scripts/all_fast_validate_checks.sh` — **17/17 pass**
- [x] `bun test` — 8,221 pass, 0 fail
- [ ] `bun run typecheck` — one pre-existing error (`PlanSchemaValidator.ts` TS2345, ajv) unrelated to this diff; **zero** errors from the new helper or tests. Local `node_modules` skew in the shared clone (`ajv` 8.20.0 vs nested `ajv-formats/node_modules/ajv` 8.17.1); CI installs with `--frozen-lockfile`.

## Notes for reviewers

Noticed but deliberately **not** fixed here: `.github/actions/setup-auto-mobile-npm-package/action.yml` declares `name: "gradle-task-run"` — a copy-paste mislabel. Pre-existing and unrelated to this change.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -607,6 +607,25 @@ jobs:
           # contain the base commit and the diff would fail.
           fetch-depth: 0
 
+      # Pure network/disk and independent of everything below, so background it
+      # here and let the JDK download and the npm-package setup (Bun + ripgrep +
+      # `bun install` + build, ~60-120s) overlap it. The `wait` barrier re-syncs
+      # before the first consumer: the aggregator's `xml` check shells out to
+      # xmlstarlet, and the BATS step needs bats. No dpkg-lock contention — this
+      # is the only apt user in the job (setup-java unpacks a tarball and
+      # setup-ripgrep installs a pre-built binary). Mechanism per #3779.
+      - name: "Install fast validation dependencies"
+        shell: bash
+        background: true
+        id: install-fast-validation-deps
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y xmlstarlet
+          if ! command -v bats &>/dev/null; then
+            git clone --depth 1 https://github.com/bats-core/bats-core.git /tmp/bats-core
+            sudo /tmp/bats-core/install.sh /usr/local
+          fi
+
       # Pin the JDK so the ktfmt check formats deterministically (project targets
       # Java 21) instead of relying on the runner image's default JDK.
       - name: "Set up JDK"
@@ -617,15 +636,7 @@ jobs:
 
       - uses: ./.github/actions/setup-auto-mobile-npm-package
 
-      - name: "Install fast validation dependencies"
-        shell: bash
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y xmlstarlet
-          if ! command -v bats &>/dev/null; then
-            git clone --depth 1 https://github.com/bats-core/bats-core.git /tmp/bats-core
-            sudo /tmp/bats-core/install.sh /usr/local
-          fi
+      - wait: install-fast-validation-deps
 
       - name: "Run fast validation checks"
         shell: bash
--- a/test/helpers/workflowSteps.ts
+++ b/test/helpers/workflowSteps.ts
@@ -1,0 +1,67 @@
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { load } from "js-yaml";
+
+/**
+ * Helpers for asserting the *structure* of a GitHub Actions job's steps.
+ *
+ * Workflow guards must parse the YAML rather than line-match it (CLAUDE.md:
+ * don't regex-parse YAML when a structured parser exists — `js-yaml` is already
+ * a direct dependency). Parsing also makes the assertions immune to reflows,
+ * comments, and text that happens to appear inside another step's block scalar.
+ *
+ * Extracted when the second consumer arrived (the parallel-steps guards for
+ * #4124 and #4125); keep the surface to what those guards actually need.
+ */
+
+export interface WorkflowStep {
+  name?: string;
+  id?: string;
+  run?: string;
+  uses?: string;
+  /** Native parallel-steps: run this step asynchronously. */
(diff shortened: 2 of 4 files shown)


kaeawc:
### Measured on CI (run 29879480522) — Fast Validation green ✅

| step | start | end | dur |
| --- | --- | --- | --- |
| **Install fast validation deps (bg)** | **00:11:31** | 00:11:41 | **10s** |
| Set up JDK | **00:11:31** | 00:11:37 | 6s |
| setup-auto-mobile-npm-package | 00:11:37 | 00:11:43 | 6s |
| **wait: install-fast-validation-deps** | 00:11:43 | 00:11:43 | **0s** |
| Run fast validation checks | 00:11:43 | 00:12:03 | 20s ✅ |
| Run BATS Tests (Ubuntu) | 00:12:03 | 00:13:35 | 92s ✅ |

The install and `Set up JDK` **started at the same instant (00:11:31)**, and the install finished inside the JDK + npm-package window, so the `wait` barrier cost **0s**.

**Delta: ~22s serial (6 + 6 + 10) → ~12s overlapped ≈ ~10s saved.**

Because this PR modifies the `fast-validation` job itself, AC3 is verified directly — this run *is* the change under test, and both consumers (`Run fast validation checks` for xmlstarlet, `Run BATS Tests (Ubuntu)` for bats) passed after the barrier.

**Honest caveat:** the install took 10s here, below the 30-60s the issue estimated — the runner image already had most of it. The saving scales with how cold that install actually is, so treat ~10s as the floor rather than the typical case.

PR #4144, opened 2026-07-22
perf(ci): background the BATS install in bats-tests (#4126)

### LINKED ISSUE #4126: CI: background BATS install across the Bun setup in bats-tests (pull_request + merge)
## Summary

The `bats-tests` job installs BATS serially after the Bun setup, in **two** workflows:

- `.github/workflows/pull_request.yml` (matrix incl. macos-latest)
- `.github/workflows/merge.yml` (matrix `[ubuntu-latest, macos-latest]`)

Order today: `Setup Bun` -> `Install Bun dependencies` -> `Install BATS` -> `Run BATS Tests`.

`Install BATS` is `brew install bats-core` on macOS (typically 25-60s) or `git clone` + `sudo install.sh /usr/local` on Linux — pure network/disk, and independent of everything before it. `BATS Shell Tests (macos-latest)` is a **required** check.

## Proposed change

In both files, move `Install BATS` to immediately after checkout with `background: true` + `id: install-bats`, and add `- wait: install-bats` immediately before `Run BATS Tests`.

## Dependency justification

- `scripts/ci/install-bun-deps.sh` runs `bun install --frozen-lockfile` into `node_modules` and appends `node_modules/.bin` to `$GITHUB_PATH`. It never uses brew or bats.
- The BATS install writes only `/usr/local/bin/bats` (or the brew prefix) and clones into `/tmp/bats-core`. Disjoint filesystem targets.
- `bats test/bats/` is the sole consumer of both.
- Keep `install-bun-deps.sh` in the foreground since it writes `$GITHUB_PATH`.

Mechanism is the proven `run:`-step form (see #3779).

## Acceptance criteria

- [ ] BATS install overlaps Setup Bun + `bun install` in both workflows.
- [ ] `- wait: install-bats` precedes `Run BATS Tests` in both.
- [ ] BATS Shell Tests green on both matrix legs.

kaeawc:
## Summary

The `bats-tests` job installed BATS serially after the Bun setup, in both `pull_request.yml` and `merge.yml`, even though the two are independent.

`Install BATS` is `brew install bats-core` on macOS (tens of seconds) or a `git clone` + source install on Linux — pure network/disk. `install-bun-deps.sh` only runs `bun install --frozen-lockfile` into `node_modules`. Disjoint targets, and `bats test/bats/` is the sole consumer of both.

Resulting order in both workflows:

```
Git Checkout -> Install BATS (background) -> Setup Bun -> Install Bun dependencies -> wait: install-bats -> Run BATS Tests
```

`BATS Shell Tests (macos-latest)` is a **required** check, and `pull_request.yml`'s matrix is macOS-only — which is exactly where the slow `brew` path lives, so that is where the win lands.

## Linked Issue

Closes #4126

## Why it's safe
- **Disjoint targets** — verified `scripts/ci/install-bun-deps.sh` contains no `brew`/`bats` usage; it is `bun install --frozen-lockfile` plus a `$GITHUB_PATH` append.
- **`install-bun-deps.sh` stays in the foreground** because it writes `$GITHUB_PATH`; only the BATS install is backgrounded. Pinned by its own test.
- **Barrier precedes the sole consumer** in both files.

## Testing

New `test/scripts/batsInstallHoist.test.ts`, parameterized over **both** workflows and using the shared `test/helpers/workflowSteps.ts` extractor (parsed YAML, not line matching):

| Acceptance criterion | Test |
| --- | --- |
| Install overlaps Setup Bun + `bun install`, both workflows | asserts `background: true` + `id`, and index before `Setup Bun` and `Install Bun dependencies` |
| `wait: install-bats` precedes `Run BATS Tests`, both workflows | asserts a `wait` targeting the id, indexed before the consumer |
| BATS Shell Tests green on both legs | not unit-testable — verified by this PR's own run |

Plus a regression guard that `Install Bun dependencies` still precedes the test step, and an anti-vacuous guard so a renamed job fails loudly.

Confirmed **red before** the change: 6 failures (3 per workflow — exactly the AC assertions), with the 4 pre-existing-invariant tests passing as expected.

**Scoping note:** the guard is scoped by job id. `pull_request.yml` also runs bats inside `fast-validation` (as `Run BATS Tests (Ubuntu)`), so a whole-file search would have conflated the two.

## Validation

- [x] `scripts/all_fast_validate_checks.sh` — **17/17 pass**
- [x] `bun test` — 8,231 pass, 0 fail
- [x] `bats test/bats/` — 490 tests, 2 failures, both reproducing on a pristine `origin/main` checkout (formatter version pin; local ffmpeg absence) and unrelated to this change
- [ ] `bun run typecheck` — one pre-existing `PlanSchemaValidator.ts` TS2345 (ajv) unrelated to this diff, which touches no `src/`. Local `node_modules` skew in the shared clone; CI installs with `--frozen-lockfile`.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -306,17 +306,15 @@ jobs:
         with:
           lfs: false
 
-      - name: "Setup Bun"
-        uses: oven-sh/setup-bun@v2
-        with:
-          bun-version: 1.3.9
-
-      - name: "Install Bun dependencies"
-        shell: bash
-        run: scripts/ci/install-bun-deps.sh
-
+      # Pure network/disk (`brew install bats-core` on macOS, a source install on
+      # Linux) and independent of the Bun setup below, which only writes
+      # node_modules. Background it so the two overlap, then re-sync at the
+      # barrier before `bats test/bats/`, the sole consumer of both. Mechanism
+      # per #3779.
       - name: "Install BATS"
         shell: bash
+        background: true
+        id: install-bats
         run: |
           if command -v bats &>/dev/null; then
             echo "bats already installed: $(bats --version)"
@@ -328,6 +326,18 @@ jobs:
             git clone --depth 1 https://github.com/bats-core/bats-core.git /tmp/bats-core
             sudo /tmp/bats-core/install.sh /usr/local
           fi
+
+      - name: "Setup Bun"
+        uses: oven-sh/setup-bun@v2
+        with:
+          bun-version: 1.3.9
+
+      # Stays in the foreground: it appends node_modules/.bin to $GITHUB_PATH.
+      - name: "Install Bun dependencies"
+        shell: bash
+        run: scripts/ci/install-bun-deps.sh
+
+      - wait: install-bats
 
       - name: "Run BATS Tests"
         run: bats test/bats/
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -749,17 +749,15 @@ jobs:
         with:
           lfs: false
 
-      - name: "Setup Bun"
-        uses: oven-sh/setup-bun@v2
-        with:
-          bun-version: 1.3.9
-
-      - name: "Install Bun dependencies"
-        shell: bash
-        run: scripts/ci/install-bun-deps.sh
-
+      # Pure network/disk (`brew install bats-core` on macOS, a source install on
+      # Linux) and independent of the Bun setup below, which only writes
+      # node_modules. Background it so the two overlap, then re-sync at the
+      # barrier before `bats test/bats/`, the sole consumer of both. Mechanism
+      # per #3779.
       - name: "Install BATS"
         shell: bash
+        background: true
+        id: install-bats
         run: |
           if command -v bats &>/dev/null; then
             echo "bats already installed: $(bats --version)"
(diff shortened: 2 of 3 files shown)


kaeawc:
### Measured on CI (macos-latest leg) — BATS Shell Tests green ✅

| step | start | end | dur |
| --- | --- | --- | --- |
| **Install BATS (bg)** | **01:59:09** | 01:59:18 | **9s** |
| Setup Bun | **01:59:09** | 01:59:12 | 3s |
| Install Bun dependencies | 01:59:12 | 01:59:17 | 5s |
| **wait: install-bats** | 01:59:17 | 01:59:18 | 1s |
| Run BATS Tests | 01:59:18 | 02:01:44 | ✅ pass |

`Install BATS` and `Setup Bun` **started at the same instant (01:59:09)**, and the install ran alongside the Bun setup + dependency install (8s combined), leaving the barrier just 1s to absorb.

**Delta: ~17s serial (3 + 5 + 9) → ~9s overlapped ≈ ~8s saved**, on a required check.

AC3 is verified directly: this PR modifies `bats-tests`, and that job ran and passed here.

**Honest caveat:** `brew install bats-core` took 9s on this runner because bats was largely present already. The issue estimated 25-60s cold; the saving scales with how cold the install is, so ~8s is the floor rather than the typical case.

Issue #3943, opened 2026-07-18
Flaky test (recurrence of #3877): videoRecordingToolSegmented "bare (by-device) stop" still times out on macOS CI


kaeawc:
### Summary

Recurrence of #3877 (closed as completed 2026-07-17). The test **`videoRecording tool segmentation branch > bare (by-device) stop finalizes the segmented session and leaves no rotation timer`** (`test/server/videoRecordingToolSegmented.test.ts:292`) timed out again on the **`Node TypeScript Build and Test (macos-latest)`** leg, despite the #3877 fix.

### Latest evidence

Run [29646285847, job 88085003583](https://github.com/kaeawc/auto-mobile/actions/runs/29646285847/job/88085003583) — **macOS-latest only**; the ubuntu-latest and windows-latest legs of the same run passed.

```
1 tests failed:
(fail) videoRecording tool segmentation branch >
       bare (by-device) stop finalizes the segmented session and leaves no rotation timer
       [20009.64ms]
  ^ this test timed out after 20000ms.

7960 pass, 13 skip, 1 fail, 1 error
```

### Why this is a recurrence, not the original

The #3877 fix landed both mitigations that issue recommended:

- The test now injects fake `startVideoRecording` / `stopVideoRecording` into the tool-created segmented session via `setSegmentedSessionRecordingDependencies` (`test/server/videoRecordingToolSegmented.test.ts:293`).
- The per-test timeout was raised to `TIMER_DRIVEN_TEST_TIMEOUT_MS = 20_000` (`:36`), with a "headroom stops a loaded CI machine failing a correct test" comment.

Yet the test still consumes the **full 20s** and reports `1 error` alongside the timeout — the same symptom profile #3877 traced to a late async continuation hanging (there, the real-DB guard #3063/#3067 rebuilding the real `VideoRecordingRepository` off the shared module-level `videoRecordingManager` singleton once the injected deps were nulled). So either the injected-deps seam doesn't cover every async path the bare-stop handler can take on a slow runner, or a different late continuation is hanging. Raising the timeout only widened the race window; it didn't close it.

### Next step to diagnose

[...]

PR #4146, opened 2026-07-22
perf(installer): overlap independent setup

kaeawc:
## Summary

- overlap the informational iOS simulator-runtime probe with the rest of installer setup
- install missing macOS development tools in one Homebrew transaction and verify every requested formula afterward
- overlap the non-interactive contributor CLI and daemon path with independent Homebrew work while preserving output and failure propagation

## Why

The macOS development installer spent most of its time waiting serially for CoreSimulator and repeated Homebrew startup. These operations do not depend on the Bun CLI/daemon path once Bun is available.

## Validation

- `bash -n scripts/install.sh`
- `shellcheck scripts/install.sh`
- `bats test/bats/`
- `git diff --check`

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -3343,12 +3343,34 @@ start_ios_runtime_probe() {
     IOS_RUNTIME_PROBE_PID=$!
 }
 
+# Seconds to wait for the background runtime probe before abandoning it.
+# Overridable so the tests can exercise the timeout path without a real stall.
+IOS_RUNTIME_PROBE_TIMEOUT_SECONDS="${IOS_RUNTIME_PROBE_TIMEOUT_SECONDS:-20}"
+
 finish_ios_runtime_probe() {
     if [[ -z "${IOS_RUNTIME_PROBE_PID}" ]]; then
         return 0
     fi
 
-    local probe_status=0
+    # `xcrun simctl list` can stall indefinitely on a loaded macOS runner -- the
+    # hazard #3943 documents and the reason FakeDeviceSessionManager exists. This
+    # probe only decorates output with the installed runtimes, so an unbounded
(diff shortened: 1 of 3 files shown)


kaeawc:
Hang fixed and validated on a real runner.

**Before:** `BATS Shell Tests (macos-latest)` cancelled at **6h0m17s**.
**After:** success in **2m28s** (09:41:28 → 09:43:56) — matching the 2.4-minute healthy median measured across 43 runs.

Root cause was the background iOS runtime probe joined with a bare `wait`:

```bash
xcrun simctl list runtimes >"${FILE}" 2>/dev/null &
wait "${IOS_RUNTIME_PROBE_PID}"
```

`xcrun simctl list` can stall indefinitely on a loaded macOS runner — the hazard #3943 documents, and the reason `FakeDeviceSessionManager` exists. The probe runs whenever the OS is macOS and `xcrun` is present, so every macOS runner took that path. It only decorates output with the installed runtime list, so an unbounded wait traded a cosmetic detail for a six-hour job.

Now polls with a deadline (`IOS_RUNTIME_PROBE_TIMEOUT_SECONDS`, default 20, overridable for tests), kills the probe on expiry, and degrades to "runtimes unknown". Regression test stubs a stalled probe; it uses `exec` so killing the probe PID actually stops it rather than orphaning a child that keeps the inherited fd open.

Verified non-vacuous **by exit code rather than output shape**: with the bound removed the test hangs (`timeout` exit 124); with it, exit 0.

Note I reset onto your amended `f399c0dd6` rather than resolving a rebase conflict blind — your commit is untouched, mine sits on top.

Separately, #4156 bounds the 29 jobs that had no `timeout-minutes` at all, so the *next* hang of any origin costs minutes instead of the 360-minute default.

PR #4206, opened 2026-07-22
fix(database): stop truncating SQLite error type and message at whitespace/comma

### LINKED ISSUE #4191: fix(database): SQLite error type and message silently truncated at whitespace and comma
## Defect

Two regexes in the failure path mangle SQLite error text
(`src/features/database/DatabaseInspector.ts:184-186`):

```ts
const errorType = output.match(/errorType=(\w+)/)?.[1] || "UNKNOWN";
const errorMatch = output.match(/error=([^,}]+)/);
```

- `\w+` stops at whitespace, so an errorType of `SQL Error` is reported as **`SQL`**.
- `[^,}]+` stops at the first comma, so `no such column: foo, bar` is truncated to
  **`no such column: foo`**.

Both then feed `ActionableError`, so the user is shown a **silently truncated** diagnostic —
the most misleading possible failure mode for an error path, since the message looks complete.

## Acceptance criteria

- [ ] Multi-word `errorType` preserved intact.
- [ ] Error text containing commas preserved intact.
- [ ] Error text containing `}` handled (whatever the real delimiter contract is — establish
      it from the actual producer rather than guessing, and state it in the test).
- [ ] Single-word / no-comma cases unchanged (control rows).
- [ ] Red before the fix, green after.

Confirmed by reading `:184-186`. The audit's 4-row table has 3 red at HEAD.

kaeawc:
## Summary

`DatabaseInspector.parseContentCallResult` matched the failure envelope with two regexes that
silently truncate the diagnostic they are supposed to surface: `errorType=(\w+)` stops at the
first non-word character, and `error=([^,}]+)` stops at the first comma or brace. SQLite
messages routinely contain both (`no such table: users, orders`, `near "}": syntax error`), so
the user was shown a truncated message that still *looked* complete — the worst failure mode
for an error path.

This replaces both regexes with a boundary-aware `extractBundleValue(output, key)` helper that
parses the printed Bundle the way the producer actually writes it.

**Wire contract (established from the producer, not guessed):**
`android/auto-mobile-sdk/src/debug/kotlin/.../database/DatabaseInspectorProvider.kt` fills a
`Bundle` with the string keys `success`, `errorType`, `error` and (on success) `result`.
`adb shell content call` prints it via `Bundle.toString()` as `Bundle[{key=value, key=value}]` —
entries joined by `", "`, the whole map closed by one trailing `"}]"`. Values are raw and
unescaped, so a value may itself contain spaces, commas and closing braces. The only reliable
terminators for a value are therefore the next `, <knownKey>=` boundary or the bundle's trailing
`}]`. The helper scans for exactly those, so entry order in the Bundle does not matter either
(`ArrayMap` ordering is hash-based, not insertion-ordered).

`success=` now goes through the same helper, and the existing balanced-brace `result=` extractor
is untouched.

## Linked Issue

Closes #4191

## Bug / Regression Context

- **Observed symptom:** `Database error (SQL): ...` for an `errorType` of `SQL Error`, and
  `no such column: foo` for an error of `no such column: foo, bar`.
- **Repro steps / conditions:** any `executeSQL` / `listTables` / `getTableData` failure whose
  SQLite message contains a comma or a brace, which is the common case for syntax errors.
- **Prior attempts:** none.

## Validation

- [x] `bun test` — 8317 pass / 0 fail (full suite)
- [x] `bun run lint` — clean (exit 0)
- [ ] `bun run typecheck` — fails on a **pre-existing** main breakage unrelated to this diff:
      a baselined `TS2339` line for `src/daemon/telemetryPushSocketServer.ts` drifted because the
      printed `severity` union member order changed (`"critical" | "high" | "medium" | "low"` in
      `scripts/typecheck-baseline.txt` vs `"high" | "medium" | "critical" | "low"` now). The same
      failure reproduces on other in-flight PRs (#4203, #4204, #4205), so it is not caused here
      and is deliberately not fixed here to avoid four PRs editing the same baseline line.
- [x] New code reuses the existing helper style in this file; no new dependency

## Acceptance criteria

| Criterion | Test |
| --- | --- |
| Multi-word `errorType` preserved intact | `preserves a multi-word errorType` |
| Error text containing commas preserved intact | `preserves error text containing commas` |
| Error text containing `}` handled per the real delimiter contract | `preserves error text containing a closing brace` (contract documented in the test) |
| Single-word / no-comma cases unchanged (control rows) | `single-word errorType and comma-free error text are unchanged`, plus the pre-existing `DISABLED` / `NotFound` / `SqlError` tests |
| Missing keys still fall back | `falls back to UNKNOWN and Unknown error when the keys are absent` |
| Entry order independence | `preserves error text when errorType follows error in the bundle` |

All four new criterion tests were confirmed **red** at HEAD before the fix (truncating to `SQL`,
`no such column: foo`, `near "`, `no such table: users` respectively) and are green after.

## Device Verification

N/A — pure parsing change, covered by unit tests against captured wire text.

### DIFF
--- a/src/features/database/DatabaseInspector.ts
+++ b/src/features/database/DatabaseInspector.ts
@@ -67,6 +67,13 @@ export interface SQLResult {
  * `adb shell content call` commands.
  */
 export class DatabaseInspector {
+  /**
+   * Keys the DatabaseInspectorProvider puts into the Bundle it returns from `call`
+   * (android/auto-mobile-sdk/src/debug/kotlin/.../database/DatabaseInspectorProvider.kt).
+   * Used to find entry boundaries when splitting the printed Bundle.
+   */
+  private static readonly BUNDLE_KEYS = ["success", "errorType", "error", "result"];
+
   constructor(
     private device: BootedDevice,
     private adb: AdbExecutor
@@ -177,13 +184,11 @@ export class DatabaseInspector {
    */
   private parseContentCallResult<T>(output: string): T {
     // Check for success
-    const successMatch = output.match(/success=(\w+)/);
-    const success = successMatch?.[1] === "true";
+    const success = this.extractBundleValue(output, "success") === "true";
 
     if (!success) {
-      const errorType = output.match(/errorType=(\w+)/)?.[1] || "UNKNOWN";
-      const errorMatch = output.match(/error=([^,}]+)/);
-      const error = errorMatch?.[1]?.trim() || "Unknown error";
+      const errorType = this.extractBundleValue(output, "errorType") || "UNKNOWN";
+      const error = this.extractBundleValue(output, "error") || "Unknown error";
       throw new ActionableError(`Database error (${errorType}): ${error}`);
     }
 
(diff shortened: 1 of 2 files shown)


kaeawc:
Follow-up for the reviewer's P2 on the residual `Bundle.toString()` ambiguity: filed as #4216. The mechanism is real and confirmed — a caller-controlled fragment such as `near "x, error=y": syntax` is still misparsed — but removing it requires changing the SDK-to-TS wire format (a JSON failure envelope from `DatabaseInspectorProvider`) plus a version-skew story and an SDK release, which is out of scope for a parser bug fix. The limitation is documented at `src/features/database/DatabaseInspector.ts:230`.

CI note: the three red checks on this head are all pre-existing main breakage reproducing identically on unrelated PRs (#4203, #4204, #4205) — `Node TypeScript Build and Test (ubuntu-latest)` is the typecheck-gate drift tracked by #4208/#4209, and `BATS Shell Tests (macos-latest)` / `Shell Tests` are the bash 3.2 `boot-simulator.sh` breakage tracked by #4212. `Installer Development (macos-latest)` failed on a runner-side `actions/checkout/v6/action.yml` load error, which is infrastructure. Nothing here is caused by this diff.

Issue #4213, opened 2026-07-22
fix(action): openLink interpolates the URL into a shell string — quote-breaking URLs are mangled or inject


kaeawc:
## Summary

`openLink` interpolates a caller-supplied URL straight into a shell command string on
both platforms, with only a pair of double quotes around it:

`src/features/action/OpenURL.ts:169`
```ts
await this.adb.executeCommand(`shell am start -a android.intent.action.VIEW -d "${url}"`);
```

`src/features/action/OpenURL.ts:201`
```ts
await simctl.executeCommand(`openurl ${this.device.deviceId} "${url}"`);
```

A URL containing a `"` closes the quote, so the remainder is parsed by the shell as
further arguments — and `$`, backticks and `\` are all still live inside double quotes.
`openLink` takes the URL directly from the MCP tool argument, so the value is arbitrary
and frequently comes from a page, a deep link table, or a model-generated string.

## Impact

Two distinct effects, in increasing severity:

- **Correctness.** Any legitimate URL containing `"` (or `$`, or a backtick) is silently
  mangled before it reaches the intent, in the same "logs look right, device sees something
  else" way as [#4166](https://github.com/kaeawc/auto-mobile/issues/4166).
- **Injection.** `openLink('https://x/" ; id ; echo "')` appends shell commands to what
  `adb shell` / `xcrun simctl` executes.

## Suggested fix

Prefer an argv-form execution that preserves argument boundaries rather than quoting into
a shell string — this is the same class the repo's `spawn(..., { shell: true })` guidance in
`CLAUDE.md` already calls out. If the exec seam must stay string-shaped here, escape the
value explicitly (single-quote and escape embedded single quotes) rather than relying on
double quotes.

A regression test belongs alongside the [#4166](https://github.com/kaeawc/auto-mobile/issues/4166) whitespace table in
`test/features/action/OpenURL.test.ts`, asserting the exact command string issued via
`FakeAdbExecutor.getExecutedCommands()` / `FakeSimCtlClient.getMethodCalls("executeCommand")`
for a URL containing `"`, `$` and a backtick.

Noticed while reviewing [#4204](https://github.com/kaeawc/auto-mobile/pull/4204) (fix for [#4166](https://github.com/kaeawc/auto-mobile/issues/4166)); deliberately left out of that
PR to keep it to the one-line trim fix.

PR #4233, opened 2026-07-22
fix(ios): issue simctl commands as argv arrays instead of re-split strings

### LINKED ISSUE #4196: fix(ios): SimCtlClient drops empty argument values and mangles escapes, turning "defaults write" into "read"
## Defect

`SimCtlClient` loses argument content when round-tripping through its string-based command
path (`src/utils/ios-cmdline-tools/SimCtlClient.ts:302`). Measured on a 12-row table: 9 pass,
**3 fail at HEAD**:

| Input | Result |
|---|---|
| empty string value | **dropped** — `defaults write` silently becomes `defaults read` |
| `\n` in a value | collapsed to a letter |
| `\t` in a value | collapsed to a letter |

The empty-value case is the worst: dropping an argument changes the *verb* of the command,
so a write silently becomes a read and reports success.

## Fix

Migrate to `executeCommandArgs` (argv array) rather than string concatenation, so no
quoting/splitting round-trip exists to lose data.

## Acceptance criteria

- [ ] Empty-string values survive; `defaults write` stays a write.
- [ ] `\n`, `\t`, `\r`, quotes, backslashes, spaces, and unicode survive the round trip.
- [ ] Table-driven over argv shapes, asserting the **argv actually issued**, not the return value.
- [ ] Red before the fix, green after.

Must land together with the `executeCommandArgs` migration — the table is red until then.
Measurements from the audit's critic; **not independently re-run by me**.

kaeawc:
## Summary

`SimCtlClient` had a string round-trip in its command path: callers built an `xcrun simctl`
command line by `JSON.stringify`-quoting values into a string (`quoteSimctlArg`), and
`splitCommandArgs` then re-split that string back into argv. The round trip is lossy.

Two confirmed defects, both reproduced as red tests before the fix:

- **An empty value is dropped entirely.** `splitCommandArgs` only pushed a token when
  `current.length > 0`, so `""` vanished. That shifts every later positional argument, so
  `spawn <udid> defaults write <domain> <key> ""` reached the simulator as
  `defaults write <domain> <key>` — a different command, silently.
- **Escapes are mangled.** `quoteSimctlArg` emits JSON escapes (`\n`, `\t`, `\r`, `\"`,
  `\\`), and `splitCommandArgs` treats `\` as "take the next character literally", so a
  value containing a newline came back as the letter `n`.

The fix removes the round trip rather than patching individual escapes: every data-bearing
simctl call site now passes an **argument array** to the existing `executeCommandArgs`
seam, which hands argv straight to `execFile`. `quoteSimctlArg` had no remaining callers
and is deleted, so the footgun cannot be reintroduced by copy-paste.

`splitCommandArgs` is also fixed to preserve empty quoted tokens, as defense in depth for
the remaining literal (non-data) string commands such as `list devices --json`.

No new helper was added — `executeCommandArgs` / `startCommandArgs` / `createExecResult`
already existed and are reused as-is.

## Linked Issue

Closes #4196

## Bug / Regression Context

- **Observed symptom:** a `defaults write` with an empty value silently executed as a
  shorter, different command and reported success; values containing `\n`/`\t` were
  corrupted before reaching the simulator.
- **Repro steps / conditions:** any `restoreIosSettings` / `get_app_container` /
  `simctl privacy` call whose value contains whitespace, a quote, a backslash, or is empty.

## Changes

- `src/utils/ios-cmdline-tools/iosSettings.ts` — `defaults read` / `defaults write` /
  `simctl ui` now issue argv arrays; the helpers take a narrow
  `Pick<SimCtlClient, "executeCommandArgs">`.
- `src/utils/ios-cmdline-tools/iosAppContainer.ts` — `get_app_container` via argv;
  `quoteSimctlArg` removed.
- `src/utils/DeepLinkManager.ts` — `get_app_container ... app` via argv.
- `src/features/action/IosSimulatorPermissions.ts` — `simctl privacy` via argv; the local
  `IosSimulatorPrivacyClient` interface now declares `executeCommandArgs`.
- `src/utils/ios-cmdline-tools/SimCtlClient.ts` — `splitCommandArgs` preserves empty
  quoted tokens.
- `test/fakes/FakeSimCtlClient.ts` — `setCommandArgsResult` / `setCommandArgsError` key
  stubs by exact argv (a joined command string cannot represent an empty or spaced value),
  plus `get_app_container` handling on the argv path.

## Testing

New table-driven suite `test/utils/ios-cmdline-tools/simctlArgvIntegrity.test.ts` asserts
the **argv actually issued** (not the return value) for 12 argument shapes: empty string,
`\n`, `\t`, `\r`, double quote, single quote, backslash, literal `C:\new\tab`, spaces,
leading/trailing spaces, shell metacharacters (`$HOME`, backtick, `;`), and unicode.
All 27 tests were red at HEAD and are green with the fix. Existing simctl string
assertions were converted to argv assertions.

## Validation

- [x] `bun run lint` passes
- [x] `bun test` — 8338 pass / 0 fail / 11 skip
- [x] `bun run typecheck` reports no new errors (baseline gate green; note it also reports
      10 baseline errors no longer occur — baseline left unmodified deliberately)
- [x] `bun run build` succeeds
- [x] New code reuses the existing `executeCommandArgs` exec seam; no new helper or dependency

## Device Verification

Not run — no simulator was touched (another session is live-testing against the shared
daemon). All coverage is through the `SimCtlClient` fake/seam.

## Follow-ups

- [#4213](https://github.com/kaeawc/auto-mobile/issues/4213) is the **same bug class** in
  `openLink` (`src/features/action/OpenURL.ts:169,201`), where a URL is interpolated into a
  shell string on both Android and iOS. It does **not** share a code path or helper with
  what this PR changed — `OpenURL` never used `quoteSimctlArg`, and the Android half goes
  through `AdbClient`, a different executor — so this PR does not close it. The iOS half is
  now a one-line change to the same `executeCommandArgs` seam this PR standardizes on; left
  out to keep this PR scoped.

### DIFF
--- a/src/features/action/IosSimulatorPermissions.ts
+++ b/src/features/action/IosSimulatorPermissions.ts
@@ -5,7 +5,6 @@ import { promisify } from "util";
 import type { BootedDevice, ExecResult } from "../../models";
 import { SimCtlClient } from "../../utils/ios-cmdline-tools/SimCtlClient";
 import { isIosSimulatorUdid } from "../../utils/ios-cmdline-tools/iosDeviceType";
-import { quoteSimctlArg } from "../../utils/ios-cmdline-tools/iosAppContainer";
 
 export type IosSimulatorPermissionAction = "grant" | "revoke" | "reset";
 
@@ -47,7 +46,7 @@ export interface IosSimulatorPermissionQueryResult {
 }
 
 export interface IosSimulatorPrivacyClient {
-  executeCommand(command: string, timeoutMs?: number): Promise<ExecResult>;
+  executeCommandArgs(args: string[], timeoutMs?: number): Promise<ExecResult>;
 }
 
 interface TccPermissionRow {
@@ -241,14 +240,13 @@ export class IosSimulatorPermissions {
     const results: IosSimulatorPermissionCommandResult[] = await Promise.all(
       normalizedPermissions.map(async permission => {
         try {
-          const command = [
+          const result = await this.simctl.executeCommandArgs([
             "privacy",
-            quoteSimctlArg(this.device.deviceId),
+            this.device.deviceId,
             action,
-            quoteSimctlArg(permission),
-            quoteSimctlArg(normalizedAppId)
-          ].join(" ");
-          const result = await this.simctl.executeCommand(command);
+            permission,
+            normalizedAppId
+          ]);
           return { permission, success: true, stdout: result.stdout, stderr: result.stderr };
         } catch (error) {
           return {
--- a/src/utils/DeepLinkManager.ts
+++ b/src/utils/DeepLinkManager.ts
@@ -16,7 +16,6 @@ import type { ElementGeometry } from "./interfaces/ElementGeometry";
 import { DefaultElementParser } from "../features/utility/ElementParser";
 import { DefaultElementGeometry } from "../features/utility/ElementGeometry";
 import { SimCtlClient } from "./ios-cmdline-tools/SimCtlClient";
-import { quoteSimctlArg } from "./ios-cmdline-tools/iosAppContainer";
 import { isIosSimulatorUdid } from "./ios-cmdline-tools/iosDeviceType";
 
 /**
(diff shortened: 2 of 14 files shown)


kaeawc:
Manual-test verification on HEAD `44f83116b`, iOS simulator (iOS 18.x) — **confirmed fixed**.

`openLink` with `https://example.invalid/?a=1&b=2` — Safari's URL field after load reads `{"resource-id":"URL","value":"https://example.invalid/?a=1&b=2"}`. The full query survived; a re-split/shell-interpolated argument would have truncated at the `&`.

Regression sweep on the same simctl-backed paths also clean: `launchApp com.apple.Preferences` → `activeWindow.appId":"com.apple.Preferences"`; `terminateApp` confirmed via `launchctl list` (no mobilesafari). Also covers #4307.

PR #4246, opened 2026-07-22
fix(cli): coerce --cli values by declared schema type instead of JSON.parse

### LINKED ISSUE #4241: --cli JSON.parse coercion makes numeric-looking strings unpassable: inputText --text 12345 and sendSms --phoneNumber 5551234567 both fail
## Summary

The `--cli` argument parser runs every value through `JSON.parse`, so a numeric-looking string argument arrives as a **number** and is then rejected by schema validation. String-typed parameters whose natural values are digits — phone numbers, PINs, OTP codes, zip codes, quantities — cannot be passed at all.

Most visibly, you cannot type a numeric string:

```
$ ... --cli inputText --platform android --text "12345"
Error: Invalid parameters for tool inputText: text expected string, received number
```

Quoting does not help — the shell strips the quotes and the parser re-interprets the bare token.

## Evidence

`origin/main` @ 92cd3baa5, Android emulator.

| call | result |
|---|---|
| `inputText --text "12345"` | **fails** — `text expected string, received number` |
| `inputText --text "abc123"` | succeeds |
| `sendSms --phoneNumber "5551234567"` | **fails** — `phoneNumber expected string, received number` |
| `sendSms --phoneNumber "+15551234567"` | succeeds |
| `sendSms --phoneNumber "555-123-4567"` | rejected by the tool's own validator (correct, different error) |
| `tapOn --selector '{"text":"12345"}'` | succeeds — nested JSON preserves the inner string |

The `sendSms` pair is the clearest demonstration: the tool's own validator accepts *"digits with optional leading `+`"*, so `5551234567` is a **valid** phone number by the tool's own contract. Only the `+` variant survives the CLI, and only because a leading `+` is not valid JSON.

## Root cause

`src/cli/index.ts:87`

```ts
// Try to parse as JSON, fallback to string
try {
  params[paramName] = JSON.parse(nextArg);
} catch {
  // If not valid JSON, treat as string
  params[paramName] = nextArg;
}
```

`JSON.parse` is doing double duty: it enables nested objects (`--selector '{"text":"Settings"}'`) and numeric params (`--count 5`, documented in the help text at `:369`), but it has no knowledge of the target parameter's declared type, so it cannot tell `--count 5` (wants a number) from `--text 12345` (wants a string).

Observed coercions:

```
"5551234567" -> number   5551234567
"12345"      -> number   12345
"007"        -> string   "007"      (leading zero is not valid JSON)
"true"       -> boolean  true
"null"       -> null
"abc123"     -> string   "abc123"
```

Note `"007"` surviving as a string while `"12345"` does not — the behaviour is inconsistent in a way that is hard to predict from the outside.

## Suggested fix

Consult the tool's input schema, which is already available to the CLI (it validates against it moments later):

- If the target property is declared `type: "string"`, use the raw token verbatim and skip `JSON.parse` entirely.
- If it is `object`/`array`, parse as JSON as today.
- If it is `number`/`boolean`, coerce as today.
- Unknown/absent property: keep the current best-effort behaviour.

That preserves every documented form (`--count 5`, `--raw true`, `--selector '{...}'`) while fixing string params.

If schema-aware parsing is judged too invasive, a narrower fix is an explicit escape hatch — e.g. `--text:string 12345`, or honouring a JSON string literal (`--text '"12345"'`). Both are worse ergonomically, and neither helps a caller who does not know the trap exists, so schema-aware parsing seems clearly preferable.

Worth noting for whoever picks this up: `"true"`/`"false"`/`"null"` are coerced by the same line, so a string param whose legitimate value is the word `true` or `null` has the same problem.

## Impact

Typing digits is a routine automation step — PINs, one-time codes, zip codes, quantities, phone numbers. Today the CLI cannot express any of them for a string-typed parameter, and the error (`expected string, received number`) points at the tool rather than at the transport that mangled the value, so it reads like a schema bug. The MCP path is unaffected; this is CLI-only.

---

*Found during a manual-test iteration over `0.0.44..main`, while exercising `sendSms` / `phoneCall` / `inputText`.*

kaeawc:
## Summary

Every `--cli` value went through `JSON.parse`, so a numeric-looking string arrived as a **number** and string-typed params rejected it. The two examples from the issue both failed:

```
--cli inputText  --text 12345               -> text expected string, received number
--cli sendSms    --phoneNumber 5551234567   -> phoneNumber expected string, received number
```

A phone number, PIN, OTP or zip code is a string whose natural value is digits. The token alone cannot say which the caller meant — only the schema can.

## Linked Issue

Closes #4241

## Bug / Regression Context

- **Observed symptom:** the above. `--phoneNumber +15551234567` worked only because a leading `+` is not valid JSON, and `"007"` survived as a string while `"12345"` did not — behaviour that is hard to predict from outside.
- **Repro:** `--cli inputText --platform android --text 12345`. Quoting does not help; the shell strips quotes and the parser re-interprets the bare token.
- **Prior attempts:** none. `cli/index.ts:87` is long-standing.

## Changes

`parseCliArgs` looks the tool up in `ToolRegistry` and coerces each value by its declared zod type — `string`/`enum` keep the raw token, `number`/`boolean` convert, `object`/`array` parse as JSON, anything unknown falls back to the previous JSON-if-possible behaviour. Enums are treated as strings so a member that happens to look numeric is not coerced.

Two things were needed to make that actually reachable, and both are worth calling out because without them the change is a silent no-op:

1. **`runCliCommand` never registered tools on the execution path.** `initializeCliTools()` was called only on the no-args and `help` branches (`:340`, `:347`), so the registry was empty exactly when a real `--cli <tool>` invocation needed it — every value would have fallen back to the old behaviour and the fix would have done nothing. `getDeclaredParamTypes` now registers first; registration is idempotent (the registry is a `Map` keyed by tool name, verified: 43 tools after one call, 43 after two).

2. **The CLI registered 9 of the 26 categories the server registers.** `sendSms` lives in `telephonyTools`, which was not among them, so the issue's own headline example had no schema to consult. All 26 are now registered.

That second point also fixes `--cli help` under-reporting: it listed **43** tools, now **61**. Execution still routes through the daemon (`runToolViaDaemon` → `proxy.callTool`), so this does not change which tools can actually run — only which the CLI can describe and type. Gated tools such as `barrier`/`criticalSection` remain absent from the daemon's `tools/list` and are still not executable.

Help text updated, since "Values are parsed as JSON if possible" stopped being true.

## Testing

End-to-end against a live daemon and Android emulator — both headline examples now work:

```
--cli sendSms --platform android --phoneNumber 5551234567 --message hi
  -> {"message":"Delivered simulated SMS from 5551234567 (2 chars)","success":true,
      "phoneNumber":"5551234567", ...}          # string, not number

--cli inputText --platform android --text 12345
  -> {"message":"Input text", ...}              # succeeds
```

- `bun test` — **8629 pass, 11 skip, 0 fail** (676 files)
- `bun run build` / `bun run lint` / `bun run typecheck` — all pass
- 9 new tests in `test/cli/parseCliArgsCoercion.test.ts`

## Acceptance criteria

| # | Criterion | Pinned by |
|---|---|---|
| AC1 | `--text 12345` sends the string `"12345"` | `keeps a numeric-looking value as a string…` |
| AC2 | `--phoneNumber 5551234567` succeeds | `keeps an all-digit phone number as a string` |
| AC3 | `--count 5` still yields a number | `still coerces a number-typed param…` |
| AC4 | `--raw true` still yields a boolean | `still coerces a boolean-typed param…` |
| AC5 | `--selector '{"text":"x"}'` still yields an object | `still parses an object-typed param as JSON` |
| AC6 | Unknown tool / undeclared param falls back | `falls back to best-effort parsing…` (×2) |

A nested string inside an object param stays a string, and bare flags with no value still become `true` — both pinned.

## Behaviour change worth noting

`"007"` previously survived as a string only because it is not valid JSON; `"12345"` did not. Both now follow the declared type. Anyone relying on the old accident sees a change — strictly more predictable, but not invisible.

## Follow-up

`--session-uuid` and the daemon flags are parsed elsewhere and keep their own handling; I checked but did not restructure them.

### DIFF
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -146,7 +146,7 @@ function getDeclaredParamTypes(toolName: string): Record<string, string> | null 
   for (const shape of shapes) {
     for (const [key, value] of Object.entries(shape)) {
       const types = byParam.get(key) ?? new Set<string>();
-      types.add(getCliHelpParameterInfo(value).typeName);
+      types.add(resolveDeclaredType(value));
       byParam.set(key, types);
     }
   }
@@ -156,6 +156,52 @@ function getDeclaredParamTypes(toolName: string): Record<string, string> | null 
       .filter(([, types]) => types.size === 1)
       .map(([key, types]) => [key, [...types][0]])
   );
+}
+
+/**
+ * The effective declared type of a parameter, seeing through wrappers.
+ *
+ * `getCliHelpParameterInfo` unwraps `optional` only, so a wrapped param reports
+ * the wrapper (`nullable`, `default`) instead of the type underneath and falls
+ * back to JSON coercion — `setKeyValue --value 12345` sends the number `12345`
+ * even though `storageTools.ts` declares `z.string().nullable()`. Wrappers are
+ * peeled here so the inner type decides the coercion.
+ *
+ * A `literal` resolves to the runtime type of its value, so `z.literal("copy")`
+ * coerces as a string rather than falling through.
+ */
+const SCHEMA_WRAPPER_TYPES = new Set([
+  "optional", "nullable", "default", "readonly", "catch", "branded", "lazy"
+]);
+
+/** The zod type name of a schema, normalized ("ZodString" -> "string"). */
+function schemaTypeName(schema: any): string {
+  const raw = schema?._def?.typeName ?? schema?._def?.type ?? "unknown";
+  return String(raw).replace(/^Zod/, "").toLowerCase();
+}
+
+/** The schema a wrapper wraps; the key varies across zod versions. */
+function unwrapSchema(schema: any): any {
+  const definition = schema?._def;
+  return definition?.innerType ?? definition?.type ?? definition?.schema ?? null;
+}
+
+function resolveDeclaredType(schema: any): string {
+  let current = schema;
+
+  for (let depth = 0; depth < 10 && current; depth++) {
+    const name = schemaTypeName(current);
+
+    if (name === "literal") {
+      return typeof (current._def?.value ?? current._def?.values?.[0]);
+    }
+    if (!SCHEMA_WRAPPER_TYPES.has(name)) {
+      return name;
+    }
+    current = unwrapSchema(current);
+  }
+
+  return "unknown";
 }
 
 /**
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -104,8 +104,17 @@ function tryParseJsonToken(raw: string): { parsed: boolean; value: unknown } {
   }
 }
 
-function coerceCliValue(raw: string, typeName: string | undefined): unknown {
+function coerceCliValue(raw: string, declared: DeclaredType | undefined): unknown {
   const bestEffort = (): unknown => tryParseJsonToken(raw).value;
+
+  // A nullable param must still be able to receive an explicit JSON null --
+  // storageTools removes a key only when `value === null`, so turning the token
+  // into the string "null" would overwrite it instead of clearing it.
+  if (declared?.nullable && raw === "null") {
+    return null;
+  }
+
+  const typeName = declared?.type;
 
   switch (typeName) {
     // Enums are string-valued here (e.g. --action start); treating them as
@@ -160,7 +169,7 @@ function asDeclaredNumber(raw: string): unknown {
  * the old JSON coercion. Registration is idempotent (the registry is a Map keyed
  * by tool name), so calling it per invocation is safe.
  */
-function getDeclaredParamTypes(toolName: string): Record<string, string> | null {
+function getDeclaredParamTypes(toolName: string): Record<string, DeclaredType> | null {
   initializeCliTools();
   const tool = ToolRegistry.getTool(toolName);
   if (!tool) {
@@ -175,19 +184,22 @@ function getDeclaredParamTypes(toolName: string): Record<string, string> | null 
   // A param declared with different types across union branches cannot be
   // coerced unambiguously from the token alone, so it is omitted and falls back
   // to best-effort parsing rather than being coerced to the wrong branch's type.
-  const byParam = new Map<string, Set<string>>();
+  // Nullability is unioned: if any branch accepts null, a JSON null must survive.
+  const byParam = new Map<string, { types: Set<string>; nullable: boolean }>();
   for (const shape of shapes) {
     for (const [key, value] of Object.entries(shape)) {
-      const types = byParam.get(key) ?? new Set<string>();
-      types.add(resolveDeclaredType(value));
-      byParam.set(key, types);
+      const declared = resolveDeclaredType(value);
+      const entry = byParam.get(key) ?? { types: new Set<string>(), nullable: false };
+      entry.types.add(declared.type);
+      entry.nullable = entry.nullable || declared.nullable;
+      byParam.set(key, entry);
     }
   }
 
   return Object.fromEntries(
     [...byParam.entries()]
-      .filter(([, types]) => types.size === 1)
-      .map(([key, types]) => [key, [...types][0]])
+      .filter(([, entry]) => entry.types.size === 1)
+      .map(([key, entry]) => [key, { type: [...entry.types][0], nullable: entry.nullable }])
   );
 }
 
@@ -219,22 +231,32 @@ function unwrapSchema(schema: any): any {
   return definition?.innerType ?? definition?.type ?? definition?.schema ?? null;
 }
 
-function resolveDeclaredType(schema: any): string {
+interface DeclaredType {
+  type: string;
+  /** True when a JSON `null` is a legal value and must survive coercion. */
+  nullable: boolean;
+}
+
+function resolveDeclaredType(schema: any): DeclaredType {
   let current = schema;
+  let nullable = false;
 
   for (let depth = 0; depth < 10 && current; depth++) {
     const name = schemaTypeName(current);
 
+    if (name === "nullable") {
+      nullable = true;
+    }
     if (name === "literal") {
-      return typeof (current._def?.value ?? current._def?.values?.[0]);
+      return { type: typeof (current._def?.value ?? current._def?.values?.[0]), nullable };
     }
     if (!SCHEMA_WRAPPER_TYPES.has(name)) {
-      return name;
+      return { type: name, nullable };
     }
     current = unwrapSchema(current);
   }
 
-  return "unknown";
+  return { type: "unknown", nullable };
 }
 
 /**
--- a/src/cli/index.ts
+++ b/src/cli/index.ts
@@ -133,13 +133,60 @@ function getDeclaredParamTypes(toolName: string): Record<string, string> | null 
   if (!tool) {
     return null;
   }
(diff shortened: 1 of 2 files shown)


kaeawc:
Confirmed and fixed in 7375e7983.

Reproduced before changing anything — `clipboard` resolves to `zodType=union` and `getCliHelpSchemaShape` returns `NULL`, so `parseCliArgs(["clipboard","--action","copy","--text","12345"])` gave `text = 12345` as a number, exactly as described.

Sweeping the registry found three affected tools, not one: `clipboard` (union), `deviceSnapshot` (union), and `postNotification` (a pipe *into* a union). The other 50 resolve fine.

`collectSchemaShapes` now walks union `options` and pipe `out`, and `getDeclaredParamTypes` merges the branch shapes. Where branches declare the same param with **different** types, the param is omitted rather than guessed — it falls back to best-effort parsing instead of being coerced to the wrong branch's type.

`getCliHelpSchemaShape` is deliberately unchanged: it drives `help` output and has its own test (`helpSchema.test.ts`), so only the typing path gained union awareness.

Verified after the change:

```
clipboard.text            = "12345"    string
deviceSnapshot.snapshotName = "20260722" string
postNotification.title    = "911"      string
```

Three tests added. Full suite 8632 pass / 0 fail; build, lint, typecheck green.

kaeawc:
Confirmed and fixed in 0758ff92b.

Reproduced first: `parseCliArgs(["setKeyValue", …, "--value", "12345"])` gave `value = 12345` as a number, because `getCliHelpParameterInfo` unwraps `optional` only and reported the `nullable` wrapper.

I checked the whole wrapper class rather than just `nullable`, to avoid fixing this one shape and leaving siblings. Five params on registered tools use `default` wrappers — `tapOn.action`, `tapAny.action`, `executePlan.startStep`, `executePlan.deviceAllocationTimeoutMs`, `executePlan.abortStrategy`. Those happened to survive `JSON.parse` by accident (`"tap"` throws → string; `"3"` parses → number), but they were relying on the same gap. They are now typed deliberately.

`resolveDeclaredType` peels `optional`/`nullable`/`default`/`readonly`/`catch`/`branded`/`lazy`, and resolves a `literal` to the runtime type of its value so `z.literal("copy")` coerces as a string.

One clarification on the example: `setKeyValue` is `embeddedSdkOnly`, so the registry only exposes it in embedded-SDK mode — which is also the only state where the tool is callable. With that enabled it now resolves correctly:

```
setKeyValue.value      = "12345"  string
clipboard.text         = "12345"  string
tapOn.action           = "tap"    string
executePlan.startStep  = 3        number
```

Three tests added (default-wrapped enum, default-wrapped number, nullable string under embedded-SDK mode). Full suite 8635 pass / 0 fail; build and typecheck green.

Lint initially flagged my first cut at complexity 23 — split into `schemaTypeName` / `unwrapSchema` / `resolveDeclaredType` rather than growing the suppressions baseline.

kaeawc:
Both confirmed and fixed in 1ed306dc3 — these were regressions the schema-aware coercion introduced, and the reproductions matched exactly.

**JSON-encoded strings.** `--text '"12345"'` returned `"\"12345\""`, a token carrying literal quote characters, where the old path yielded `12345`. `asDeclaredString` now unwraps a JSON string and keeps the raw token otherwise:

```
--text '"12345"' -> "12345"      --text '"a b"' -> "a b"
--text 12345     -> "12345"      --text plain   -> "plain"
```

**Non-JSON number tokens.** `Number()` accepted what JSON does not, so a bad argument silently became a valid-looking value instead of being rejected:

```
before: --timeout-ms ""  -> 0     --timeout-ms 0x10 -> 16
after:  --timeout-ms ""  -> ""    --timeout-ms 0x10 -> "0x10"     (schema rejects, as before)
        --timeout-ms 12  -> 12    --timeout-ms 1e3  -> 1000       (still coerced)
```

`asDeclaredNumber` accepts only a finite JSON number and returns the raw token otherwise, preserving the rejection the old path gave.

While fixing these I consolidated the three `JSON.parse` sites into `tryParseJsonToken`, which logs the non-JSON case at debug per the catch convention in CLAUDE.md — a token that is not JSON is ordinary input (`--text plain`, `--action tap`), not an error. That also kept `cli/index.ts` inside its `catch-convention` suppression budget rather than growing the baseline.

10 tests added (25 total in the file). Full suite 8645 pass / 0 fail; build, lint, typecheck green.

kaeawc:
Confirmed and fixed in 42402d6db — and this one was a regression this PR introduced, not a pre-existing gap.

Verified the mechanism end to end: `storageTools.ts:169,176` removes the key only when `args.value === null`, and `--value null` was resolving to the literal string `"null"`, so the key would have been overwritten with that text instead of cleared. The pre-change `JSON.parse` path passed `null` through correctly, so this was a behaviour I broke by stripping the wrapper.

`resolveDeclaredType` now reports nullability alongside the unwrapped type — unioned across union branches, so a param that is nullable in any branch keeps that — and `coerceCliValue` passes a bare `null` token through for nullable params only.

```
setKeyValue --value null      -> null       (clears, as intended)
setKeyValue --value 12345     -> "12345"    (string, unchanged)
setKeyValue --value null-ish  -> "null-ish" (untouched)
inputText   --text  null      -> "null"     (non-nullable: literal string)
```

Four tests added, 29 in the file. Full suite 8649 pass / 0 fail; build, lint, typecheck green.

Worth recording for anyone reading this thread later: across four review rounds on this PR, five of the six findings were correct, and three of those were regressions the schema-aware coercion introduced rather than gaps in the original bug. The change touches every CLI invocation, and the edge cases clustered exactly where a token's textual form and its declared type disagree — JSON-encoded scalars, non-JSON numerics, and explicit nulls. That is a useful shape to remember for the next change in this area.

kaeawc:
Verified, and **deferred to #4247** rather than fixed here — the finding is correct, but it is pre-existing rather than something this PR introduced.

Reproduced it:

```
--embedded-sdk --cli setKeyValue … -> Tool "setKeyValue" is advertised by this
AutoMobile client but the connected daemon does not provide it
```

Two things establish the provenance:

- `src/index.ts:721` (`runCliCommand(cliArgs, { safeAreaWarnings })`) is **byte-identical on `origin/main`**. This PR does not touch it.
- The error originates in `daemonMcpProxy.ts:141`, which compares the requested tool against the **daemon's** advertised list, not the local registry. So registering storage tools client-side neither causes nor worsens it — the same call failed the same way before.

In other words this has never worked, and #4246 does not change that in either direction.

I also found the gap is wider than the report: `DaemonOptions` declares `embeddedSdk`, `networkMockable` **and** `safeAreaWarnings` (`types.ts:188,190,232`), but only the last is threaded — so `--network-mockable` drops too, gating `mockNetwork`/`clearMockNetwork` the same way.

#4247 covers the fix, plus two things worth deciding alongside it: whether a warm daemon running *without* the flag should be reconciled rather than silently reused (as `DaemonMcpProxy` already does for build skew), and that the current error blames "a wrong-build daemon" when the build actually matches and only the *mode* differs.

Keeping it out of this PR: threading daemon startup flags is a different concern from CLI value coercion, and folding it in would mean this change also has to answer the warm-daemon reconciliation question.

Issue #4248, opened 2026-07-22
chore(ci): ratchet down the stale typecheck and shell-sete baseline entries


kaeawc:
## Summary

`bun run typecheck` currently reports ~10 baselined signatures in `scripts/typecheck-baseline.txt` that no longer occur, and `scripts/shellcheck/validate_shell_sete.sh` reports ~8 of its own. Both baselines are one-way ratchets: entries that no longer fire should be pruned so the gate tightens.

## Why this was deliberately deferred

During the July 22 bug-fix batch, five separate agents independently hit this and each declined to act, for the same correct reason: roughly 15 PRs were in flight simultaneously, all touching `src/`, and regenerating a shared baseline file would have collided across them. Several also noted that regenerating from a single platform can delete entries that still fire on another OS, converting them into "new" errors and re-reddening the gate for everyone.

That queue is now clear, so the ratchet is safe to take.

## Acceptance criteria

1. Run `bun run typecheck:update` and commit the smaller `scripts/typecheck-baseline.txt`. The tool refuses to grow the baseline without `-- --allow-grow`; do not pass that flag.
2. Run the shell equivalent (`validate_shell_sete.sh --update` or the documented command) and commit the smaller suppression set.
3. **Verify cross-platform before committing.** Baselines are platform-sensitive: confirm the pruned entries do not still fire on the other runners. `Node TypeScript Build and Test` runs on ubuntu (this is the leg carrying the typecheck gate), macos, and windows; the shell gates run on ubuntu and macos. A pruned entry that still fires elsewhere turns green CI red on merge.
4. State in the PR body how many entries each baseline lost.

## Context

- The typecheck gate's signature matching was recently made union-order-insensitive ([#4211](https://github.com/kaeawc/auto-mobile/issues/4211) via [#4224](https://github.com/kaeawc/auto-mobile/pull/4224)) and then made idempotent ([#4229](https://github.com/kaeawc/auto-mobile/pull/4229)). Regenerating now exercises that fixed path — worth confirming a regenerate/check round-trip is stable, since non-idempotence was exactly #4229's bug.
- Do not fold any unrelated code fix into this PR. It should be baseline files only, so a reviewer can see the ratchet is purely subtractive.

Issue #4234, opened 2026-07-22
fix(ios): migrate the remaining SimCtlClient internal commands off the string path to argv


kaeawc:
## Problem

[#4233](https://github.com/kaeawc/auto-mobile/pull/4233) (fixing [#4196](https://github.com/kaeawc/auto-mobile/issues/4196)) moved every *data-bearing* simctl call site off the
string command path and onto the argv seam `executeCommandArgs`, and deleted the
`quoteSimctlArg` helper that made the pattern easy to copy. A call-site sweep during that PR
found the remaining string-built commands are all inside `SimCtlClient` itself, and several
still interpolate values that can legitimately contain a space or a quote:

- `src/utils/ios-cmdline-tools/SimCtlClient.ts:1136` — `create "${name}" "${deviceType}" "${runtime}"`
  (`name` is caller-supplied)
- `src/utils/ios-cmdline-tools/SimCtlClient.ts:1288` — `install ${targetDevice} "${appPath}"`
- `src/utils/ios-cmdline-tools/SimCtlClient.ts:1379` — `push ${deviceId} ${bundleId} "${file}"`
- plus the fixed-shape ones (`shutdown`, `erase`, `delete`, `launch`, `terminate`, `uninstall`,
  `bootstatus`, `io ... enumerate`, `ui ... appearance`) where the interpolated value is a UDID
  or bundle id

These are **not** a shell-injection risk — `executeCommand` ends at `execFile("xcrun", argv)`,
so no shell ever parses the string. The exposure is the correctness half of [#4196](https://github.com/kaeawc/auto-mobile/issues/4196): a value
containing a double quote or a backslash is still round-tripped through `splitCommandArgs` and
comes back different from what the caller passed. [#4233](https://github.com/kaeawc/auto-mobile/pull/4233) fixed the empty-token drop in
`splitCommandArgs`, so the "silently becomes a different command" failure mode is closed; what
remains is per-character mangling of unusual-but-legal values.

## Definition of done

- The `SimCtlClient` internal call sites above build argv arrays and call `executeCommandArgs`.
- `executeCommand(command: string)` either has no production callers left, or its remaining
  callers are only fixed literal commands (`list devices --json`, `list runtimes --json`), at
  which point it can be narrowed or removed.
- A table-driven test in the style of `test/utils/ios-cmdline-tools/simctlArgvIntegrity.test.ts`
  covers a simulator name and an app path containing a space, a quote, and a backslash,
[...]

Issue #4166, opened 2026-07-22
fix(action): OpenURL dispatches the untrimmed URL despite trimming for validation and logs


kaeawc:
## Summary

`OpenURL` validates, normalizes and logs a **trimmed** URL, then dispatches the **raw,
untrimmed** one to the device.

## Root cause

`src/features/action/OpenURL.ts`

```ts
:79    const trimmedUrl = url.trim();          // computed
:83    if (trimmedUrl.startsWith("package:"))  // used
:92    url: trimmedUrl,                        // used
:133   logger.info(`… ${trimmedUrl}`);         // used
…
:141   this.executeAndroidOpenURL(url)         // <- RAW url dispatched
:145   this.executeiOSOpenURL(url)             // <- RAW url dispatched
```

and the raw value reaches the shell verbatim:

```ts
:169   await this.adb.executeCommand(`shell am start -a android.intent.action.VIEW -d "${url}"`);
```

Every branch between `:79` and `:133` uses `trimmedUrl`; only the two dispatch sites regress
to `url`. The trim is therefore cosmetic — it affects validation and logs but not behavior.

## Impact

`openLink(" https://example.com ")` passes validation, logs the clean URL, and then issues
`am start -d " https://example.com "`. The whitespace reaches the intent. The logs show the
*correct* URL, so the failure is actively misleading during debugging.

Affects both the Android and iOS dispatch paths.

## Acceptance criteria

- [ ] Both dispatch sites pass `trimmedUrl`.
- [ ] Assert the **command string actually issued** (via `FakeAdbExecutor.getExecutedCommands()`),
      not just the returned result object — the return value already looks correct today,
      which is exactly why this went unnoticed.
- [ ] `test.each` table over leading/trailing/both/inner whitespace, tab, newline, and the
      no-whitespace control row.
- [ ] Regression test **red before the fix, green after**.

Note for the implementer: an audit pass proposed a test here using
`FakeAdbExecutor.getAllCommands()` — that method does not exist. The real accessor is
`getExecutedCommands()` (`test/fakes/FakeAdbExecutor.ts:145`).

Found by an automated unit-test-quality audit.

PR #4249, opened 2026-07-22
fix(action): issue openLink URLs as argv instead of interpolating into a shell string

### LINKED ISSUE #4213: fix(action): openLink interpolates the URL into a shell string — quote-breaking URLs are mangled or inject

kaeawc:
## Summary

`openLink` interpolated the caller-supplied URL into a command **string** on both
platforms, wrapped in nothing but a pair of double quotes. Both call sites now issue
**argv arrays**, and the Android one quotes the URL for the shell that actually parses
it.

The two halves are not equally severe, and the fix differs accordingly:

- **Android — real command injection.** The `am start …` string goes to `adb shell`.
  Host-side there is no shell (`execFile`), but adb concatenates its arguments and hands
  the result to `/system/bin/sh` **on the device**, which does full word splitting and
  expansion. Double quotes leave `"`, `$`, backticks and `\` live. Reproduced as a red
  test: `openLink('https://x/" ; id ; echo "')` issued
  `am start … -d "https://x/" ; id ; echo ""` — the device shell parsed `; id ;` as a
  second command and ran it. The URL is now wrapped with `shellQuote()` (POSIX
  single-quoting, `'` → `'\''`), so the whole value is one literal shell word.
- **iOS simulator — mangling, not injection.** `SimCtlClient.executeCommand` ends at
  `execFile("xcrun", argv)` with no shell anywhere, so nothing can be injected. What it
  does do is re-split the string it was handed (`splitCommandArgs`), so a URL containing
  a quote, a space or a backslash reaches the simulator different from what the caller
  passed. Fixed by calling the existing argv seam `executeCommandArgs`, exactly as
  #4233 did for the other simctl call sites.

## Reuse check

No new exec primitive was invented. `AdbClient.execute(args)` and
`SimCtlClient.executeCommandArgs(args)` (standardized in #4233, built on
`createExecResult`) already existed and are reused as-is — `AdbClient` needed **no** new
argv path. The one new file is `src/utils/shellQuote.ts`: device-shell quoting is a
distinct concern from argv construction and the repo had **five** private copies of the
same one-liner (`AppPreferences.ts`, `appFileService.ts`, `DeviceAppManager.ts`,
`AndroidSystemConfigurationAdapter.ts`, `WebpBinaryResolver.ts`) and no shared module.
Rather than add a sixth copy, this promotes one canonical `shellQuote` and uses it here;
migrating the five existing copies is deliberately left to a follow-up so this PR stays
reviewable.

## Linked Issue

Closes #4213

## Bug / Regression Context

- **Observed symptom:** a URL containing `"` was silently mangled before reaching the
  intent (same "logs look right, device sees something else" class as #4166); a URL
  containing `"` plus `;` executed arbitrary commands on the device as the shell user.
- **Repro steps / conditions:** any `openLink` call whose URL contains `"`, `$`,
  a backtick, `;`, `&&`, `|` or a space. `openLink` takes the URL straight from the MCP
  tool argument, so the value is arbitrary and often comes from a page, a deep-link
  table, or a model-generated string.

## Changes

- `src/features/action/OpenURL.ts` — Android dispatch goes through
  `adb.execute(["shell", …])` with the URL `shellQuote`d for the device shell; iOS
  simulator dispatch goes through `simctl.executeCommandArgs(["openurl", udid, url])`.
- `src/utils/shellQuote.ts` — new canonical POSIX single-quote helper, documented with
  *why* `adb shell` needs it when host-side argv does not.
- `test/fakes/FakeAdbExecutor.ts` — implements the `AdbExecutor.execute(args)` member it
  was missing, recording argv verbatim and exposing `getExecutedArgv()` so tests can
  assert argument boundaries rather than a re-joined string. The joined form is still
  appended to the existing command history, so every existing pattern-based stub keeps
  matching.

## Testing

New `test/features/action/openUrlShellInjection.test.ts` — a 14-row table of
hostile-but-legal URLs (control row, embedded `"`, the quote-break `" ; id ; echo "`
payload, `$(…)`, backtick, `$HOME`, `;`, `&&`, `|`/`>`, backslash, inner space, single
quote, unicode, percent-encoding) run against both the Android and the iOS-simulator
dispatch. Assertions are on the **argv actually issued** via
`FakeAdbExecutor.getExecutedArgv()` / `FakeSimCtlClient.getMethodCalls("executeCommandArgs")`,
never the log line or the returned result — both of those looked correct before the fix,
which is why this went unnoticed. Expected quoted forms are written out literally so they
cannot drift with the implementation. One dedicated test pins that the injection payload
yields exactly one single-quoted word (`split("'")` has length 3), so nothing the caller
supplied can become a new shell word.

**Red before, green after:** all 29 new tests failed against `main` (Android: no argv
issued at all; iOS: still on the string path), all 29 pass after.

The existing #4166 whitespace-table assertions in `test/features/action/OpenURL.test.ts`
were updated to the new argv shape — same cases, same coverage, now boundary-aware.

## Validation

- [x] `bun run lint` passes
- [x] `bun test` — 8683 pass / 0 fail
- [x] `bun run typecheck` reports no new errors (baseline gate). It also reports 10
      baseline errors no longer occur — those are the `FakeAdbExecutor` missing-member
      errors this PR fixes. `scripts/typecheck-baseline.txt` is deliberately **not**
      ratcheted here; #4248 owns that file.
- [x] New code uses interfaces + fakes; the new suite runs in ~160ms
- [x] New generic code reuses existing project helpers (see Reuse check)

## Device Verification

Not applicable — this PR is verified entirely with fakes, and deliberately so: the whole
point is asserting the exact bytes handed to the exec seam, which a real device cannot
show you.

## Follow-ups

- Not addressed here, tracked separately: #4234 (the remaining string-built commands
  *inside* `SimCtlClient` — `create`/`install`/`push` and the fixed-shape ones). Those
  are the correctness half only, share no helper with `openLink`, and are explicitly
  scoped to their own issue.
- Consolidating the five private `shellQuote`/`quoteShellArg` copies onto
  `src/utils/shellQuote.ts` — filed as #4255.

### DIFF
--- a/src/features/action/OpenURL.ts
+++ b/src/features/action/OpenURL.ts
@@ -7,6 +7,7 @@ import { DeviceAppManager, DeviceUrlLauncher } from "../../utils/ios-cmdline-too
 import { isIosSimulatorUdid } from "../../utils/ios-cmdline-tools/iosDeviceType";
 import { IOSCtrlProxyManager } from "../../utils/IOSCtrlProxyManager";
 import { logger } from "../../utils/logger";
+import { shellQuote } from "../../utils/shellQuote";
 import { LaunchApp } from "./LaunchApp";
 import { createGlobalPerformanceTracker } from "../../utils/PerformanceTracker";
 
@@ -163,10 +164,18 @@ export class OpenURL extends BaseVisualChange {
    * @returns Result of the URL opening operation
    */
   private async executeAndroidOpenURL(url: string): Promise<OpenURLResult> {
-    // Pass URL through as-is to am start without any reformatting
+    // Pass URL through as-is to am start without any reformatting.
     // Android's Intent system handles both hierarchical (scheme://authority/path)
-    // and opaque (scheme:scheme-specific-part) URIs correctly
-    await this.adb.executeCommand(`shell am start -a android.intent.action.VIEW -d "${url}"`);
+    // and opaque (scheme:scheme-specific-part) URIs correctly.
+    //
+    // The `am start …` line is issued as ONE argv element, so no host shell ever
+    // sees it; adb hands that element to the device's `sh`, which does parse it,
+    // so the URL is single-quoted for that shell (issue #4213). Double quotes
+    // would leave `"`, `$`, backticks and `\` live.
+    await this.adb.execute([
+      "shell",
+      `am start -a android.intent.action.VIEW -d ${shellQuote(url)}`,
+    ]);
 
     return {
       success: true,
@@ -197,8 +206,10 @@ export class OpenURL extends BaseVisualChange {
   private async executeiOSSimulatorOpenURL(url: string): Promise<OpenURLResult> {
     try {
       const simctl = this.simctl ?? new SimCtlClient();
-      // Use simctl openurl command: xcrun simctl openurl <device> <url>
-      await simctl.executeCommand(`openurl ${this.device.deviceId} "${url}"`);
+      // xcrun simctl openurl <device> <url>, issued as argv so the URL reaches
+      // execFile byte-for-byte. The string path re-splits its command, which
+      // mangles quotes and backslashes (issue #4213 / #4196).
+      await simctl.executeCommandArgs(["openurl", this.device.deviceId, url]);
 
       return {
         success: true,
--- a/src/utils/shellQuote.ts
+++ b/src/utils/shellQuote.ts
@@ -1,0 +1,24 @@
+/**
+ * POSIX shell quoting for values that must survive a shell that is NOT ours.
+ *
+ * Host-side execution in this repo goes through `execFile`/`spawn` with an argv
+ * array, so no host shell is involved. But `adb shell <cmd>` is different: adb
+ * concatenates its remaining arguments with spaces and hands the result to
+ * `/system/bin/sh` ON THE DEVICE, which then does full word splitting and
+ * expansion. Any caller-supplied value embedded in such a command must therefore
+ * be quoted for that device shell.
+ *
+ * Double quotes are not enough — `$`, backticks and `\` stay live inside them.
+ * Single quotes are the only POSIX construct that suppresses every expansion.
+ */
+
+/**
+ * Wrap `value` in single quotes so a POSIX shell sees it as exactly one literal
+ * word. A single quote cannot appear inside a single-quoted string, so each one
+ * is emitted as `'\''` — close, escaped literal quote, reopen.
+ * @param value - Raw value to quote
+ * @returns The value as a single, fully literal shell word
+ */
+export function shellQuote(value: string): string {
+  return `'${value.replace(/'/g, "'\\''")}'`;
+}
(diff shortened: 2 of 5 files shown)


kaeawc:
Manual-test verification on HEAD `44f83116b`, Android emulator API 35 — **confirmed fixed**.

`openLink` with `"  https://example.com/?a=1&b=2  "` (leading/trailing whitespace plus an unescaped `&`) delivered, per `dumpsys activity activities`:

```
act=android.intent.action.VIEW dat=https://example.com/?a=1&b=2 cmp=com.android.chrome/...ChromeTabbedActivity
```

The `&` survived intact — shell interpolation would have truncated at `?a=1` — and the URL is trimmed (also covers #4204).

Nit, cosmetic only: the response `message` echoes the raw untrimmed input (`"Opened link   https://example.com/?a=1&b=2  "`) while the dispatched value is correctly trimmed.

PR #4252, opened 2026-07-22
fix(setUIState): bound the scroll search so its own error surfaces before the request timeout

### LINKED ISSUE #4242: setUIState hangs to the 30s request timeout on an unmatched selector, and reports "Unknown field type: unknown" when it matches a label
## Summary

Two problems with `setUIState`, both reproducible on an Android emulator:

1. **A selector that matches nothing hangs until the caller's request timeout (~30 s)** instead of failing fast. `tapOn` with an equally-bogus selector fails in ~2 s.
2. **When the selector matches a non-editable node, the error is `Unknown field type: unknown`** — which does not tell the caller that the problem is their selector, or what to select instead.

## Evidence

`origin/main` @ 92cd3baa5, Android emulator (`Medium_Phone_API_35`, API 35), playground app.

### Slow failure vs the fast-failing equivalent

```
# no "Email" anywhere on the current screen (onboarding)
setUIState --fields '[{"selector":{"text":"Email"},"value":"..."}]'
  -> exit 1 after 30s: "MCP error -32001: Request timed out"     (reproduced twice)

# same situation, different tool
tapOn --selector '{"text":"NoSuchElementXYZ"}'
  -> exit 1 after 2s: "Failed to perform tap on element: Element not found with provided text 'NoSuchElementXYZ'"
```

`tapOn` returns a typed, actionable error quickly. `setUIState` consumes the whole request budget and surfaces only a generic transport timeout, which does not identify the tool, the field, or the reason.

### Unhelpful error on a non-editable match

On the login screen, where `Email` *is* present:

```
setUIState --fields '[{"selector":{"text":"Email"},"value":"test@example.com"}]'
  -> exit 1 after 6s
  {"success": false,
   "fields": [{"selector": {"text": "Email"}, "success": false,
               "attempts": 3, "error": "Unknown field type: unknown",
               "fieldType": "unknown"}],
   "totalAttempts": 3}
```

The matched node is the field's **label**, not its input:

```json
{"text": "Email", "view-id": "s-3d76c79e0f0c6c1d",
 "bounds": {"left": 84, "top": 881, "right": 195, "bottom": 944}}
```

No class, no editable attributes — so classification failing is understandable. But `Unknown field type: unknown` describes an internal state rather than the actionable fact: *the selector matched a non-editable node (a label); select the input instead*. It also retried 3 times before giving up on something that could not have succeeded.

## Suggested fix

1. **Bound the not-found path.** Give the field search its own timeout well inside the request budget and return a typed `element not found` error naming the failing selector, the way `tapOn` already does. A caller currently cannot distinguish "your selector is wrong" from "the daemon is wedged".
2. **Make the classification error actionable.** When a node matches but is not editable, say so and include what was matched (class / view-id / bounds), e.g. `selector {"text":"Email"} matched a non-editable node (no editable class); select the input field rather than its label`.
3. **Do not retry an unclassifiable node.** `attempts: 3` on a node whose type is `unknown` spends time on an outcome that cannot change unless the UI itself changes.

## Impact

The 30 s hang is the practical problem: a plausible-but-wrong selector stalls an automation run and reports a transport timeout rather than a tool-level cause, which sends debugging in the wrong direction. Both behaviours make `setUIState` noticeably harder to use than the sibling interaction tools.

---

*Found during a manual-test iteration over `0.0.44..main`.*

kaeawc:
## Summary

`setUIState` was not hanging — it was **scroll-searching**. When no field matches it swipes, re-observes and repeats, `MAX_FUTILE_SCROLLS` in each direction before reversing: roughly **eight device round trips**. That lands at about the caller's request timeout, so the caller saw a transport timeout instead of the tool's own error.

The tool already produced the right message (`Fields not found after scrolling: …`, `SetUIState.ts:210`) — it just finished at the same moment the request budget expired, so it never reached the caller.

## Linked Issue

Closes #4242

## Bug / Regression Context

- **Observed symptom:** `setUIState` with a selector that matches nothing → `exit 1 after 30s: MCP error -32001: Request timed out` (reproduced twice). `tapOn` with an equally bogus selector fails in ~2s.
- **Repro:** `setUIState --fields '[{"selector":{"text":"Email"},"value":"x"}]'` on a screen with no `Email`.
- **Correction to the issue:** #4242 said this should "fail fast like `tapOn`". That is wrong — `tapOn` does not scroll-search, and making `setUIState` fail fast would remove its ability to reach **off-screen** fields. The fix is to bound the search so its own error surfaces, not to delete the search.

## Changes

- **`SEARCH_BUDGET_MS = 20_000`** — the scroll loop checks a deadline each iteration and stops, reporting `Fields not found within the 20s search budget: …`. Time comes from the injected `Timer`, so it is testable without real waiting.
- **`Unknown field type: unknown` replaced.** That string describes an internal state; the message now names what was matched and says to select the input rather than its label, which is the usual cause (a selector matching a field's *label* — no class, no editable attributes).
- **No retries on an unclassifiable node.** The type comes from the element itself, so the remaining two attempts would fail identically and only cost round trips.

## Testing

- `bun test` — **8630 pass, 11 skip, 0 fail** (675 files)
- `bun run build` / `bun run lint` / `bun run typecheck` — all pass
- 5 new tests in `test/features/action/SetUIState.test.ts`

The budget test simulates real round-trip cost by advancing `FakeTimer` inside the observe fake — fakes are instant, so without that the unit layer cannot reproduce a wall-clock timeout at all. That is precisely why this escaped existing coverage.

## Acceptance criteria

| # | Criterion | Pinned by |
|---|---|---|
| AC1 | An unmatched selector returns the tool's own error | `an unmatched selector returns the tool's own error…` |
| AC2 | The search is bounded in time | `the search stops once its time budget is spent…` |
| AC3 | The search is bounded in scrolls | `the search is bounded — it does not scroll indefinitely` |
| AC4 | A non-editable match says what it matched | `a matched but unclassifiable node reports what it matched…` |
| AC5 | No retry on an unclassifiable node | `an unclassifiable node is not retried` |

## Follow-up

Whether `MAX_FUTILE_SCROLLS = 3` in **both** directions is the right search shape is a separate question — this PR only stops it from outliving the request.

## Note on the budget

20s is chosen to sit inside the ~30s request timeout with room for the initial observe and the per-field work that follows a successful find. A field that today would be found on a very late scroll under a slow device could now be missed; the error names the budget so that case is diagnosable rather than mysterious.

### DIFF
--- a/src/features/action/SetUIState.ts
+++ b/src/features/action/SetUIState.ts
@@ -129,14 +129,10 @@ export class SetUIState extends BaseVisualChange {
     let scrollsWithoutProgress = 0;
     let currentDirection: "up" | "down" = scrollDirection;
     let triedReverse = false;
-    const searchDeadline = this.timer.now() + SEARCH_BUDGET_MS;
+    let searchDeadline = this.timer.now() + SEARCH_BUDGET_MS;
     let budgetSpent = false;
 
     while (processed.size < options.fields.length) {
-      if (this.timer.now() >= searchDeadline) {
-        budgetSpent = true;
-        break;
-      }
       // Find all unprocessed fields visible in the current hierarchy, sorted by bounds.top
       const visibleFields = this.findVisibleFieldsInScreenOrder(
         options.fields,
@@ -161,6 +157,10 @@ export class SetUIState extends BaseVisualChange {
         fieldResults[fieldIndex] = result;
         processed.add(fieldIndex);
         totalAttempts += result.attempts;
+        // The budget bounds futile searching, not successful work: a form with
+        // many fields legitimately takes longer than one scroll search, and the
+        // next field may already be on screen (#4252 review).
+        searchDeadline = this.timer.now() + SEARCH_BUDGET_MS;
 
         // Refresh observation after each success
         if (result.success) {
@@ -183,6 +183,11 @@ export class SetUIState extends BaseVisualChange {
         }
       } else {
         // No visible matches — scroll to find more
+        if (this.timer.now() >= searchDeadline) {
+          budgetSpent = true;
+          break;
+        }
+
         scrollsWithoutProgress++;
 
         if (scrollsWithoutProgress > MAX_FUTILE_SCROLLS) {
--- a/src/features/action/SetUIState.ts
+++ b/src/features/action/SetUIState.ts
@@ -66,6 +66,17 @@ interface SetUIStateDependencies {
 const DEFAULT_MAX_RETRIES = 3;
 const DEFAULT_SCROLL_DIRECTION = "down";
 const MAX_FUTILE_SCROLLS = 3;
+/**
+ * Wall-clock ceiling for the scroll search.
+ *
+ * Each futile scroll costs a swipe plus an observe against the device, and the
+ * loop tries MAX_FUTILE_SCROLLS in each direction — roughly eight round trips,
+ * which lands at about the caller's request timeout. The caller then sees a
+ * transport timeout instead of this tool's own "Fields not found" error, which
+ * points at the wrong thing (#4242). Stopping short of that keeps the actionable
+ * error reachable.
+ */
+const SEARCH_BUDGET_MS = 20_000;
 
 /**
  * SetUIState - Declarative form field population tool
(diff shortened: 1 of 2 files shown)


kaeawc:
Confirmed and fixed in the pushed commit — a regression this PR introduced, and the reasoning was exactly right.

I put the deadline at the top of the loop, so time spent **successfully setting earlier fields** counted against a budget that was only ever meant to bound futile scrolling. Reproduced it: a two-field form, both visible the whole time, with a slow first field — the second was reported `Fields not found within the 20s search budget` despite being on screen.

Two changes:

- the deadline check moved into the no-visible-match (scroll) branch, so it only ever bounds searching;
- the deadline resets whenever a field is processed, so the budget measures **time without progress** rather than total execution.

That is what a search ceiling should mean — a long form is legitimately slower than a single scroll search, and the original defect (#4242) was only ever about futile scrolling outliving the request.

Test added: `a visible field is still set even after earlier fields consumed the budget`. It advances `FakeTimer` by 9s per observe so two fields alone exceed the 20s budget; before the fix it failed with "not found", now both are set.

23 tests in the file, full suite 8631 pass / 0 fail; build, lint, typecheck green.

PR #4271, opened 2026-07-22
fix(observe): budget the hierarchy fresh wait and re-read the iOS stale fallback

### LINKED ISSUE #4261: CtrlProxy hierarchy: fresh-wait ignores elapsed budget (Android) and stale fallback returns a pre-await snapshot (iOS)
## Problem

Two unresolved codex findings on merged PRs, both in the CtrlProxy hierarchy delegates.

### 1. Android: fresh-wait is not budgeted against the remaining deadline

`src/features/observe/android/CtrlProxyHierarchy.ts:310`

```ts
const available = await perf.track("checkAvailable", () => ... isAvailable());
...
const freshWaitMs = timeoutMs === undefined
  ? DEFAULT_FRESH_WAIT_MS
  : Math.max(0, Math.min(DEFAULT_FRESH_WAIT_MS, timeoutMs));
```

`timeoutMs` is the caller's **overall** budget (see the doc comment: it exists so the keyboard-state confirmation poll cannot be blocked past its own deadline). But `freshWaitMs` is derived from the original `timeoutMs`, not from what is left of it. When the read starts from a slow availability check or a reconnect, the subsequent `getLatestHierarchy` wait can still consume another full `DEFAULT_FRESH_WAIT_MS` on top of the time already spent. Only the sync fallback (`syncTimeoutMs`, line 329) subtracts elapsed time.

From codex review of merged PR [#4239](https://github.com/kaeawc/auto-mobile/pull/4239).

### 2. iOS: the stale fallback returns a pre-await snapshot

`src/features/observe/ios/CtrlProxyHierarchy.ts:134`

`cachedHierarchy` is captured at the top of `getLatestHierarchy`, before the awaited `requestHierarchySync`. When an invalidated cache takes the `skipWaitForFresh` override and that sync request times out or fails, an unsolicited `hierarchy_update` push that arrived meanwhile (`IOSCtrlProxyClient.processMessage` replaces the cache) is discarded — the stale fallback still returns the pre-await, explicitly-invalidated snapshot.

Fix is to re-read the cache after the await. Clearing the cache instead is wrong on iOS: under `skipWaitForFresh` a missing cache yields no hierarchy at all rather than a refetch, because the delegate never fetches on a cache miss there. The [#4230](https://github.com/kaeawc/auto-mobile/pull/4230) invariants stand — no writer stores `fresh: false` except `invalidateCache()`, and `collectRaw` was the victim, not a caller written around the bug.

From codex review of merged PR [#4230](https://github.com/kaeawc/auto-mobile/pull/4230).

## Acceptance criteria

1. Android: the fresh wait is bounded by `timeoutMs` minus elapsed time at the point the wait starts, floored at 0.
2. Android: a unit test with `FakeTimer` where the availability check consumes part of the budget shows the fresh wait shrinking accordingly (red before, green after).
3. iOS: the stale fallback returns the cache as it stands **after** the awaited sync, so a push that landed during the await wins over the invalidated snapshot.
4. iOS: a unit test drives a failing sync concurrent with a cache-replacing push and asserts the returned hierarchy is the pushed one (red before, green after).
5. No change to the `#4230` invalidation semantics; the existing `ctrlProxyHierarchyCache.test.ts` suite stays green.

kaeawc:
## Summary

Two unresolved codex findings on merged PRs, both hierarchy-read budget/staleness bugs in the CtrlProxy delegates.

**Android ([#4239](https://github.com/kaeawc/auto-mobile/pull/4239)) — the fresh wait was not charged against the caller's deadline.** `timeoutMs` is documented as the *overall* budget for the read; it exists so a caller polling against its own deadline (keyboard-state confirmation) cannot be blocked past it. But `freshWaitMs` was derived from the original `timeoutMs`, so a slow availability check or a reconnect could be followed by another full `DEFAULT_FRESH_WAIT_MS` wait on top of the time already spent. Only the sync fallback subtracted elapsed time. The wait now gets what is left of the budget.

**iOS ([#4230](https://github.com/kaeawc/auto-mobile/pull/4230)) — the stale fallback returned a pre-await snapshot.** `cachedHierarchy` is captured at entry. When an invalidated entry takes the `skipWaitForFresh` override and the awaited sync then fails, an unsolicited `hierarchy_update` push that `IOSCtrlProxyClient.processMessage` wrote during the await was discarded: the fallback still served the explicitly-invalidated snapshot. The fallback now re-reads the cache after the await.

The iOS change **re-reads only** — it never nulls the cache. Under `skipWaitForFresh` a missing cache yields no hierarchy at all rather than a refetch, because the delegate does not fetch on a cache miss there. The [#4230](https://github.com/kaeawc/auto-mobile/pull/4230) invalidation semantics are untouched and `ctrlProxyHierarchyCache.test.ts` stays green.

## Linked Issue

Closes #4261

## Bug / Regression Context

- **Prior attempts:** [#4239](https://github.com/kaeawc/auto-mobile/pull/4239) introduced the bounded read; [#4230](https://github.com/kaeawc/auto-mobile/pull/4230) fixed `invalidateCache()` being an observable no-op. Both left unresolved codex feedback, filed as #4261.
- **Observed symptom:** Android — with a 1000ms budget and an availability check costing 800ms, `getLatestHierarchy` was still handed a 1000ms fresh wait. iOS — a push landing during a failing sync was thrown away in favour of the invalidated entry.
- **Repro:** both new test files are red against unmodified `main` (see below).

## Validation

- [x] Red-before / green-after verified per fix, by reverting each source file individually:
  - unmodified source: Android 2 fail / 2 pass, iOS 1 fail / 1 pass
  - Android fix reverted (iOS fix applied): Android suite 2 fail
  - iOS fix reverted (Android fix applied): iOS suite 1 fail
  - both applied: 6 pass / 0 fail
- [x] `bun test` — 8720 pass, 0 fail (681 files)
- [x] `bun run lint`
- [x] `bun run typecheck` — no new errors (baseline untouched)
- [x] New tests use interface + fakes + `FakeTimer`; both files run in single-digit ms
- No device, emulator, or simulator interaction.

## Notes

The Android test drives the elapsed time through a `FakeTimer` advanced inside a stubbed `isAvailable()`, so the budget arithmetic is exercised without any real availability probe. The iOS test resolves the sync request with no hierarchy while replacing the cache in the same `send` — the write lands after the pre-await capture and before the await resumes, which is exactly the race being fixed. A second iOS test pins the no-push case so the fallback cannot be "fixed" by dropping it.

### DIFF
--- a/src/features/observe/android/CtrlProxyHierarchy.ts
+++ b/src/features/observe/android/CtrlProxyHierarchy.ts
@@ -307,9 +307,13 @@ export class CtrlProxyHierarchy {
 
       // Get hierarchy from WebSocket service
       const waitForFresh = !skipWaitForFresh && (cachedHierarchy === null || !cachedHierarchy.fresh);
+      // `timeoutMs` is the caller's overall budget, not a per-step allowance, so
+      // the wait gets what is LEFT of it. Starting from the original value would
+      // let a slow availability check or reconnect be followed by another full
+      // fresh wait, blowing the deadline this parameter exists to protect.
       const freshWaitMs = timeoutMs === undefined
         ? DEFAULT_FRESH_WAIT_MS
-        : Math.max(0, Math.min(DEFAULT_FRESH_WAIT_MS, timeoutMs));
+        : Math.max(0, Math.min(DEFAULT_FRESH_WAIT_MS, timeoutMs - (this.context.timer.now() - startTime)));
       const response = await perf.track("getHierarchy", () =>
         this.getLatestHierarchy(waitForFresh, freshWaitMs, perf, skipWaitForFresh, minTimestamp, signal)
       );
--- a/src/features/observe/ios/CtrlProxyHierarchy.ts
+++ b/src/features/observe/ios/CtrlProxyHierarchy.ts
@@ -149,20 +149,29 @@ export class CtrlProxyHierarchy {
 
     const reconnectStatus = this.context.getReconnectStatus?.() ?? undefined;
 
+    // Re-read the cache: `cachedHierarchy` was captured before the awaited sync,
+    // and an unsolicited hierarchy_update push handled by
+    // IOSCtrlProxyClient.processMessage can have replaced it while that request
+    // was in flight. Serving the pre-await snapshot would hand back the very
+    // entry an invalidation was meant to retire even though newer data arrived.
+    // Note this only re-reads; it must never null the cache, because under
+    // skipWaitForFresh a missing cache yields no hierarchy at all (#4193/#4230).
+    const fallbackHierarchy = this.context.getCachedHierarchy() ?? cachedHierarchy;
+
     // Return cached (stale) data if available
-    if (cachedHierarchy) {
+    if (fallbackHierarchy) {
       // Update tracking from cache — it may have been refreshed by a WebSocket push
-      if (cachedHierarchy.hierarchy.packageName) {
-        if (this.lastKnownPackageName && cachedHierarchy.hierarchy.packageName !== this.lastKnownPackageName) {
-          logger.warn(`[CTRL_PROXY] Stale cache packageName differs: cached=${cachedHierarchy.hierarchy.packageName}, lastKnown=${this.lastKnownPackageName}`);
+      if (fallbackHierarchy.hierarchy.packageName) {
+        if (this.lastKnownPackageName && fallbackHierarchy.hierarchy.packageName !== this.lastKnownPackageName) {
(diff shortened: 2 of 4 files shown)


kaeawc:
Manual-test verification on HEAD `44f83116b`, iOS simulator (iOS 18.x) — **confirmed delivered**.

Three timed `observe` calls: 1.23s / 1.60s / 1.48s (the cold first call was 8.4s including interpreter start). `viewHierarchy` was 2326–10077 bytes and **never empty** — the documented post-cold-boot empty-hierarchy retry was not needed once.

Issue #4223, opened 2026-07-22
test(observe): capture real dumpsys fixtures and teach parseTasks the modern Task{...} header


kaeawc:
Follow-up from [#4197](https://github.com/kaeawc/auto-mobile/issues/4197) / [PR #4222](https://github.com/kaeawc/auto-mobile/pull/4222).

#4197 fixed `parseActivities` in `src/features/observe/GetBackStack.ts` — the activity regex expected
the component *after* the closing brace of `ActivityRecord{...}` when AOSP puts it *inside*. That fix
was provable from format samples already committed to this repo, so it landed without a device.

Two things were deliberately left out of that PR because they cannot be settled without real captures:

**1. `parseTasks` does not recognize the modern task-header line.**

`GetBackStack.parseTasks` matches only:

```
/Task\s+id\s+#(\d+)|TaskRecord.*#(\d+)/
```

Newer Android prints the header as:

```
* Task{d19dee2 #61 type=standard A=10164:com.example U=0 visible=true mode=fullscreen sz=2}
```

Neither alternative matches that (`TaskRecord` was renamed `Task`), so on those API levels
`result.tasks` is likely empty and `affinity` / `realActivity` / `numActivities` never get attached.
`depth` and `currentTaskId` are unaffected — after #4197 both derive from the activity lines, which
carry their own `tNN` — so this is a metadata gap, not a correctness regression.

This was **not** fixed in #4197 because no capture of the `Task{...}` header is committed to this
repo and no device was available to confirm the exact shape (field order, whether the `#id` is always
the second token). Fixing it blind risks pinning a guess.

**2. Capture real fixtures, as the original issue asked.**

#4197 asked for `dumpsys activity activities` captured from a current emulator **and** a physical
device on at least two API levels, committed as fixtures. The parsing suite added in
`test/features/observe/GetBackStackParsing.test.ts` uses hand-written fixtures derived from the
format samples already in the repo. Real captures would let us confirm the `Task{...}` shape above
and pin per-API-level differences.

## Work

- [ ] Capture `dumpsys activity activities` from an emulator and a physical device, on two API levels
- [ ] Commit them as fixtures and extend `test/features/observe/GetBackStackParsing.test.ts`
- [ ] Teach `parseTasks` the `Task{<hash> #id ...}` header form, driven by those captures
- [ ] Confirm whether `affinity=` still appears as its own line on modern output, or whether
      `A=<uid>:<pkg>` inside the `Task{...}` header is now the only source

Issue #4197, opened 2026-07-22
investigate(observe): GetBackStack dumpsys parser returns zero activities against real output


kaeawc:
## Defect

The `dumpsys activity activities` parser in `src/features/observe/GetBackStack.ts:51` does
not match real `dumpsys` output. Against a real capture it yields:

```
{ depth: 0, activities: [], tasks: [{ id: 123 }] }
```

`parseTasks` finds the task; `parseActivities` returns empty; `depth` is therefore always 0.
The audit's assessment is that **the regex has never matched** current output.

## Status — needs a decision, not just a fix

I have **not** reproduced this myself: it needs a real device capture, which I do not have.
The audit's critic reproduced it, but also flagged that the proposed regression test "pins
backwards semantics" — its `isTaskRoot` row treats the topmost activity as the task root,
which is inverted.

So this issue is filed as **investigate + decide**, not as a known fix.

## Work

- [ ] Capture real `dumpsys activity activities` output from a current emulator **and** a
      physical device, on at least two API levels. Attach them as fixtures.
- [ ] Establish the actual back-stack semantics: which end of the list is the task root, and
      what `depth` is meant to count. Write it down in the test as prose before coding.
- [ ] Fix the parser against the captured fixtures.
- [ ] Table-driven tests over the captured fixtures plus degenerate rows: no tasks, one task
      no activities, multiple tasks, malformed line mid-block.
- [ ] Red before the fix, green after.

If it turns out the parser does work against some output shape we still see in the wild,
close this as not-a-bug with the capture that proves it.

Source: `w3-critique-observe-core` A1 (rank 10 of the audit's top-25).

PR #4273, opened 2026-07-22
fix(observe): anchor legacy task headers and source task root from Hist #0

### LINKED ISSUE #4263: GetBackStack: legacy inline TaskRecord refs read as headers; modern A= tasks lack packageName/rootActivity
Follow-up to [PR #4253](https://github.com/kaeawc/auto-mobile/pull/4253), which taught `GetBackStack` the modern `Task{...}` header. Two P2 review findings were left unresolved on that PR. Both are in `src/features/observe/GetBackStack.ts`.

## Ground truth used

The 13 committed captures under `test/features/observe/windowDumps/` are `dumpsys window windows` output. They contain **zero** task headers (`Task{`, `TaskRecord`, `Task id #`), zero `Hist #N` rows, and zero `A=`/`I=`/`aI=`/`realActivity=` tokens - verified by grep. They cannot reproduce either finding, and the previous agent said so explicitly rather than pretending otherwise.

The finding-1 shapes below ARE grounded in real `dumpsys activity activities` output - the capture linked from the review comment (`gist.github.com/ganadist/0102b683fb8a873feee7f0568b9a0382`, an Android 9-era device). The modern `Task{...}` shapes remain AOSP-sourced and **unverified against a real capture**; that is unchanged from #4253 and is not something this issue can fix.

## Finding 1 - legacy inline `TaskRecord{...}` references are misread as task headers

`LEGACY_TASK_HEADER = /Task\s+id\s+#(\d+)|TaskRecord.*#(\d+)/` is unanchored. In the real capture, two non-header lines match it:

```
          frontOfTask=false task=TaskRecord{1d3813b #14418 A=com.facebook.katana U=0 StackId=436 sz=2}
```

(inside an activity detail block, **between** two `Hist #N` rows), and

```
    Running activities (most recent first):
      TaskRecord{1d3813b #14418 A=com.facebook.katana U=0 StackId=436 sz=2}
```

Consequences:

- `parseActivities` - #4253 added `currentTaskAffinity = header.affinity`, and legacy headers never populate `affinity`, so the inline ref between `Hist #1` and `Hist #0` clears the affinity. `Hist #1` keeps `taskAffinity`, `Hist #0` loses it.
- `parseTasks` - the header branch calls `flush()` and then replaces `currentTask` with a bare `{ id }`. The `Running activities` repeat therefore **overwrites** the fully-parsed task in the map with an empty one, discarding `affinity`, `packageName`, `rootActivity` and `numActivities`.

Also note `TaskRecord.*#(\d+)` is greedy: it captures the *last* `#`-number on the line, not the task id.

Real headers are line-anchored, printed by `ActivityStack.dumpActivitiesLocked` as `"    * " + task`:

```
    Task id #14418
    * TaskRecord{1d3813b #14418 A=com.facebook.katana U=0 StackId=436 sz=2}
```

## Finding 2 - modern `A=` tasks have no `packageName`/`rootActivity`

For the common modern app-task header `* Task{<hash> #61 type=standard A=10164:com.example ...}` the parser records only `affinity`. `header.component` is unset (that comes from `I=`/`aI=`), and the modern path has no `realActivity=` line, so `observe` returns `TaskInfo` entries with `packageName`/`rootActivity` undefined for ordinary app tasks.

**`A=` is not a usable source for `packageName`.** It is `Task.affinity`, which (a) is uid-prefixed on Android 11+ (`10164:com.example`, per `ActivityRecord.computeTaskAffinity`), (b) is an app-declarable `android:taskAffinity` string that need not be a package name, and (c) is empty for `taskAffinity=""`. Stripping a `uid:` prefix would be a guess, not a derivation.

The honest source is the task's own activity rows, which the modern path does print:

```
    * Hist  #0: ActivityRecord{7ff01aa u0 com.example/com.example.MainActivity t61}
```

`Hist #0` is the task root (pinned by #4197), and its component has the same `pkg/cls` shape as legacy `realActivity=`.

## Acceptance criteria

1. An inline `task=TaskRecord{...}` reference inside an activity detail block is not treated as a task header, on either the activity path or the task path.
2. A repeated `TaskRecord{...}` line in the `Running activities` section does not overwrite the already-parsed task with an empty one.
3. The legacy task id is read from the token immediately after the identity hash, not from the last `#`-number on the line.
4. Legacy `TaskRecord{... A=/I=/aI=...}` headers populate `affinity`/`component` the same way modern headers do.
5. A modern `A=` task gets `packageName` and `rootActivity` from its `Hist #0` activity row, as a fallback that never overrides `I=`/`aI=` or a legacy `realActivity=` line.
6. A task with no printed `Hist #0` row leaves `packageName`/`rootActivity` undefined, and that is documented rather than guessed at.
7. Existing `Task id #`, `TaskRecord{}`, modern `Task{}`, `TaskFragment{}` and inline-`Task{}`-reference behaviour from #4253 stays green.

kaeawc:
Closes [#4263](https://github.com/kaeawc/auto-mobile/issues/4263).

Follow-up to [PR #4253](https://github.com/kaeawc/auto-mobile/pull/4253), addressing the two P2 review findings left unresolved there. Both are in `src/features/observe/GetBackStack.ts`.

## Provenance, stated up front

This parser has been wrong three times, each time by being too tight or too loose about brace/token shapes, so it is worth being explicit about what is evidence and what is not.

The 13 committed captures under `test/features/observe/windowDumps/` are `dumpsys window windows` output. Grep says they contain zero `Task{`, zero `TaskRecord`, zero `Task id #`, zero `Hist #N` and zero `A=`/`I=`/`aI=`/`realActivity=` tokens. They cannot exercise task-header parsing at all. They remain the ground truth for the `ActivityRecord{...}` shapes (including the API 33 early-brace form), and that regex is unchanged here.

The legacy shapes fixed below come from a real `dumpsys activity activities` capture of an Android 9-era device - the one linked from the review comment (`gist.github.com/ganadist/0102b683fb8a873feee7f0568b9a0382`). The relevant excerpt is committed verbatim as `LEGACY_REAL_CAPTURE` in the new test file.

The modern `Task{...}` shapes are still AOSP-sourced and **unverified against any real capture**. That is unchanged from #4253 and this PR does not fix it.

## Finding 1 - inline `TaskRecord{...}` refs read as task headers

`LEGACY_TASK_HEADER` was `/Task\s+id\s+#(\d+)|TaskRecord.*#(\d+)/`, unanchored. Two non-header lines in the real capture match it:

```
      * Hist #1: ActivityRecord{7c63edf u0 com.facebook.katana/.activity.FbMainTabActivity t14418}
          frontOfTask=false task=TaskRecord{1d3813b #14418 A=com.facebook.katana U=0 StackId=436 sz=2}
      * Hist #0: ActivityRecord{7f67c32 u0 com.facebook.katana/.activity.FbMainTabActivity t14418}

    Running activities (most recent first):
      TaskRecord{1d3813b #14418 A=com.facebook.katana U=0 StackId=436 sz=2}
```

Observed against unfixed code:

- `parseActivities` - the `frontOfTask=` line sits between `Hist #1` and `Hist #0`, and #4253's `currentTaskAffinity = header.affinity` reset assigns `undefined` because legacy headers carry no affinity. `Hist #1` keeps `taskAffinity`, `Hist #0` comes back `undefined`.
- `parseTasks` - the `Running activities` repeat calls `flush()` and then replaces `currentTask` with a bare `{ id }`, so the finished task is **overwritten by an empty one**, losing `affinity`, `packageName`, `rootActivity` and `numActivities`.
- `TaskRecord.*#(\d+)` is greedy, so it captured the last `#`-number on the line rather than the task id.

Fixes:

- The regex is line-anchored and reads the id from the token right after the identity hash. That alone rejects the `frontOfTask=... task=` ref, which is never at line start.
- The `Running activities` repeat **is** at line start, differing only by the missing `* ` bullet - and the bullet-less form is itself pinned as a valid header by the committed #4223 suite (`GetBackStackTaskHeader.test.ts:178`). So the bullet stays optional and the repeat is defused differently: `parseTasks` now *resumes* an already-seen task id instead of restarting it. That also fixes the legacy `Task id #N` / `* TaskRecord{... #N}` pair, which prints two headers for one task.
- Because a task can now be flushed and resumed, `numActivities` and the hist-derived root are filled in a single pass after the loop rather than at flush time, so a resumed task is not mistaken for one with explicit values.
- A legacy `TaskRecord{...}` header now reads its own `A=`/`I=`/`aI=` tokens through the same helper as the modern header - the real capture carries them, and this is the second half of the reviewer's suggestion.

## Finding 2 - modern `A=` tasks had no `packageName`/`rootActivity`

**`packageName` is not derivable from `A=`.** `A=` is `Task.affinity`, which is uid-prefixed on Android 11+ (`10164:com.example`, per `ActivityRecord.computeTaskAffinity`), is an app-declarable `android:taskAffinity` string that need not name a package, and is empty for `android:taskAffinity=""`. Stripping a `uid:` prefix off it would be a guess dressed up as a derivation, so this PR does not do that. `affinity` keeps the verbatim value, uid prefix included.

The honest source is the task's own activity rows, which the modern path does print, and which the reviewer pointed at:

```
    * Hist  #0: ActivityRecord{7ff01aa u0 com.example/com.example.MainActivity t61}
```

`Hist #0` is the task root (pinned by #4197) and its component has the same `pkg/.Cls` shape legacy `realActivity=` prints. It is applied as a **fallback only**, so `I=`/`aI=` and a legacy `realActivity=` line still win - which matters, because the real capture's task 14414 prints `origActivity=InitActivity` as `Hist #0` while `realActivity=` is `MainActivity`.

A task with no printed `Hist #0` row keeps `packageName`/`rootActivity` undefined. Nothing else in this dump identifies its package; that is documented in code rather than guessed at, and is pinned by a test.

The `Hist` row regex was extracted to a shared `parseHistRow` so `parseActivities` and `parseTasks` cannot drift apart on the shape - it is the exact regex from #4253, unchanged.

## Testing

New `test/features/observe/GetBackStackTaskAnchoring.test.ts`, 12 tests.

Red-before / green-after, verified by checking out `origin/main`'s `GetBackStack.ts` under the new tests:

- red: 6 pass / 6 fail. The Finding 1 failure is exactly the reviewer's prediction - `["com.facebook.katana", undefined]` where both should be `com.facebook.katana`. The Finding 2 failure is `packageName` / `rootActivity` undefined for the modern `A=` task.
- green with the fix: 44 pass / 0 fail across all four `GetBackStack*` suites, i.e. every #4197 / #4222 / #4223 behaviour still holds.

Full local gates: `bun test` 8726 pass / 0 fail, `bun run typecheck` clean (214 baseline errors, none new), `bun run lint` clean. No baselines touched. No device interaction - committed fixtures only.

## Acceptance criteria

All seven from #4263 are satisfied and each has a test.

### DIFF
--- a/src/features/observe/GetBackStack.ts
+++ b/src/features/observe/GetBackStack.ts
@@ -7,10 +7,91 @@ import type { BackStack } from "./interfaces/BackStack";
 import { Timer, defaultTimer } from "../../utils/SystemTimer";
 
 /**
- * Legacy task header, printed by `TaskRecord.toString()` / the pre-Android-10
- * `dumpsys` layout: "Task id #123" or "TaskRecord{task123 #123 A=... sz=3}".
+ * Legacy task header, printed by the pre-Android-10 `dumpsys` layout as either
+ * "Task id #123" or `ActivityStack.dumpActivitiesLocked`'s `"    * " + task`,
+ * i.e. "* TaskRecord{1d3813b #14418 A=com.example U=0 StackId=436 sz=2}".
+ *
+ * Anchored to the start of the line (issue #4263). `TaskRecord{...}` is printed
+ * verbatim in at least two NON-header positions in real legacy output:
+ *
+ *     * Hist #1: ActivityRecord{...}
+ *         frontOfTask=false task=TaskRecord{1d3813b #14418 A=... sz=2}
+ *
+ *     Running activities (most recent first):
+ *       TaskRecord{1d3813b #14418 A=... sz=2}
+ *
+ * The first sits BETWEEN two `Hist` rows, so read as a header it splits one
+ * task's activities across two headers and (before #4263) blanked the affinity
+ * of every activity below it. Line-anchoring rejects it: it is always preceded
+ * by `frontOfTask=<bool> task=`.
+ *
+ * The second IS at the start of its line, only without the bullet, so anchoring
+ * alone cannot tell it apart from a real header -- and the bullet-less form is
+ * itself pinned as a header by the #4223 suite. It is instead defused in
+ * `parseTasks`, which resumes an already-seen task id rather than restarting
+ * it, so a repeat cannot discard what was already parsed.
+ *
+ * The id is taken from the token right after the identity hash rather than by
+ * the previous greedy scan, which captured the last "#number" on the line.
  */
-const LEGACY_TASK_HEADER = /Task\s+id\s+#(\d+)|TaskRecord.*#(\d+)/;
+const LEGACY_TASK_HEADER = /^\s*(?:Task\s+id\s+#(\d+)\b|\*?\s*TaskRecord\{\S+\s+#(\d+)\b(.*))$/;
+
+/**
+ * `A=` / `I=` / `aI=` out of a task-header tail. Space-anchored so "aI=" is not
+ * read as "I=" and so "visibleRequested=" and friends cannot contribute a stray
+ * key match; the value excludes "}" so a field printed last yields the bare
+ * value rather than one with the closing brace glued on.
+ */
+const HEADER_AFFINITY = /(?:^|\s)A=([^\s}]+)/;
+const HEADER_COMPONENT = /(?:^|\s)a?I=([^\s}]+)/;
+
+/**
+ * One printed activity: "* Hist  #0: ActivityRecord{2b2ce0f u0 pkg/.Cls t61}".
+ *
+ * AOSP's `ActivityRecord.toString()` always puts the user id and the component
+ * INSIDE the braces, but it does NOT agree across versions on where the task id
+ * goes. Both shapes are present in this repo's committed captures under
+ * test/features/observe/windowDumps:
+ *
+ *   API 30-32, 34-36: ActivityRecord{2b2ce0f u0 com.example/.Main t61}
+ *   API 33:           ActivityRecord{a9cf40f u0 com.example/.Main} t8}
+ *
+ * On API 33 the component is closed off by its own brace and the task id trails
+ * outside it. So the component group is bounded to exclude "}" (otherwise the
+ * API 33 shape yields a name ending in "}"), and the task id is scanned for
+ * anywhere in the tail rather than assumed to sit at a fixed offset.
+ *
+ * Other trailing tokens are tolerated deliberately: a finishing activity prints
+ * " f" and an activity with no task prints "t??". Requiring the task id to be
+ * the last token is what made the original regex drop every line, so an
+ * unrecognized tail leaves the task id to fall back to the enclosing header
+ * rather than rejecting the whole activity.
+ */
+const ACTIVITY_LINE = /Hist\s+#(\d+):\s+ActivityRecord\{\S+\s+u\d+\s+([^\s}]+)([^\n]*\})/;
+
+/** A parsed "Hist #N" row: its index within the task and its component. */
+interface HistRow {
+  /** Counts UP from the task root, so #0 is the root, not the top. */
+  histIndex: number;
+  /** Verbatim "pkg/.Cls" -- the same shape legacy `realActivity=` prints. */
+  component: string;
+  /** The row's own "tNN" token, when it carries one. */
+  taskId?: number;
+}
+
+function parseHistRow(line: string): HistRow | undefined {
+  const match = line.match(ACTIVITY_LINE);
+  if (!match) {
+    return undefined;
+  }
+  // Standalone "tNN" token in the tail, on either side of a closing brace.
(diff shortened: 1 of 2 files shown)


kaeawc:
Manual-test verification on HEAD `44f83116b`, Android emulator API 35.

Task-header anchoring and the `Hist #0` task-root sourcing both parse correctly, but verification surfaced an adjacent gap this PR did not close: `isTaskRoot` is still derived as `histIndex === 0`, while `dumpsys` prints the authoritative `rootOfTask=true|false` (API 30+), which is never parsed. Live on API 35 the launcher activity reports `isTaskRoot: false` despite `rootOfTask=true`, and the committed #4332 captures reproduce it in both directions (api35/api36 miss a real root; api34 invents one).

Filed as https://github.com/kaeawc/auto-mobile/issues/4340. Everything else this PR targets verified good.

Issue #4282, opened 2026-07-23
Android: recover from a secure keyguard (PIN entry / prompt the user)


kaeawc:
Follow-up to #4235 (Part 4 of 4). PR #4279 added the `deviceLock` signal (including `secure`), which is the prerequisite for this.

## Problem

On a secure (PIN/pattern/password) lock an agent cannot self-recover: it either loops on swipe attempts or proceeds against the lock screen. The PIN has to come from the user, and nothing today surfaces that requirement or accepts the credential.

## Design question (not yet decided)

Should AutoMobile accept a PIN/pattern/password to dismiss a secure keyguard — via a tool parameter, an env var, or an interactive prompt — and if so, how is it handled safely (never logged, never persisted)? This is a security-sensitive design decision separate from the observability that #4279 shipped.

With `deviceLock.secure === true` now observable, an agent can at least detect the state and stop to ask the user. This issue tracks the recovery mechanism itself.

## Acceptance (to refine once the mechanism is chosen)

- A caller can supply a credential to dismiss a secure keyguard, or AutoMobile surfaces a clear "prompt the user for a PIN" requirement.
- The credential is never written to logs, telemetry, or disk.

kaeawc:
Closing as satisfied by the device-lock series, under the agreed **prompt-only, no credential handling** scope.

For a secure keyguard, prompt-only means one thing: surface that the device is secure-locked and tell the agent to stop and ask the user to unlock it — AutoMobile never handles the PIN/pattern/password itself. That is now delivered across three surfaces:

- **observe** (#4279 / PR #4279, merged): `deviceLock.secure === true` alongside `locked === true`, so an agent inspecting before acting knows it is a secure lock.
- **actions** (#4280 / PR #4288, merged): a locked interaction carries a `deviceLockWarning` whose secure branch says "ask the user to unlock (PIN/pattern/password)".
- **hierarchy failure** (#4281 / PR #4297): a secure keyguard blocking the accessibility service produces "unlock the device… you may need to ask the user" instead of the generic error.

There is nothing left to build for secure recovery: dismissing a PIN would require handling the credential, which is explicitly out of scope, and "surface the prompt" is done on every surface an agent would hit. A swipe-lock (non-secure) self-dismiss helper was considered and declined — it is outside this issue's secure-keyguard scope, and an agent can already dismiss a swipe lock with a full-screen `swipeOn`.

[...]

Issue #4281, opened 2026-07-23
Android: distinct fresh-boot error when keyguard blocks accessibility-service binding


kaeawc:
Follow-up to #4235 (Part 3 of 4). PR #4279 added the `deviceLock` observability signal.

## Problem

On a fresh emulator boot, before the keyguard is first dismissed, Android refuses to bind non-encryption-aware accessibility services (`AccessibilityManagerService: Ignoring non-encryption-aware service`). Every Android tool then fails with the generic `Failed to retrieve view hierarchy from accessibility service` — which points at the service (installed, enabled, healthy) and hides the real cause: the device is locked.

This is **indistinguishable from the transport failure in #4039**, which emits the identical string for a completely different cause.

## Proposed change

When the accessibility service is enabled but unbound *and* the keyguard is showing, emit a distinct message — "device is locked; dismiss the keyguard so the accessibility service can bind" — rather than the generic hierarchy-retrieval failure. The `deviceLock`/keyguard read from #4279 supplies the signal to distinguish this from #4039.

## Acceptance

- Fresh-boot locked state produces a lock-specific, actionable error, not the generic hierarchy-retrieval string.
- The message is distinguishable from #4039's transport-failure path.

Issue #4235, opened 2026-07-22
Android: no structured signal that the device is asleep/locked — observe and tapOn silently succeed against the lock screen, and a PIN lock leaves agents stuck


kaeawc:
## Summary

When an Android device is asleep and/or behind the keyguard, AutoMobile gives an agent **no structured signal that the device is locked**. Tools do not fail — they succeed against lock-screen content. An agent has no way to distinguish "my app is on screen" from "I am looking at the lock screen", so it proceeds confidently against the wrong UI. When the lock is PIN-protected the agent cannot recover on its own at all, but nothing tells it to stop and ask the user.

This should be observable enough that an agent can detect the state, and either recover (swipe lock) or **prompt the user to unlock / supply a PIN** (secure lock), instead of silently doing the wrong thing or getting stuck.

## Evidence

Android emulator (`Medium_Phone_API_35`, API 35), `origin/main` @ 92cd3baa5.

### `observe` succeeds and returns the lock screen

Asleep + keyguard showing (`mWakefulness=Asleep`, `isKeyguardShowing=true`, `mScreenState=OFF`):

```
observe --platform android   ->  exit 0, no error
  screenSize:  {width: 1080, height: 2400}
  elements:    {clickable: 7, scrollable: 0, text: 13, media: 9}
  activeWindow: {appId: "com.android.systemui", activityName: "android.widget.FrameLayout"}
  visible text: ["Account action required", "2 minutes ago", "jason.d.pearson@gmail.com",
                 "Expand", "11:11", "Wed, Jul 22", "T-Mobile",
                 "Screen is locked in portrait orientation.", "Battery 100 percent."]
```

There is a populated element tree, so an agent sees a plausible screen and keeps going.

### Actions report success against lock-screen content

```
tapOn --selector '{"text":"Expand"}'   ->  success: true, "Tapped on element"
```

The tap landed on a lock-screen notification. An agent counts that as a successful interaction with its app under test.

### No structured lock signal exists

`observe` top-level keys:

```
accessibilityState, activeWindow, backStack, elements, focusedElement, freshness,
insets, intentChooserDetected, notificationPermissionDetected, performanceAudit,
rotation, screenSize, systemInsets, updatedAt, viewHierarchy, wakefulness
```

[...]

PR #4288, opened 2026-07-23
feat(action): warn when an interaction runs on the Android keyguard (#4280)

### LINKED ISSUE #4280: Android: actions should refuse (not report success) when deviceLock.locked is true
Follow-up to #4235 (Part 2 of 4). PR #4279 added the observable `deviceLock` signal on `observe`; this covers the behavioral half.

## Problem

`tapOn` (and other actions) return `success: true` for a tap that lands on the keyguard — the most actively-misleading part of the locked-device experience, because an agent counts it as a successful interaction with its app under test.

## Proposed change

When `observe` reports `deviceLock.locked === true`, actions should refuse (or at minimum warn) rather than reporting a false success. An `ActionableError` naming the state ("device is locked; dismiss the keyguard or supply a PIN") is far better than a silent `success: true`. With `deviceLock` now on the observe result, actions have a structured signal to gate on.

## Notes

- Decide refuse-vs-warn per action; `tapOn` false-success is the priority case.
- With `deviceLock.secure === false` an agent can self-recover (swipe dismiss); with `secure === true` it should stop and prompt — see #4235 Part 4 (PIN entry) for the recovery side.

kaeawc:
## Summary

Closes #4280. A tap that landed on the Android keyguard used to return a clean `success: true`, so an agent counted a tap on lock-screen furniture as a real interaction with its app under test. This annotates the action result so that can never silently happen — **without** blocking the gesture, since swipe-to-dismiss and PIN-digit taps are exactly how an agent recovers.

Follows #4235 / #4279, which added the `deviceLock` observability this builds on.

## Behavior (approved: warn + annotate, still execute)

When an interaction runs while the Android device is locked at action time, `observedInteraction` attaches to the result:
- `deviceLock` — the pre-action `{ locked, keyguardShowing, secure? }`, so an agent can branch on `secure`.
- `deviceLockWarning` — a human-readable message: secure lock → "ask the user to unlock"; swipe lock → "dismiss the keyguard"; unknown → generic.

The gesture still executes and the result keeps its real `success`, so recovery flows are unaffected. A blanket refuse-when-locked was explicitly rejected because it would block the very swipe/PIN gestures needed to recover.

## Changes

- `BaseVisualChange.observedInteraction` — single chokepoint every interaction tool (tapOn, swipeOn, inputText, pressButton, dragAndDrop, …) flows through; added `annotateDeviceLock` + `buildDeviceLockWarning`. Android-only (`deviceLock` is only collected on Android observe); reads the pre-action observe result, so **no extra adb call**.
- Docs updated (`docs/design-docs/mcp/observe/index.md`).

## Testing

- `BaseVisualChange.deviceLockWarning.test.ts` — secure lock (PIN-oriented warning + still succeeds), swipe lock (dismiss-oriented warning), unlocked (no annotation), no-signal (no annotation), iOS (never annotated). Verified red before the annotation.
- Full suite green (8791 pass, 0 fail); lint + typecheck clean.
- `getDeviceLock` itself was verified end-to-end on a live PIN-secured emulator in #4279.

## Acceptance criteria

- ✅ An interaction on a locked device carries a structured lock signal + warning instead of silent clean success.
- ✅ The message distinguishes swipe ("dismiss the keyguard") from secure ("ask the user to unlock").
- ✅ Recovery gestures (swipe-dismiss, PIN entry) are not blocked.
- ✅ iOS is unaffected.

## Follow-ups

Part of the #4235 device-lock series: #4281 (distinct fresh-boot accessibility-bind message) and #4282 (secure-keyguard recovery, prompt-only) are next.


kaeawc:
Manual-test verification on HEAD `44f83116b`, Android emulator API 35 — **confirmed delivered**.

With the swipe keyguard showing (`dumpsys window policy` → `showing=true secure=false`), `pressButton menu` returned:

- `deviceLock` = `{"locked":true,"keyguardShowing":true,"secure":false}`
- `deviceLockWarning` = "Device is locked; this interaction likely did not reach the app under test. The lock is a swipe lock — dismiss the keyguard (e.g. swipe up) before continuing."

Correct `secure:false` branch selected, and the gesture was still dispatched rather than blocked, matching the documented intent.

PR #4357, opened 2026-07-24
fix(observe): derive isTaskRoot from dumpsys rootOfTask, not the Hist index

### LINKED ISSUE #4340: fix(observe): GetBackStack infers isTaskRoot from the Hist index instead of dumpsys rootOfTask
Found during a manual-test iteration over `92cd3baa5..main` (HEAD `44f83116b`), verifying [#4273](https://github.com/kaeawc/auto-mobile/pull/4273) / [#4332](https://github.com/kaeawc/auto-mobile/pull/4332).

## Summary

`GetBackStack` derives `isTaskRoot` from the printed history index:

https://github.com/kaeawc/auto-mobile/blob/main/src/features/observe/GetBackStack.ts#L234

```ts
isTaskRoot: histIndex === 0
```

The `Hist #N` index is not a task-root indicator. `dumpsys activity activities` prints the answer explicitly as `rootOfTask=true|false` inside each `ActivityRecord` block, and `rootOfTask` is **never parsed anywhere in `src/`** (`grep -rn rootOfTask src/` returns nothing). The index heuristic disagrees with the device in **both** directions.

## Observed on a live device

Emulator `am-api35-ga-arm64` (API 35), launcher foregrounded. `observe` returned:

```json
"backStack": {
  "currentActivity": { "name": "com.google.android.apps.nexuslauncher.NexusLauncherActivity", "taskId": 29 },
  "activities": [ { "name": "com.google.android.apps.nexuslauncher.NexusLauncherActivity", "taskId": 29, "isTaskRoot": false } ]
}
```

Ground truth from the same `dumpsys activity activities`:

```
* Task{90f2a0f #29 type=home I=com.google.android.apps.nexuslauncher/.NexusLauncherActivity U=0 rootTaskId=1 ... sz=2}
      * Hist  #1: ActivityRecord{1e6036e u0 com.google.android.apps.nexuslauncher/.NexusLauncherActivity t29}
        ...
        rootOfTask=true task=Task{90f2a0f #29 type=home I=com.google.android.apps.nexuslauncher/.NexusLauncherActivity}
```

The activity is the task root (`rootOfTask=true`, and it matches the task header's own `I=` root-intent component), but it prints as `Hist #1`, so the parser reports `isTaskRoot: false`.

## Reproduces on the committed fixtures

This is not a live-device artifact — the captures added by [#4332](https://github.com/kaeawc/auto-mobile/pull/4332) under `test/features/observe/activityActivitiesDumps/` already contain both failure directions. Running the parser over them:

| Fixture | Activity | `Hist #N` | `rootOfTask` (truth) | parsed `isTaskRoot` |
|---|---|---|---|---|
| `api35-home-settings-secondapp.log` | `...nexuslauncher/.NexusLauncherActivity` | `#1` | `true` | **`false`** |
| `api36-home-settings-secondapp.log` | `...nexuslauncher/.NexusLauncherActivity` | `#1` | `true` | **`false`** |
| `api34-home-settings-secondapp.log` | `com.android.settings/.Settings$WifiSettingsActivity` | `#0` | `false` | **`true`** |

So it both misses real task roots (API 35/36) and invents them (API 34). The fixtures are load-bearing for other assertions but nothing currently asserts `isTaskRoot` against `rootOfTask`, which is why the gap survived #4273.

`rootOfTask=` is present in the captures from **API 30 onward** (absent on API 24–29):

```
api24..api29: rootOfTask=true occurrences = 0
api30,31,33,34,35,36: rootOfTask=true occurrences = 4   (api32: 2)
```

## Impact

`isTaskRoot` is user-visible in `observe`'s `backStack.activities[]`. Anything branching on it — deciding whether a back press exits the app vs. pops within the task — gets the wrong answer on the current-generation API levels (34/35/36), which is where most testing happens.

## Suggested fix

Parse `rootOfTask=(true|false)` from within each `ActivityRecord` block and use it as the authoritative value, keeping the `histIndex === 0` result only as the fallback for API ≤ 29 where the field is not printed. The block is already being scanned line-by-line for `packageName=` etc., so this is the same pass; terminate the lookahead at the next `Hist #N:` line so a value never bleeds into the following record.

Secondary: the task header's `I=<component>` gives an independent cross-check for the task root on modern dumps.

## Acceptance

- `isTaskRoot` matches `rootOfTask` for every activity in every fixture that prints the field (API 30–36).
- A regression test asserts that pairing directly off the committed captures, so the API 34 and API 35/36 shapes above are both pinned.
- API 24–29 fixtures keep their current index-derived behavior.

kaeawc:
## Summary

`GetBackStack.parseActivities` derived `isTaskRoot` from the printed `Hist #N` index (`histIndex === 0`), but that index is not a task-root indicator. `dumpsys activity activities` prints the answer explicitly as `rootOfTask=true|false` inside each `ActivityRecord` block from API 30 onward, and the heuristic disagrees with it in both directions on the committed captures. This PR parses that field as the authoritative value in the same line-by-line pass, keeping the index heuristic only as the fallback for API ≤ 29 where the field is not printed. The pending-record reference is cleared at the next `Hist #N:` row or task header, so a value can never bleed into the following record.

## Linked Issue

Closes #4340

## Bug / Regression Context

- **Prior attempts:** [#4273](https://github.com/kaeawc/auto-mobile/pull/4273) / [#4332](https://github.com/kaeawc/auto-mobile/pull/4332) added the real captures but nothing asserted `isTaskRoot` against `rootOfTask`, which is why the gap survived.
- **Observed symptom:** `observe`'s `backStack.activities[]` reported the launcher (`rootOfTask=true`, printed as `Hist #1`) as `isTaskRoot: false` on API 35/36, and `Settings$WifiSettingsActivity` (`rootOfTask=false`, printed as `Hist #0`) as `isTaskRoot: true` on API 34.
- **Repro steps / conditions:** run the parser over the committed fixtures under `test/features/observe/activityActivitiesDumps/` — the new fixture-pairing tests fail on `main` for exactly api34/api35/api36.

## Acceptance criteria → tests

| Criterion | Test |
|---|---|
| `isTaskRoot` matches `rootOfTask` for every activity in every fixture that prints the field (API 30–36) | `GetBackStackRealCaptures.test.ts` — per-fixture "isTaskRoot matches the dump's own rootOfTask field (#4340)": independently extracts the `(Hist → rootOfTask)` pairing from the raw capture and asserts it 1:1 against parsed `activities[]` |
| The API 34 and API 35/36 shapes are both pinned directly off the committed captures | "isTaskRoot vs rootOfTask, exact values (#4340)": api35/api36 launcher `Hist #1` → `true`; api34 `WifiSettingsActivity` `Hist #0` → `false` |
| API 24–29 fixtures keep their current index-derived behavior | Per-fixture "no rootOfTask printed; isTaskRoot stays index-derived (#4340)" asserts the fixture has no `rootOfTask=` and the parsed values equal the index-derived truth |

Plus two synthetic parser tests in `GetBackStackParsing.test.ts`: override in both directions, and no bleed of a `rootOfTask` value into a following record that lacks one.

## Validation

- [x] `scripts/all_fast_validate_checks.sh` passes (`bun run lint` + `bun test`) — via `turbo run lint build test`: 9032 pass / 0 fail
- [x] New code uses interfaces + fakes + FakeTimer; unit tests run in <100ms
- [x] New generic code reuses the standard library or an existing project helper; no new dependencies
- [x] Based on latest `main` (`44f83116b`)

## Follow-ups

- The issue's secondary suggestion — cross-checking the task root against the task header's `I=<component>` — is intentionally not built: `rootOfTask` is authoritative on every fixture that prints it, and a second source of truth adds its own edge cases. Noted here rather than filed, per the issue marking it secondary.

### DIFF
--- a/src/features/observe/GetBackStack.ts
+++ b/src/features/observe/GetBackStack.ts
@@ -267,6 +267,14 @@ export class GetBackStack implements BackStack {
         continue;
       }
 
+      if (HIST_LINE.test(line)) {
+        // A Hist row in a shape parseHistRow does not recognize still starts a
+        // new record, so the previous record's block is over; without this, the
+        // unrecognized record's rootOfTask= would land on the previous activity.
+        openActivity = undefined;
+        continue;
+      }
+
       // The current record's own rootOfTask= field (see ROOT_OF_TASK_LINE).
       const rootOfTask = line.match(ROOT_OF_TASK_LINE);
       if (rootOfTask && openActivity) {
@@ -297,8 +305,10 @@ export class GetBackStack implements BackStack {
     // for the same task ("Task id #N" then "* TaskRecord{... #N ...}"), so the
     // second must resume the first's count rather than restart it.
     const histCounts: Map<number, number> = new Map();
-    // The component of each task's "Hist #0" row. `Hist #0` is the task root
-    // (see #4197), and its "pkg/.Cls" is the same shape legacy `realActivity=`
+    // The component of each task's "Hist #0" row. `Hist #0` is normally the
+    // task root (see #4197; the authoritative rootOfTask= field can disagree
+    // on API 30+, see #4340 -- close enough for the package-name fallback this
+    // feeds), and its "pkg/.Cls" is the same shape legacy `realActivity=`
     // prints. This is the ONLY package source on the modern path (issue #4263):
     // a modern header's `A=` is `Task.affinity`, which is uid-prefixed on
     // Android 11+, is an app-declarable string that need not name a package,
--- a/src/features/observe/GetBackStack.ts
+++ b/src/features/observe/GetBackStack.ts
@@ -118,6 +118,25 @@ const MODERN_TASK_HEADER = /^\s*\*\s*Task\{\S+\s+#(\d+)\b(.*)$/;
 /** "* Hist  #0: ActivityRecord{...}" -- one printed activity. */
 const HIST_LINE = /\bHist\s+#\d+:/;
 
+/**
+ * `rootOfTask=<bool>` out of an ActivityRecord block (issue #4340). Printed by
+ * `ActivityRecord.dump` from API 30 on as its own field line a few lines below
+ * the record's `Hist #N:` row:
+ *
+ *     * Hist  #1: ActivityRecord{c87b532 u0 com.google...nexuslauncher/.NexusLauncherActivity t7}
+ *         ...
+ *         rootOfTask=true task=Task{97acb18 #7 type=home I=...}
+ *
+ * This is the authoritative task-root answer; the `Hist #N` index is not. Real
+ * captures disagree with the index in both directions: the launcher is the root
+ * of its task yet prints as `Hist #1` on API 35/36, and API 34 prints a
+ * `Hist #0` row whose block says `rootOfTask=false`. Anchored to the start of
+ * the line so the inline `task=Task{...}` tail of the same line -- or any other
+ * mention -- cannot match. Absent on API <= 29, where the index heuristic
+ * remains the only source.
+ */
+const ROOT_OF_TASK_LINE = /^\s*rootOfTask=(true|false)\b/;
+
 /** Fields carried by a modern task header line. */
 interface TaskHeaderFields {
   id: number;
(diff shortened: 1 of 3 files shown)


kaeawc:
**Manual-test verification (v0.0.45 → main, `c3994f57c`) — FIXED on device.**

Compared `observe` back-stack output on `emulator-5554` against `dumpsys activity activities`.

- back-stack activities: Settings (task 8) `isTaskRoot:true` and NexusLauncher (task 7) `isTaskRoot:true` — matching their `rootOfTask=true` records.
- Source confirmed: `GetBackStack.ts` parses `rootOfTask=(true|false)` and prefers it over the index-derived `firstHist0`, with comments citing the exact API 34/35/36 cases where `Hist #0` ≠ the true root.

**Result: FIXED.**

PR #4363, opened 2026-07-24
feat(action): cross-platform wakeAndUnlock tool (#4360)

### LINKED ISSUE #4360: fix(action): inputText unlocks a secure keyguard but reports setText failure; tapOn cannot fit the lock-screen budget
Found during a manual-test iteration on HEAD `2c3468867` (v0.0.45), Android emulator `am-api35-ga-arm64` (API 35), with a secure PIN keyguard (`1234`).

## Summary

AutoMobile can *observe* a secure keyguard correctly, and can *raise* the PIN bouncer, but the two entry paths behave badly in opposite directions:

- **`inputText` succeeds and reports failure** (false negative) — the primary defect.
- **`tapOn` cannot work on a keyguard at all** — a single call takes ~13s against a lock-screen display budget of ~7–8s.

There is currently no first-class way to get past a keyguard, even though every primitive needed already exists.

## Defect 1 (primary): `inputText` unlocks the device but returns an error

With the bouncer up, `inputText --platform android --text 1234`:

```
message: "Input text"
error  : "Accessibility service setText failed: No focused editable node found"
```

…and the device **unlocks**. Verified immediately after, via both `dumpsys` and `observe`:

```
KeyguardServiceDelegate  showing=false  occluded=false
deviceLock   = {"locked": false, "keyguardShowing": false, "secure": true}
activeWindow = com.google.android.apps.nexuslauncher
```

The error text originates in the runner's accessibility path
(`android/control-proxy/.../CtrlProxy.kt:2952`). That leg genuinely fails — the PIN
field is not an editable a11y node — but the keystrokes still reach the bouncer via a
fallback leg, which succeeds. Only the first leg's failure is surfaced.

**Why this matters more than a cosmetic wording bug.** The whole manual-test discipline
is "ground the result in observed state, never the tool's `success` flag." This is that
hazard running in reverse: an agent that correctly distrusts a success flag and reads
this error will conclude the unlock failed, and then either abort or retry an action
that already completed. During this session it caused exactly that — the unlock was
reported as impossible for several rounds before the device state was re-checked.

**Suggested fix:** when a fallback leg satisfies the request, do not surface the earlier
leg's error. Either return success (optionally noting the degraded path in a `warnings`
field), or attribute the error to the specific leg so it is not read as the outcome of
the whole call.

## Defect 2: `tapOn` is structurally unusable on a keyguard

Measured on the same device, bouncer confirmed up (`pinEntry` present via `uiautomator`):

| Measurement | Value |
|---|---|
| Single `tapOn` on `com.android.systemui:id/key1`, wall-clock | **13s** |
| Screen state when the call returned | `mScreenState=OFF` |
| Lock-screen display budget (measured, see below) | **~7–8s** |

Tapping four digits sequentially therefore cannot ever complete. The observed symptom is
misleading: the first tap returns a plain `Tapped on element`, and subsequent taps return
`Tapped on element (0 view hierarchy changes over ~29 requests within 15s)` — which reads
like a missed target rather than "the screen switched off during the settle loop."

### The lock-screen timeout is not `screen_off_timeout`

This is the non-obvious part, and worth recording regardless of what is done with the
rest of this issue. With `screen_off_timeout=1800000` (30 min) and `sleep_timeout=-1`,
confirmed effective in `dumpsys power`
(`mScreenOffTimeoutSetting=1800000`, `mSleepTimeoutSetting=-1`, no device-admin cap):

| Elapsed | Unlocked (launcher) | Locked (keyguard) |
|---|---|---|
| 6s | `Awake` / `ON` | `Awake` / `ON` |
| 9s | `Awake` / `ON` | **`Dozing` / `OFF`** |
| 12s | `Awake` / `ON` | `Asleep` / `OFF` |
| 30s | `Awake` / `ON` | `Asleep` / `OFF` |

The keyguard uses `config_lockScreenDisplayTimeout`, a framework **config resource baked
into the build**, not a settable `Settings` key. So no amount of `settings put` widens
this window; any keyguard interaction must fit inside roughly 7 seconds.

## Proposal: teach AutoMobile the keyguard sequence

Every primitive needed already works. What is missing is that AutoMobile does not know
the sequence. The combination verified end-to-end this session:

```bash
adb shell input keyevent 224   # KEYCODE_WAKEUP
adb shell input keyevent 82    # KEYCODE_MENU -> raises the PIN bouncer
```
then a single fast text input of the PIN.

Notes from testing the alternatives:

- **Raising the bouncer:** keyevents `82` (MENU), `8` (a digit) and `66` (ENTER) all
  raise it, confirmed via `uiautomator` (`pinEntry`, `key1`–`key9` present).
  `swipeOn --direction up --autoTarget false --speed slow` also raises it — AutoMobile's
  own `observe` sees `pinEntry` + `key1`–`key9` immediately afterward.
- **Power-cycling the screen** (`keyevent 26` twice) does **not** raise it.
- **`wm dismiss-keyguard`** does not get past a *secure* lock (it is the existing call at
  `AndroidEmulatorClient.ts:1559`); it is sufficient only for a swipe lock.
- **Speed is the binding constraint**, so the entry step must be a single call — which
  rules out per-digit tapping regardless of how the defects above are resolved.

An `unlock`-style capability (or an documented recipe) taking the credential would make
this reachable. It also unblocks a real dependency: `videoRecording` cannot function on a
locked device because `/sdcard` is credential-encrypted and `screenrecord` cannot create
its output file, so today a PIN-locked device silently blocks that whole surface.

## Acceptance

- `inputText` that succeeds via a fallback leg does not surface the accessibility leg's
  error as the call's outcome.
- Keyguard interaction has a documented, tested path that fits inside the
  ~7s lock-screen display budget.
- The `config_lockScreenDisplayTimeout` constraint is recorded wherever keyguard handling
  lives, so the next investigation does not re-derive it from `tapOn` timing.

kaeawc:
## Summary

Adds a dedicated cross-platform **`wakeAndUnlock`** MCP tool that wakes a device and gets past its keyguard, and **removes all situation-specific wake/unlock logic from `inputText`** (it returns to clean a11y-only text input). This supersedes this PR's earlier approach (an inputText keyguard fallback), which is reverted here.

Grounded in on-device investigation of both platforms (see Testing).

## Linked Issue

Closes #4360

## `wakeAndUnlock({ pin?, platform })`

`pin` is optional in the schema but **logically required to unlock a secure Android device** — if omitted there, a PIN remembered earlier this session is used, else an `ActionableError` asks for one. Ignored on iOS.

- **Android**: reads wakefulness + lock state via `dumpsys` (no `observe`, so it works on a screen about to power off). Wakes a sleeping device; `wm dismiss-keyguard` dismisses a swipe lock or raises a secure bouncer; for a secure lock, types the PIN as key events and submits with `KEYCODE_ENTER` (a11y `setText` cannot type into a secure bouncer — the original #4360 symptom). The outcome is grounded in a **bounded poll** of the lock state (~250 ms × up to 2.5 s), covering a **measured ~1.3 s lag** between `ENTER` and the keyguard clearing in `dumpsys`.
- **iOS**: simulators cannot set a passcode, so it wakes and swipe-dismisses the non-secure lock screen via existing gesture primitives; `pin` is ignored.

## Remembering how to unlock

New nullable `lock_type` / `lock_credential` columns on `device_sessions` (migration `2026_07_24_000_device_session_lock`). A freshly-supplied PIN that unlocks a secure device is remembered; a recorded PIN is reused and **not** re-persisted; a PIN that failed to unlock is never stored. The credential is stored **plaintext in the local single-user `~/.auto-mobile` DB**. The boot path (`AndroidEmulatorClient.wakeAndUnlock`) delegates to the same feature and reuses a remembered PIN, leaving a secure device with no record locked (non-fatal).

## Changes

- `src/features/action/WakeAndUnlock.ts` — feature with injected seams (adb, timer, credential store, iOS unlocker); fully unit-tested.
- `src/features/action/asciiKeyEvents.ts` — `buildAsciiKeyEventPlan` extracted from `InputText` so text input and PIN entry share one canonical mapping table.
- `src/features/action/{DeviceSessionLockStore,IosLockScreenUnlocker}.ts` — DB-backed credential store; iOS wake+swipe over `PressButton`/`SwipeOn`.
- `wakeAndUnlock` tool registration + schema; `schemas/tool-definitions.json` regenerated (71 → 72).
- DB: migration + `device_sessions` columns + `getDeviceLockCredential` / `rememberDeviceLock` repo methods.
- Boot: `AndroidEmulatorClient.wakeAndUnlock` delegates to the feature (now only dismisses the keyguard when actually locked).
- `InputText` / `BaseVisualChange`: keyguard fallback and `deviceUnlocked` guard reverted; `BaseVisualChange` is byte-identical to `main`.
- Docs: `keyguard.md` documents the tool, the `wm dismiss-keyguard` matrix, the iOS findings, and the ~1.3 s poll.

## Risk

**Higher-risk** — touches a **DB migration** (append-only) and a **public tool schema**, plus the boot path. Not for auto-merge; handing the merge decision to the maintainer.

## Testing

- On-device investigation (findings drove the design):
  - iOS (iPhone 16 Pro, iOS 18.6): no `simctl` lock/passcode, no "Face ID & Passcode" in Settings, swipe dismisses with no prompt → simulators have no secure lock.
  - Android (API 35, secure PIN): `wm dismiss-keyguard` raises the bouncer on a secure lock (doesn't unlock) and fully dismisses a swipe lock; measured the ~1.3 s post-`ENTER` settle lag.
- On-device integration: the exact Android sequence the feature emits unlocked a real secure emulator (`showing=true` → `false`, launcher foreground) within the budget, cleared after ~750 ms of polling.
- `bun test` full suite green (9113 pass / 0 fail); `bun run typecheck`, `bun run lint`, `validate_mkdocs_nav.sh` all pass. New unit tests: `WakeAndUnlock` (10), `asciiKeyEvents` (7), repo lock methods (5), boot delegation (6), updated `InputText`.
- Tool wiring not exercised through a live daemon to avoid disrupting the shared daemon/fleet; the underlying recipe is verified on-device and the emitted command sequence is unit-pinned.

## Follow-ups

- #4366 — `tapOn` cannot fit the ~7 s lock-screen budget (Defect 2 of #4360), out of scope here.

### DIFF
--- a/docs/design-docs/plat/android/keyguard.md
+++ b/docs/design-docs/plat/android/keyguard.md
@@ -1,0 +1,112 @@
+# Keyguard Handling
+
+<kbd>✅ Implemented</kbd> <kbd>🧪 Tested</kbd>
+
+> **Current state:** `inputText` recovers a secure keyguard on Android by
+> delivering the credential as key events when the accessibility path fails.
+> Lock state is surfaced on `observe` and annotated on interactions
+> (issues [#4235](https://github.com/kaeawc/auto-mobile/issues/4235),
+> [#4280](https://github.com/kaeawc/auto-mobile/issues/4280),
+> [#4360](https://github.com/kaeawc/auto-mobile/issues/4360)).
+> See the [Status Glossary](../../status-glossary.md) for chip definitions.
+
+## The binding constraint: `config_lockScreenDisplayTimeout`
+
+**Any keyguard interaction must complete inside roughly 7 seconds.** This is the
+single most important — and least obvious — fact about driving a locked Android
+device, so it is recorded here rather than left to be re-derived from tool
+timing.
+
+A locked keyguard powers its display off far sooner than an unlocked device, and
+the window is **not** governed by the settable `screen_off_timeout` /
+`sleep_timeout` keys. It is `config_lockScreenDisplayTimeout`, a framework config
+**resource baked into the build**. No amount of `settings put` widens it.
+
+Measured on `am-api35-ga-arm64` (API 35) with `screen_off_timeout=1800000`
+(30 min) and `sleep_timeout=-1`, both confirmed effective in `dumpsys power`
+(`mScreenOffTimeoutSetting=1800000`, `mSleepTimeoutSetting=-1`, no device-admin
+cap):
+
+| Elapsed | Unlocked (launcher) | Locked (keyguard) |
+|---|---|---|
+| 6s  | `Awake` / `ON`  | `Awake` / `ON`   |
+| 9s  | `Awake` / `ON`  | `Dozing` / `OFF` |
+| 12s | `Awake` / `ON`  | `Asleep` / `OFF` |
+| 30s | `Awake` / `ON`  | `Asleep` / `OFF` |
+
+The practical consequence: **credential entry must be a single, fast, un-observed
+key-event burst.** A per-digit `tapOn` (or any observe/settle loop between
+digits) cannot fit the budget — a single `tapOn` against a keyguard key was
+measured at ~13s wall-clock, by which point the screen had already switched off.
+
+## `inputText` on a keyguard (#4360)
+
+A secure PIN/pattern/password bouncer exposes **no editable accessibility node**,
+so the default `a11y` mode of `inputText` (a single `ACTION_SET_TEXT`) can never
+type into it — it fails with `No focused editable node found`. Reporting that
+error as the call's outcome is a false negative: an agent that (correctly)
+distrusts a bare success flag and reads the error concludes the unlock failed,
+then aborts or retries an action that had not actually run.
+
+`inputText` therefore falls back automatically. The fallback fires **only** when
+the accessibility leg has already failed **and** a pre-check
+(`AdbClient.getDeviceLock`) confirms the device is locked **and** every character
+of the text maps to a key event. The sequence:
+
+```bash
+adb shell input keyevent KEYCODE_WAKEUP   # wake the display
+adb shell input keyevent KEYCODE_MENU     # raise the secure PIN bouncer
+adb shell input keyevent KEYCODE_<digit>  # ...one per credential character
+adb shell input keyevent KEYCODE_ENTER    # submit
+```
+
+The outcome is **grounded in a re-read of the lock state**, never in the fact
+that the key events were sent:
+
+- **Unlocked afterward** → `success: true`, `method: "eventAll"`, and the
+  accessibility error is demoted to a non-fatal `warnings` entry.
+- **Still locked afterward** → `success: false` with the error attributed to the
+  keyguard leg (wrong credential or entry failure), never to accessibility.
+- **Not locked at the pre-check, or text not key-event-mappable** → the original
+  accessibility failure is returned unchanged; behavior is identical to before.
+
+Because the credential is submitted, a wrong value counts as a failed unlock
+attempt against the device's retry throttle. This path is reached only when a
+caller has explicitly asked to input text at a device already known to be locked.
+
+## What does NOT work (do not retry these approaches)
+
+Verified during the #4360 session on the same secure PIN keyguard:
+
+- **`wm dismiss-keyguard`** — dismisses only a *swipe* lock; it does not get past
+  a secure lock. This is the existing call at `AndroidEmulatorClient.ts`
+  (`wakeAndUnlock`), and it is why a freshly booted emulator with a PIN is not
+  automatically usable.
+- **Power-cycling the screen** (`keyevent 26` twice) does **not** raise the
+  bouncer.
+- **Per-digit `tapOn`** on `com.android.systemui:id/key1`…`key9` — cannot fit the
+  ~7s budget (a single tap measured ~13s); subsequent taps return
+  `0 view hierarchy changes` because the screen switched off mid-settle, which
+  misleadingly reads as a missed target.
+
+## What works to raise the bouncer
+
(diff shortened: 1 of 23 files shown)


kaeawc:
**Manual-test verification (v0.0.45 → main, `c3994f57c`) — VERIFIED on device.**

Exercised `wakeAndUnlock` against a real Android emulator (`emulator-5554`) and an iPhone 16 Pro simulator, grounding each result in device state rather than the tool's `success` flag.

**Android:**
- Put the screen to sleep (`KEYCODE_SLEEP`) → `dumpsys power` `mWakefulness=Asleep`, keyguard `showing=true`.
- `wakeAndUnlock` **woke** the device (Asleep→Awake, `SCREEN_STATE_OFF`→`ON`) and **correctly detected the secure lock**, refusing without a pin — ground truth confirmed the lock is real (`locksettings get-disabled=false`, `secure=true`, `deviceLocked=1`), so this is a correct refusal, not a false negative.
- With `pin`, it unlocked: result `{wasAsleep:true, wasLocked:true, secure:true, unlocked:true}` and ground truth flipped to keyguard `showing=false`, `deviceLocked=0`.

**iOS:** correct cross-platform routing — `{platform:"ios", wasAsleep:false, wasLocked:false, unlocked:true}`. The swipe-dismiss branch wasn't triggerable (a booted simulator has no scriptable lock screen), but the iOS wake path returns well-formed.

**Note (not a defect):** the remembered-credential path (`DeviceSessionLockStore`) reads/writes only `status='active'` `device_sessions` rows, so it can only be exercised within a single long-lived session — consistent with the tool's "for later in this session" contract.

**Result: delivered / PASS.**

PR #4415, opened 2026-07-24
fix: warm and deliver iOS screen capture helper

kaeawc:
## Summary

- Keep one simulator ScreenCaptureKit helper warm behind TTL-bounded, target-identity-safe leases, replaying the latest frame to new streams and stopping deterministically on daemon shutdown.
- Report capture readiness phases, fail simulator target resolution within two seconds, and escalate stuck helper shutdown from `SIGTERM` to `SIGKILL` with macOS process-group cleanup.
- Deliver the signed universal helper exclusively from checksum-pinned GitHub Release assets. The release workflow builds, Developer ID-signs, notarizes, verifies, and attaches the archive; npm no longer copies Swift sources or build output.

## Impact

Warm simulator streams reuse a capture session while retaining stream-scoped encoders. Missing Screen Recording permission, an unresolved Simulator window, a missing helper release asset, or a checksum mismatch now produces an actionable error instead of a generic no-video timeout.

Automatic release delivery starts with the first release minted through this workflow, because older release registry entries intentionally have no `screenCaptureHelperSha256`. Local Swift builds remain available only through the explicit `AUTOMOBILE_IOS_SCREEN_CAPTURE_HELPER` development override.

## Validation

- `bun run lint`
- `bun run build`
- `bun run validate:yaml`
- Focused Bun tests: 164 passing
- Release and packaging BATS tests: 77 passing
- `swift test` in `ios/screen-capture`: 41 passing
- `shellcheck` for changed release scripts
- `npm pack --dry-run --json` confirms no ScreenCaptureKit helper sources or copied build tree in the npm package


kaeawc:
@coderabbitai review

Issue #3350, opened 2026-07-06
Implement client drag-to-swipe via daemon input API


kaeawc:
## Overview
Implement drag-to-swipe in `DeviceScreenView` using the daemon input API.

In device-control mode, a pointer drag on the rendered device screen should send a coordinate swipe through the typed daemon input helper. Viewport pan must remain available through the control-mode contract.

Part of #1099.
Depends on #3348.

## Who this is for
The **desktop app** and **third-party MCP daemon clients** using `input/swipe`. The **IDE plugin is not a consumer**; drag continues to mean viewport pan there.

The drag threshold and cancellation rules are **client-side policy, not daemon behavior** — the daemon will faithfully execute whatever swipe it is handed. Document the chosen threshold semantics so other clients converge on comparable behavior instead of each inventing one.

## Scope
- Use coordinate mapping from #3346.
- Use typed daemon swipe helper from #3345.
- Convert drag start/end positions to device coordinates.
- Define drag threshold and cancellation behavior, and document it as client-side policy.
- Preserve viewport pan behavior according to the mode contract.
- Reuse the post-input refresh behavior from #3348.
- Add focused tests with fake clients.

## Out of Scope
- Multi-finger gestures.
- Pinch/zoom forwarding.
- Text or keyboard forwarding.
- Daemon swipe endpoint changes.
- Any `ide-plugin` change.

## Acceptance Criteria
- Dragging in control mode sends one daemon swipe request with mapped start/end coordinates.
- Short movements below the threshold do not accidentally send swipes.
- Cancelled or out-of-bounds drags do not send malformed input.
- Dragging outside control mode preserves viewport pan/inspector behavior.
- Successful swipes use the refresh policy from #3348.
- The threshold/cancellation policy is documented for other daemon clients.
- Tests use fakes and do not require real daemon/device connections.

## Test Plan
- Add UI/fake-client tests for control-mode drag-to-swipe.
- Add threshold and cancellation tests.
- Add regression tests for viewport pan outside control mode.
- Run the narrowest relevant Gradle test task for `android/desktop-core`.
- Before PR, run the validation required by `/ship-issue`.

## /ship-issue Notes
- Start by reading #3346, #3348, and the typed swipe helper from #3345.
- Keep this issue swipe-only; defer pinch and multi-finger input.
- Stop after the implementation plan for approval before editing.

kaeawc:
Latest `main` status before `/ship-issue`:

[...]

Issue #3348, opened 2026-07-06
Refresh client inspector state after forwarded input


kaeawc:
## Overview
Refresh `DeviceScreenView` inspector state after forwarded input actions.

After a client sends input through the daemon API, the inspector should update predictably instead of leaving stale hierarchy, selection, or screenshot state on screen.

Part of #1099.
Depends on #3347.

## Who this is for
The **desktop app** and **third-party MCP daemon clients**. The refresh question is not desktop-specific: any client that forwards input and also renders observed state has to decide when to re-observe. The policy defined here should therefore be written as **client guidance backed by the daemon's existing observation stream**, with the desktop app as the reference implementation. The **IDE plugin is not a consumer** and its inspector refresh behavior must not change.

## Scope
- Define the refresh behavior after successful forwarded input.
- Use existing observe/stream update mechanisms where appropriate; prefer consuming the daemon's existing stream signals over client-side polling, so third-party clients can follow the same policy.
- Handle failed input without incorrectly refreshing to a misleading state.
- Decide how selection/hover state should behave after input changes the screen.
- State the policy in a form a non-desktop client can follow.
- Add focused tests for success and failure paths.

## Out of Scope
- Adding new input actions.
- Changing daemon input endpoints.
- Video streaming transport changes.
- Broad layout inspector refactors.
- Any `ide-plugin` change.

## Acceptance Criteria
- After a successful click-to-tap action, the inspector refreshes or consumes the next stream update according to a documented policy.
- The policy is expressed in terms of daemon signals a third-party client also receives, not desktop-only internals.
- Failed input actions do not clear useful state unless the existing UI pattern requires it.
- Selection and hover state after input are deterministic and tested.
- Tests use fake clients/state and do not require real devices.
- The behavior is documented in code or nearby docs, since later input actions and other clients depend on it.
- Inspector-only hosts (`ide-plugin`) see no behavior change.

## Test Plan
- Add focused Kotlin tests for successful input refresh behavior.
- Add focused Kotlin tests for failed input behavior.
- Add regression coverage for selection/hover clearing or preservation.
- Run the narrowest relevant Gradle test task for `android/desktop-core`.
- Before PR, run the validation required by `/ship-issue`.

## /ship-issue Notes
- Start by reading the click-to-tap implementation from #3347.
- Treat refresh semantics as the main deliverable; do not add new gestures here.
- Stop after the implementation plan for approval before editing.

kaeawc:
[...]

Issue #1099, opened 2026-01-30
Screen Streaming: Touch input forwarding (planned)


kaeawc:
## Overview
Enable touch input forwarding from the IDE plugin's device screen view to the actual device, allowing interactive control.

## Status
**Planned, not yet scheduled.** This is a future enhancement after core streaming is stable.

## Prerequisites
- #1031 Android video server implemented
- #1033 IDE plugin video display implemented
- Screen streaming stable and performant

## Future Feature
This is planned for future implementation. See [Screen Streaming Design Doc](docs/design-docs/mcp/observe/screen-streaming.md) - "Future: Touch Input (Planned, Not Implemented)".

### Click-to-Tap
- [ ] Capture mouse clicks on DeviceScreenView
- [ ] Map click coordinates to device coordinates
- [ ] Account for scaling and aspect ratio
- [ ] Send tap command to device

### Drag Gestures
- [ ] Capture mouse drag on DeviceScreenView
- [ ] Convert to swipe gesture with start/end points
- [ ] Support multi-point drag for complex gestures

### Keyboard Input
- [ ] Capture keyboard events when DeviceScreenView focused
- [ ] Forward key events to device
- [ ] Support special keys (back, home, etc.)
- [ ] Handle IME composition

### Touch Protocol
```kotlin
// IDE Plugin → MCP Server → Device
sealed class TouchEvent {
    data class Tap(val x: Float, val y: Float) : TouchEvent()
    data class Swipe(val x1: Float, val y1: Float, val x2: Float, val y2: Float, val duration: Long) : TouchEvent()
    data class KeyPress(val keyCode: Int) : TouchEvent()
}
```

### Implementation Notes
- Android: Use accessibility service or video server for injection
- iOS: Need to research input injection options
- Latency is critical for interactive feel
- May need visual feedback for touch points

## Acceptance Criteria (Future)
- Tapping on video view triggers device tap
- Dragging on video view triggers device swipe
- Keyboard input forwarded to device
- Latency under 50ms for responsive feel

PR #4513, opened 2026-07-26
feat(desktop): client keyboard, text and button forwarding via the daemon input API (#3351)

### LINKED ISSUE #3351: Implement client keyboard and button forwarding via daemon input API
## Overview
Implement keyboard, text, and device-button forwarding from `DeviceScreenView` using the daemon input API.

When the screen view is focused in device-control mode, safe keyboard input should forward to the active device through the typed daemon helpers without swallowing the host application's normal shortcuts.

Part of #1099.
Depends on #3350.

## Who this is for
The **desktop app** and **third-party MCP daemon clients** using `input/pressButton`, `input/typeText`, and `input/key`. The **IDE plugin is not a consumer** — no keyboard forwarding is added there, which also removes the original concern about competing with IDE keymaps.

The host-shortcut question does not disappear, it changes owner: the desktop app has its own menu accelerators and window shortcuts, and any third-party client will have its own. So the deliverable is a **forwarding policy defined in terms of focus and mode** — which keys forward, which stay with the host — that a client author can apply to whatever host they are embedding in, rather than a rule tuned to one IDE's keymap.

## Scope
- Forward supported device/navigation buttons through the typed daemon `pressButton` helper.
- Forward printable text through the typed daemon text helper when appropriate.
- Forward discrete keys through the typed daemon key helper if implemented.
- Define focus and mode rules for when keyboard events go to the device.
- Preserve the host application's normal shortcuts and text navigation outside the approved forwarding rules; state the rule generically (modifier-bearing chords stay with the host unless explicitly listed) rather than enumerating one host's keymap.
- Reuse the post-input refresh behavior from #3348 where appropriate.
- Add focused tests with fake clients.

## Out of Scope
- Full IME composition support.
- Global keyboard capture outside `DeviceScreenView` focus.
- New daemon input endpoints.
- Complex shortcut remapping UI.
- Any `ide-plugin` change.

## Acceptance Criteria
- Keyboard forwarding only occurs when `DeviceScreenView` is focused and in control mode.
- Host application shortcuts are not swallowed by device forwarding, per a documented, host-agnostic policy.
- Supported button keys route to the typed daemon press-button helper.
- Printable text routes to the typed daemon text helper according to the approved policy.
- Unsupported keys are ignored or reported according to the approved policy.
- The forwarding policy is documented for third-party client authors embedding control mode in a different host.
- Tests use fakes and do not require real daemon/device connections.

## Test Plan
- Add focus/mode gating tests.
- Add fake-client tests for button forwarding.
- Add fake-client tests for printable text forwarding.
- Add regression tests for shortcuts/modifiers that must not be forwarded.
- Run the narrowest relevant Gradle test task for `android/desktop-core`.
- Before PR, run the validation required by `/ship-issue`.

## /ship-issue Notes
- Start by reading #3343, #3345, and the current `DeviceScreenView` key handling.
- Do not implement IME composition in this issue.
- Stop after the implementation plan for approval before editing.

kaeawc:
## Summary

Item 5/6 of milestone 28 (Client Screen Control), built directly on the seam
[#3348](https://github.com/kaeawc/auto-mobile/issues/3348) established and
[#3350](https://github.com/kaeawc/auto-mobile/issues/3350) extended.

Keyboard, text and device-button forwarding become **three more variants of the
existing sealed `DeviceControlInputCommand`**, dispatched through the same
bounded ordered queue, the same error-ordering token and the same post-input
refresh tracker as taps and swipes. No second queue, no new gate, no parallel
error path — `DeviceControlSession.key()` sits beside `tap()` and `swipe()`.

### What was added

- **`DeviceKeyboardInputPolicy`** (`desktop-domain`, pure, Compose-free) — what a
  keystroke sends:
  - Modifier-bearing chords (Ctrl / Alt / Meta) **stay with the host** unless
    explicitly opted in via `forwardedChordKeys` (empty by default). Shift is
    deliberately *not* a chord modifier — it is how capitals are typed.
  - `Escape` -> one `input/pressButton` with `back`.
  - `Enter` / `Tab` / `Backspace` / `Delete` / arrows -> one `input/key`.
  - A printable character -> one `input/typeText`.
  - Anything else -> nothing, and the event is left **unconsumed**.
- **`DeviceKeyboardEventTranslator`** (`desktop-core`) — the only Compose-aware
  piece, so a third-party client reuses the policy verbatim with its own
  translation.
- **`DeviceScreenView`** gains `onControlKey` and takes keyboard focus when it
  *enters* control mode. Forwarding is gated by stock Compose focus routing (a
  key event only reaches the handler when the view is on the focus path) plus the
  control-mode check. The handler returns the client's own "did you forward it?"
  answer as its `onKeyEvent` result, so a declined chord is never swallowed — the
  guarantee is structural rather than a separate pass-it-on path.

### Focus + mode rule

Forwarding happens only when the device-screen view **holds keyboard focus** AND
is in **Control** mode, and is additionally inert without a frame snapshot (the
same fail-closed rule taps and drags use).

### Host-shortcut policy (host-agnostic)

Stated in terms of modifiers rather than any one host's keymap, because the host
is not knowable from the policy: the desktop app has its own menu accelerators
and a third-party embedder has different ones. Ctrl/Alt/Meta chords are refused
by default and left unconsumed; a client that knows its host leaves a particular
chord unclaimed opts that key in explicitly. The `ide-plugin` is not a consumer.

## What was tested

All against fakes and injected dispatchers — no real daemon, device, socket or
timer.

- `DeviceKeyboardInputPolicyTest` (10) — the chord rule pinned from **both**
  directions (every chord modifier must stay with the host; Shift must not, or
  capitals become untypable), key-over-character precedence, the
  control-character filter, and the explicit chord allowlist.
- `DeviceKeyboardEventTranslatorTest` (4) — toolkit-key translation, letters
  carrying no device key, supplementary code points dropped.
- `DeviceControlSessionKeyboardTest` (8) — the exact daemon payloads
  (`inputPressButton("back", "android", "emulator-5554")`,
  `inputTypeText("a", …, submit = null)`, `inputKey("enter", …)`), a
  tap-then-type-then-Escape sequence draining in order on the one queue, the
  reused `AwaitingSnapshot` refresh transition, a daemon rejection surfacing
  verbatim, and `reset()` dropping a queued keystroke.
- `DeviceScreenViewKeyboardTest` (8) — observes **both** what the view forwarded
  and what an ancestor host handler still received, so "host shortcuts are not
  swallowed" is checked rather than assumed. Covers focus gating, mode gating,
  the no-snapshot fail-closed case, and one press forwarding exactly once.
- Extended `DeviceControlInputDispatcherTest` and `DeviceControlInputForwarderTest`.

Every significant rule was mutation-checked (revert the rule, confirm a test goes
red, restore): the chord guard, Shift-as-chord, key-over-character precedence,
the control-character filter, unconditional consumption, the mode gate, the
KeyDown-only gate, the control-mode focus request, button-vs-text routing, and
returning consumed for an ignored keystroke — all ten produced failures.

### Validation

`:desktop-core:test` (738 tests, 0 failures), `:desktop-core:detektMain`,
`:desktop-core:detektTest`, `:desktop-domain:detektMain`,
`:desktop-domain:detektTest`, `:desktop-app:compileKotlin`,
`:ide-plugin:compileKotlin` — all green with `--rerun-tasks`. ktfmt
`--google-style` 0.64 clean on all 13 touched Kotlin files. No TypeScript changed.

**Zero `ide-plugin` changes** — it is inspector-only and never opts into control
mode, so it produces no daemon input and its focus behavior is untouched.

Closes [#3351](https://github.com/kaeawc/auto-mobile/issues/3351)


kaeawc:
Review addressed in a12c27665 — all seven findings, one commit.

## P1-A — Android text input was destructive: **fixed** (not disabled)

`input/typeText`'s default Android path is `ACTION_SET_TEXT`, which replaces the
focused field. Confirmed all four existing `InputTextMode`s are replace-shaped
(`a11y` sets the whole string; `eventAll`/`eventOnly` clear first), so none was
usable per-keystroke.

Added a non-destructive **`append`** mode:
- `InputText.executeAndroidAppendTextInput` — builds an ASCII key-event plan per
  character and executes it. **No clear, no `requestSetText`.** Reuses the
  existing `buildAsciiKeyEventPlan` / `executeKeyEventPlan` machinery.
- Unmappable characters **fail** rather than falling back to `a11y` — that
  fallback is precisely the destruction the mode exists to prevent.
- `InputText.appendText()` bypasses `execute()`'s observe round trip, so a
  per-keystroke call stays one round trip.
- `input/typeText` accepts `mode: "append"`; any other value is rejected.
- Kotlin `inputTypeText(..., append: Boolean)`; the forwarder always passes
  `append = true`.

**iOS: printable text is not forwarded at all.** Its control proxy exposes only
`requestSetText`, so there is no non-destructive path; the daemon rejects
`mode: "append"` on iOS and `DeviceKeyboardInputPolicy` refuses text via a
`textSupported` gate, returning a new `TextUnsupported` rejection. Buttons and
discrete keys still work there. Follow-up filed:
[#4519](https://github.com/kaeawc/auto-mobile/issues/4519), linked to
[#1099](https://github.com/kaeawc/auto-mobile/issues/1099).

## P1-B — Shell preview handler swallowed five keys: fixed

`ThreePaneShell` gained `deviceControlCapturesKeys: () -> Boolean`; while true its
preview handler declines every **un-chorded** keystroke, so the event reaches the
focused canvas. Chorded shortcuts (Cmd-0, Cmd-/, Cmd-K) are untouched, which
matches the policy's own "chords belong to the host" rule. `DeviceScreenView`
gained `onControlFocusChanged`, and `AutoMobileContent` feeds
`deviceCanvasFocused && interactionSnapshot != null` into the shell, read at event
time.

Regression coverage, per your note that the test matters more than the fix:
`ThreePaneShellDeviceKeyTest` composes the **real** `ThreePaneShell` and asserts
against its actual navigation callbacks — Up/Down/Enter/Escape claimed by none of
them while capturing, all four still claimed when not, and Cmd-0 still toggling
the pane. `DeviceScreenViewKeyboardTest` adds the same contract against a
preview-consuming stand-in.

## P2-C, P2-D, P2-E, Minor-F, Minor-G

- **C** — control-mode tap now calls `focusRequester.requestFocus()`.
- **D** — `DeviceKeyModifiers.isAltGraphShape` (Ctrl+Alt, no Meta); a Ctrl+Alt
  keystroke that produced a **printable character** types. A real accelerator
  produces no character, so it still stays with the host; adding Meta never
  qualifies. Pinned both ways.
- **E** — `forwardedChordKeys` → `forwardedChords: Set<DeviceChordAllowance>`
  with `OfKey` and `OfCharacter` (case-insensitive), so Ctrl-S is opt-in-able.
- **F** — KDoc and the doc section now describe the Boolean as *"should the view
  consume this event"*, explicitly noting a queue rejection still consumes.
- **G** — asserts against the `S` stroke specifically and checks
  `modifiers.meta`, not just `hasChordModifier`.

## Validation (all `--rerun-tasks`)

| Check | Result |
| --- | --- |
| `:desktop-core:test` | **750 tests, 0 failures** (6 pre-existing skips) |
| detekt desktop-core main + test, desktop-domain main + test | pass |
| `:desktop-app:compileKotlin`, `:ide-plugin:compileKotlin` | pass |
| ktfmt `--google-style` 0.64, full tree | clean |
| `bun test` (action + daemon + docs suites) | **984 tests, 0 failures** |
| `bun run lint` | pass, no new suppressions |
| `bun run typecheck` | no new errors (206 baselined) |

**Zero `ide-plugin` changes.** Tap/swipe/inspector regressions
(`DeviceScreenViewControlTest`, `DeviceControlSessionTest`,
`DeviceControlInputDispatcherTest`) all green.

## Mutation checks — 13 mutations, 13 failed a test

| Mutation | Test that went red |
| --- | --- |
| forwarder sends `append=false` | `a printable character reaches the daemon as typeText` |
| policy ignores `textSupported` | policy + session iOS text tests |
| append clears the field first | 2 `InputText` tests + socketServer routing test |
| append falls back to a11y | `append fails rather than falling back…` |
| socketServer ignores `append` | `append mode never reaches the destructive requestSetText path` |
| shell never stands down | `the shell stops claiming device keys…` (real shell) |
| shell stands down for chords too | `chorded shell shortcuts survive the stand-down` |
| no click-to-focus | `clicking the control canvas restores keyboard focus` |
| no AltGr allowance | `AltGr composed characters type…` |
| AltGr allowance too loose | `a real Ctrl-Alt shortcut still stays with the host` + 2 |
| allowlist ignores characters | `a printable chord can be opted in by its character` |
| translator drops modifiers | `the chord modifiers reach the client…` (the Minor-G fix) |
| focus changes not reported | `focus changes are reported to the host` |

## Remaining honest gaps

- The append path's **device behavior** is unit-tested (key events issued, no
  clear, no setText) but not verified against real hardware.
- The socketServer append test asserts the destructive path is *not* taken; it
  cannot assert success end-to-end, because the append path shells out to adb
  and the socket harness provides none.
- Bounded-queue-full still consumes the keystroke by design, still untested (no
  blocking-consumer harness).

PR #4583, opened 2026-07-27
fix: keep cleanup from blocking startup

### LINKED ISSUE #4581: Daemon startup gates JSON-RPC readiness on unbounded cleanup sweeps
## Summary

The daemon's JSON-RPC readiness is gated by two unbounded best-effort cleanup sweeps. On long-lived or busy hosts, those startup sweeps can exceed a client's fixed startup budget, leaving the stdio subprocess alive but unable to answer `initialize`.

## Observed behavior

A client spawns AutoMobile as a long-lived stdio subprocess and runs the usual startup sequence:

1. `initialize`
2. Read `serverInfo.version`
3. Call `doctor`
4. Probe `listDeviceImages`

With a roughly 60-second timeout, this intermittently times out while the subprocess remains running, without an exit code or stderr. The frequency increases with host uptime. It did not reproduce with `0.0.44`.

This is a startup hang, not a version mismatch or crash: the daemon has not become ready enough to answer `initialize`.

## Cause

`main()` awaits both cleanup operations before the daemon can accept JSON-RPC requests:

- [`src/index.ts`](https://github.com/kaeawc/auto-mobile/blob/main/src/index.ts) awaits `AndroidCtrlProxyManager.sweepStalePrefetchDirsOnStartup()`. The sweep reads `os.tmpdir()` and sequentially `stat`s and removes every matching `auto-mobile-prefetch-*` directory. Its cost grows with accumulated scratch directories.
- [`src/index.ts`](https://github.com/kaeawc/auto-mobile/blob/main/src/index.ts) awaits `IOSCtrlProxyManager.reapOrphanedRunnerProcessesOnStartup()` on macOS. The reaper enumerates candidate processes and walks their process trees, so its cost grows with process count and host load.

Neither operation has an overall deadline or work cap, and both are best-effort cleanup of artifacts from prior daemon runs. In contrast, APK, iOS build, and video-server prefetches already start asynchronously.

The reported `0.0.44` to `0.0.45` package growth can further slow cache-cold consumers, but it is secondary: the readiness regression is the awaited unbounded local work.

## Proposed fix

Do not let janitorial sweeps block daemon readiness.

1. Start both sweeps asynchronously from `src/index.ts`, and handle failures as non-fatal warnings.
2. Apply a small timeout to the caller-owned wait so neither sweep can delay `initialize`.
3. Bound work inside the sweep implementations as defense in depth:
   - Process at most a fixed number of stale prefetch directories per startup, with useful logging for skipped work.
   - Apply a deadline and iteration cap while finding and reaping iOS runner trees.

The cleanup may finish later, but daemon startup and RPC readiness must not wait for it.

## Acceptance criteria

- A daemon answers `initialize`, `doctor`, and `listDeviceImages` within the normal startup budget even with many stale `auto-mobile-prefetch-*` directories or a heavily loaded process table.
- Cleanup errors, timeouts, and skipped excess work are logged but never prevent startup.
- Tests cover that startup reaches request readiness while each cleanup promise is still pending or times out.
- Tests cover the internal directory/process work caps or deadline behavior using the existing timer/process seams.

## Immediate mitigation

Until a release contains the fix, restarting the host or clearing stale `auto-mobile-prefetch-*` directories from the operating system temp directory restores normal startup time.

kaeawc:
## Summary

- Start Android stale-prefetch cleanup and macOS iOS-runner cleanup in the background so they cannot delay stdio JSON-RPC readiness.
- Log cleanup failures and five-second timeout observations without failing startup.
- Cap Android prefetch candidates and iOS runner candidates at 20 per startup; bound iOS tree discovery with a fake-clock deadline.

## Root Cause

`src/index.ts` awaited unbounded best-effort cleanup before connecting the MCP server to stdio. Accumulated temp directories or a busy process table could therefore leave a healthy subprocess unable to answer `initialize`.

## Validation

- `AUTOMOBILE_DATA_DIR=/private/tmp/auto-mobile-4581-green bun test test/utils/startupMaintenance.test.ts test/utils/CtrlProxyManager.test.ts test/utils/IOSCtrlProxyManager.test.ts`
- `bun run lint`
- `bun run typecheck`
- `bun run turbo:validate`

Closes #4581

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -188,7 +188,7 @@ async function main() {
     startStartupMaintenance({
       platform: process.platform,
       startAndroidSweep: () => AndroidCtrlProxyManager.sweepStalePrefetchDirsOnStartup(),
-      startIosReap: () => IOSCtrlProxyManager.reapOrphanedRunnerProcessesOnStartup(),
+      startIosReap: () => IOSCtrlProxyManager.startOrphanRunnerReapOnStartup(),
     });
     if (skipCtrlProxyDownload) {
       logger.info(`CtrlProxy downloads disabled (${SKIP_CTRL_PROXY_DOWNLOAD_FLAG} or ${SKIP_CTRL_PROXY_DOWNLOAD_ENV})`);
--- a/src/index.ts
+++ b/src/index.ts
@@ -39,6 +39,7 @@ import {
   setFatalProcessHandler,
   setProcessShutdownHandler,
 } from "./processLifecycle";
+import { startStartupMaintenance } from "./utils/startupMaintenance";
 
 interface FatalLogger {
   error(...args: unknown[]): void;
@@ -181,10 +182,14 @@ async function main() {
     serverConfig.setNetworkMockableEnabled(networkMockable);
     serverConfig.setDismissKeyboardAfterInputEnabled(dismissKeyboardAfterInput);
     serverConfig.setEventAllMarkers(eventAllMarkers);
-    // Reclaim any auto-mobile-prefetch-* scratch dirs leaked by a previously
-    // crashed/killed daemon before its shutdown cleanup ran (#4334). Runs even
-    // when downloads are disabled, since leaked dirs still need reclaiming.
-    await AndroidCtrlProxyManager.sweepStalePrefetchDirsOnStartup();
+    // Reclaim artifacts leaked by previous daemons without gating stdio
+    // readiness. The sweeps are best-effort maintenance, not a prerequisite for
+    // serving a new client (issue #4581).
+    startStartupMaintenance({
+      platform: process.platform,
+      startAndroidSweep: () => AndroidCtrlProxyManager.sweepStalePrefetchDirsOnStartup(),
+      startIosReap: () => IOSCtrlProxyManager.reapOrphanedRunnerProcessesOnStartup(),
+    });
     if (skipCtrlProxyDownload) {
       logger.info(`CtrlProxy downloads disabled (${SKIP_CTRL_PROXY_DOWNLOAD_FLAG} or ${SKIP_CTRL_PROXY_DOWNLOAD_ENV})`);
     } else {
@@ -193,10 +198,9 @@ async function main() {
       AndroidCtrlProxyManager.prefetchApk();
     }
 
-    // Reap orphaned iOS runners before the daemon accepts iOS work, then start
-    // the build prefetch asynchronously so it can be ready for first use.
+    // Start the iOS build prefetch asynchronously so it can be ready for first
+    // use. Runner cleanup above is also asynchronous and bounded.
     if (process.platform === "darwin") {
-      await IOSCtrlProxyManager.reapOrphanedRunnerProcessesOnStartup();
       if (skipCtrlProxyDownload) {
         logger.info(`CtrlProxy iOS prefetch disabled (${SKIP_CTRL_PROXY_DOWNLOAD_FLAG} or ${SKIP_CTRL_PROXY_DOWNLOAD_ENV})`);
       } else {
--- a/src/utils/CtrlProxyManager.ts
+++ b/src/utils/CtrlProxyManager.ts
@@ -29,6 +29,8 @@ import {
   DefaultAndroidPrerequisiteDetector
 } from "./android-cmdline-tools/AndroidPrerequisiteDetector";
 
+export const MAX_STALE_PREFETCH_DIRS_PER_STARTUP = 20;
+
 /**
  * Android-specific accessibility-service lifecycle, extending the
  * platform-agnostic {@link ProxyManager}.
@@ -424,24 +426,42 @@ export class AndroidCtrlProxyManager implements CtrlProxyManager {
       return;
     }
 
+    const candidates = entries.filter(
+      entry => entry.isDirectory() && entry.name.startsWith(AndroidCtrlProxyManager.PREFETCH_DIR_PREFIX)
+    );
     const now = timer.now();
-    let reclaimed = 0;
-    for (const entry of entries) {
-      if (!entry.isDirectory() || !entry.name.startsWith(AndroidCtrlProxyManager.PREFETCH_DIR_PREFIX)) {
-        continue;
-      }
+    const staleDirectories: string[] = [];
+    for (const entry of candidates) {
       const dir = path.join(tempRoot, entry.name);
       try {
         const stats = await fs.stat(dir);
-        if (now - stats.mtimeMs < AndroidCtrlProxyManager.STALE_PREFETCH_MAX_AGE_MS) {
-          continue;
+        if (now - stats.mtimeMs >= AndroidCtrlProxyManager.STALE_PREFETCH_MAX_AGE_MS) {
+          staleDirectories.push(dir);
         }
+      } catch (error) {
+        // Per-entry best-effort: a concurrent process may remove the same dir,
+        // or it may vanish mid-sweep. Skip it and keep going.
+        logger.debug(`[CTRL_PROXY] Failed to inspect stale prefetch dir ${dir}: ${error}`);
+      }
+    }
+
+    const skippedCandidateCount = Math.max(0, staleDirectories.length - MAX_STALE_PREFETCH_DIRS_PER_STARTUP);
+    if (skippedCandidateCount > 0) {
+      logger.warn(
+        `[CTRL_PROXY] Prefetch sweep skipped ${skippedCandidateCount} stale prefetch dir(s) after ` +
+        `reaching the ${MAX_STALE_PREFETCH_DIRS_PER_STARTUP}-directory startup cap`
+      );
+    }
+
+    let reclaimed = 0;
+    for (const dir of staleDirectories.slice(0, MAX_STALE_PREFETCH_DIRS_PER_STARTUP)) {
+      try {
         await fs.rm(dir, { recursive: true, force: true });
         reclaimed++;
       } catch (error) {
         // Per-entry best-effort: a concurrent process may remove the same dir,
         // or it may vanish mid-sweep. Skip it and keep going.
-        logger.debug(`[CTRL_PROXY] Failed to sweep stale prefetch dir ${dir}: ${error}`);
+        logger.debug(`[CTRL_PROXY] Failed to remove stale prefetch dir ${dir}: ${error}`);
       }
     }
 
--- a/src/utils/CtrlProxyManager.ts
+++ b/src/utils/CtrlProxyManager.ts
@@ -30,6 +30,7 @@ import {
 } from "./android-cmdline-tools/AndroidPrerequisiteDetector";
 
 export const MAX_STALE_PREFETCH_DIRS_PER_STARTUP = 20;
+export const STALE_PREFETCH_SWEEP_DEADLINE_MS = 5_000;
 
 /**
  * Android-specific accessibility-service lifecycle, extending the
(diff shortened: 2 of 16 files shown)


src/utils/CtrlProxyManager.ts
```diff
@@ -424,12 +426,20 @@ export class AndroidCtrlProxyManager implements CtrlProxyManager {
...
+      );
+    }
+
     const now = timer.now();
     let reclaimed = 0;
-    for (const entry of entries) {
-      if (!entry.isDirectory() || !entry.name.startsWith(AndroidCtrlProxyManager.PREFETCH_DIR_PREFIX)) {
-        continue;
-      }
+    for (const entry of candidates.slice(0, MAX_STALE_PREFETCH_DIRS_PER_STARTUP)) {
```
kaeawc: The cap is applied before the eligibility filter, so ineligible entries can consume the entire budget — `candidates` is selected by name prefix only, and the `mtimeMs` check that decides whether a dir is actually stale happens inside the loop at line 446. Twenty directories younger than an hour spend the whole cap and the stale ones are never `stat`ed. Because `readdir` order is stable for a given directory, the same stale dir stays starved across restarts rather than being reclaimed on the next one.

Each daemon lifecycle leaks one dir, so twenty dirs younger than 60 minutes is just a machine that restarted the daemon twenty times in an hour — a reconnecting MCP client or a CI loop. That is the same "accumulated scratch directories" state the issue is about, which makes this the case where the sweep most needs to work.

Reproduce: plant 20 dirs aged 5 minutes and 1 aged 2 hours, run `sweepStalePrefetchDirsOnStartup`, and the 2-hour dir survives.

The counting is also slightly off independent of the ordering: the warning at line 434 calls the skipped entries "stale prefetch dir candidate(s)" when their staleness was never evaluated.

Suggested change for guaranteeing progress on stale entries — cap the destructive work rather than the candidate list, which keeps `rm` bounded at 20 while letting the cheap `stat` see every candidate. Replaces lines 432-456:

```ts
    const now = timer.now();
    let reclaimed = 0;
    let skippedCandidateCount = 0;
    for (const entry of candidates) {
      if (reclaimed >= MAX_STALE_PREFETCH_DIRS_PER_STARTUP) {
        skippedCandidateCount++;
        continue;
      }
      const dir = path.join(tempRoot, entry.name);
      try {
        const stats = await fs.stat(dir);
        if (now - stats.mtimeMs < AndroidCtrlProxyManager.STALE_PREFETCH_MAX_AGE_MS) {
          continue;
        }
        await fs.rm(dir, { recursive: true, force: true });
        reclaimed++;
      } catch (error) {
        // Per-entry best-effort: a concurrent process may remove the same dir,
        // or it may vanish mid-sweep. Skip it and keep going.
        logger.debug(`[CTRL_PROXY] Failed to sweep stale prefetch dir ${dir}: ${error}`);
      }
    }
    if (skippedCandidateCount > 0) {
      logger.warn(
        `[CTRL_PROXY] Prefetch sweep skipped ${skippedCandidateCount} stale prefetch dir candidate(s) after ` +
        `reaching the ${MAX_STALE_PREFETCH_DIRS_PER_STARTUP}-directory startup cap`
      );
    }
```

This keeps the existing cap test green: 21 stale dirs still reclaims 20, leaves 1, and warns "skipped 1 stale prefetch dir".

The regression it risks is that a directory with tens of thousands of `auto-mobile-prefetch-*` entries now pays one `stat` each instead of twenty total. That is cheap relative to the `rm` calls it replaces, and the sweep is off the readiness path now anyway.

Worth adding a test with mixed ages, since the current cap test builds all 21 dirs at `2 * HOUR_MS` and so passes either way.

src/utils/IOSCtrlProxyManager.ts
```diff
@@ -385,7 +389,22 @@ export class IOSCtrlProxyManager implements CtrlProxyIosManager {
       return;
     }
 
-    for (const pid of pids) {
+    const candidates = pids.slice(0, MAX_STARTUP_ORPHAN_RUNNER_CANDIDATES);
```
kaeawc: The reorder in `findStartupCandidatePids` is the right fix for the case that matters — finding the `CtrlProxyUITests-Runner` child is enough, because `findDaemonManagedRunnerTreeRoot` walks up to the orphaned `xcodebuild` root and `terminateProcessTree` takes the whole tree. Good, and the new test discriminates properly.

One residual, narrower than the original: an orphaned CtrlProxy-shaped `xcodebuild` whose `Runner` child has already exited is discovered only by `pgrep -x xcodebuild`, which now sorts last, so it can still sit behind 20 unrelated builds. That is the shape modelled by the sibling cap test at line 2312, where every candidate is an `xcodebuild ... -only-testing:CtrlProxyUITests` command with no `Runner` process.

Not a blocker — the 5 s deadline bounds the cost either way, and a parentless `xcodebuild` with no runner is a narrow leftover. If you want it closed, capping on inspected-and-matched candidates rather than raw PIDs would do it, and would make this file consistent with whatever lands on the Android side.

src/index.ts
```diff
@@ -181,10 +182,14 @@ async function main() {
...
-    // crashed/killed daemon before its shutdown cleanup ran (#4334). Runs even
-    // when downloads are disabled, since leaked dirs still need reclaiming.
-    await AndroidCtrlProxyManager.sweepStalePrefetchDirsOnStartup();
+    // Reclaim artifacts leaked by previous daemons without gating stdio
+    // readiness. The sweeps are best-effort maintenance, not a prerequisite for
+    // serving a new client (issue #4581).
+    startStartupMaintenance({
+      platform: process.platform,
+      startAndroidSweep: () => AndroidCtrlProxyManager.sweepStalePrefetchDirsOnStartup(),
+      startIosReap: () => IOSCtrlProxyManager.reapOrphanedRunnerProcessesOnStartup(),
```
kaeawc: The comment this replaced stated an invariant — "Reap orphaned iOS runners *before the daemon accepts iOS work*" — and nothing restores it. The reap now runs concurrently with request serving.

The collision is concrete. In `start()`, `defaultPortIsHealthyForDevice` (`src/utils/IOSCtrlProxyManager.ts:742-748`) adopts a process on `DEFAULT_PORT` on the strength of a health probe alone, inspecting no process identity. A runner orphaned by a crashed daemon that is still healthy on 8765 satisfies that probe — and is simultaneously a valid reap target, since its `ppid === 1` passes `requireOrphanedRoot`. So a first iOS request landing inside the reap window can adopt a runner and have `terminateProcessTree` kill it mid-request. The same adoption exists at lines 810-821. Under the old ordering this was structurally impossible.

To be fair about the size: the daemon's own freshly spawned runner is safe (`detached: true` with the daemon as parent, so `ppid !== 1`), and the window is a few hundred ms up to the 5 s deadline. But it only opens after a crash-restart, which is exactly the scenario the reap exists to serve.

Suggested change for keeping the reap off the readiness path while restoring the ordering where it matters: hold the promise `startStartupMaintenance` already creates (or a `IOSCtrlProxyManager.startupReapPromise` static) and `await` it once at the top of the iOS setup path, before the adoption probe. It is already bounded by `STARTUP_ORPHAN_RUNNER_REAP_DEADLINE_MS`, so the wait is capped at 5 s and paid only by an actual iOS request — `initialize`, `doctor`, and `listDeviceImages` stay unblocked, which is all #4581 asks for.

I would not call this blocking on its own; landing this and following up is defensible. The false negative if left: an intermittent, load-dependent iOS failure right after a crash-restart, presenting as a runner that dies mid-request with no obvious cause.

src/utils/IOSCtrlProxyManager.ts
```diff
@@ -1726,6 +1754,9 @@ export class IOSCtrlProxyManager implements CtrlProxyIosManager {
     }
 
     while (parentPid !== undefined && parentPid > 1 && !visitedPids.has(parentPid)) {
+      if (options.shouldContinue && !options.shouldContinue()) {
```
kaeawc: On budget exhaustion this returns `null`, which is the same value that means "not a daemon-managed tree". Harmless in the startup caller, which re-checks the deadline immediately after the call and bails — but the instance wrapper at line 1734 does `(await findDaemonManagedRunnerTreeRoot(...)) ?? process.pid`, reading `null` as "this process is its own tree root" and proceeding to terminate it. The moment anyone threads `shouldContinue` through that caller, a timeout silently becomes a kill order.

This branch is also unexercised. Both new fixtures give every candidate `ppid: 1`, so the `while (parentPid !== undefined && parentPid > 1 && ...)` loop body never runs; in the deadline test the clock is pushed past the deadline inside the `ps -p` handler, so what actually fires is the post-call re-check at line 427, not this check.

A candidate with multi-level ancestry — `ppid` pointing at a shell or `xcodebuild` parent whose `getProcessInfo` advances the fake clock — would cover it. Separately, returning a distinct sentinel or a discriminated result for "budget exhausted" would remove the trap rather than document it. Not a blocker.

src/utils/startupMaintenance.ts
```diff
@@ -0,0 +1,70 @@
+import { defaultTimer, type Timer } from "./SystemTimer";
+import { logger, type Logger } from "./logger";
+
+export const STARTUP_MAINTENANCE_TIMEOUT_MS = 5_000;
```
kaeawc: Nit: this doesn't time anything out — `startBackgroundCleanup` logs a warning at 5 s and lets the work continue, which is correct for the design but not what the name says. Worth renaming so the next reader doesn't assume the sweeps are bounded by it; the Android sweep in particular has no deadline at all, only the directory cap.

```suggestion
export const STARTUP_MAINTENANCE_SLOW_WARNING_MS = 5_000;
```

test/utils/startupMaintenance.test.ts
```diff
@@ -0,0 +1,77 @@
...
+      },
+      startIosReap: () => {
+        iosStarted = true;
+        return pendingPromise();
+      },
+      timer,
+      logger: log,
+    });
+
+    await expect(Promise.resolve("request-ready")).resolves.toBe("request-ready");
```
kaeawc: Nit: this asserts a literal promise resolves to its own value, which is true regardless of the code under test. The real claim — that `startStartupMaintenance` returned synchronously while both cleanups are still pending — is carried entirely by the two `expect(...Started).toBe(true)` lines below, which can only run if the call returned. Dropping this line makes the test's actual claim legible.

```suggestion
```

kaeawc (commented):
Reviewed at `02fb0f297`. Three lenses: runtime behavior, delivery/enforcement, and a generated third on startup-path completeness and process lifecycle (chosen because the change converts awaited startup work into background work, which the two constant lenses cover only obliquely).

**This does fix the reported problem, and I verified it empirically rather than by reading.** With 400 aged `auto-mobile-prefetch-*` dirs planted in `TMPDIR`, the stdio server answered `initialize` in 351 ms and reclaimed exactly 20 dirs. The two sweeps were the only unbounded best-effort work between `main()` and `server.connect` — this is the root cause, not the symptom.

I also checked the concurrency hazard the backgrounding introduces, since the sweeps now overlap the prefetches they used to precede. Both are safe for non-obvious reasons worth recording: the Android sweep skips anything younger than `STALE_PREFETCH_MAX_AGE_MS`, and `startStartupMaintenance` runs before `prefetchApk()` so the `readdir` snapshot predates the new dir; the iOS reap requires `ppid === 1`, and a runner spawned by the live daemon has the daemon as parent. The new tests also genuinely discriminate — neutering the cap and deadline logic in `src/` while keeping the exports fails exactly the three new tests and nothing else.

Findings inline. The Android one is the only one I'd want addressed before merge; the rest are smaller or are follow-ups.

Two things outside the diff, for follow-up issues rather than this PR:

- **`initialize` still sits behind a 60-second ceiling on the stdio path.** Every non-`--daemon-mode` run awaits `runDirectModeStartup` → `applyFeatureFlagStartup` → `ensureMigrations()` → `FileMigrationLock.acquire()`, bounded only by `DEFAULT_MIGRATION_LOCK_TIMEOUT_MS = 60_000` (`src/db/migrationLock.ts:85`). Reproduced: with a live-PID `auto-mobile.db.migrate.lock` held, the server emits no `initialize` response for 60 s. This is pre-existing (`directModeStartup.ts` last changed in `694af7c53`, #3271) and I do not think it explains the reported regression — that symptom scales with host uptime, which matches accumulated tmp dirs, not lock contention. But #4581's acceptance criterion is "answers `initialize` within the normal startup budget", and that is not yet unconditionally true. The stdio front-end opens and migrates the shared DB purely to read feature flags, while proxy mode has a daemon that owns the DB anyway.
- **Cross-process prefetch-dir sweep (pre-existing, from #4335).** A daemon up more than 60 minutes still owns a retained prefetch dir whose mtime is its creation time, so a second AutoMobile process starting up will sweep it out from under the live one. The victim's `consumePrefetchedApk` hits `copyFile` ENOENT and falls back to a normal download — degraded, not fatal. Flagging so it is not mistaken for something the 60-minute guard already covers.

PR #4712, opened 2026-07-30
ci(desktop): build and publish signed desktop installers on release

kaeawc:
## Summary

Adds release engineering for the Compose Desktop app (`android/desktop-app`). The release pipeline shipped every artifact except the desktop app, and PR CI only ran `:desktop-app:build`; nothing ever produced or published the native installers. This wires the DMG/MSI/Deb into the release flow.

- **New reusable workflow** [`.github/workflows/build-desktop-app-installers.yml`](https://github.com/kaeawc/auto-mobile/blob/claude/desktop-app-release-engineering-c0o2df/.github/workflows/build-desktop-app-installers.yml): a per-OS matrix (macOS/Windows/Linux) packages each installer via the existing `gradle-task-run` composite (`:desktop-app:packageDmg` / `packageMsi` / `packageDeb`). On macOS the DMG is signed by Compose (Developer ID cert imported into a temporary keychain) and then **notarized + stapled with `notarytool`** using the App Store Connect API key — the same secrets and pattern as the ScreenCaptureKit helper release. A `sign` input gates the Apple path so callers that don't need signing skip it.
- **`release.yml`** calls it with `sign: true`, downloads the three installers into the release job, and attaches them to the GitHub release (plus a Desktop App notes section). These are direct user downloads, so they are intentionally **not** routed through the runtime-artifact source-checksum provenance gate.
- **`prepare-release.yml`** calls it with `sign: false` as an unsigned packaging smoke gate, so a desktop-packaging break fails before the tag is pushed rather than mid-release.
- **Version-driven packaging**: `packageVersion` is now driven from the release version via `-PdesktopPackageVersion` (falling back to `VERSION_NAME` with any `-SNAPSHOT`/build-metadata suffix stripped), so installer identity matches the npm/Maven coordinates instead of the hardcoded `1.0.0`. The macOS bundle floors the major version at `1` (`0.0.47` → `1.0.47`) because jpackage rejects a `0` major on macOS; MSI/Deb keep the real version.

## Linked Issue

No pre-existing issue tracked this; opened as the desktop app's release-engineering gap. (No `Closes #N`.)

## Validation

- [x] Workflow YAML parses cleanly (all three files); `ktfmt --google-style` applied to the touched Kotlin
- [ ] `scripts/all_fast_validate_checks.sh` — could not fully run in this environment (`shellcheck`, `xmlstarlet`, `lychee`, and the bun/TypeScript-based boundary checks are not installed here); the relevant `ktfmt` check passes and the YAML parses. Full suite runs in CI.
- [x] Gradle DSL changes use documented Compose Multiplatform 1.11.1 APIs (`macOS.signing { }`, `bundleID`, per-format `packageVersion`)
- Real end-to-end packaging (jpackage across all three OSes + macOS notarization) exercises on the runners; PR CI already builds `:desktop-app` on macOS/Windows/Linux.

## Kotlin Reuse Check

- [x] Only `build.gradle.kts` changed; version parsing uses Kotlin stdlib (`substringBefore`, `split`), no new helper files or dependencies
- [x] macOS signing/notarization reuses the existing `scripts/ios/setup-macos-signing-keychain.sh` and the established `notarytool` flow rather than introducing a new mechanism

## Follow-ups

- [ ] macOS package version floors the major at `1` for pre-1.0 releases; revisit the mapping when the project reaches a real `1.0.0` (noted inline in `build.gradle.kts`).
- [ ] The app icons in `desktop-app/src/main/resources/icons/` are small placeholders — worth replacing with real assets before wide distribution.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mbhf2LovdeZmjSvsg1cBsY)_

### DIFF
--- a/.github/workflows/build-desktop-app-installers.yml
+++ b/.github/workflows/build-desktop-app-installers.yml
@@ -83,9 +83,32 @@ jobs:
       - name: "Git Checkout"
         uses: actions/checkout@v6
         with:
+          # persist-credentials stays enabled (default) so the Git LFS step below
+          # can authenticate; a full `lfs: true` would pull every LFS object in
+          # the repo (videos, archives), so fetch only the icons instead.
           lfs: false
           ref: ${{ inputs.checkout-ref || github.sha }}
-          persist-credentials: false
+
+      # app-icon.png (Linux) and app-icon.ico (Windows) are Git LFS-tracked
+      # (.gitattributes), so a non-LFS checkout leaves them as pointer files that
+      # jpackage would embed as invalid icon data. Materialize just the icons.
+      - name: "Materialize desktop app icons (Git LFS)"
+        shell: bash
+        run: git lfs pull --include="android/desktop-app/src/main/resources/icons/**"
+
+      # Fail closed on a malformed version before packaging. Also prevents any
+      # `${{ inputs.version }}` from carrying shell metacharacters into later run
+      # steps. Mirrors the semver shape accepted by bump-versions.sh.
+      - name: "Validate installer version"
+        shell: bash
+        env:
+          VERSION: ${{ inputs.version }}
+        run: |
+          set -euo pipefail
+          if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+([-+][0-9A-Za-z.-]+)?$ ]]; then
+            echo "Invalid installer version: $VERSION (expected MAJOR.MINOR.PATCH semver)" >&2
+            exit 1
+          fi
 
       # jpackage builds the .deb with dpkg-deb under fakeroot; fakeroot is not
       # preinstalled on the ubuntu runner image.
@@ -167,6 +190,10 @@ jobs:
       - name: "Stage installer"
         shell: bash
         working-directory: android
+        env:
+          # Consumed through the environment, not interpolated into the script
+          # body, so the version can never expand into runner-side shell code.
+          VERSION: ${{ inputs.version }}
         run: |
           set -euo pipefail
           SRC=$(find "desktop-app/build/compose/binaries/main/${{ matrix.format }}" \
@@ -177,7 +204,7 @@ jobs:
           fi
           DEST_DIR="desktop-app/build/desktop-dist"
           mkdir -p "$DEST_DIR"
-          DEST="${DEST_DIR}/AutoMobile-${{ inputs.version }}-${{ matrix.platform }}.${{ matrix.format }}"
+          DEST="${DEST_DIR}/AutoMobile-${VERSION}-${{ matrix.platform }}.${{ matrix.format }}"
           cp "$SRC" "$DEST"
           echo "Staged $SRC -> $DEST"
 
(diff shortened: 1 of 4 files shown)


kaeawc:
Triaged the automated review findings and pushed fixes in 5bbeb1f:

**Fixed**
- **LFS icons (Codex P2).** Confirmed real — `app-icon.png`/`app-icon.ico` are Git LFS pointers, so the non-LFS checkout would have fed jpackage 127-byte pointer text as the Linux/Windows icons. Added a scoped `git lfs pull --include=…/icons/**` (not a full `lfs: true`, which would pull every video/archive in the repo).
- **Version template injection (CodeRabbit/zizmor Major).** Added a strict-semver validation gate on `version` and route it through an env var in the staging step, so it can't expand into runner-side shell code.

**Documented (accepted limitation)**
- **Prerelease version collapse (Codex P2).** jpackage has no field for a prerelease qualifier, so an RC packages with the same OS-level version as its final. Noted inline in `build.gradle.kts`; prereleases are rare and npm/tag identity still distinguishes them.

**Skipped (out of scope)**
- **`required: "false"` string in `.github/actions/gradle-task-run/action.yml` (CodeRabbit Major).** That's a pre-existing shared composite action (13 such fields) that I didn't touch and that works in production; `actionlint` isn't a CI gate in this repo, so it isn't blocking this PR. Happy to fix it in a separate change if you'd like the whole file cleaned up.

---

kaeawc:
Handled the two new review findings on `5bbeb1f` (pushed `5b28d6e`):

**Fixed — broken placeholder icons (Codex P2).** Verified: `app-icon.icns` is actually a 1×1 PNG (magic `89 50 4e 47`), and `app-icon.png`/`.ico` are LFS pointers to tiny stubs. Packaging is the first code path to exercise them, and jpackage can reject an invalid `.ico`/`.icns` and fail the Windows/macOS build. Since these carry no real branding and valid replacements aren't available in this change, I dropped the custom `iconFile` config so jpackage uses its default icon and packaging stays green. This also made the earlier "materialize LFS icons" step unnecessary, so it's removed (real branded icons + `iconFile` remain a follow-up).

**Won't fix here — Windows MSI Authenticode signing (Codex P1).** Correct that the MSI is unsigned, but there's no Windows code-signing certificate in this repo (only Apple secrets), so `signtool` isn't wireable here. The release notes scope the "signed & notarized" claim to the macOS DMG only — Windows/Linux are attached unsigned. Windows Authenticode signing is a separate follow-up gated on obtaining a cert.

Still-open bot threads on the pre-existing `required: "false"` composite field and the prerelease-version note were addressed in my earlier comment (out of scope / documented limitation).

---

kaeawc:
Fixed in `1350a5d` — pinned a stable `windows.upgradeUuid` in `build.gradle.kts` (Codex P1). Without it jpackage mints a fresh Windows Installer UpgradeCode per build, so consecutive MSIs wouldn't be recognized as upgrades and would install side-by-side. The committed UUID must stay constant across future releases (noted inline).

---

Issue #4754, opened 2026-07-30
Add a socket seam and tests for VideoStreamWriter recovery paths (plus untested pure logic)


kaeawc:
## Summary

`VideoStreamWriter` — the class that concentrates the video-server's write-failure, client-reconnect-race, and keyframe-command-channel recovery logic — has **no tests reaching past its inner helpers**, because it news up `LocalServerSocket` directly and has no injectable socket/stream seam. Several other pure/robustness paths in the module are also untested, and one load-bearing contract (`writePacket`'s return value) is effectively dead code that nothing pins.

## Highest-leverage gap: `VideoStreamWriter` has no socket seam

`android/video-server/src/main/kotlin/dev/jasonpearson/automobile/video/VideoStreamWriter.kt`:
- Constructor takes a `socketName: String` and constructs `LocalServerSocket` directly (`:101`) — no injection point.
- The existing `VideoStreamWriterTest` only exercises the inner `ReconnectWindow` (`VideoStreamWriterTest.kt:9-24`); `ReconnectWindow` expiry is even tested twice (also `VideoStreamProtocolTest.kt:238-257`), while the writer proper has zero coverage.

Untested recovery paths in the writer:
- Write IOException → `closeClientLocked()` + `reconnectWindow.onClientDetached()`, **returns `true` anyway** (`:256-261`).
- No-client short-circuit `outputStream ?: return true` — writes silently dropped, not failed (`:251`).
- Client-attach rollback when `writeStreamHeaderLocked`/`replayCachedVideoLocked` throw (`:145-150`).
- `serverSocket.accept()` IOException / null return (`:127-133`).
- Replacement-client displacement + command-reader `clientSocket === client` guard — the reconnect-race core (`:137`, `:176`, `:264`).
- Command-reader read-loop / `command < 0` break / IOException swallow / finally-detach (`:162-187`).

**Fix:** introduce a narrow interface seam (server-socket factory + client socket/streams) so the accept/write/replay/command paths can be driven by fakes, matching how `DisplayControl` was wrapped.

## Fragile contract worth pinning

`writePacketDataLocked` returns `true` even after a write fails (`:260`) and even with no client (`:251`). The encode loop's `if (!success) break` (`VideoServer.kt:289-291`) is therefore effectively dead — loop termination actually relies solely on `reconnectWindowExpired()` (`VideoServer.kt:310`). Either make the return value meaningful or remove the dead branch, and add a test that pins whichever contract is chosen.

[...]

Issue #4743, opened 2026-07-30
perf(android-video): unbuffered LocalSocket writes — two syscalls per packet


kaeawc:
## Summary

Each encoded packet is written to the `LocalSocket` as two separate `write()` calls
(header, then payload) on the raw, unbuffered output stream. At 30-60fps that is
60-120+ write syscalls/second plus extra TCP segmentation that splits each frame's
header away from its payload — the classic small-header pessimization.

## Evidence

- `output.write(VideoStreamProtocol.packetHeader(...))` then `output.write(data)` — [VideoStreamWriter.kt:253-254](android/video-server/src/main/kotlin/dev/jasonpearson/automobile/video/VideoStreamWriter.kt).
- No `BufferedOutputStream`; the raw `LocalSocket.outputStream` is used directly (`outputStream = client.outputStream`, [VideoStreamWriter.kt:139](android/video-server/src/main/kotlin/dev/jasonpearson/automobile/video/VideoStreamWriter.kt)).

## Proposed change

Either wrap the output in a `BufferedOutputStream` and flush once per packet, or
assemble header + payload into a single buffer and issue one `write`. Combining the
write also tightens on-wire framing. Preserve the existing per-client cleanup and the
config/IDR replay path.

## Impact / effort / risk

Medium (syscall count + framing) · low effort · low risk.

Issue #4749, opened 2026-07-30
perf(android-video): encode loop writes to socket synchronously under lock (no drop-latest)


kaeawc:
## Summary

The device encode loop dequeues an output buffer, writes it to the socket
**synchronously**, then releases the buffer — all on one thread. A slow or stalled
host reader blocks the dequeue, so the encoder's input queue backs up and frames are
delayed rather than dropped-latest. There is no latest-frame-wins drop policy on the
device side.

## Evidence

- Serial loop: `dequeueOutputBuffer(bufferInfo, 100_000)` → `streamWriter.writePacket(...)` (blocking) → `releaseOutputBuffer(index)` — [VideoServer.kt:283-314](android/video-server/src/main/kotlin/dev/jasonpearson/automobile/video/VideoServer.kt).
- The host has a drop-latest primitive (`LatestFrameQueue`, src/features/screen-stream/LatestFrameQueue.ts) and the daemon relay implements per-subscriber pause/drain (`videoStreamSocketServer.ts`), but neither protects the on-device encode loop.

## Proposed change

Decouple encode from transport with a dedicated writer thread fed by a bounded,
drop-oldest single-slot handoff, so a transport stall cannot back-pressure the
encoder. Preserve the existing reconnect window, cached SPS/PPS + IDR replay, and
keyframe-command reader semantics.

## Impact / effort / risk

Medium (latency under slow readers) · medium effort · medium risk — introduces
concurrency; keep the reconnect/replay behavior intact and covered by tests.

PR #4812, opened 2026-07-30
perf(android-video): write inter frames directly from the codec ByteBuffer

### LINKED ISSUE #4745: perf(android-video): per-frame heap allocation in the socket write path
## Summary

The write path allocates a fresh `ByteArray` and copies the entire encoded frame out
of the codec `ByteBuffer` on **every** frame, creating steady GC pressure on the
device at 30-60fps. The copy is only actually required for the packets the cache
retains (codec config + latest IDR); inter frames could be written straight from the
`ByteBuffer`.

## Evidence

- `val data = ByteArray(bufferInfo.size); buffer.get(data, 0, bufferInfo.size)` per frame — [VideoStreamWriter.kt:193-196](android/video-server/src/main/kotlin/dev/jasonpearson/automobile/video/VideoStreamWriter.kt).
- Only config/IDR packets are cached for replay to a replacement client — `VideoPacketCache.remember(...)` at [VideoStreamWriter.kt:296-307](android/video-server/src/main/kotlin/dev/jasonpearson/automobile/video/VideoStreamWriter.kt) keeps just `config` and `idr`.

## Proposed change

Write inter (non-config, non-keyframe) packets directly from the codec `ByteBuffer`,
copying only when a packet must be cached. Alternatively reuse a growable scratch
buffer across frames. Mind the buffer lifetime relative to `releaseOutputBuffer`
([VideoEncoder.kt:111](android/video-server/src/main/kotlin/dev/jasonpearson/automobile/video/VideoEncoder.kt)) — the payload must be fully written before the output buffer is released.

## Impact / effort / risk

Medium (GC/jank on device) · low-medium effort · low risk.

kaeawc:
Closes [#4745](https://github.com/kaeawc/auto-mobile/issues/4745)

## Summary

The socket write path in `VideoStreamWriter.writePacket` allocated a fresh `ByteArray` and copied the entire encoded frame out of the codec `ByteBuffer` on **every** frame, creating steady GC pressure on the device at 30-60fps. Only config and IDR packets are actually cached for replay to a replacement client, so only those need an owned copy.

This PR writes inter (non-config, non-keyframe) packets from the codec `ByteBuffer` through a **reused growable scratch buffer** instead of a fresh per-frame allocation:

- **Config / keyframe packets** keep their exact-size `ByteArray` copy — that copy is required for the replay cache anyway (`VideoPacketCache.remember(...)`), so it does double duty as the write source.
- **Inter frames** copy into a scratch buffer that grows only when a larger frame arrives (`growScratch`) and is reused otherwise. `writePacketDataLocked` now takes an explicit `length`, and writes exactly `length` bytes (`output.write(data, 0, length)`), so a larger reused buffer never leaks stale trailing bytes onto the wire.

Buffer lifetime is safe: `VideoServer`'s encoding loop calls `writePacket` fully synchronously **before** `encoder.releaseOutputBuffer(index)` ([VideoServer.kt:288-294](https://github.com/kaeawc/auto-mobile/blob/main/android/video-server/src/main/kotlin/dev/jasonpearson/automobile/video/VideoServer.kt#L288-L294)), and the payload is fully written within `writePacket`, so the codec-owned `ByteBuffer` stays valid for the duration of the call. The scratch buffer is only ever touched under `lock` from the single encoder thread.

## Changes

- `VideoStreamWriter.kt` — reused `scratch` field; `writePacket` splits cached (exact copy) vs. inter (scratch) paths; `writePacketDataLocked` gains a defaulted `length` param and writes only `length` bytes; new pure `internal growScratch(...)` helper.
- `VideoStreamWriterTest.kt` — added `scratchBufferGrowsOnceAndIsReusedForSmallerFrames`, verifying the buffer grows once, is reused for smaller frames (no realloc), and grows again only for a larger frame.

A full end-to-end `writePacket` unit test isn't feasible in this JVM-only module — it depends on Android's `MediaCodec.BufferInfo` and a socket `OutputStream`, and the module has no Robolectric/MockK (JUnit only). Rather than add a dependency or a wider refactor (this file also has sibling PRs #4743 and #4754 in flight), the allocation-avoidance crux was extracted into the pure `growScratch` helper and tested deterministically.

## Validation

- `./gradlew :video-server:test` (JDK 21) — BUILD SUCCESSFUL, all tests pass (including the new test).
- ktfmt (`scripts/ktfmt/validate_ktfmt.sh`) — "All Kotlin source files are properly formatted."
- Diff kept surgical to minimize conflicts with sibling PRs #4743 / #4754 that also edit `VideoStreamWriter.kt`.

### DIFF
--- a/android/video-server/src/main/kotlin/dev/jasonpearson/automobile/video/VideoStreamWriter.kt
+++ b/android/video-server/src/main/kotlin/dev/jasonpearson/automobile/video/VideoStreamWriter.kt
@@ -66,6 +66,14 @@ class VideoStreamWriter(
   private val packetCache = VideoPacketCache()
   private val reconnectWindow = ReconnectWindow(nowMs, CLIENT_RECONNECT_WINDOW_MS)
 
+  /**
+   * Growable scratch buffer reused across inter frames to move the payload out of the codec
+   * [ByteBuffer] without a per-frame heap allocation. Only ever touched under [lock] from the
+   * single encoder thread. Config/keyframe packets bypass it because they are copied for the cache
+   * anyway.
+   */
+  private var scratch = ByteArray(0)
+
   @Volatile private var stopped = false
 
   companion object {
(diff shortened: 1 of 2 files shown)


kaeawc:
## Subsumed by the #4749 encode/transport decoupling ([#4821](https://github.com/kaeawc/auto-mobile/pull/4821))

This PR's optimization is no longer applicable after [#4821](https://github.com/kaeawc/auto-mobile/pull/4821) (issue [#4749](https://github.com/kaeawc/auto-mobile/issues/4749)) landed on `main`, and cannot be rebased forward without regressing that architecture. Leaving it open for you to close.

### Why

This PR reused a single growable `scratch` buffer for inter frames and wrote them **synchronously** from `writePacket` — the correctness argument was *"the payload is fully written before the caller releases the output buffer, so the codec-owned bytes stay valid for the duration of this call."*

[#4821](https://github.com/kaeawc/auto-mobile/pull/4821) removed that synchronous write. `writePacket` now copies each access unit into an **owned** `ByteArray`, wraps it in an `EncodedVideoFrame`, and offers it to a bounded drop-oldest `FrameHandoff`; a dedicated `video-transport-writer` thread drains the handoff and performs the socket write **later**. That decoupling is the whole point of [#4749](https://github.com/kaeawc/auto-mobile/issues/4749) — it stops a stalled reader from back-pressuring MediaCodec's Surface input.

Under that model a reused scratch buffer is unsafe: the encode loop would overwrite `scratch` with frame N+1 before the writer thread has drained frame N (they can also both reference the same array through the slot), producing a data race and torn frames on the wire. **The per-frame owned copy this PR set out to eliminate is now load-bearing** — it is the ownership transfer into the handoff, not wasted work.

Reconciling would force an either/or:
- keep the handoff (required — it is the merged [#4749](https://github.com/kaeawc/auto-mobile/issues/4749) architecture) and drop the scratch reuse, leaving no change; or
- restore synchronous scratch writes and undo [#4749](https://github.com/kaeawc/auto-mobile/issues/4749)'s decoupling — a regression.

The only piece that would survive is the pure `growScratch` helper and its unit test, but with the handoff in place it has no caller, so shipping it would be dead code.

### If the per-frame allocation is still worth chasing

It would need a different design than a single scratch buffer — e.g. a small pool of buffers recycled back to the encode loop **after** the writer thread finishes the socket write (returning ownership post-drain), rather than reused eagerly on the producer side. That is materially larger than this PR and out of scope for a straight rebase. Suggest closing this and opening a fresh issue if the GC pressure proves to matter in the [#4815](https://github.com/kaeawc/auto-mobile/pull/4815) throughput baseline.

kaeawc:
Closing unmerged — subsumed by #4749 (#4821); the reused-scratch-buffer approach is unsafe under the new async FrameHandoff (the per-frame copy is now the load-bearing ownership transfer). Full analysis in the earlier comment; issue #4745 closed with the residual buffer-pool path noted.

PR #4976, opened 2026-08-02
fix(android): harden eventOnly device timestamp fallback

### LINKED ISSUE #4720: eventOnly fresh-focus: harden getDeviceTimestampMs degraded modes (lenient %3N parse / second truncation / host-timer fallback)
## Summary

The eventOnly fresh-focus no-`updatedAt` fallback derives its freshness lower bound from `getDeviceTimestampMs()` ([PR #4719](https://github.com/kaeawc/auto-mobile/pull/4719) / #4617). The common millisecond path is correct, but the function's degraded modes leave three edges — all verified at `main` @ `cd06a74e0`:

1. **Lenient ms-probe parse (worst, fix first).** `AdbClient.ts:553-557` parses `date +%s%3N` output with `parseInt`, which stops at the first non-digit. A device whose `date` prints `%3N` literally (e.g. `1754063999%3N`) yields the *seconds* value returned **as milliseconds** — a ~1970 `minTimestamp` that accepts any cached hierarchy. Full freshness-gate bypass, and it likely preempts the seconds fallback on exactly the devices that fallback was written for.
2. **Second-granularity truncation.** When the ms probe genuinely fails, the `date +%s` fallback (`AdbClient.ts:563-572`) truncates the bound down by up to 999 ms; the Android gate is inclusive (`updatedAt >= minTimestamp`, `CtrlProxyHierarchy.ts:638`), so any stale device-authored hierarchy stamped within the same second is wrongly accepted as fresh.
3. **Host-timer final fallback.** When both device probes fail, `AdbClient.ts:575` returns the host timer — reintroducing the original #4617 cross-clock skew for that degraded case instead of failing closed.

## Where

- `src/utils/android-cmdline-tools/AdbClient.ts:551-576` (`getDeviceTimestampMs`)
- Consumer: `src/features/action/InputText.ts:669-671` (`refreshFocusedTextInputHierarchy`) — degraded modes only matter on the no-`updatedAt` fallback path
- Gate: `src/features/observe/android/CtrlProxyHierarchy.ts:638` (inclusive `>=`)

## Proposed fix (with known design traps)

1. Strict-validate the ms probe: require the full trimmed stdout to match `/^\d+$/` so digits-plus-junk falls through to the seconds path instead of mis-parsing.
2. **Degraded modes are invisible to callers today** — `getDeviceTimestampMs()` returns a bare `number`, so "add margin" / "fail closed" cannot be implemented in `InputText` without an API change. Add a variant returning `{ timestampMs, source: "device-ms" | "device-seconds" | "host" }` on `AdbExecutor` (+ `FakeAdbClient`), keeping the existing method as a wrapper.
3. In `InputText.refreshFocusedTextInputHierarchy` only: on `device-seconds`, add a sub-second margin (or fail closed for that refresh); on `host`, fail closed (treat as not-fresh).
4. **Do not make `getDeviceTimestampMs()` itself throw or fail closed** — other consumers (`BaseVisualChange.ts:136`, `LaunchApp.ts:713/727/746`, `systemTrayHelpers.ts:52`) legitimately accept a best-effort host fallback. Fail-closed policy belongs in the freshness consumer, keyed off `source`.

## Acceptance criteria

- Literal-`%3N` stdout (digits + junk) never produces a seconds-scale value in the ms domain; unit test with a `FakeAdbClient`/fake-exec stdout of `"1754063999%3N"`.
- When device time is only second-granular, a same-second stale focused hierarchy is rejected.
- When both device probes fail, the eventOnly refresh fails closed rather than comparing against the host timer.
- All tests deterministic via the existing fake seams (`FakeAdbClient`, FakeTimer) — no real devices, no real waits.

## Severity

Low for edges 2–3 (narrow degraded modes on a fallback-of-a-fallback path). Edge 1 is the priority: on affected devices it silently voids the freshness guarantee entirely, and the fix is a two-line validation change.

## Not in scope

- iOS freshness path (`ios/CtrlProxyHierarchy.ts:114`) — different timestamp source, no ADB `date` probes.
- Changing the semantics of `getDeviceTimestampMs()` for its other consumers.

---
*Body rewritten after design review — see the "Design review (fable-review)" comment for the claim-by-claim verification record.*

kaeawc:
## Summary

Hardens the Android `inputText` `eventOnly` fresh-focus fallback for degraded device timestamp modes.

Closes [#4720](https://github.com/kaeawc/auto-mobile/issues/4720)

## Root cause

`getDeviceTimestampMs()` accepted partially numeric `%s%3N` output such as `1754063999%3N`, returned second-granularity timestamps without identifying the loss of precision, and finally fell back to the host clock. The eventOnly freshness consumer could therefore accept stale device-authored hierarchies.

## Changes

- Strictly validate both device timestamp probes and report whether the result came from device milliseconds, device seconds, or the host fallback.
- Reject seconds values whose conversion to milliseconds exceeds JavaScript's safe-integer range.
- Preserve `getDeviceTimestampMs()` as a compatibility wrapper for existing consumers.
- Make the eventOnly no-`updatedAt` refresh reject host-clock fallback and advance second-granularity bounds by 1000 ms.
- Extend the existing fakes and add deterministic regression coverage for all three degraded modes.

## Validation

- `bun test test/features/action/InputText.test.ts test/utils/android-cmdline-tools/AdbClient-executionContract.test.ts test/lint/fakeHygiene.test.ts` — 169 pass / 0 fail
- `bun run lint` — passed
- `bun run build` — passed
- `bun run typecheck` — no new errors
- `git diff --check` — passed
- `bun run turbo:validate` was attempted; its repository-wide test phase remained CPU-bound for 10 minutes because of unrelated long-running test processes in other worktrees and was stopped without a result. Focused and static gates are green.

### DIFF
--- a/src/features/action/InputText.ts
+++ b/src/features/action/InputText.ts
@@ -29,6 +29,8 @@ import { Keyboard } from "./Keyboard";
 
 export type InputTextMode = "a11y" | "eventLast" | "eventAll" | "eventOnly" | "append";
 
+const DEVICE_TIMESTAMP_SECOND_GRANULARITY_MARGIN_MS = 1000;
+
 export type AppendKeyEventValidator = (timeoutMs?: number) => Promise<{ success: boolean; error?: string }>;
 
 interface KeyboardCloser {
@@ -665,10 +667,19 @@ export class InputText extends BaseVisualChange {
     // clock (`this.timer.now()`) lets a device clock running ahead of the host
     // accept an older cached focused hierarchy as fresh, defeating the freshness
     // guarantee (issue #4617). Derive the fallback from the device clock via
-    // `getDeviceTimestampMs`, the same source the rest of the freshness logic uses.
-    const minTimestamp = typeof viewHierarchy.updatedAt === "number"
-      ? viewHierarchy.updatedAt + 1
-      : await this.adb.getDeviceTimestampMs();
+    // `getDeviceTimestampMsWithSource`, and reject degraded host-clock results.
+    let minTimestamp: number;
+    if (typeof viewHierarchy.updatedAt === "number") {
+      minTimestamp = viewHierarchy.updatedAt + 1;
+    } else {
+      const timestampResult = await this.adb.getDeviceTimestampMsWithSource();
+      if (timestampResult.source === "host") {
+        return undefined;
+      }
+      minTimestamp = timestampResult.source === "device-seconds"
+        ? timestampResult.timestampMs + DEVICE_TIMESTAMP_SECOND_GRANULARITY_MARGIN_MS
+        : timestampResult.timestampMs;
+    }
     const refreshedObserveResult = await this.observeScreen.execute({
       skipWaitForFresh: false,
       minTimestamp
--- a/src/utils/android-cmdline-tools/AdbClient.ts
+++ b/src/utils/android-cmdline-tools/AdbClient.ts
@@ -577,8 +577,9 @@ export class AdbClient implements AdbExecutor {
       const trimmed = result.stdout.trim();
       if (/^\d+$/.test(trimmed)) {
         const parsed = Number(trimmed);
-        if (Number.isSafeInteger(parsed) && parsed > 0) {
-          return { timestampMs: parsed * 1000, source: "device-seconds" };
+        const timestampMs = parsed * 1000;
+        if (Number.isSafeInteger(parsed) && parsed > 0 && Number.isSafeInteger(timestampMs)) {
+          return { timestampMs, source: "device-seconds" };
         }
       }
     } catch (error) {
--- a/src/utils/android-cmdline-tools/AdbClient.ts
+++ b/src/utils/android-cmdline-tools/AdbClient.ts
@@ -13,6 +13,7 @@ import {
   type AdbExecuteOptions,
   type AdbProcess,
   type AdbSpawnOptions,
+  type DeviceTimestampResult,
 } from "./interfaces/AdbExecutor";
 import { getAbortSignal } from "../AbortContext";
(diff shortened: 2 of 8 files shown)


kaeawc:
Manual-test verification (`0.0.48`..`main` @ 9c868f2cd, on-device):

**Confirmed FIXED** — Android emulator-5554 (am-api35-ga-arm64, API 35), fresh dist with `getDeviceTimestampMsWithSource` + `DEVICE_TIMESTAMP_SECOND_GRANULARITY_MARGIN_MS` compiled in.

Repro path: focused `EditText` pre-seeded with text "to", then `inputText { mode: "eventOnly", text: "wifi7x" }`.

Result: the field was cleared and re-typed via key events only — the tool's fresh-focus field snapshot was exactly `wifi7x` (no `to` prefix), with a populated `observationDiff`. The hardened fresh-focus path returned a fresh focused hierarchy with no stale-hierarchy hang and no degraded host-clock rejection on the positive path.

Note: the degraded-mode edges (host-only clock → `undefined`, lenient `%3N` parse, second-granularity margin) are covered by the change's unit tests; only the positive device path was exercised on hardware here.

kaeawc:
Verified on current main (HEAD 59caa18) against Android `emulator-5554` (API 35).

`inputText` landed correctly in both paths, confirmed via `uiautomator dump` ground truth (not just the tool's `success` flag):
- **default a11y mode** — typed `battery` → field showed `text="battery"` and search surfaced `App battery usage`.
- **`eventOnly` mode** (the hardened path) — `clearText` then typed `wifi` via key events only → field showed `text="wifi"` and matched `Wi-Fi`.

No regression in the `eventOnly` path from the device-timestamp fallback hardening. (Manual-test iteration, range 0.0.48..main.)

PR #5141, opened 2026-08-08
fix(android): calculate CPU usage over sampling intervals

### LINKED ISSUE #5107: PerformanceMonitor: Android CPU% is cumulative/uptime-averaged, not interval-based
## Summary

`PerformanceMonitor.collectCpuMetrics` (Android) derives CPU usage from **cumulative** process ticks divided by **total system uptime**, so a long-lived device makes even a busy process report a tiny percentage. On a device up for 10,000s, a process that has consumed 10 CPU-seconds reports ~0.1% regardless of its *recent* load. This understates CPU in every short window.

Surfaced by review on #5094 (the `perfSnapshot` feature exposes `cpu.avg/latest`), but the calculation is **pre-existing** and shared with the live IDE performance stream (`PerformanceStreamData.cpuUsagePercent`), so it's out of scope for that PR.

## Proposed fix

Track prior `(processTicks, sampleTime)` per monitored identity and record the **interval delta**: `cpu% = (Δprocess_ticks / Δelapsed_ticks) * 100`. First sample per identity has no delta → report null. Update `MonitoredDevice` to hold the previous readings (reset on package switch), and adjust `PerformanceMonitor` tests + any consumers that assume the first tick yields a CPU value.

## Notes

- Affects both `perfSnapshot` (`PerfWindowBuffer`) and the IDE stream, so verify both.
- iOS CPU comes from host `ps` `%CPU` which is already a recent-usage figure — this is Android-specific.

kaeawc:
## Summary

Android `PerformanceMonitor` now reports CPU usage from process-tick and uptime deltas instead of cumulative process time divided by device uptime. The first sample and samples after a package or process reset remain `null` until a valid interval is available.

## Linked Issue

Closes [#5107](https://github.com/kaeawc/auto-mobile/issues/5107)

## Bug / Regression Context

- **Observed symptom:** Android CPU usage was uptime-averaged, understating recent process load on long-lived devices.
- **Repro conditions:** Collect two medium-interval samples from the same PID while process ticks and `/proc/uptime` advance.

## Validation

- [x] Focused `PerformanceMonitor` suite: 47 passed
- [x] `bun run typecheck`: no new errors
- [x] `bun run lint`: passed with existing ratcheted warnings
- [x] `scripts/all_fast_validate_checks.sh`: 32 passed
- [x] Full `bun test`: 12,916 passed, 13 skipped; the one failure was an environment-path assertion caused by the temporary `AUTOMOBILE_DATA_DIR` under `/tmp/` and the isolated test passes without that override
- [x] `git diff --check`

### DIFF
--- a/src/features/performance/PerformanceMonitor.ts
+++ b/src/features/performance/PerformanceMonitor.ts
@@ -6,12 +6,18 @@ import {
   DEFAULT_THRESHOLDS,
   PerformancePushSocketServer,
 } from "../../daemon/performancePushSocketServer";
-import { getDeviceDataStreamServer, PerformanceStreamData } from "../../daemon/deviceDataStreamSocketServer";
+import {
+  getDeviceDataStreamServer,
+  PerformanceStreamData,
+} from "../../daemon/deviceDataStreamSocketServer";
 import { RecompositionTracker } from "./RecompositionTracker";
 import { getPerfWindowBuffer, PerfWindowBuffer } from "./PerfWindowBuffer";
 import { getSdkFrameMetricsStore, SdkFrameMetricsStore } from "./SdkFrameMetricsStore";
 import { TelemetryRecorder } from "../telemetry/TelemetryRecorder";
-import { defaultAdbClientFactory, AdbClientFactory } from "../../utils/android-cmdline-tools/AdbClientFactory";
+import {
+  defaultAdbClientFactory,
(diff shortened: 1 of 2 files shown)


kaeawc:
Closing as a duplicate of #5140. The reviewed implementation is being carried on #5140; this branch is not the delivery path for #5107.

PR #5195, opened 2026-08-09
fix(android): parse unit-suffixed AVD RAM

### LINKED ISSUE #5194: fix(android): support unit-suffixed AVD RAM settings
## Summary

`startDevice` rejects valid Android AVDs when `hw.ramSize` uses a size suffix such as `2G`.

## Reproduction

1. Create or configure an Android AVD with:

   ```ini
   hw.ramSize=2G
   ```

2. Use a modern Play Store system image.
3. Call `startDevice` for that AVD.

## Actual behavior

The value is interpreted as 2 MB and device startup fails with an error stating that the AVD needs at least 2048 MB.

## Expected behavior

Unit-suffixed RAM settings should be normalized before validation. In particular, `2G` should be treated as 2048 MB so the AVD can start.

## Impact

Valid AVD configurations using gigabyte notation cannot be started through AutoMobile.

kaeawc:
## Summary

Normalize unit-suffixed Android AVD RAM settings before enforcing the modern Play system-image memory floor. This lets valid settings such as `hw.ramSize=2G` start correctly and rejects malformed values rather than interpreting a numeric prefix.

## Linked Issue

Closes [#5194](https://github.com/kaeawc/auto-mobile/issues/5194)

## Bug / Regression Context

- **Prior attempts:** None.
- **Observed symptom:** `startDevice` reported that a 2 GiB AVD had only 2 MB of RAM and refused to start it.
- **Repro steps / conditions:** Set `hw.ramSize=2G` in an AVD using a modern Play Store image, then call `startDevice`.

## Validation

- [x] `bun test test/utils/android-cmdline-tools/AvdConfigReader.test.ts`
- [x] `bun run lint`
- [x] `bun run build`
- [x] `bun test` (13,020 passed; 13 integration tests skipped)
- [x] `bun run typecheck` reports no new errors
- [x] `scripts/all_fast_validate_checks.sh` (33 checks passed)
- [x] No new interfaces, dependencies, or generic helpers were introduced
- [x] Branch is based on current `origin/main`

### DIFF
--- a/src/utils/android-cmdline-tools/AndroidEmulatorClient.ts
+++ b/src/utils/android-cmdline-tools/AndroidEmulatorClient.ts
@@ -585,6 +585,11 @@ export class AndroidEmulatorClient implements AndroidEmulator {
     if (!avdConfig) {return;}
     const isModernPlayImage = avdConfig.tag?.toLowerCase().includes("play")
       && (avdConfig.apiLevel ?? 0) >= MODERN_PLAY_IMAGE_MIN_API_LEVEL;
+    if (isModernPlayImage && avdConfig.ramSizeInvalid) {
+      throw new ActionableError(
+        `Cannot start AVD '${avdName}': hw.ramSize is invalid. Use a whole number in MB or a K, M, or G size suffix and retry.`
+      );
+    }
     if (isModernPlayImage && avdConfig.ramSizeMb !== undefined && avdConfig.ramSizeMb < MIN_AVD_RAM_MB) {
       throw new ActionableError(
         `Cannot start AVD '${avdName}': hw.ramSize is ${avdConfig.ramSizeMb} MB, below the minimum ${MIN_AVD_RAM_MB} MB needed for a modern system image. Increase hw.ramSize in the AVD config and retry.`
--- a/src/utils/android-cmdline-tools/AvdConfigReader.ts
+++ b/src/utils/android-cmdline-tools/AvdConfigReader.ts
@@ -14,6 +14,8 @@ export interface AvdConfig {
   screenDensity?: number;
   /** Configured guest memory in megabytes. */
   ramSizeMb?: number;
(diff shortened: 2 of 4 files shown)


kaeawc:
**Manual-test verification (0.0.51..main @ `ecc9e546a`) — CONFIRMED FIXED ✅**

Exercised the fixed `validateAvdMemory` gate on a real Android emulator (Apple Silicon, `emulator 36.4.5.0`) via `startDevice --platform android --name <avd>` against a modern play-store AVD (`Medium_Phone_API_35`, api 35, `tag.id=google_apis_playstore`). Three discriminating cases, temporarily editing `hw.ramSize` in `config.ini` (restored afterward):

| `hw.ramSize` | Result (this fix) | Old `parseInt` behavior |
|---|---|---|
| `notanumber` | ❌ rejected: *"hw.ramSize is invalid. Use a whole number in MB or a K, M, or G size suffix and retry."* | `NaN` → `{}` → gate skipped (no error) |
| `1G` | ❌ rejected: *"hw.ramSize is **1024 MB**, below the minimum 2048 MB…"* | would report *"is **1 MB**"* |
| `2G` | ✅ accepted → emulator launched (`Increasing RAM size to 2048MB`, boots to `sys.boot_completed=1`) | `parseInt("2G")`=2 → falsely rejected as *"2 MB below minimum"* |

The `1G`→"1024 MB" and `2G`→accepted cases directly prove the unit-suffix multiplier (`G`→×1024) and that unit-suffixed RAM on a modern play image is no longer falsely blocked. The `notanumber` case proves the new `ramSizeInvalid` path. Verified in the compiled `dist` (`ramSizeInvalid` present).

Note: this change is daemon-side TS only — no APK / iOS-runner rebuild needed to reach users.

_Unrelated robustness issue observed on the same `startDevice` path while testing (stale QEMU lock from the arch-compat `-verbose` probe → full-timeout hang): filed as #5202._

PR #5272, opened 2026-08-16
fix(observe): report real freshness and recover from a frozen iOS hierarchy cache

kaeawc:
## Summary

Fixes the iOS `observe` failure mode where a visibly changing screen—especially a WebView—could keep returning the same hierarchy while claiming `freshness.isFresh: true`.

The change fixes three parts of that contract:

1. Plain `observe` now measures freshness instead of treating its lack of a public `minTimestamp` input as proof of freshness.
2. Platform delegates' verified/unverified cache verdict reaches the public observation result; an unverified cache entry is never reported fresh, and Android never mistakes device/host clock skew for staleness.
3. The iOS cache now ages a capture using the host time it was first received, retained across re-delivery of the same `updatedAt` capture. This avoids both indefinite reuse of unchanged native hierarchies and mixing a device wall clock with the host clock. Once the age budget expires, the default glance path forces a synchronous re-verification.

Hierarchy-retrieval failures and platform-rejected/scrubbed hierarchies are explicitly unavailable and therefore report `isFresh: false` with an explanatory warning. A newer iOS push that wins a re-verification race also retains its verified verdict.

## Regression coverage

- Plain observation of a stale hierarchy is not fresh.
- An unverified cache entry is not fresh at any age.
- A current-timestamp hierarchy retrieval error is not fresh.
- Re-delivery of an unchanged hierarchy after a device clock one hour ahead still expires by host capture age and forces re-verification.
- iOS client cache updates preserve the first host receipt time for an unchanged capture.
- Android preserves a freshly verified delegate verdict despite a device clock behind the host.
- A scrubbed cross-platform hierarchy is unavailable rather than fresh.
- A newer iOS push that supersedes a failed re-verification remains fresh.
- Existing `waitFor` / `requestedAfter` semantics remain covered unchanged.

## Validation

- [x] `bun run turbo:validate` — 13,392 passed, 13 intentionally skipped, 0 failed
- [x] Focused freshness/cache suite — 92 passed, 347 assertions, 0 failed
- [x] `bun run lint`, `bun run typecheck`, and `bun run build` after the final rebase
- [x] `git diff --check origin/main...HEAD`
- [x] Manual iOS simulator check: a live Safari Wikipedia WebView was observed as device-verified; a later synchronous-hierarchy failure returned an honest `isFresh: false` fallback rather than falsely claiming freshness

## CI note

The previous head was 73 commits behind `main`; its failed iOS simulator check was a now-obsolete Reminders launch failure. This branch is rebased onto current `main`, whose workflow removes/rewires that launch path, so CI must be evaluated from this pushed head.


kaeawc:
CI red on **XCTestRunner Simulator Tests** (and the **iOS** umbrella job, which only fails because that child did). This is a pre-existing simulator flake, not a regression from this PR — a re-run should clear it.

Evidence:
- The single failing case is `RemindersLaunchPlanTests.testLaunchRemindersPlan`, failing with `open "com.apple.reminders" failed` / `Executed: 0/3 steps`. The app never launched, so no test body ran.
- This is a real-simulator XCUITest (Swift). This PR's diff is **TypeScript-only** (host-side `observe` freshness reporting) plus TS tests — it cannot affect Swift XCUITest app-launch behavior.
- The same iOS suite flaked on an unrelated PR in the same batch (`pull_request.yml` run #7544) and went green on re-run (#7545–7547) with no relevant code change.

I don't have permission to re-run failed jobs from here (`rerun-failed-jobs` returns 403 for the integration). Could a maintainer re-run the failed jobs, or it'll clear on the next push. Flagging rather than papering over it — I won't skip/quarantine the test.

Note this PR is still **draft** pending on-device verification of the original WebView repro (see the PR description's Device Verification item); that on-simulator check is the more important gate than this flake.

---

kaeawc:
I checked all three Codex findings against the code (I've tracked this PR since the original fork). Triage:

**#2 — `ObserveScreen.ts:351`, "mark scrubbed hierarchies unavailable" — confirmed, real.** This is a genuine stale-as-fresh gap, the exact class this PR targets. `enforceHierarchyPlatform()` runs at line 289 (before the freshness computation) and on a cross-platform rejection calls `discardHierarchyDerivedData()`, which sets `result.viewHierarchy = undefined` and attaches a **top-level** `result.error`. But `unavailable` is derived from `result.viewHierarchy?.hierarchy?.error`, which is then `undefined !== undefined → false`; `resolveObservationTimestampMs()` falls back to `result.updatedAt` (the recent observation time); so `computeFreshness` lands on the final `isFresh: true` branch for an observation carrying no hierarchy at all. Suggested direction — treat an absent hierarchy or a top-level error as unavailable, e.g.
```ts
unavailable: result.viewHierarchy?.hierarchy === undefined || result.error !== undefined,
```
worth a regression test alongside the platform-scrub path.

**#1 — `observationFreshness.ts:129`, Android clock skew — valid, edge case.** For Android there's no `verified` plumbing, so age (device-authored `updatedAt` vs host `now`) is the sole signal; a device clock more than the 5s budget behind the host would report a current hierarchy as `isFresh: false`. Worth noting this clock-domain mix is pre-existing — the `minTimestamp`/`waitFor` path already compares host-derived against device-derived timestamps — so this generalizes an existing assumption rather than introducing a new one. Emulators share the host clock; real-device skew is the exposure. A host-receipt timestamp for the age basis would close it if it's worth the plumbing.

**#3 — `CtrlProxyHierarchy.ts:~222`, superseding push — valid, but the safe direction.** In the fallback, `getCachedHierarchy() ?? cachedHierarchy` can be a fresh push-replacement, yet it's hardcoded `fresh: false`. That's a false-negative (fresh reported as unverified), which errs conservative — unlike #2. Low harm; fine to defer.

Happy to push the #2 fix + a regression test on top of `9897132` if you'd like — holding off so I don't step on your rebase.

---

kaeawc:
CI red again on **XCTestRunner Simulator Tests** + the **iOS** umbrella — but this is a *different*, environmental failure from the earlier `testLaunchRemindersPlan` flake, and not from the freshness changes. Causality from the job log:

1. **"Boot iOS Simulator for CtrlProxy UI tests: Failed"** — `startDevice timeout exhausted while starting the device; remainingBudgetMs=0` (`deviceBootService.ts:340`). The simulator didn't boot within the 300s budget.
2. That left the `simulator_udid` step output empty, so the next step ran `./scripts/ios/navigation-graph-sdk-event-integration.sh ""` → `usage: navigation-graph-sdk-event-integration.sh <simulator-udid>` → exit 1. That empty-arg "usage" line is a *symptom* of the failed boot, not a script-arg bug.

The CtrlProxy build itself succeeded (370s), and the freshness diff is host-side TS — it can't affect `simctl boot`. So this is an iOS-simulator-boot infra timeout; a re-run should clear it. (I still don't have permission to trigger re-runs from here — `rerun-failed-jobs` 403s for the integration.)

Separately — the P2 triage in my previous comment still stands; #2 (scrubbed-hierarchy reports `isFresh: true`) is the one real correctness item and I'm happy to push that fix + a regression test whenever you want it.

---

kaeawc:
Good catch by Codex on the round-2 findings — and #A is on me: the `result.error !== undefined` clause I suggested for the `unavailable` fix is too broad.

**#A (`ObserveScreen.ts:353`) — confirmed regression.** `result.error` is an **aggregate**, not a hierarchy signal — `ObserveError.ts` builds it as `result.errors.map(e => e.message).join("; ")` across every phase. So a best-effort failure in back-stack / active-window / screenshot collection now marks a perfectly valid, device-verified observation `isFresh: false` with "View hierarchy could not be retrieved." The minimal fix is to **drop the `result.error` clause entirely** rather than narrow it:
```ts
unavailable: result.viewHierarchy?.hierarchy === undefined ||
  result.viewHierarchy.hierarchy.error !== undefined,
```
The platform-scrub case (original #2) stays covered — `discardHierarchyDerivedData()` sets `result.viewHierarchy = undefined`, so the first clause already catches it. So this both fixes the new false-negative and keeps #2 closed, no separate platform-validation check needed.

**#B (`observationFreshness.ts:189`) — valid, edge case.** A within-TTL cache hit returns `fresh: true` before the age-budget/`cacheStale` path runs, so with `AUTOMOBILE_MAX_OBSERVATION_AGE_MS` set below the 500ms iOS / 1s Android TTL it propagates `verified: true` / `isFresh: true` past the configured budget. Real but only bites when the override is set below the platform TTL; the deeper point — `verified: true` slightly overclaims "checked against the device this call" for a cache hit — is the part worth deciding on.

**#C (`CtrlProxyHierarchy.ts:200`) — valid, low harm.** `SdkHierarchyRefreshPublisher` rebroadcasts an SDK-enriched tree while `HierarchyMerger` preserves the original XCUITest `updatedAt`, so a genuine replacement can share the prior `updatedAt` and the timestamp-inequality check misses it → reported `fresh: false`. Errs conservative (fresh-as-stale) like the original #3, so lowest priority; cache-identity rather than timestamp would close it.

#A is the one I'd fix before merge. Happy to push it + a regression test (a valid hierarchy plus a non-hierarchy `result.error` must stay `isFresh: true`) if you want it — otherwise it's a one-line delete.

---

kaeawc:
Confirmed on `6e2c50c` — all three round-2 findings look correctly addressed:

- **#A**: `unavailable` now keys off the explicit `hierarchyPlatformValid` flag (the `enforceHierarchyPlatform` return) instead of the aggregate `result.error`, so a best-effort non-hierarchy error no longer flips a valid, device-verified observation to stale. Cleaner than the drop-the-clause version I suggested.
- **#B**: cache-hit acceptance is now `cacheAge < Math.min(cacheFreshTtlMs, maxObservationAgeMs())`, so a configured budget below the platform TTL is honored.
- **#C**: superseding-push detection now compares hierarchy+frameContext via `isDeepStrictEqual` rather than `updatedAt` inequality, so a same-`updatedAt` SDK rebroadcast is recognized as a real replacement.

I ran the focused freshness/cache/stale-fallback suites locally — 29 pass, 0 fail. The only CI risk I'd expect on this head is the same iOS-simulator infra flake (boot timeout / codeload 429), which is environmental. Nice work hardening this.

---

kaeawc:
Heads up on this run (32072157878), since auto-merge is on: the iOS check failed for a **different** reason than the earlier flakes, and it's not the freshness change.

- The **XCUITests booted and passed** this time — `Test Suite 'CtrlProxyUITests.xctest' passed … Executed 4 tests, with 0 failures`, `** TEST EXECUTE SUCCEEDED **`. So the sim-boot flake didn't recur.
- The failure is an unrelated TS integration test: `test/integration/iosVideoRecordingStartStop.integration.test.ts:191` → `legacy iOS session auto-start automation runner readiness failed … phase=runner-setup … remainingBudgetMs=0: readiness phase exceeded the remaining deadline`. A **video-recording runner-setup readiness timeout** on a slow runner (the observe warm-up in the same job needed 3 retries over ~2 min before `CtrlProxy ready`).

This is untouched by the freshness diff — it's the video-recording session runner bring-up, not the hierarchy cache path, and observe itself worked later in the same run (nav-graph integration passed on attempt 1). So: environmental/timing, a re-run should clear it. I still can't trigger re-runs from here (`rerun-failed-jobs` 403s for the integration), so if the iOS job is a required check you'll need to re-run it (or it'll pick up on the next push).

---

kaeawc:
**Manual-test verification (post-`0.0.58`, `main` @ `7f6d641`, dist rebuilt from that commit)**

✅ **iOS — fully verified.** `iPhone 16 Pro E2E` sim:
- Real freshness reported: `observe` returns `ageMs: 108-112, verified: true, isFresh: true`, no spurious warning.
- **Frozen-cache recovery works:** after idling 7s on an unchanging springboard (no structural-hash push expected from the debouncer), `observe` still returned `ageMs: 108, verified: true` — the delegate forced a synchronous re-verification rather than serving a 7s-frozen tree. This is exactly the behavior this PR set out to fix.

⚠️ **Android — verdict correct, but `ageMs`/`warning` are skew-contaminated.** On emulator-5554 (clock ~24s behind host), `observe` reports `ageMs ≈ 25600` and the warning "…already 25597ms old (budget 5000ms) — the rest of the observation was slow" for a hierarchy captured ~0.4s ago in device time. `isFresh` stays correct (protected by `CtrlProxyHierarchy.ts:382-385`'s `fresh = isFresh`), but `ageMs` and the warning are computed as `hostNow − deviceUpdatedAt`, mixing clock domains. Filed as #5377.

PR #5340, opened 2026-08-16
fix(daemon): rekey device epoch on adb transport change

### LINKED ISSUE #5267: feat(daemon): key device-session epoch on adb transportId so fast same-serial reconnects re-mint (follow-up to #5257)
Epic: #5256 · follow-up to #5257 / PR #5265

Raised by Codex (P1) on PR #5265. The `DeviceSessionRegistry` (#5257) keys a device epoch on the
pool's monotonic `PooledDevice.incarnation`, which the pool assigns only when it allocates a fresh
`PooledDevice`. But `DevicePool.matchesRuntimeIdentity()` compares only serial + name + platform
(`src/daemon/devicePool.ts:~2127-2135`) and **not** `transportId`. So when an Android device
disconnects and reconnects under the same serial *before the disconnect monitor confirms its miss
threshold*, the pool retains the existing pooled entry (unchanged incarnation) even though adb has
handed out a new `transportId` — a genuinely new ADB connection epoch.

Consequence: `DeviceSessionRegistry.onDeviceConnected` sees the unchanged incarnation and returns the
**previous** `deviceSessionUuid`, so two real ADB connection epochs alias to one UUID. A consumer
keyed on the UUID cannot tell it should invalidate stale routing/cache state for that reconnect.

This is the "fast same-serial reconnect the pool never dropped" case already acknowledged in the
`DeviceConnectedInput` docstring — it's a **pool device-identity limitation**, not a registry bug:
the registry faithfully reflects the pool's own view that this is the same device. Fixing it means
making the pool's identity finer-grained, which is why it's out of scope for the keystone.

## Options
1. **Persist and compare `transportId` in the pool.** Store `transportId` on `PooledDevice` and include
   it in `matchesRuntimeIdentity`, so a transport change forces a fresh incarnation (and thus a fresh
   epoch). Cleanest, but changes device-recovery/identity semantics broadly — needs its own risk review.
2. **Bump incarnation on transport change only for the registry.** Narrower, but splits "epoch" from the
   pool's device identity, which the registry deliberately aligns to today.

Prefer option 1. Coordinate with the recovery paths and the `intentionalShutdowns` incarnation gating.

## Acceptance criteria
- A same-serial reconnect with a changed `transportId` (no confirmed disconnect) mints a **new**
  `deviceSessionUuid`; the prior UUID stops resolving.
- Existing device-recovery / same-serial-replacement behavior is unchanged for the genuine cases.
- Tests cover transport-change → new epoch without a full disconnect-monitor cycle.

kaeawc:
## Summary

Fixes device-session identity for an Android device that reconnects under the
same serial before the disconnect monitor confirms it disappeared.

The device pool now bypasses the short ADB device-list cache during refresh,
persists the resulting `transportId`, and includes it in runtime identity
comparison. A changed transport evicts the prior pooled connection,
which retires its `deviceSessionUuid` through the existing removal callback and
creates a fresh pooled incarnation and UUID through the ready callback.

## Root cause

The pool previously compared only serial, device name, and platform. A fast
same-serial ADB reconnect therefore reused the existing pooled entry and its
incarnation, causing `DeviceSessionRegistry` to retain the old epoch UUID.

## Validation

- `bun test test/daemon/devicePool.test.ts test/daemon/deviceSessionRegistry.test.ts test/daemon/daemonIdGenerator.test.ts`
- `bun run lint`
- `bun run typecheck`
- `bun run turbo:validate` — 13,199 passed; 13 expected device-dependent skips; 0 failures

Closes [#5267](https://github.com/kaeawc/auto-mobile/issues/5267)

### DIFF
--- a/src/daemon/devicePool.ts
+++ b/src/daemon/devicePool.ts
@@ -300,11 +300,6 @@ export class DevicePool {
         `Device discovery completed in ${discoveryTime}ms, found ${bootedDevices.length} devices`,
       );
 
-      if (refreshGeneration !== this.refreshGeneration) {
-        logger.info("Discarding an out-of-date device discovery snapshot");
-        return 0;
-      }
-
       const now = this.seedLastUsedAt(this.timer.now());
       let addedCount = 0;
       let removedCount = 0;
@@ -312,11 +307,18 @@ export class DevicePool {
       const bootedPlatforms = new Set(bootedDevices.map((device) => device.platform));
 
       perf.startOperation("poolUpdate");
-      removedCount = await this.removeDevicesMissingFrom(
+      const removed = await this.removeMissingDevicesForRefresh(
+        assignmentLockHeld,
+        refreshGeneration,
         bootedDeviceIds,
         bootedPlatforms,
         discovery.succeededPlatforms,
       );
+      if (removed === undefined) {
+        logger.info("Discarding an out-of-date device discovery snapshot");
+        return 0;
+      }
+      removedCount = removed;
 
       for (const device of bootedDevices) {
         this.clearAutoStartSuppressionForBootedDevice(device);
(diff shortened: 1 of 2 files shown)


kaeawc:
Heads-up: adding `transportId` to `DevicePool.matchesRuntimeIdentity` here appears to break cold-boot `startDevice` — the verification path builds `expectedIdentity` from a `Pick` that omits `transportId`, so it compares `undefined` against the discovered pooled transportId and always throws `Device pool identity mismatch`. Filed with repro + root cause in [#5369](https://github.com/kaeawc/auto-mobile/issues/5369).

PR #5393, opened 2026-08-18
fix(test): eliminate the never-settling-wait hang class behind the macos watchdog

kaeawc:
Follow-up cure to the defensive wall-clock watchdog in [#5391](https://github.com/kaeawc/auto-mobile/pull/5391): that PR made a wedged macos "Run Tests" step fail fast and red; this PR removes the hang class itself — tests whose waits can never settle. `bun test --timeout` only bounds test/hook bodies, so a never-settling wait or a leaked socket/child handle can still wedge a run the way [run 32156898919](https://github.com/kaeawc/auto-mobile/actions/runs/32156898919) wedged mid-suite for ~24 minutes.

## The wedge that is live on main right now

While proving this PR with a full local `bun test --isolate`, the suite wedged deterministically — and bisecting found the culprit is **not** one of the original real-I/O suspects:

**`test/server/deviceTools.startDevice.test.ts` deadlocks the entire run** since [#5368](https://github.com/kaeawc/auto-mobile/pull/5368) (merged today). `runPhase` now holds an externally-aborted phase's rejection until the operation gets `ABORT_SETTLEMENT_GRACE_MS` (1s) to settle. In the *"does not reach runner readiness or session binding after an externally cancelled boot"* test, the fake image listing deliberately never settles **and** the grace timer runs on the suite's injected `FakeTimer`, which nobody advances — a three-way circular wait (test awaits rejection ← rejection awaits settlement ← settlement awaits a clock nobody advances) that not even bun's per-test timeout interrupts. Reproduces 100% on a single-file run, plain or `--isolate`; the latest merge run's "Node TypeScript Build and Test (ubuntu-latest)" job is being **cancelled** the same way. The fix advances the injected fake clock past the grace deterministically and aligns the two cancellation tests with #5368's deliberate reason-propagation semantics (their real assertions — no runner readiness, no session binding, owned process killed — are unchanged). Suite: 42 pass in ~200ms, 20/20 loop green.

## Audit verdicts on the original leads

| Suspect | Verdict | Blocking construct |
| --- | --- | --- |
| `test/daemon/socketServer*` suites | **Confirmed** | Hand-rolled `sendRequest` promises resolved only on a full JSON line over a real unix socket — no deadline, and the client socket destroyed only on the response path, so an unanswered request pended forever and leaked its handle (old `socketServerHandshake.test.ts:25-61`, `inputSocketHarness.ts:83-129`, plus 8 more copies; `openPersistentConnection` sends, `PersistentSocketClient.request` in the mcpReconnect / mcpForwardSerialization suites, unbounded `waitForFrames` in the notifications suite, and 3 error-handler-less `once("connect")` awaits in the videoStream suite) |
| `test/features/webrtc/WebRtcPublisher.loopback.test.ts` | **Confirmed** | The only real-werift suite: DTLS/ICE handshake over real loopback UDP inside `publisher.start()`, plus unbounded `publisher.stop()` / `receiver.close()` in teardown — under contention the handshake can stall and the werift UDP sockets leak past the test timeout |
| `test/integration/mediamtxWebRtcPublisher.integration.test.ts:304/316/326` | **Confirmed** | Three helper tests gated only by `skipIf(win32)` run on every macos CI pass; `await once(child.stdout, "data")` and the other `once()` waits had no deadline, and the SIGTERM-ignoring `while :; do sleep 1; done` child was not killed on failure paths |
| Broader sweep (real `spawn`/UDP/`Bun.listen`, >100ms sleeps, other integration files) | **Cleared** | Other daemon suites use `FakeSocket`/resolve-gates; publisher/trickle webrtc suites use fake peer connections; all other `test/integration` files are env-gated; the "npm 503" WARNs seen mid-run come from a fakes-based metrics test, not the network |

## What changed

- **New canonical bounded helper** `test/daemon/helpers/socketRequest.ts`: `sendRawSocketRequest` / `sendSocketRequest` / `connectBounded` enforce a hard 10s client-side deadline (via `defaultTimer`, per the no-raw-timer rule) and destroy the socket on **every** settle path — response, error, close, deadline — so a server bug becomes a fast red failure with a diagnostic instead of a leaked handle.
- `inputSocketHarness` delegates to it; `openPersistentConnection` gains a bounded connect and per-send deadlines (all gesture-suite sends are awaited before `close()`, so no orphan rejections).
- The nine hand-rolled unbounded `sendRequest` copies now use the canonical helper: handshake, toolCapabilities, keyValue, mcpReconnect, featureFlags, ideStatus, inputPressButton, inputKey (drain mode preserves its frame-count assertion), mcpForwardSerialization (including its persistent client and the pipelined two-request helper).
- `socketServerNotifications.waitForFrames` and the videoStream connect awaits are bounded with diagnostics.
- **Loopback suite gated + bounded**: `WebRtcPublisher.loopback.test.ts` moves behind `AUTOMOBILE_MEDIAMTX_WEBRTC_INTEGRATION=1`, with `start`/`stop`/`close` deadline-wrapped. It still runs in CI: `scripts/webrtc/run-mediamtx-publisher-integration.sh` now includes it, so the dedicated **"WebRTC Publisher Integration (MediaMTX)"** job runs it on every webrtc-touching PR and on merge (and locally via `bun run test:integration:webrtc-mediamtx` — verified 20/20 green with the gate armed). Default-suite coverage of the publisher logic stays with the fakes-based `WebRtcPublisher.test.ts` / `.trickle.test.ts`. The BATS contract test for the runner is updated to match.
- The three always-on child-process helper tests bound their `once()` waits (`boundedOnce`, 10s) and kill the child in `finally`.
- `test/server/deviceTools.startDevice.test.ts`: deterministic fake-clock advance past the abort-settlement grace + expectation alignment (see above).

## Proof

- 20/20 loop of all 17 daemon/webrtc/integration touched files: 0 failures, ~1.5s per iteration.
- 20/20 loop of the gated loopback suite with the integration gate armed: 0 failures (~1s each).
- 20/20 loop of `test/server/deviceTools.startDevice.test.ts` under `--isolate`: 0 failures (~200ms each; previously wedged forever).
- `bun test test/daemon/ test/features/webrtc/ test/integration/`: 1892 pass / 14 skip / 0 fail in 3.0s.
- Full `bun test --isolate`: **before** this PR's second commit it wedged indefinitely (two runs killed at 12-13 min, plus two 2-day-old `bun test --isolate` processes found spinning on this machine from before the fix — live evidence of the class); **after**, green (wall-clock in the first comment below).
- `bun run lint` (ratchet unchanged, boundaries pass), `bun run typecheck` (no new baseline errors), `bun run build`, `bats test/bats/mediamtx-webrtc-integration.bats`, `shellcheck`: all green.

## Deliberately left alone

- The [#5391](https://github.com/kaeawc/auto-mobile/pull/5391) watchdog itself stays as defense in depth.
- `src/utils/deviceBootService.ts` is untouched: #5368's settlement-then-rethrow semantics are days old and deliberate; the deadlock leg that belongs to the test (an un-advanced fake clock) is what this PR fixes. If the phase-labeled cancellation message is preferred over the platform AbortError for bare `abort()`, that is a product decision for a follow-up.
- `stopProcess`'s post-SIGKILL `once(process, "exit")` wait in the mediamtx file: SIGKILL delivery is reliable, and the helper tests around it are now bounded and kill children in `finally`.
- `socketServerClose.test.ts` / `BaseSocketServerOwnership.test.ts` listen/close callback waits: local unix-socket `listen`/`close` callbacks always fire, and `connectClient` uses `events.once`, which rejects on `error`.
- `deviceDataStreamSocketServer`'s never-resolving promise is an in-process fake return (deterministic), not real I/O.

### DIFF
--- a/scripts/webrtc/run-mediamtx-publisher-integration.sh
+++ b/scripts/webrtc/run-mediamtx-publisher-integration.sh
@@ -9,6 +9,10 @@ set -euo pipefail
 
 readonly MEDIAMTX_VERSION="1.19.2"
 readonly TEST_FILE="${AUTOMOBILE_MEDIAMTX_TEST_FILE:-test/integration/mediamtxWebRtcPublisher.integration.test.ts}"
+# The werift loopback suite shares the opt-in gate: its DTLS/ICE handshake runs
+# over real UDP sockets, so it lives in this integration lane rather than the
+# default `bun test` run (macos real-I/O hang class, #5391).
+readonly LOOPBACK_TEST_FILE="test/features/webrtc/WebRtcPublisher.loopback.test.ts"
 
 repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 cache_dir="${AUTOMOBILE_MEDIAMTX_CACHE_DIR:-${repo_root}/scratch/mediamtx}"
@@ -92,4 +96,4 @@ mediamtx_binary="$(cd "$(dirname "${mediamtx_binary}")" && pwd -P)/$(basename "$
 cd "${repo_root}"
 AUTOMOBILE_MEDIAMTX_WEBRTC_INTEGRATION=1 \
 AUTOMOBILE_MEDIAMTX_BINARY="${mediamtx_binary}" \
-  exec bun test "${TEST_FILE}"
+  exec bun test "${TEST_FILE}" "${LOOPBACK_TEST_FILE}"
--- a/test/bats/mediamtx-webrtc-integration.bats
+++ b/test/bats/mediamtx-webrtc-integration.bats
@@ -5,6 +5,7 @@
 
 SCRIPT="scripts/webrtc/run-mediamtx-publisher-integration.sh"
 TEST_FILE="test/integration/mediamtxWebRtcPublisher.integration.test.ts"
+LOOPBACK_TEST_FILE="test/features/webrtc/WebRtcPublisher.loopback.test.ts"
 
 setup() {
   TEST_DIR="$(mktemp -d)"
@@ -54,7 +55,7 @@ SCRIPT
     bash "${ABS}"
 
   [ "${status}" -eq 0 ]
-  [ "$(cat "${INVOCATIONS_FILE}")" = "gate=1 binary=${resolved_binary} args=test ${TEST_FILE}" ]
+  [ "$(cat "${INVOCATIONS_FILE}")" = "gate=1 binary=${resolved_binary} args=test ${TEST_FILE} ${LOOPBACK_TEST_FILE}" ]
 }
 
 @test "runner resolves an injected relative MediaMTX binary before forwarding it" {
@@ -75,7 +76,7 @@ SCRIPT
     bash "${ABS}"
 
   [ "${status}" -eq 0 ]
-  [ "$(cat "${INVOCATIONS_FILE}")" = "gate=1 binary=${resolved_binary} args=test ${TEST_FILE}" ]
+  [ "$(cat "${INVOCATIONS_FILE}")" = "gate=1 binary=${resolved_binary} args=test ${TEST_FILE} ${LOOPBACK_TEST_FILE}" ]
 }
 
 @test "download path rejects a MediaMTX archive whose pinned checksum does not match" {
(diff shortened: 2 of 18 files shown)


kaeawc:
Full `bun test --isolate` wall-clock with both commits: **13507 pass / 14 skip / 0 fail across 957 files in 3m11s** (real 3m11.3s, user 3m46.6s) on a heavily contended local arm64 mac. Before the second commit the same command wedged indefinitely (two runs killed at 12-13 minutes with the runner idle at ~4% CPU, stuck in `deviceTools.startDevice.test.ts`).

PR #5451, opened 2026-08-22
fix: resolve Android AVD name from system property

### LINKED ISSUE #5450: Fall back to Android AVD system property during emulator discovery
On some Android emulator environments, `adb emu avd name` can return empty even after the device has fully booted. The emulator still exposes its AVD identity through `adb shell getprop ro.boot.qemu.avd_name`.

When the console query is empty or unavailable, use that property before treating the emulator name as unknown. Add focused coverage for the fallback so readiness can associate the launched emulator with its requested AVD.

kaeawc:
Closes #5450

## Summary

- resolve an Android emulator's AVD name from `ro.boot.qemu.avd_name` when the emulator-console query is empty or unavailable
- retain the emulator-console query as the primary lookup
- cover the fallback and command ordering with a focused discovery test

## Root cause

Some emulator environments can report an empty result for `adb emu avd name` after boot. AutoMobile then treated the serial as an unknown AVD name, preventing readiness from associating the launched emulator with its requested AVD.

## Validation

- `bun test test/utils/android-cmdline-tools/AndroidEmulatorClient-getBootedDevices.test.ts`
- `bun run build`
- `bun run lint`
- `bun run turbo:validate` (13,768 pass, 14 skipped, 0 fail)

### DIFF
--- a/src/utils/android-cmdline-tools/AndroidEmulatorClient.ts
+++ b/src/utils/android-cmdline-tools/AndroidEmulatorClient.ts
@@ -1322,6 +1322,51 @@ export class AndroidEmulatorClient implements AndroidEmulator {
     }
   }
 
+  private async getRunningAVDName(
+    device: BootedDevice,
+    infoTimeoutMs: number,
+    signal?: AbortSignal,
+  ): Promise<string> {
+    const deviceId = device.deviceId;
+    const adbWithDevice = this.adbFactory.create(device);
+    try {
+      const result = await adbWithDevice.executeCommand(
+        "emu avd name",
+        infoTimeoutMs,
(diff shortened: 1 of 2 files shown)


kaeawc:
Hosted CI note: the initial macOS TypeScript job reached its 12-minute suite watchdog while running `test/integration/webrtcDeviceCapture.integration.test.ts`; it did not report a test assertion failure. The Android AVD-discovery change does not modify WebRTC code or tests, and the Ubuntu/Windows TypeScript jobs passed. Rerunning the failed macOS job once to distinguish the known real-I/O runner hang from a reproducible failure.

PR #5507, opened 2026-08-22
feat(tap): support precise targets within Android elements

### LINKED ISSUE #5500: Android: support precise taps within text elements containing ClickableSpans
## Summary

`tapOn` can match a text element but activates its center. This cannot invoke an actionable span positioned elsewhere inside the same Android `TextView`, such as a mention or link embedded in longer text.

## Reproduction

1. Render a single Android text element containing ordinary text plus a clickable span near one edge.
2. Locate the containing element with `observe`.
3. Call `tapOn` for that element.
4. The center tap does not activate the span.

## Expected behavior

A caller should be able to target an actionable sub-element within a matched text element without relying on an accidental center tap.

## Acceptance criteria

- Provide a backward-compatible way to request an explicit coordinate or validated relative offset for a matched Android element.
- Preserve the current center-tap behavior when no precise target is supplied.
- Validate the requested point against the resolved element and screen bounds.
- Report the resolved target coordinates in the tool result.
- Add coverage for a `ClickableSpan` near both edges of a wider text element and for rejected out-of-bounds targets.

kaeawc:
## Summary

- add an optional normalized `relativePosition` for precise Android taps within the final resolved element
- validate the resolved point against half-open element and screen bounds while preserving center taps by default
- report resolved `x`/`y` target coordinates and force coordinate execution under TalkBack so embedded `ClickableSpan` targets are honored
- publish and drift-check the updated tool schemas, document TalkBack behavior, and cover edge spans, clipping, rotation, stability, invalid targets, and routing

## Linked Issue

Closes https://github.com/kaeawc/auto-mobile/issues/5500

## Validation

- [x] Focused tap and schema suite: 211 passed
- [x] `bun run turbo:validate -- --env-mode=loose`: 7/7 tasks, 13,802 tests passed, 0 failed
- [x] `bun run typecheck`: no new errors
- [x] `scripts/all_fast_validate_checks.sh --skip ktfmt`: 33/33 applicable checks passed; no Kotlin files changed
- [x] New code uses existing interfaces/fakes and deterministic unit tests complete well under 100ms each
- [x] No new helper package or direct dependency
- [x] Integrated latest `main`

## Device Verification

- [ ] Verified on a real Android device
- [ ] Screenshots or before/after attached

The existing Android coordinate transport already accepts arbitrary points; focused fakes cover dispatch and TalkBack routing without requiring a protocol or Kotlin change.

### DIFF
--- a/docs/design-docs/mcp/a11y/talkback-voiceover.md
+++ b/docs/design-docs/mcp/a11y/talkback-voiceover.md
@@ -126,7 +126,7 @@ When TalkBack is active, Android reserves certain gestures:
 
 ### Per-Tool Adaptations
 
-**tapOn**: Use `ACTION_CLICK` on the target element instead of a coordinate-based tap. Optionally set accessibility focus first to mimic user behavior and trigger TalkBack announcement. Long press uses `ACTION_LONG_CLICK`. When Android-only `relativePosition: { x, y }` is present, bypass semantic activation and dispatch the normalized coordinate through CtrlProxy so distinct targets within one accessibility node remain addressable; the option is unsupported with `focus`.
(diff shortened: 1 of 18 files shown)


kaeawc:
we're going to use another approach

PR #5589, opened 2026-08-24
fix(desktop): detect severe active stream drops

### LINKED ISSUE #5582: Screen streaming: severe-drop detection for auto-adjust via on-device dropped-frame stats
Follow-up from [#1098](https://github.com/kaeawc/auto-mobile/issues/1098) / PR #5569.

The desktop live-mirror quality auto-adjustment shipped in #5569 infers stream health from **inter-frame timing** in `QualityController`. That signal is fundamentally confounded on this workload (UI automation is "mostly static frames"):

- Android's encoder **repeats** the previous frame after ~100ms on a static screen (`VideoEncoder.KEY_REPEAT_PREVIOUS_FRAME_AFTER`) → a still screen streams at ~10fps.
- iOS ScreenCaptureKit **suppresses** idle buffers → a still screen makes no frame progress.

From timing alone, a static screen's ~10fps repeat cadence is **indistinguishable from a severe real drop to ~10fps**. #5569 deliberately chose the safe horn: intervals at/over the idle cadence (~90ms) are treated as idle and never downgrade, so the shipped auto-adjust covers the **moderate** degradation band (~12fps up to target) but does **not** catch *severe* (<~11fps) sustained drops.

## Ask
Give auto-adjust a health signal that timing cannot supply, so severe drops are caught without false-downgrading a static screen:

- Preferred: plumb the on-device `VideoStatsAccumulator` **dropped-frame count** (already computed in `android/video-server/.../VideoStatsAccumulator.kt`, currently only emitted as a `VIDEO_STATS` log line) to the desktop client over the stream socket, and drive downgrades off `dropped`/encoder-overload rather than inferred fps.
- Alternative/complement: client-side content-change detection to distinguish a repeated (idle) frame from a new one.

This pairs naturally with the already-noted server-side **shared-capture reconfigure** follow-up (a client-side re-subscribe only applies to a sole subscriber / the next subscribe; a live shared capture keeps the first subscriber's encode).

## Acceptance
- A sustained severe drop (e.g. 5fps under load) on an *actively changing* screen downgrades quality.
- A static screen (Android repeat cadence / iOS idle suppression) never downgrades.
- Signal is source-aware (works for both Android repeat-frame and iOS suppress-frame behavior).

Context: `QualityController` KDoc "Known limit" section; PR #5569 review threads.

kaeawc:
## Summary

- Carry the source encoder's cumulative dropped-frame counter from Android and iOS through the relay to the desktop controller.
- Downshift for sustained source-side encoder drops, while retaining timing-only logic for moderate degradation and recovery.
- Keep Android repeat frames and iOS idle suppression neutral: an unchanged source counter is never severe degradation.

## Root cause

Frame timing alone cannot distinguish a static Android screen's repeated ~10fps output from an actively changing stream at the same rate. A source-side encoder drop counter resolves that ambiguity without inferring content from decoded pixels: it remains flat during Android repeats and iOS idle suppression, and advances only when the encoder is overloaded.

## Validation

- `bun run turbo:validate`
- `bun run typecheck`
- `(cd android && ./gradlew :desktop-core:test)`
- `(cd android && ./gradlew :desktop-core:detekt --offline)`
- `scripts/ktfmt/validate_ktfmt.sh --only-changed-since origin/main`
- `(cd ios/screen-capture && swift test)`

Closes https://github.com/kaeawc/auto-mobile/issues/5582


kaeawc:
## Next steps

This PR was re-architected onto encoder-drop telemetry and now has **5 unresolved P1s and ~20 failing checks**. The telemetry is wired at the edges but doesn't actually reach production capture, so nothing ships as-is. Suggested order — the wiring bugs first, since the later fixes are unobservable until telemetry flows:

1. **Forward the callback through the default source factory.** `defaultDependencies().createCaptureSource` (`src/daemon/videoStreamSocketServer.ts:695–705`) rebuilds capture options but omits `onDroppedFrames`, so the callback never reaches the Android or iOS source. No telemetry is emitted in normal daemon use — the tests only pass because they inject a factory that consumes the callback directly. Fix this first; the rest is dead code until it's done.
2. **Forward the native iOS counter.** `src/features/webrtc/IosH264Source.ts` forwards the TS raw-frame-queue `frameMetrics` counter instead of the native `captureMetrics` (`recordEncoderDroppedFrame()` from `FrameWriter.metrics()`). The encoded Simulator path bypasses `enqueueFrame`, so VideoToolbox overload never reaches the relay. Forward from the `captureMetrics` handler and add a test on the encoded-Simulator path.
3. **Separate encoder-drop hysteresis from frame timing.** In `QualityController.kt`, `lowStreak` is incremented on counter updates but reset by ordinary frames in `classify`. On an actively-delivering 10fps stream the resets outpace the 3-increment threshold, so quality stays High indefinitely. Track encoder-drop evidence in its own streak (or a cumulative delta), not the cadence streak.
4. **Exclude iOS idle resumes from severe-drop streaks.** Post-idle iOS frames arrive with `contentChanged=true` (ScreenCaptureKit suppressed the idle buffers) and are misclassified as severe drops; five spaced changes downgrade High→Medium→Low on a healthy stream. The current iOS test masks this by supplying `contentChanged=false` — it needs active-cadence evidence or platform-aware handling.
5. **Recut the shipped screen-capture helper.** The iOS encoder-behind accounting is only in the rebuilt `screen-capture-helper`, but runtime resolves it via `ScreenCaptureHelperProvider.ensure()` against the checksum pinned in `src/constants/release.ts`, which is unchanged here. Default installs keep the old binary, so the iOS fix ships to no one. Recut the artifact and bump the pin — do this last, once the behavior is correct.

Then get the 20 red checks green. Until 1–2 land, the whole feature is unobservable in production regardless of the classifier work.

PR #5593, opened 2026-08-24
feat(android): stage session-bound shared storage fixtures

### LINKED ISSUE #5587: feat(android): add session-bound shared-storage fixture transfer and media indexing
## Summary

Add an Android MCP surface for staging files in shared device storage and making them visible to the system file and media pickers. Existing app-container file operations are useful for app-owned storage, but they cannot seed a scoped shared Downloads-style location or request a media refresh after writes.

## Use case

Automation needs to prepare a small, known set of files before an interaction that opens an Android system picker. The operation should be bound to the resolved device context, create a deterministic per-run directory, write supplied content, and report whether the files became discoverable by the relevant Android system surfaces.

## Proposed capability

Extend the existing file-management surface or add a focused Android tool that:

- accepts a host file, UTF-8 content, or base64 content using the existing input conventions;
- targets an allowlisted shared-storage scope, initially a caller-named subdirectory beneath Downloads;
- prepares that scope deterministically, including a safe replace/reset option limited to the declared namespace;
- triggers Android media indexing for written files when appropriate and reports the resulting per-file status;
- binds the request to the resolved device/session context and returns stable metadata such as device identity, logical destination, byte count, and indexing outcome;
- exposes actionable failures without requiring callers to construct platform shell commands or select transport endpoints themselves.

## Package observation

The existing public installed-app resources should remain the primary package-inventory surface. As part of this work, document the supported package-present or package-absent diagnostic path for workflows that need to correlate a shared-storage setup failure with the target app's installation state. Avoid adding a second package-listing API when the existing resource can provide the needed information.

## Acceptance criteria

- [ ] A discoverable MCP operation can write one or more files into a bounded Android shared-storage namespace on the resolved device.
- [ ] Resetting a namespace cannot delete outside the caller-declared shared-storage scope.
- [ ] A successful write returns structured metadata for every file, including destination, byte count, and media-indexing result or a clear reason indexing was not requested.
- [ ] Files written for media or document use become visible to the applicable Android system picker after the reported indexing step completes.
- [ ] The request supports the same host-path and inline-content modes as the existing file API.
- [ ] Package-presence diagnostics reuse or clearly document the existing public app-inventory surface.
- [ ] Unit coverage exercises validation, namespace containment, result handling, and indexing outcomes without requiring a physical device; a device-backed test covers picker visibility where feasible.
- [ ] Documentation includes a minimal Android shared-storage example and states the supported locations and limits.

## Non-goals

- Replacing app-container file operations.
- Arbitrary filesystem access or unbounded recursive deletion.
- Expanding to non-Android platforms in the first version.

kaeawc:
## Summary
- Add the session-bound `stageSharedStorageFixtures` Android tool for bounded Download namespaces.
- Reuse file source preparation for host paths, UTF-8 text, and base64 content.
- Add media-scan dispatch metadata and make installed-app inventory completion explicit.

## Validation
- `bun run typecheck`
- `bash scripts/oxlint-baseline.sh`
- Focused shared-storage and app-file tests (34 passing)
- `bun build.ts`
- Full `bun test --bail` reached 8437 tests but hit an unrelated flaky `CtrlProxyClient` singleton cleanup failure; the failing file passes independently.

Closes #5587


kaeawc:
## Next steps

The unresolved P1 ("Restore the advertised shared-storage tool") was written against an earlier revision where only the CtrlProxy test change was present. The feature is now restored on the head — `stageSharedStorageFixtures` is present in `src/server/sharedStorageTools.ts`, `schemas/tool-definitions.json`, `docs/design-docs/mcp/tools.md`, and the tests — so that finding is stale.

Remaining to land:
- Re-request review so the stale P1 thread resolves against the current SHA (it doesn't reflect head).
- Only red check is CodeQL, which looks like an infra flake — re-run it.
- Confirm `stageSharedStorageFixtures` is actually registered (not just defined) and appears in the generated tool count.

PR #5619, opened 2026-08-24
feat(ios): list installed apps on physical devices via devicectl

### LINKED ISSUE #2883: feat(ios): listApps/ListInstalledApps physical-device parity via devicectl device info apps
## Summary

`ListInstalledApps.executeiOS` ([`src/features/observe/ListInstalledApps.ts`](src/features/observe/ListInstalledApps.ts)) is **simctl-only**: it calls `simctl.listApps()`, which returns `[]` for a physical device. This makes `listApps` (and every installed-app check built on it) blind on real hardware.

Downstream impact:
- `launchApp` (#2487 / PR #2876) works around it by skipping the installed-app fallback check on physical devices (the devicectl launch error is authoritative there).
- `terminateApp` (see the terminateApp follow-up) mis-reports `wasInstalled: false` → false `success: true` no-op on device.
- `uninstallApp` uses the same `ListInstalledApps` for its pre/post checks.

## Proposed work

Back the `ListInstalledApps` iOS branch with `xcrun devicectl device info apps --device <udid> --json-output <file>` for physical UDIDs (selected via `isIosSimulatorUdid`). The repo already parses that envelope — `findBundleEntry` / `extractBundlePath` in `DeviceAppInspector.ts` — so this is mostly wiring + an injected fake and unit tests. This was called out as "option (b)" in #2487.

## References
- Issue #2487, PR #2876
- `src/features/observe/ListInstalledApps.ts`, `src/utils/ios-cmdline-tools/DeviceAppInspector.ts`

kaeawc:
`ListInstalledApps.executeIosDetailed` was simctl-only, so it returned `[]` on a physical device and every installed-app check built on it (`uninstallApp` pre/post checks, `terminateApp`'s `wasInstalled`) was blind on real hardware.

## What changed

- **`DeviceAppManager.listInstalledApps(deviceUdid)`** — runs `xcrun devicectl device info apps --device <udid> --json-output <file> --quiet`. Omitting `--bundle-id` is what turns the module's existing lookup into a full listing. macOS-only; devicectl exec/JSON failures are wrapped in `ActionableError` and **propagate**, so an empty list can only mean "no apps".
- **`extractInstalledAppEntries`** — pulls records out of devicectl's nested `result.apps` envelope or the flat `{ apps: [...] }` shape the existing fixtures pin, and drops non-object members.
- **`ListInstalledApps`** routes the iOS branch to an injectable `IosPhysicalAppLister` when the device id positively looks like a physical UDID (`isIosPhysicalUdid`). Anything else — simulator UUID or a non-UDID id — keeps the simctl path, so an unrecognized id degrades to today's behavior instead of shelling out to a tool that cannot serve it. The lister is constructed lazily, so simulator-only callers never build a `DeviceAppManager`.

Failure stays honest end to end: a devicectl error surfaces as `{ apps: [], successful: false }`, which is what keeps the resource cache from retaining a degraded listing.

## Tests

10 new unit tests, all with fakes — no device, no real devicectl:
- envelope extraction (nested, flat, junk entries, empty)
- the devicectl invocation shape (`device info apps`, `--device`, **no** `--bundle-id`), failure propagation, macOS guard
- physical UDID routes to devicectl and not simctl; metadata preserved; devicectl failure → `successful:false`; simulator UDID still uses simctl; duplicate bundle ids deduped

Verified failing before the source change (4 of the 5 `ListInstalledApps` cases fail on the parent commit; the simulator-path case passes by construction).

`bunx turbo run lint build test` green (14235 pass / 0 fail), `bun run typecheck` reports no new type errors.

## Manual verification needed

The listing itself cannot be proven without hardware. On a real iOS 17+ device, confirm `listApps` returns the installed set and that the records carry the bundle-id key `getIosInstalledAppBundleId` expects (`bundleIdentifier`). Also worth confirming on hardware: this uses devicectl's **default** app scope (no `--include-all-apps`/`--include-default-apps`), which is user-installed apps — narrower than simctl's listing. If parity needs the system apps too, that is a one-flag follow-up.

Closes #2883

### DIFF
--- a/src/features/observe/ListInstalledApps.ts
+++ b/src/features/observe/ListInstalledApps.ts
@@ -23,6 +23,8 @@ import {
   getIosInstalledAppBundleId,
   type IosInstalledAppRecord,
 } from "../../utils/ios-cmdline-tools/iosInstalledApp";
+import { DeviceAppManager } from "../../utils/ios-cmdline-tools/DeviceAppManager";
+import { isIosPhysicalUdid } from "../../utils/ios-cmdline-tools/iosDeviceType";
 
 const INSTALLED_APPS_CACHE_TTL_MS = 5 * 60 * 1000;
 
@@ -38,10 +40,20 @@ export interface IosInstalledAppsDetailedResult {
   successful: boolean;
 }
 
+/**
+ * Physical-device app listing seam (`devicectl device info apps`), narrowed to
+ * the single call this feature makes so tests can inject a fake without a
+ * `DeviceAppManager`.
+ */
+export interface IosPhysicalAppLister {
+  listInstalledApps(deviceUdid: string): Promise<IosInstalledAppRecord[]>;
+}
+
 interface ListInstalledAppsOptions {
   cacheEnabled?: boolean;
   installedAppsRepository?: InstalledAppsStore;
   timer?: Timer;
+  iosPhysicalAppLister?: IosPhysicalAppLister;
 }
 
 export class ListInstalledApps {
@@ -51,6 +63,7 @@ export class ListInstalledApps {
   private installedAppsRepository: InstalledAppsStore;
   private cacheEnabled: boolean;
   private timer: Timer;
+  private iosPhysicalAppLister: IosPhysicalAppLister | null;
   /**
    * Create an ListInstalledApps instance
    * @param device - Device to run ADB commands against
@@ -72,6 +85,17 @@ export class ListInstalledApps {
     const defaultCacheEnabled = adbFactory === defaultAdbClientFactory;
     this.cacheEnabled = options.cacheEnabled ?? defaultCacheEnabled;
     this.timer = options.timer ?? defaultTimer;
+    this.iosPhysicalAppLister = options.iosPhysicalAppLister ?? null;
+  }
+
+  /**
+   * Physical iOS devices have no simctl; `devicectl device info apps` is the
+   * only listing available there. Constructed lazily so simulator-only callers
+   * never build a DeviceAppManager.
+   */
+  private getIosPhysicalAppLister(): IosPhysicalAppLister {
+    this.iosPhysicalAppLister ??= new DeviceAppManager();
+    return this.iosPhysicalAppLister;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


kaeawc:
Review triage (pushed as 424efa537):

**Fixed**
- *No `apps` array in the payload* (Codex P2). `extractInstalledAppEntries` now returns null for a payload it does not recognize and `listInstalledApps` throws, instead of degrading to `[]` — which callers read as "no apps installed", exactly the conflation the method exists to avoid. A present-but-empty `apps` array still returns `[]`. Two tests pin the distinction.
- *Cleanup masking the primary error* (CodeRabbit). The `finally` block's `rm` failure replaced the devicectl/JSON diagnostic; it is now logged and swallowed, with a test that the devicectl message survives a failing cleanup.

**Filed as follow-ups** — both accurate, both pre-existing, neither a regression from this PR:
- #5620 — physical devices are never resolved by the app resources (`getBootedIosDevicesIfAvailable` returns simulators only), so the resource path cannot reach this listing even now that it exists. Verified at `src/utils/deviceUtils.ts:337-345`.
- #5621 — `UninstallApp`/`terminateApp` pre-check via `execute()`, which discards the `successful` flag, so a failed listing reads as "app absent" and no-ops with `success: true`. Verified at `src/features/action/UninstallApp.ts:109-114`. Reachable in a new way now that a physical listing can genuinely fail, but the mis-report predates this change (the physical listing previously returned `[]` unconditionally).

**Not taking**
- *Fake filesystem for the new tests* (CodeRabbit). `DeviceAppManager.test.ts` uses real temp dirs throughout — that is the file's established pattern, and it is what exercises the JSON-output round trip that this code's contract is about. Introducing a fake FS for four new cases would leave the file inconsistent; if the suite should move off real I/O, that is a whole-file change, not a rider on this PR. The full file runs in ~200ms.

kaeawc:
Review triage round 2 (pushed as 8235b49ab):

**Fixed**
- *Reject nonempty listings with no valid entries* (Codex P2). Correct, and the same conflation as the missing-envelope case: an `apps` array that carried members but none we can read is a payload shape we do not understand, not an empty device. `extractInstalledAppEntries` now returns `[]` only for an originally empty array and null when a nonempty array filters down to nothing; `listInstalledApps` throws on null. Tests pin both.
- *Preserve physical app paths in the apps resource* (Codex P2). Also correct. Path reading and `file://` normalization now live in `iosInstalledApp.ts` as one primitive (`getIosInstalledAppPath` / `normalizeIosDevicePath`), shared by `appResources.extractIosPath()` and `DeviceAppManager`, so `url`/`bundleURL` reach the resource as filesystem paths. One deliberate side effect worth flagging: the merged key list prefers an explicit bundle path (`bundlePath`, then the devicectl URL keys, then `path`/`Path`) over a container (`bundleContainer`, `dataContainer`), where the old resource list put `bundleContainer` second. For a record carrying both, the `.app` path is the better answer; full suite is green either way.
- *Reject JSON payloads without an apps array* (Codex P2) — already fixed in 424efa537; the thread predates that push. Resolving it.

**Still open, deliberately**
- #5620 (physical devices are never resolved by the app resources) and #5621 (a failed listing reads as "app absent" in the uninstall pre-check) stay as filed follow-ups. Both are accurate and neither is a regression from this PR; leaving those two threads open so the reviewer can see them against the issues.
- Fake filesystem for the new tests (CodeRabbit): unchanged position from the previous triage — real temp dirs are this file's established pattern and are what exercise the JSON round trip; moving off host I/O is a whole-file change, not a rider here.

Validation: `bun run lint` (ratchet clean), `bun run build`, `bun test` (14,247 pass / 0 fail), `bun run typecheck` (no new errors).

Issue #5674, opened 2026-08-25
fix(ios): CtrlProxy WebSocket ignores FIN bit and drops fragmented messages


kaeawc:
## Summary

The iOS CtrlProxy WebSocket server ignores the frame FIN bit and does not handle
continuation frames (opcode `0x0`). A fragmented WebSocket message is silently
truncated to its first fragment, and continuation frames are read but their
payload discarded.

## Where

- `ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift` —
  `WebSocketConnection.parseWebSocketFrame(_:)` and `readPayload(length:isMasked:opcode:)`.

## Root cause

`parseWebSocketFrame` extracts `opcode = byte0 & 0x0F` but never inspects the FIN
bit (`byte0 & 0x80`). `readPayload` only delivers a message when
`opcode == 0x01 || opcode == 0x02`:

```swift
if opcode == 0x01 || opcode == 0x02 {
    self.onMessage(payload)
}
```

So a message split across frames (RFC 6455 §5.4) — first frame FIN=0 with a
text/binary opcode, subsequent FIN=0 frames with the continuation opcode `0x0`,
final frame FIN=1 with `0x0` — is handled as: first fragment delivered alone
(truncated), continuation frames consumed but dropped. There is no reassembly and
no rejection of the unsupported case.

## Impact

Latent/dormant today: the current TS `ws` client sends these small JSON control
messages as single unfragmented frames, so fragmentation isn't exercised. But any
peer that fragments (large payloads, a proxy, a future streaming client) gets a
truncated/dropped message with no error signal. Low-med correctness bug.

Runner-side change: ships as a pinned, checksummed artifact; the source fix does
not reach production until the runner is re-cut (tracked as a separate follow-on).

## Acceptance criteria

1. A fragmented text/binary message (initial frame FIN=0 with opcode `0x1`/`0x2`,
   zero or more continuation frames FIN=0 opcode `0x0`, final frame FIN=1 opcode
   `0x0`) is reassembled in order and delivered once, complete, via `onMessage`.
2. The FIN bit is honored: a single-frame message (FIN=1) is delivered exactly as
   today (no regression), and a non-final frame does not trigger delivery.
3. Interleaved control frames (ping/close) between fragments do not corrupt the
   reassembly buffer (control frames are handled independently per RFC §5.4).
4. Reassembly is bounded by the existing `maxFramePayloadLength` (64 MiB) applied
[...]

Issue #5669, opened 2026-08-24
fix(ios): CtrlProxy WebSocket ping frame desyncs the wire (mask/payload not consumed)


kaeawc:
## Summary

The iOS CtrlProxy WebSocket server mishandles inbound **ping** frames: it sends a
pong but never consumes the ping frame's masking key (and any ping payload) from
the stream before reading the next frame header, desynchronizing the wire. It
also violates RFC 6455 §5.5.2/§5.5.3 by not echoing the ping payload in the pong.

## Where

- `ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift` —
  `WebSocketConnection.parseWebSocketFrame(_:)`, the `opcode == 0x09` branch.

## Root cause

Only the 2-byte frame header has been read when the ping branch runs:

```swift
// Handle ping
if opcode == 0x09 {
    let pongFrame = createWebSocketFrame(data: Data(), opcode: 0x0A)
    connection.send(content: pongFrame, completion: .contentProcessed { _ in })
    receiveWebSocketFrame()   // ← immediately reads the NEXT 2 bytes as a header
    return
}
```

A client→server frame is **always masked** (RFC 6455 §5.3), so after the 2-byte
header there are 4 masking-key bytes (plus any ping payload) still unread in the
stream. `receiveWebSocketFrame()` then reads those masking-key bytes as if they
were a new frame header → protocol desync → subsequent frames are misparsed and
the connection garbles/drops.

Note the asymmetry that confirms the defect: pong (`0x0A`) and other non-control
opcodes fall through to `readPayload`, which correctly consumes `payload + mask`;
only the ping branch short-circuits before consuming those bytes.

## Impact

Latent/dormant today: the current TS `ws` client does not send ping frames, so
this path isn't exercised in production. But any pinging WebSocket peer (a browser
inspector, a future keepalive) desyncs the connection. Low-med correctness bug.

This is a change to the iOS control-proxy **runner**, which ships as a pinned,
checksummed release artifact — the source fix here does not reach production until
the runner is re-cut (tracked separately as a follow-on).

## Acceptance criteria

1. On receiving a masked ping frame, the server consumes the full frame
   (4-byte masking key + `payloadLength` payload bytes) before reading the next
   frame header — no stream desync.
2. The pong reply echoes the (unmasked) ping application data, per RFC 6455
   §5.5.3 (empty payload → empty pong, which is already the case).
3. Ping payload length is validated with the same bound already applied to data
[...]

PR #5675, opened 2026-08-25
fix(ios): honor WebSocket FIN bit and reassemble fragmented CtrlProxy messages

kaeawc:
## Summary

The iOS CtrlProxy WebSocket server ignored the frame FIN bit and did not handle continuation frames (opcode `0x0`), so a fragmented message (RFC 6455 §5.4) was silently truncated to its first fragment and continuation payloads were read then discarded — with no reassembly and no error signal.

This threads the FIN bit through the frame path and adds per-connection reassembly, built on top of the control-frame refactor from #5671 (the pure `frameAction` helper and `isValidControlFramePayloadLength`).

## Linked Issue

Closes #5674

## Bug / Regression Context

- **Observed symptom:** any peer that fragments a message (large payloads, a proxy, a future streaming client) got a truncated/dropped message with no error. Latent today because the current TS `ws` client sends unfragmented frames.
- **Root cause:** `parseWebSocketFrame` extracted `opcode` but never inspected the FIN bit; delivery happened per-frame for opcode `0x1`/`0x2` only, so continuation frames were dropped.

## Changes

- New pure `WebSocketConnection.reassemble(...)` state machine (mirrors #5671's `frameAction`): decides `deliver` / `buffered` / `protocolError` from opcode + FIN + in-progress state. Unit-testable without a socket.
- Per-connection reassembly state (`fragmentedOpcode`, `fragmentBuffer`) mutated only on the serialized server `queue` — no new locks.
- FIN threaded from `parseWebSocketFrame` → `readExtendedLength` → `readPayload`; data/continuation frames route through reassembly, control frames stay on the independent `frameAction` path and never touch the buffer.
- Total reassembled size bounded by the existing `maxFramePayloadLength` (64 MiB) applied cumulatively; overflow and malformed sequences close the connection with a trace.

## Acceptance criteria

1. Fragmented message reassembled in order, delivered once — `testTwoFragmentMessageReassembles`, `testThreeFragmentMessageReassemblesInOrder`, `testFragmentedCommandReassemblesEndToEnd` (real socket).
2. FIN honored; single-frame unchanged; non-final does not deliver — `testSingleUnfragmentedFrameDeliversUnchanged`, `testNonFinalStartFrameDoesNotDeliver`.
3. Interleaved control frames don't corrupt reassembly — `testControlFrameBetweenFragmentsDoesNotCorruptReassembly` + interleaved ping in the end-to-end test.
[...]

PR #5679, opened 2026-08-25
fix(ios): CtrlProxy drops WebSocket frames pipelined with the upgrade handshake

### LINKED ISSUE #5678: fix(ios): CtrlProxy drops WebSocket frames pipelined with the upgrade handshake
## Summary

After completing the WebSocket upgrade handshake, the iOS CtrlProxy server issues
a fresh socket `receive` and never consults bytes already buffered in
`pendingHTTPRequest` beyond the handshake request. Any WebSocket frame that
arrives in the **same TCP segment** as the upgrade request (pipelined) is silently
dropped.

## Where

- `ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift` —
  `WebSocketConnection.receiveHTTPUpgrade()` (the
  `pendingHTTPRequest.removeFirst(requestLength)` slice) and the transition into
  `receiveWebSocketFrame()` after `handleWebSocketUpgrade`.

## Root cause

`receiveHTTPUpgrade()` accumulates bytes into `pendingHTTPRequest`, finds one
complete HTTP request, slices it out with
`pendingHTTPRequest.removeFirst(requestLength)`, and dispatches it. On the
WebSocket-upgrade branch, `handleWebSocketUpgrade` sends the `101` response and
then calls `receiveWebSocketFrame()`, which starts a brand-new
`connection.receive(...)`. Any residual bytes left in `pendingHTTPRequest` after
the slice (a first WS frame delivered in the same read) are never handed to the
frame parser — they are stranded in `pendingHTTPRequest` and lost.

The frame-reading path is pull-based (`receive` exact byte counts) and has no
notion of the handshake's leftover buffer, so the two stages don't share an
inbound byte stream.

## Impact

Latent/dormant today: real WebSocket clients (the TS `ws` library) wait for the
`101` response before sending any frame, so a frame is not normally pipelined with
the handshake. But a client that pipelines (or a TCP stack that coalesces
segments under load) loses its first frame with no error. Low severity.

Runner-side change: ships as a pinned, checksummed artifact; the source fix does
not reach production until the runner is re-cut (tracked as a separate follow-on).

## Acceptance criteria

1. Any bytes remaining in the connection's inbound buffer after the upgrade
   request is sliced out are processed as WebSocket frame data — a frame pipelined
   in the same TCP segment as the upgrade request is delivered, not dropped.
2. The normal case (client waits for `101`, then sends frames in later segments)
   is unchanged — no double-processing, no regression to handshake parsing.
3. The residual-carry path is bounded by the same limits already applied
   (`maximumHTTPRequestLength` for handshake accumulation; `maxFramePayloadLength`
   for frames) — no unbounded buffering.
4. Non-WebSocket branches (`GET /health`, `POST /sdk-events`, `GET /sdk-events`)
   are unaffected; they already close the connection after responding, so residual
   handling there is out of scope (document that explicitly).
5. Swift unit tests in `CtrlProxyTests/WebSocketServerTests.swift` pin: a buffer
   containing an upgrade request immediately followed by a complete masked WS frame
   yields both the upgrade AND the delivered frame; and the frame-arrives-later
   case still works.

## Notes

- The clean implementation is a single per-connection inbound byte buffer that
  both the handshake reader and the frame reader drain from (refilling from the
  socket as needed), rather than two disjoint read paths. If the FIN/fragmentation
  fix (issue #5674) already introduced per-connection inbound buffering, build on
  that; otherwise this is the issue that introduces it. All receive callbacks run
  serialized on the server `queue`, so the buffer needs no extra locking — confirm
  that invariant.
- Reuse existing helpers (`completeHTTPRequestLength`, `frameReadLength`,
  `unmaskFrame`); do not add a parallel parse path.
- Last of five WebSocket-inspection findings shipped independently; the Swift
  framing re-cut is captured as a single follow-on issue.

kaeawc:
## Summary

After the WebSocket upgrade handshake, CtrlProxy sliced one complete HTTP request out of its inbound buffer and then started a *fresh* socket `receive` for frames. Any WebSocket frame that arrived in the **same TCP segment** as the upgrade request (pipelined) was stranded in the leftover buffer and silently dropped — the handshake reader and the frame reader did not share an inbound byte stream.

This makes the inbound buffer a single per-connection byte buffer (`pendingHTTPRequest` → `inboundBuffer`) that **both** the handshake reader and the frame reader drain from. A new `receiveFrameBytes(_:onData:)` helper consumes any buffered bytes first and refills from the socket only for the shortfall, so a pipelined first frame is parsed instead of lost.

## Linked Issue

Closes #5678

## Bug / Regression Context

- **Prior attempts:** builds on the shared-buffer direction of the recent WebSocket framing work — ping desync (#5669) and FIN/fragmentation reassembly (#5674/#5675). The fragmentation fix added *message* reassembly state (`fragmentedOpcode`/`fragmentBuffer`) but not an *inbound byte* buffer, so this PR introduces that single shared buffer.
- **Observed symptom:** a first WS frame pipelined with the handshake is dropped with no error.
- **Repro steps / conditions:** latent today — real clients (`ws`) wait for the `101` before sending — but a client that pipelines, or a TCP stack that coalesces segments under load, loses its first frame.

## Changes

- `inboundBuffer` is drained by both the handshake reader (`receiveHTTPUpgrade`) and the frame reader.
- New `receiveFrameBytes(_:onData:)` drains buffered bytes first, refilling from the socket only when short; all three frame-read sites (`receiveWebSocketFrame`, `readExtendedLength`, `readPayload`) go through it — no parallel parse path.
- Reuses `completeHTTPRequestLength` / `frameReadLength` / `unmaskFrame`. Bounds unchanged: handshake accumulation by `maximumHTTPRequestLength`, per-frame by `frameReadLength`/`maxFramePayloadLength`.
- HTTP branches (`/health`, `POST`/`GET /sdk-events`) close after responding, so residual-carry there is out of scope by design — documented inline (AC4).
- All receive completions run serialized on the server `queue`, so the buffer needs no lock (same invariant as the fragmentation reassembly state).

## Acceptance criteria

1. Residual bytes after the upgrade slice are processed as frame data — pipelined frame delivered, not dropped ✅ (`testFramePipelinedWithUpgradeIsDelivered`, real socket, upgrade+frame in one send).
2. Normal case (frame in a later segment) unchanged, no double-processing ✅ (`testFrameArrivingAfterUpgradeStillDelivered`).
3. Residual-carry bounded by the existing limits (no unbounded buffering) ✅ (`readPayload` rejects over-cap frames before buffering; helper appends only the exact shortfall).
4. Non-WebSocket branches unaffected — they already close after responding ✅ (documented inline).
5. Swift tests pin both the pipelined and frame-arrives-later cases ✅.

## Runner note

This is a Swift change to the iOS control-proxy **runner**, which ships as a pinned, checksummed release artifact. The source fix does **not** reach production until the runner is re-cut — tracked separately as a follow-on and **out of scope** for this PR.

## Validation

- [x] `./scripts/ios/swift-build.sh` and `./scripts/ios/swift-test.sh` green (control-proxy 45 WebSocketServer tests incl. the 2 new; all Swift packages pass)
- [x] `bun run turbo:validate` (lint + build + test + boundary checks) exit 0 — 14340 TS tests pass, 0 fail
- [x] SwiftLint: 0 serious, no new warnings (the 2 remaining warnings pre-exist on `main`)
- [x] XcodeGen project files in sync (no file added/removed/renamed; no `.pbxproj` change)
- [x] TDD: pipelined test confirmed red (timeout, frame dropped) before the fix, green after

## Device Verification

- [ ] Not run on-device — runner re-cut is a separate follow-on; behavior pinned by real-socket Swift unit tests over loopback.

### DIFF
--- a/ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift
+++ b/ios/control-proxy/Sources/CtrlProxy/WebSocketServer.swift
@@ -644,7 +644,16 @@ class WebSocketConnection: WebSocketResponding {
     /// can detect a runner/client port mismatch (issue #2735).
     private let boundPort: UInt16
     private var isWebSocketUpgraded = false
-    private var pendingHTTPRequest = Data()
+    /// Single per-connection inbound byte buffer drained by **both** the handshake
+    /// reader (`receiveHTTPUpgrade`) and the frame reader (`receiveFrameBytes`).
+    /// The handshake reader accumulates socket reads here and slices out one
+    /// complete HTTP request; any bytes left after that slice — most importantly a
+    /// WebSocket frame pipelined in the same TCP segment as the upgrade request —
+    /// stay in the buffer and are handed to the frame parser rather than being
+    /// stranded and lost (issue #5678). Mutated only from `connection.receive`
+    /// completions, which all run on the server `queue`, so no lock is needed
+    /// (same invariant as `fragmentedOpcode`/`fragmentBuffer`).
+    private var inboundBuffer = Data()
     private static let maximumHTTPRequestLength = 1_000_000
 
     /// In-progress fragmented-message reassembly state (RFC 6455 §5.4). Both are
@@ -728,25 +737,32 @@ class WebSocketConnection: WebSocketResponding {
                 return
             }
 
-            self.pendingHTTPRequest.append(data)
-            guard self.pendingHTTPRequest.count <= Self.maximumHTTPRequestLength else {
+            self.inboundBuffer.append(data)
+            guard self.inboundBuffer.count <= Self.maximumHTTPRequestLength else {
                 print("[WebSocketConnection] HTTP request exceeds maximum length")
                 self.connection.cancel()
                 return
             }
 
-            guard let requestLength = Self.completeHTTPRequestLength(in: self.pendingHTTPRequest) else {
+            guard let requestLength = Self.completeHTTPRequestLength(in: self.inboundBuffer) else {
                 self.receiveHTTPUpgrade()
                 return
             }
 
-            let requestData = Data(self.pendingHTTPRequest.prefix(requestLength))
-            self.pendingHTTPRequest.removeFirst(requestLength)
+            let requestData = Data(self.inboundBuffer.prefix(requestLength))
+            self.inboundBuffer.removeFirst(requestLength)
             guard let request = String(data: requestData, encoding: .utf8) else {
                 self.connection.cancel()
                 return
             }
 
+            // Only the WebSocket-upgrade branch consumes bytes left in
+            // `inboundBuffer` after the slice: `receiveWebSocketFrame` drains them
+            // first (issue #5678). The HTTP branches (`GET /health`,
+            // `POST /sdk-events`, `GET /sdk-events`) each send their response and
+            // then `connection.cancel()`, so any residual bytes after their request
+            // are discarded with the connection — residual-carry there is out of
+            // scope by design (AC4).
             if request.contains("Upgrade: websocket") || request.contains("upgrade: websocket") {
                 self.handleWebSocketUpgrade(request)
             } else if request.contains("GET /health") {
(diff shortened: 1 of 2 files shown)


kaeawc:
Nightly manual-test — code **verified working in a local runner build**, but **not yet delivered to users**.

Verification (local build, `AUTOMOBILE_CTRL_PROXY_IOS_USE_LOCAL_BUILD=true`, derived data `/tmp/automobile-ctrl-proxy`, daemon log confirms `Local-build mode`):

- `observe` on iPhone 17 Pro / iOS 26.2 returns a well-formed hierarchy (SpringBoard, `screenSize 402x874`).
- After `launchApp com.apple.Preferences`, `observe` transfers a **164,392-byte** hierarchy intact, with `activeWindow.appId = com.apple.Preferences`. A payload that size cannot fit in a single WebSocket frame, so the fragmentation/FIN-reassembly path was necessarily exercised end to end with no truncation or decode error.

**Delivery gate.** The runner re-cut `990f58b73` ("chore: update CtrlProxy APK, iOS IPA, and iOS runner SHA256") landed *before* this fix in `d74377207..HEAD`:

```
1ec216bae fix(ios): consume masked ping frame + echo pong payload (#5671)
990f58b73 chore: update CtrlProxy APK, iOS IPA, and iOS runner SHA256   <-- re-cut here
70e4c953e fix(ios): honor WebSocket FIN bit and reassemble fragmented CtrlProxy messages (#5675)
16ce03d4b fix(ios): CtrlProxy drops WebSocket frames pipelined with the upgrade handshake (#5679)
```

So the pinned runner `runnerSha256: e392dd7f...` (`src/constants/release.ts:71`) contains #5671 but **not** #5675 or #5679. Against the shipped runner the daemon refuses to launch a locally built binary (SHA guard), so users stay on the pre-fix runner until the next re-cut.

**Release-checklist item: re-cut the iOS runner so #5675 and #5679 actually reach users.**

PR #5755, opened 2026-08-26
fix(ios): retry UserDefaults defaults commands

### LINKED ISSUE #5754: fix(ios): give simulator UserDefaults commands a cold-start-safe timeout
## Summary

`setPreference` can fail on a booted iOS Simulator because its `defaults` read-back command has a fixed 5-second deadline. A simulator that is temporarily busy during app install or early startup can exceed that deadline even though it becomes responsive shortly afterward.

## Current behavior

1. A caller writes an iOS `userDefaults` preference with `setPreference`.
2. The tool performs a `defaults` read-back verification through `xcrun simctl spawn`.
3. If that command runs longer than 5 seconds, the whole operation fails. The write may already have succeeded, leaving the caller with a false-negative result.

## Proposed timing policy

The normal path should complete quickly; this policy is a bounded recovery allowance, not an expected latency target.

- Give each iOS UserDefaults `defaults` read, write, or type-read command a 10-second deadline.
- Retry once only after a timeout or a transient CoreSimulator connection failure.
- Bound the complete `setPreference` operation, including write and verification, to 30 seconds.
- Preserve the final command error and do not retry missing-key responses or input-validation failures.

## Reproduction

1. Start or freshly prepare an iOS Simulator under normal host load.
2. Invoke `setPreference` with `scope: "userDefaults"`.
3. Observe that the `simctl spawn … defaults read …` verification can exceed five seconds and the tool reports failure.

## Acceptance criteria

- iOS UserDefaults read, write, and type-read commands use the 10-second per-command bound.
- A timeout or transient CoreSimulator connection failure gets at most one retry.
- The complete preference operation remains bounded to 30 seconds.
- A successful write is not reported as failed solely because a 5-second read-back budget was too short.
- Unit tests cover the timing and retry policy for write, read, and type-read paths.
- Existing missing-key behavior remains unchanged.

kaeawc:
## Summary

Extends iOS UserDefaults `defaults` commands to a 10-second per-command deadline, retries one transient timeout or CoreSimulator connection failure, and shares a 30-second deadline across a write and its read-back verification.

## Linked Issue

Closes [#5754](https://github.com/kaeawc/auto-mobile/issues/5754)

## Bug / Regression Context

- **Observed symptom:** A temporarily busy booted simulator could make `setPreference` fail during a five-second `defaults` read-back despite a successful write.
- **Repro steps / conditions:** Set an iOS `userDefaults` preference while the simulator is under early-startup or installation load.

## Validation

- [x] `bun test test/features/preferences/AppPreferences.test.ts`
- [x] `bun run typecheck`
- [x] Tests use an injected FakeTimer for the operation deadline.

### DIFF
--- a/src/features/preferences/AppPreferences.ts
+++ b/src/features/preferences/AppPreferences.ts
@@ -103,7 +103,7 @@ export class AppPreferences {
       return this.getAndroidSharedPreference(input);
     }
 
-    return this.getIosUserDefault(input, this.iosDefaultsDeadline());
+    return this.getIosUserDefault(input);
   }
 
   async setPreference(input: SetPreferenceInput): Promise<PreferenceResult> {
@@ -223,7 +223,7 @@ export class AppPreferences {
 
   private async getIosUserDefault(
     input: GetPreferenceInput,
-    deadlineMs: number,
+    deadlineMs?: number,
   ): Promise<PreferenceResult> {
     if (!isIosSimulatorDevice(this.device)) {
       throw unsupportedPhysicalIosUserDefaultsError();
@@ -269,7 +269,7 @@ export class AppPreferences {
 
   private async readIosDefaultsType(
     input: GetPreferenceInput,
-    deadlineMs: number,
+    deadlineMs?: number,
   ): Promise<PreferenceValueType | undefined> {
     const domain = iosDefaultsDomain(input);
     try {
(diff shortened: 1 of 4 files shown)


kaeawc:
### Nightly `manual-test` — **VERIFIED FIXED** (iOS 26.2 sim, iPhone 17 Pro, HEAD `87e15e05d`)

Read-back verification is present and working. `setPreference` now returns
`verified: true`, and the value round-trips:

```
setPreference userDefaults com.apple.Preferences AutoMobileMT2 = "v2" (string)
  -> success:true found:true verified:true value:"v2"
getPreference userDefaults com.apple.Preferences AutoMobileMT2
  -> success:true found:true value:"v2" type:"string"
```

A `bool` write verified the same way. Dist freshness was confirmed by grepping
the compiled bundle for this PR's strings (`Failed to read iOS UserDefaults type
with defaults`, `defaults type read found no value`) before the run, not by the
daemon version string.

One adjacent defect surfaced while exercising this path: `getPreference` on a
**missing** key throws instead of returning `found:false`. That is a pre-existing
`$`-anchor bug in `looksLikeMissingIosDefault` (from #2717), not a regression
here — this PR only widened its blast radius by adding a second throw site on the
same predicate. Filed as #5763.

PR #5758, opened 2026-08-26
fix(tap): prefer actionable dialog text matches

### LINKED ISSUE #5638: bug(tap): prefer an actionable text match over an identically named modal title
# bug(tap): prefer an actionable text match over an identically named modal title

## Labels

- `bug`
- `ios`
- `a11y`

## Milestone

`iOS Platform Foundation` (2)

## Relationships

- Relates to [#5618](https://github.com/kaeawc/auto-mobile/issues/5618) because both issues involve `tapOn` behavior across dialog or window hierarchies.
- Not blocked by [#5618](https://github.com/kaeawc/auto-mobile/issues/5618): this issue concerns ordinary text-selector ranking, rather than semantic-link owner uniqueness.
- No duplicate identified.

## Summary

When a confirmation dialog gives its static title and its confirmation button the same visible label, an unindexed `tapOn` text selector chooses the first visible text match. That can select the non-actionable title instead of the confirmation button, forcing callers to guess hierarchy indices.

This is especially brittle for destructive confirmations: an index can change when background content, overlays, or accessibility hierarchy order changes.

## Reproduction

1. Open an iOS confirmation dialog whose title and confirm button have the same text, such as `Sign Out`.
2. Call `tapOn` with `{ "selector": { "text": "Sign Out" }, "platform": "ios" }`.
3. Observe that the title is selected before the actionable `UIButton`.

## Expected behavior

For the default, unindexed text selection path, `tapOn` should prefer a visible actionable node with the requested text over a non-actionable text node with the same label. A node with a click accessibility action must count as actionable even when it has no `clickable` boolean.

## Acceptance criteria

- An unindexed `tapOn` text selector chooses the visible confirmation button in a dialog containing both a static title and a button with the same text.
- Ranking prefers the active/topmost dialog window before background content when otherwise comparable.
- If no actionable match exists, retain the current first-visible-text fallback.
- Explicit `index` keeps its current hierarchy-order semantics; callers can still deliberately select a non-actionable match.
- `selectionStrategy: "random"` behavior remains unchanged.
- The selected-element metadata identifies the actual actionable target.
- Add deterministic unit coverage for a hierarchy with a static dialog title plus a button that exposes `actions: ["click"]` but omits `clickable`.

## Suggested implementation direction

Keep the existing matching API, but change the default unindexed text-selection path to rank visible matching nodes by dialog/window priority and actionability before falling back to the existing document-order result. Reuse the canonical click predicate so iOS accessibility actions and explicit `clickable` flags behave consistently.

kaeawc:
## Summary

Fixes unindexed `tapOn` text selection when a dialog title and its confirmation button share the same label. Default first selection now searches topmost window content and prefers actionable matches, including iOS nodes exposing `actions: ["click"]` without a `clickable` flag.

## Linked Issue

Closes #5638

## Bug / Regression Context

- **Prior attempts:** None.
- **Observed symptom:** A static confirmation-dialog title could be selected before the actionable confirmation button.
- **Repro steps / conditions:** Call unindexed `tapOn` with a text selector matching both an iOS dialog title and confirmation button.

## Validation

- [x] Focused selector, finder, parser, tap, and accessibility tests
- [x] `bun test`
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] `bun run format:check`
- [ ] `scripts/all_fast_validate_checks.sh` (blocked locally by an environment ktfmt version mismatch; no Kotlin files changed)

## Follow-ups

No follow-ups identified.

### DIFF
--- a/src/features/utility/DefaultElementSelector.ts
+++ b/src/features/utility/DefaultElementSelector.ts
@@ -11,7 +11,7 @@ function shouldIncludeWindowsForTextSelection(
   index: number | undefined,
   strategy: ElementSelectionStrategy,
 ): boolean {
-  return index !== undefined || strategy === "first";
+  return index === undefined && strategy === "first";
 }
 
 export class DefaultElementSelector implements ElementSelector {
--- a/src/features/utility/DefaultElementSelector.ts
+++ b/src/features/utility/DefaultElementSelector.ts
@@ -7,6 +7,13 @@ import type { ElementFinder } from "../../utils/interfaces/ElementFinder";
 import { defaultRandom } from "../../utils/Random";
 import { DefaultElementFinder } from "./ElementFinder";
 
+function shouldIncludeWindowsForTextSelection(
+  index: number | undefined,
+  strategy: ElementSelectionStrategy,
+): boolean {
+  return index !== undefined || strategy === "first";
+}
+
 export class DefaultElementSelector implements ElementSelector {
   private finder: ElementFinder;
   private random: () => number;
@@ -31,6 +38,7 @@ export class DefaultElementSelector implements ElementSelector {
     },
   ): ElementSelectionResult {
     const strategy = options?.strategy ?? "first";
+    const includeWindows = shouldIncludeWindowsForTextSelection(options?.index, strategy);
     const matches = this.finder.findElementsByText(
       viewHierarchy,
       text,
@@ -38,6 +46,7 @@ export class DefaultElementSelector implements ElementSelector {
       options?.partialMatch ?? true,
       options?.caseSensitive ?? false,
       options?.index !== undefined,
+      includeWindows,
     );
     return this.pickMatch(matches, strategy, viewHierarchy, options?.index);
   }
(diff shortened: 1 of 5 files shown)


kaeawc:
### Nightly `manual-test` — **VERIFIED, no regression** (Android API 34 + API 31 emulators, HEAD `87e15e05d`)

Exercised against real system dialogs (ANR "isn't responding" alerts, which are
genuine topmost modal windows over a live background app):

- `tapOn {text:"Wait"}` resolved to `android:id/aerr_wait`, `class=android.widget.Button`,
  `clickable=true`, bounds `[70,1264,1010,1390]` — the clickable control in the
  topmost dialog window, and the ANR was dismissed.
- `tapOn {text:"Close app"}` resolved to `android:id/aerr_close` the same way.
- On API 31 the same path recovered a wedged `Process system isn't responding`
  dialog, after which `observe` went from 3 elements (`appId:"android"`, dialog
  masking everything) to 22 with `appId:"com.android.settings"`.
- Text-collision check on the playground Semantic Links screen: with
  "Privacy Policy" present both as a standalone link and inside
  "Last activated: Privacy Policy", `tapOn {text:"Privacy Policy"}` selected the
  exact match at `[42,399,135,420]` (`totalMatches` picked the link, not the
  partial).

Caveat on strength of evidence: these confirm correct behavior and no regression,
but they are not strict reproduce-then-confirm — I could not construct a live
case where a *covered background clickable* competed with a topmost
*non-clickable* dialog match, which is the exact ordering this PR changed. That
discrimination remains covered by the unit tests in
`DefaultElementSelector.test.ts`.

PR #5759, opened 2026-08-26
fix(tap): preserve dialog window z-order

kaeawc:
## Summary

Preserve dialog/window z-order while applying the actionable-text preference introduced for #5638. Text matches are now ranked for actionability within each window group, while topmost windows remain ahead of lower and background windows.

## Linked Issue

Follow-up to #5638

## Bug / Regression Context

- **Observed symptom:** A clickable match in a covered lower window could outrank a static match in the topmost modal.
- **Cause:** Exact matches from all windows were flattened and globally sorted by actionability, discarding window layering.
- **Fix:** Rank each topmost-first window group independently, then concatenate groups in z-order.

## Validation

- [x] Added deterministic regression coverage for layered windows
- [x] Focused `DefaultElementSelector` and `ElementFinder` tests
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] Formatting check

### DIFF
--- a/src/features/utility/ElementFinder.ts
+++ b/src/features/utility/ElementFinder.ts
@@ -336,6 +336,18 @@ export class DefaultElementFinder implements ElementFinder {
     return isClickableElementProperties(props);
   }
 
+  private selectAndRankTextMatches(matches: {
+    exactMatches: Element[];
+    partialMatches: Element[];
+  }): Element[] {
+    const selectedMatches =
+      matches.exactMatches.length > 0 ? matches.exactMatches : matches.partialMatches;
+    selectedMatches.sort(
+      (a, b) => Number(this.isClickableNode(b)) - Number(this.isClickableNode(a)),
+    );
+    return selectedMatches;
+  }
+
   private isCollectionNode(props: Record<string, unknown>): boolean {
     const className = typeof props.class === "string" ? props.class : "";
     const scrollable = props.scrollable === "true" || props.scrollable === true;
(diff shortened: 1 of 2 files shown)


kaeawc:
### Nightly `manual-test` — **VERIFIED, no regression** (Android API 34 + API 31 emulators, HEAD `87e15e05d`)

Exercised against real system dialogs (ANR "isn't responding" alerts, which are
genuine topmost modal windows over a live background app):

- `tapOn {text:"Wait"}` resolved to `android:id/aerr_wait`, `class=android.widget.Button`,
  `clickable=true`, bounds `[70,1264,1010,1390]` — the clickable control in the
  topmost dialog window, and the ANR was dismissed.
- `tapOn {text:"Close app"}` resolved to `android:id/aerr_close` the same way.
- On API 31 the same path recovered a wedged `Process system isn't responding`
  dialog, after which `observe` went from 3 elements (`appId:"android"`, dialog
  masking everything) to 22 with `appId:"com.android.settings"`.
- Text-collision check on the playground Semantic Links screen: with
  "Privacy Policy" present both as a standalone link and inside
  "Last activated: Privacy Policy", `tapOn {text:"Privacy Policy"}` selected the
  exact match at `[42,399,135,420]` (`totalMatches` picked the link, not the
  partial).

Caveat on strength of evidence: these confirm correct behavior and no regression,
but they are not strict reproduce-then-confirm — I could not construct a live
case where a *covered background clickable* competed with a topmost
*non-clickable* dialog match, which is the exact ordering this PR changed. That
discrimination remains covered by the unit tests in
`DefaultElementSelector.test.ts`.

Issue #5872, opened 2026-08-29
Action tools return the full raw hierarchy with no opt-out (392 KB to fill 3 contact fields) while observe already has a compact skeleton form


kaeawc:
Filed by the MCP dogfood loop. Two real end-user tasks on `emulator-5554`
(Android 14), local build `0.0.66+ga99f379ce400`:

| task | MCP calls | total tool output |
|---|---|---|
| set a 7:00 AM alarm | 8 | ~230 KB |
| add contact "Ada Lovelace" + phone | 6 | ~392 KB |

392 KB of context to type three short strings into three fields. Both tasks
succeeded — this is purely about what it cost to do them.

## The same screen comes back in two formats, 3-5x apart

`observe` returns the compact `skeleton` (one affordance-tagged line per element).
`tapOn` and `inputText` return the full raw `viewHierarchy` for the *same screen*:

```
observe                    -> 19,091 / 13,739 / 18,128 / 18,207 bytes   (skeleton)
tapOn                      -> 34,466 / 50,672 / 55,451 / 68,706 / 96,510 / 98,636 bytes  (full viewHierarchy)
inputText                  -> 45,916 / 49,871 / 51,008 bytes            (full viewHierarchy)
```

`launchApp` is the same story and additionally inconsistent — the identical call
returned 28,204 bytes on one invocation and 336 bytes on another, with no
parameter difference and nothing in the response explaining which shape I got.

Neither `tapOn`, `inputText`, nor `launchApp` has any parameter to control this.
`observe` has `raw` / `scope` / `project`; the action tools have none of them. So
the cheap representation exists, is already implemented, and is simply not
reachable from the tools that produce the most bytes.

## The key name changes with the producer

`observe` puts the compact form at top-level `skeleton` and has **no
`viewHierarchy` key at all** — even though the tool's description is literally
"Get screen view hierarchy". `launchApp` puts a full one at
`observation.viewHierarchy`. `tapOn` puts a full one at `viewHierarchy`. Three
producers, three shapes, for one concept.

## `inputText` has no selector, so every field costs two calls

```
tapOn     {"selector":{"text":"First name"}}   -> 49,708 bytes
inputText {"text":"Ada"}                       -> 51,008 bytes
tapOn     {"selector":{"text":"Last name"}}    -> 98,636 bytes
inputText {"text":"Lovelace"}                  -> 49,871 bytes
tapOn     {"selector":{"text":"Phone"}}        -> 96,510 bytes
[...]

PR #5895, opened 2026-08-29
fix(inputText): dismiss the Android soft keyboard via the confirmed Keyboard.close() route (#5887)

### LINKED ISSUE #5887: inputText dismissKeyboard:true does not dismiss the Android soft keyboard (follow-up to #5872)
Follow-up from [#5872](https://github.com/kaeawc/auto-mobile/issues/5872) / [PR #5885](https://github.com/kaeawc/auto-mobile/pull/5885).

While filing #5872, the reporter noted a minor, separate bug (verbatim):

> (Minor, same call: `dismissKeyboard: true` didn't dismiss it — the next `observe` reported `activeWindow.activityName: android.inputmethodservice.SoftInputWindow`.)

`inputText` with `dismissKeyboard: true` on Android did not actually dismiss the soft keyboard — a subsequent `observe` still showed the IME (`SoftInputWindow`) as the active window.

This is orthogonal to #5872's three response-shape changes (skeleton default, key-name consistency, `inputText` selector), so it was intentionally left out of PR #5885 to keep that change focused.

## Scope

- Reproduce on Android (emulator-5554, Android 14): `inputText` with `dismissKeyboard: true`, then `observe` and check `activeWindow.activityName`.
- Investigate the Android dismiss path in `src/features/action/InputText.ts` (the `dismissKeyboard` handling and the `Keyboard` close route) and why the IME stays foregrounded.
- Add a regression test pinning that the keyboard is dismissed (or that the dismiss request is issued and confirmed) when `dismissKeyboard: true`.

Observed on local build `0.0.66+ga99f379ce400`.

kaeawc:
## Summary

`inputText` with `dismissKeyboard: true` on Android did not actually dismiss the soft keyboard — a subsequent `observe` still reported the IME (`SoftInputWindow`) as the active window.

The default `a11y` mode (and `eventLast`/`eventAll`) implemented `dismissKeyboard: true` by forwarding the flag to the runner's `requestSetText`, whose only dismiss mechanism is `softKeyboardController.setShowMode(SHOW_MODE_HIDDEN)` (`CtrlProxy.kt`). That flag suppresses the accessibility service from *re-showing* the keyboard but does **not** dismiss an already-visible IME window, leaving it foregrounded. Meanwhile `eventOnly`/`append` already dismissed reliably via `Keyboard.close()` (KEYCODE_BACK + state-confirmation poll).

This change routes the a11y-family dismiss through that same confirmed `Keyboard.close()` route, via a shared `dismissKeyboardViaCloser()` helper (generalized from the former `dismissKeyboardAfterAppend`). The runner's `dismissKeyboard` flag is still forwarded to `requestSetText`, so when `SHOW_MODE_HIDDEN` does hide the keyboard the closer detects it and short-circuits to "already closed" without sending a stray Back. All five Android modes now dismiss through one confirmed path.

## Linked Issue

Closes #5887

## Bug / Regression Context

- **Prior attempts:** Follow-up to #5872 / #5885, which intentionally scoped this dismiss bug out.
- **Observed symptom:** After `inputText(..., dismissKeyboard: true)`, `observe` reported `activeWindow.activityName: android.inputmethodservice.SoftInputWindow` — the keyboard was never dismissed.
- **Repro steps / conditions:** Android (emulator-5554, Android 14): `inputText` with `dismissKeyboard: true`, then `observe`. Root cause: `SHOW_MODE_HIDDEN` alone does not dismiss a visible IME window.

## Validation

- [x] `bun test` — full suite green except two pre-existing, environment-only failures unrelated to this change (missing `node_modules/.bin/oxlint` in a `.claude` worktree; a WebRTC integration test)
- [x] `oxfmt --check` clean; `oxlint` complexity ratchet count unchanged for the file
- [x] New/updated unit tests use interfaces + fakes; run in <100ms
- [x] Regression tests pin that `dismissKeyboard: true` issues and confirms the dismissal for a11y / eventLast / eventAll (and surfaces a dismissal failure)

## Device Verification

- [ ] Not run on hardware in this session — no device attached. Fix is pinned by unit tests; the dismiss route reuses the already-device-verified `Keyboard.close()` path used by `eventOnly`/`append`.

## Follow-ups

- [ ] None.

### DIFF
--- a/src/features/action/InputText.ts
+++ b/src/features/action/InputText.ts
@@ -284,9 +284,14 @@ export class InputText extends BaseVisualChange {
       if (dismissKeyboard) {
         const dismissError = await this.dismissKeyboardViaCloser("a11y", signal);
         if (dismissError) {
+          // Text (and any imeAction) already landed — only the cleanup dismiss
+          // failed. Carry imeAction so a consumer can tell a post-submit cleanup
+          // failure from an operation that never ran and must not blindly retry
+          // (issue #5887 review).
           return {
             success: false,
             text,
+            imeAction,
             error: dismissError,
             method: "a11y",
           };
@@ -395,9 +400,12 @@ export class InputText extends BaseVisualChange {
     if (dismissKeyboard) {
       const dismissError = await this.dismissKeyboardViaCloser("eventLast", signal);
       if (dismissError) {
+        // imeAction already ran — carry it so a dismiss-only failure is not
+        // mistaken for a no-op and retried (issue #5887 review).
         return {
           success: false,
           text,
+          imeAction,
           error: dismissError,
           method: "eventLast",
         };
@@ -504,9 +512,12 @@ export class InputText extends BaseVisualChange {
     if (dismissKeyboard) {
       const dismissError = await this.dismissKeyboardViaCloser("eventAll", signal);
       if (dismissError) {
+        // imeAction already ran — carry it so a dismiss-only failure is not
+        // mistaken for a no-op and retried (issue #5887 review).
         return {
           success: false,
           text,
+          imeAction,
           error: dismissError,
           method: "eventAll",
         };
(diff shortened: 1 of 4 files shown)


kaeawc:
Device-verified on the nightly manual-test run — HEAD `fc89f5422`, emulator-5554 (API 34).

`inputText { ..., dismissKeyboard: true }` into the Settings search field, then device-side ground truth:

- `uiautomator dump` after the call contains **no** `com.google.android.inputmethod.latin:id/key_pos_*` nodes — the IME window is gone from the hierarchy.
- `dumpsys input_method` reports `mInputShown=false`.

(`mIsInputViewShown` still reads `true` in IMMS bookkeeping, but the window is genuinely gone from the hierarchy, so that is stale IMMS state rather than a failed dismiss.) Same result on API 36. **PASS.**

PR #5941, opened 2026-08-30
fix(inputText): surface selector target and failures

### LINKED ISSUE #5902: inputText returns a success envelope (message: "Input text", no isError) when its selector doesn't match — the #5868 postcondition fix didn't cover it
Found while re-verifying [#5872](https://github.com/kaeawc/auto-mobile/issues/5872)
against `main` @ `fc06cb7ef`. The new `inputText` `selector` from
[#5885](https://github.com/kaeawc/auto-mobile/pull/5885) is a real improvement —
three fields now cost three calls instead of six, and two of them came back at
**222 bytes** where the pre-fix pair was ~95 KB. No complaints there.

But `inputText` did not get the postcondition treatment that
[#5877](https://github.com/kaeawc/auto-mobile/pull/5877) gave `launchApp` and
`tapOn`. When its selector doesn't match, it returns a **success envelope**:

```json
{"content":[{"type":"text","text":"{\"message\":\"Input text\",\"success\":false,\"text\":\"Grace\",\"error\":\"Failed to perform tap on element: Element not found with provided text 'First name'\",\"method\":\"a11y\"}"}]}
```

`isError` is not set, and `message` reads `"Input text"` — the same string a
successful call returns. The only signal that nothing was typed is `success:false`
/ `error` nested inside the text payload.

This is precisely the shape #5868 was filed about, and the fix for it is already
established in the two sibling tools:

- `tapOn` now returns `"Tapped on element (matched id=android:id/title text=\"Network & internet\"; 1 match)"`
- `launchApp` now returns `"Launched app com.android.settings (foreground verified)"`, and a real MCP error when the postcondition fails

`inputText` should do the same: set `isError` when the selector doesn't resolve,
and on success say what it typed into (`matchedId` / `matchedText`), so a client
can tell "typed into the right field" from "typed into nothing" without parsing a
nested `success` flag out of a success response.

Reproduced by calling `inputText {"platform":"android","text":"Grace","selector":{"text":"First name"}}`
on a screen with no `First name` field.

kaeawc:
## Summary

`inputText` selector failures were returned as a success envelope with only a nested `success:false` signal. This change preserves the resolved field identity, gives successful selector calls a useful message, and marks failed input operations with `isError: true`.

## Linked Issue

Closes #5902

## Bug / Regression Context

- **Prior attempts:** The selector optimization from #5885 did not receive the postcondition response treatment added to `tapOn` and `launchApp`.
- **Observed symptom:** A selector miss returned `message: "Input text"` without `isError`, requiring clients to parse nested JSON.
- **Repro steps / conditions:** Call `inputText` with an Android selector that does not match any field.

## Validation

- [x] Focused inputText and response-message tests pass
- [x] `bun run lint` passes
- [x] `bun run build` passes
- [x] `bun run typecheck` reports no new errors
- [x] `scripts/all_fast_validate_checks.sh` passes

## Follow-ups

- [x] No follow-up issues needed.

### DIFF
--- a/src/features/action/InputText.ts
+++ b/src/features/action/InputText.ts
@@ -48,7 +48,12 @@ export interface TextInputTargetFocuser {
   focus(
     selector: TextInputTargetSelector,
     signal?: AbortSignal,
-  ): Promise<{ success: boolean; error?: string }>;
+  ): Promise<{
+    success: boolean;
+    error?: string;
+    matchedId?: string;
+    matchedText?: string;
+  }>;
 }
 
 type TextInputTargetFocuserFactory = (device: BootedDevice) => TextInputTargetFocuser;
@@ -60,7 +65,12 @@ const defaultTargetFocuserFactory: TextInputTargetFocuserFactory = (device) => (
       undefined,
       signal,
     );
-    return { success: result.success, error: result.error };
+    return {
+      success: result.success,
+      error: result.error,
+      matchedId: result.selectedElement?.resourceId,
+      matchedText: result.selectedElement?.text,
+    };
   },
 });
 
@@ -145,6 +155,7 @@ export class InputText extends BaseVisualChange {
     // Focus before the observedInteraction captures its baseline, so the "before"
     // snapshot already reflects the focused field. A focus failure short-circuits —
     // typing into an unknown field would be worse than reporting the miss.
+    let targetMetadata: Pick<SendTextResult, "matchedId" | "matchedText"> = {};
     if (selector) {
       assertInputNotAborted(signal);
       const focusResult = await this.targetFocuser.focus(selector, signal);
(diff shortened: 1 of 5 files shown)


kaeawc:
Device-verified on the nightly manual-test run — HEAD `fc89f5422`, emulator-5554 (API 34) and emulator-5560 (API 36).

`inputText` with a selector now reports what it actually matched:

```json
"matchedId": "com.google.android.settings.intelligence:id/open_search_view_edit_text",
"matchedText": "Search settings"
```

Device-side ground truth via `uiautomator dump` confirmed the field held the typed value afterwards:

```
<node text="bluetooth" resource-id="com.google.android.settings.intelligence:id/open_search_view_edit_text" ... />
```

Also verified the failure direction on iOS 26.2, where the diagnostic is genuinely actionable rather than a bare failure:

```
No element has keyboard focus ... | app.label=" " | app.identifier="" | textFields.count=0
  | searchFields.count=0 | app.keyboards.count=0 | springboard.keyboards.count=0
```

That diagnostic is what let me root-cause #5991 in one call. **PASS.**

PR #5943, opened 2026-08-30
feat(tapOn): report post-tap screen effect

### LINKED ISSUE #5910: tapOn returns success:true for a tap that had no effect — client cannot distinguish a no-op tap from an effective one
Filed by the MCP dogfood loop while completing a real end-user task ("set an alarm") on `emulator-5554` (API 34) against `main` @ d8949029c.

## What happened

Tapping the Clock app's "Add alarm" FAB with `tapOn` returned success **twice** and the time picker never opened:

```
tapOn { platform: "android", deviceId: "emulator-5554", selector: { text: "Add alarm" } }
-> { "success": true,
     "message": "Tapped on element (matched id=com.google.android.deskclock:id/fab text=\"Add alarm\"; 1 match)",
     "selectedElement": { "bounds": { ..., "centerX": 540, "centerY": 1959 }, "totalMatches": 1 } }
```

Ground truth after each call: screenshot shows the unchanged alarm list, and
`dumpsys activity activities` shows `mFocusedWindow=Window{... DeskClock}` with no dialog.

The element match was correct. `adb shell input tap 540 1959` — the exact `centerX/centerY`
`tapOn` reported — opens the "Select time" dialog immediately. So the coordinates were right
and the dispatch was the part that did nothing.

`tapOn` on other elements in the same app works (`{text:"AM"}`, `{text:"6"}`, `{text:"45"}`,
`{text:"OK"}` all took effect), so this is not a broken device or runner — it is a specific
tap that silently no-ops.

## Why this matters as a client

`success: true` here means "I matched an element and dispatched to it", not "the tap did
anything". As the MCP client I had no field to distinguish a no-op tap from an effective one,
so I re-issued the same call, got the same confident success, and only found out by taking a
screenshot outside the tool surface. Every tool call after that would have been built on a
wrong belief about the screen.

The one affordance that would have told me — `observationDiff` — degraded to
`{"mode":"full","reason":"missing_session"}`, so it reported nothing about whether the screen
changed. (Filed separately.)

## What I expected

Either the tap takes effect, or `tapOn` tells me it did not.

## Smallest change

`tapOn` already re-observes after the tap (it embeds an `observation`). Compare the pre-tap
and post-tap screen identity it already computes and surface one field, e.g.:

```
"effect": { "screenChanged": false, "basis": "activeWindow+layoutSeqSum unchanged" }
```

`success` can stay as-is (dispatch succeeded); `effect.screenChanged: false` is the signal a
client needs to decide whether to retry, fall back, or stop. That is strictly additive and
costs nothing — both observations are already in hand.

Separately worth investigating why the CtrlProxy dispatch no-ops on
`com.google.android.deskclock:id/fab` when `input tap` at the same point works.

kaeawc:
## Summary
- Add an additive `effect` signal to successful `tapOn` results.
- Compare screen identity, active-window metadata, or hierarchy hashes to report whether the post-tap screen changed.
- Keep `success` and existing retry behavior unchanged so clients can distinguish dispatch success from visible effect.

## Root cause
`tapOn` already captured a post-action observation, but the comparison result was only used internally for retry behavior and was not exposed to MCP clients.

## Validation
- `bun test test/features/action/TapOnElement.retryIfNoChange.test.ts test/server/interactionTools.tapOnMessage.test.ts test/server/tools/tapOnTapAnySchema.test.ts`
- `bun run typecheck`
- `bun run build`
- `bun run lint`
- `bunx oxfmt --check ...`

Closes #5910

### DIFF
--- a/schemas/tool-definitions.json
+++ b/schemas/tool-definitions.json
@@ -10275,6 +10275,28 @@
           "required": ["mode", "reason"],
           "additionalProperties": {}
         },
+        "effect": {
+          "type": "object",
+          "properties": {
+            "screenChanged": {
+              "type": "boolean"
+            },
+            "basis": {
+              "type": "string",
+              "enum": [
+                "screenIdentity changed",
+                "screenIdentity unchanged",
+                "activeWindow+layoutSeqSum changed",
+                "activeWindow+layoutSeqSum unchanged",
+                "viewHierarchy changed",
+                "viewHierarchy unchanged",
+                "insufficient observation data"
+              ]
+            }
+          },
+          "required": ["screenChanged", "basis"],
+          "additionalProperties": {}
+        },
         "selectedElement": {
           "type": "object",
           "properties": {
--- a/src/features/action/TapOnElement.ts
+++ b/src/features/action/TapOnElement.ts
@@ -480,6 +480,84 @@ export class TapOnElement extends BaseVisualChange {
     }
   }
 
+  private compareScreenIdentity(
+    previousObservation: ObserveResult,
+    currentObservation: ObserveResult,
+  ): TapOnElementResult["effect"] | undefined {
+    const previous = previousObservation.screenIdentity;
+    const current = currentObservation.screenIdentity;
+    if (!previous || !current || previous.platform !== current.platform) {
+      return undefined;
+    }
+    const changed = previous.key !== current.key;
+    return {
+      screenChanged: changed,
+      basis: changed ? "screenIdentity changed" : "screenIdentity unchanged",
+    };
+  }
+
+  private compareActiveWindow(
+    previousObservation: ObserveResult,
+    currentObservation: ObserveResult,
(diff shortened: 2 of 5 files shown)


kaeawc:
Device-verified on the nightly manual-test run — HEAD `fc89f5422`, freshly built dist, emulator-5554 (API 34).

`tapOn { selector: { elementId: "com.android.settings:id/search_action_bar" } }` returned:

```json
"effect": { "screenChanged": true, "basis": "activeWindow changed" }
```

and on API 35 a same-activity tap returned `"basis": "viewHierarchy changed"`, so both bases are reachable in practice. The field is populated and its value matched adb ground truth (`ResumedActivity` moved to `SubSettings`) in every case. **PASS.**

One caveat that belongs to the observation path rather than to this PR: on API 34/35 the observation the tap *returns alongside* `effect` is sometimes captured pre-navigation, so `screenChanged: true` can arrive with a stale or empty diff. Filed separately as #5992.

PR #5949, opened 2026-08-30
fix(observe): make stale window warning actionable

### LINKED ISSUE #5915: observe still serves a persistent stale wrong-window capture and offers the client no recovery (residual of #5867)
Follow-up to #5867 (closed). The half that issue's title named is fixed: `freshness.verified`
is now `false` and carries an explicit warning instead of claiming `true`. The half its body
named — *"and a client cannot recover from it"* — still reproduces on `main` @ d8949029c.

## What happened

Hit twice in one dogfood run on `emulator-5554` (API 34). `observe` served a
`com.google.android.calendar` hierarchy while Settings was foreground:

```json
"activeWindow": { "appId": "com.google.android.calendar",
                  "activityName": "com.google.android.calendar.allinone.AllInOneCalendarActivity",
                  "layoutSeqSum": 7092 },
"freshness": { "ageMs": 1657, "verified": false, "isFresh": false,
  "warning": "Observed hierarchy is from com.google.android.calendar, but the device's current
              top resumed activity is com.android.settings. This is a stale wrong-window
              capture; it was not verified against the foreground app." },
"skeleton": [ 6 entries — all com.android.systemui status-bar icons, nothing else ]
```

Ground truth: `topResumedActivity=ActivityRecord{... com.android.settings/.SubSettings t101}`.
Note Calendar was not even in the back stack — it is a sticky window the runner keeps serving.

The state is persistent, not a one-frame race: repeating `observe` returned the identical
stale Calendar capture. What cleared it was `pressButton home` — after which the next
`observe` came back `verified: true` with a real 21-entry skeleton.

## Why this matters as a client

The warning correctly tells me the capture is wrong, which is a real improvement over #5867.
But it is the end of the sentence. It does not tell me the tree is unusable, it does not tell
me what to do, and `observe` has no parameter I could find to force a re-capture. My only
options were to guess (I tried `observe` again — same stale tree), or to reach outside the
tool surface. A client that trusts `skeleton` without reading `freshness.warning` sees six
status-bar icons and concludes the screen is empty.

## What I expected

A stale wrong-window capture is either retried for me, or the response tells me the one thing
that clears it.

## Smallest change (either one)

1. **Retry internally.** `observe` already detects the mismatch to write the warning. On
   detection, re-capture once (the runner recovers within a second in practice) and only
   surface the warning if the retry is also stale.
2. **Make the warning actionable.** Append the recovery that actually works, e.g.
   `"... not verified against the foreground app. The runner is serving a stale window;
   call pressButton { button: \"home\" } (or relaunch the target app) and observe again."`

(1) is what I want as a client; (2) is the cheap version. Either beats a warning that
describes the problem and stops.

Refs #5867.

kaeawc:
## Summary
- Make `observe`'s stale wrong-window freshness warning actionable.
- Tell clients to call `pressButton { button: "home" }` or relaunch the target app, then observe again.
- Add regression assertions covering both recovery paths.

## Root cause
The window-identity check correctly detected and reported a persistent stale capture, but its warning stopped after describing the mismatch. Clients had no recovery guidance in the tool response.

## Validation
- `bun test test/features/observe/ObserveScreenWindowIdentity.test.ts`
- `bun run format:check -- src/features/observe/observationFreshness.ts test/features/observe/ObserveScreenWindowIdentity.test.ts`
- `bunx oxlint src/features/observe/observationFreshness.ts test/features/observe/ObserveScreenWindowIdentity.test.ts`
- `bun run build`
- `bun run typecheck`
- Whole-repository `bun run turbo:validate` was stopped after exceeding the requested 60-second limit.

Closes #5915

### DIFF
--- a/src/features/observe/observationFreshness.ts
+++ b/src/features/observe/observationFreshness.ts
@@ -189,7 +189,7 @@ export function computeFreshness(inputs: FreshnessInputs): FreshnessVerdict {
       ageMs,
       verified: false,
       isFresh: false,
-      warning: `Observed hierarchy is from ${windowIdentityMismatch.observed}, but the device's current top resumed activity is ${windowIdentityMismatch.foreground}. This is a stale wrong-window capture; it was not verified against the foreground app. The runner is serving a stale window; call pressButton { button: "home" } (or relaunch the target app) and observe again.`,
+      warning: `Observed hierarchy is from ${windowIdentityMismatch.observed}, but the device's current top resumed activity is ${windowIdentityMismatch.foreground}. This is a stale wrong-window capture; it was not verified against the foreground app. The runner is serving a stale window; call pressButton { platform: "android", button: "home" } (or relaunch the target app) and observe again.`,
     };
   }
 
--- a/src/features/observe/observationFreshness.ts
+++ b/src/features/observe/observationFreshness.ts
@@ -189,7 +189,7 @@ export function computeFreshness(inputs: FreshnessInputs): FreshnessVerdict {
       ageMs,
       verified: false,
       isFresh: false,
-      warning: `Observed hierarchy is from ${windowIdentityMismatch.observed}, but the device's current top resumed activity is ${windowIdentityMismatch.foreground}. This is a stale wrong-window capture; it was not verified against the foreground app.`,
+      warning: `Observed hierarchy is from ${windowIdentityMismatch.observed}, but the device's current top resumed activity is ${windowIdentityMismatch.foreground}. This is a stale wrong-window capture; it was not verified against the foreground app. The runner is serving a stale window; call pressButton { button: "home" } (or relaunch the target app) and observe again.`,
     };
   }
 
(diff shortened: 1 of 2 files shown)


kaeawc:
Device-verified on the nightly manual-test run — HEAD `fc89f5422`, iOS 26.2 simulator (iPhone 17 Pro).

Hit the cache-served path for real and the warning was exactly what a client needs:

```json
"freshness": { "actualTimestamp": 1788157295987, "ageMs": 4158, "verified": false, "isFresh": false,
  "warning": "Hierarchy was served from the host-side cache without being re-verified against the device (captured 4158ms ago). The runner did not answer a synchronous hierarchy request." }
```

It names the source (host-side cache), the age, and the reason (the runner did not answer a synchronous request) — enough to decide to retry, which is what I did, and the retry returned a verified capture. **PASS.**

Worth noting where it does *not* fire: on the very next call the same broken screen came back with `verified: true` and only three status-bar nodes, i.e. the honesty gate covered the cache path but not the empty-application path. Filed as #5991.

PR #5978, opened 2026-08-30
feat(apps): add bounded cross-platform crashApp tool

### LINKED ISSUE #5977: feat: intentionally crash a target app for resilience testing
## Summary

Add an explicit AutoMobile capability to intentionally crash a specified target application. This should create a genuine application crash suitable for exercising crash detection, diagnostic capture, recovery, and resilience-test flows.

## Why

Force-stopping an app is not equivalent to a crash: it bypasses the OS crash path and may not produce the logs, reports, or lifecycle signals that consumers need to validate. Test authors need a deterministic way to cause a real crash without arbitrary shell access.

## Proposed API

Expose an MCP action such as:

```text
crashApp({ appId })
```

The action must apply only to the explicitly named application. It should not become a general process-killing or command-execution interface.

## Expected behavior

- Confirm the requested app is running or return a clear, typed result.
- Trigger an OS-observable crash rather than a normal termination.
- Return structured metadata: platform, app identifier, process identifier when available, mechanism used, timestamp, and whether the crash was confirmed.
- Preserve or make available the immediate diagnostic evidence needed by existing crash-monitoring capabilities, including stack traces, native tombstones, thread state, or freeze/ANR evidence where the platform provides it.
- Fail explicitly when the operation is unsupported on a platform or device class; do not silently fall back to a force-stop.
- Keep the capability bounded to the target app and suitable for automated test environments.

## Platform notes

- **Android:** resolve the target process for the package and induce a real abort/signal-based crash; verify it through the platform crash path.
- **iOS simulator:** support the equivalent real-crash behavior where the simulator tooling permits it.
- **iOS physical devices:** document the supported mechanism or return an explicit unsupported result until a safe, reliable mechanism is available.

## Acceptance criteria

- A caller can crash a specified running app through one supported AutoMobile action.
- Android tests demonstrate that the operation is distinguishable from normal app termination and leaves crash-detectable evidence.
- iOS support and limitations are covered by tests and documentation.
- The action has no arbitrary process, package discovery, or shell-command escape hatch.
- The response contract is stable and documented for automated test clients.

## Related work

This complements #1020, which focuses on detecting and collecting crash and freeze evidence.

kaeawc:
## Summary
- add a bounded `crashApp` MCP tool that uses Android ActivityManager's real crash path and iOS Simulator `SIGABRT`
- confirm crashes with process-specific OS evidence while never falling back to force-stop or `SIGKILL`
- keep physical iOS explicitly unsupported and wire schemas, timeout handling, plan tooling, docs, and generated definitions

## Test plan
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun run build`
- [x] `bun test --isolate` (15,069 passed, 14 skipped)
- [x] `scripts/all_fast_validate_checks.sh` (35 passed)
- [x] Android `:test-plan-validation:test` and `:ide-plugin:test`
- [x] Manual Android emulator crash and iOS Simulator crash verification

Closes https://github.com/kaeawc/auto-mobile/issues/5977

Made with [Cursor](https://cursor.com)

### DIFF
--- a/android/ide-plugin/src/main/kotlin/dev/jasonpearson/automobile/ide/yaml/TestPlanToolCategories.kt
+++ b/android/ide-plugin/src/main/kotlin/dev/jasonpearson/automobile/ide/yaml/TestPlanToolCategories.kt
@@ -3,7 +3,7 @@ package dev.jasonpearson.automobile.ide.yaml
 object TestPlanToolCategories {
   private val categoryByTool: Map<String, String> =
     mapOf(
-        "App management" to setOf("launchApp", "terminateApp", "installApp"),
+        "App management" to setOf("launchApp", "terminateApp", "crashApp", "installApp"),
         "UI interactions" to setOf("tapOn", "swipeOn", "pinchOn", "dragAndDrop"),
         "Input" to
           setOf(
--- a/android/test-plan-validation/src/main/kotlin/dev/jasonpearson/automobile/validation/ValidTools.kt
+++ b/android/test-plan-validation/src/main/kotlin/dev/jasonpearson/automobile/validation/ValidTools.kt
@@ -17,6 +17,7 @@ object ValidTools {
       "clearMockNetwork",
       "clearText",
       "clipboard",
+      "crashApp",
       "criticalSection",
       "debugSearch",
       "deleteSnapshot",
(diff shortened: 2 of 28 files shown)


kaeawc:
Device-verified on the nightly manual-test run — HEAD `fc89f5422`, emulator-5554 (API 34).

`crashApp { appId: "com.android.settings", platform: "android" }`:

```json
{ "mechanism": "android_am_crash", "confirmed": true, "wasRunning": true, "processId": 22804,
  "evidence": { "source": "android_logcat",
    "summary": "... E AndroidRuntime: android.app.RemoteServiceException$CrashedByAdbException: shell-induced crash" } }
```

The crash is real (logcat carries the `CrashedByAdbException`) and the response proves it rather than asserting it — `confirmed`, the PID, and the raw logcat line. **PASS.**

PR #6020, opened 2026-09-01
feat(desktop): make workspace overlays modal to keyboard/a11y focus (#4846)

### LINKED ISSUE #4846: Desktop workspace: make overlay panels modal to keyboard/a11y focus
Both full-window overlays — the command palette (`CommandPalette.kt`, merged) and the health sheet (`WorkspaceShell.HealthSheetOverlay`, #4844) — use the same sibling-`Box` scrim pattern. Neither traps focus nor removes the underlying workspace from the focus/semantics trees, so a keyboard/screen-reader user can Tab or traverse *behind* the scrim to controls that are visually dimmed.

Raised in codex review of PR #4844 (thread on `WorkspaceShell.kt:194`). Deferred because it's a shared property of both overlays and better fixed once, consistently.

Fix: trap focus within the open overlay and clear/hide the background from a11y traversal while it's shown (e.g. `Modifier.focusProperties`/a focus trap + `clearAndSetSemantics {}` on the dimmed background, or hosting the panel in a focusable `Popup`/`Dialog`). Apply to both overlays.

kaeawc:
## What

Closes #4846.

The desktop workspace's full-window overlays each draw a dimmed scrim as a **sibling** of the workspace content (`WorkspaceShell.kt`'s `HealthSheetOverlay` / `CompareOverlay` / `OfflineBrowseOverlay`, and the ⌘K `CommandPalette` hosted alongside `WorkspaceShell` in `AutoMobileDesktopApp`). The scrim never trapped focus nor removed the underlying workspace from the focus/semantics trees, so a keyboard or screen-reader user could **Tab or traverse behind the scrim** to the visually-dimmed controls — exactly the defect raised in the codex review of #4844 (`WorkspaceShell.kt:194`).

## How

One shared mechanism instead of per-overlay one-offs, per the issue's "better fixed once, consistently":

```kotlin
fun Modifier.isolatedBehindOverlay(active: Boolean): Modifier =
  if (active) focusProperties { canFocus = false }.focusGroup().clearAndSetSemantics {} else this
```

- **Focus trap** — `focusProperties { canFocus = false }` on a `focusGroup()` deactivates the background focus subtree, so a Tab has nowhere to land behind the scrim.
- **A11y isolation** — `clearAndSetSemantics {}` drops the whole background subtree from the accessibility tree, so assistive tech can't traverse behind the overlay.

`WorkspaceShell` applies it to its content column whenever any of its three scrimmed panels is open (`overlayActive` mirrors the three render conditions one-for-one). The app root applies the same isolation to the whole shell while it hosts the sibling command-palette overlay, so every active-workspace overlay is modal through the one modifier.

Scope note: this covers the overlays of the **current** desktop root (`WorkspaceShell`). The deprecated `ThreePaneShell` path is no longer the app root and is left untouched.

## Tests

`WorkspaceShellUiTest` gains four cases asserting that opening each overlay removes the background controls (`Open Devices`, `Open command palette`, a pane's `Close <device>`) from the accessibility tree, and that dismissing the health sheet restores them. Verified as genuine regression guards: all four fail when the modifier is neutralized to a no-op, and pass with the fix.

- `:desktop-core:test` — full module green (46 in `WorkspaceShellUiTest`, 0 failures)
- `:desktop-app:compileKotlin` green
- `:desktop-core:detekt` + `:desktop-app:detekt` green
- ktfmt clean (942 files)

## Risk

Runtime UI behavior change in `android/` (focus + semantics of the workspace behind an open overlay), so labeled `needs-human` with auto-merge off. Purely additive: a no-op when no overlay is open, so the un-overlaid workspace keeps its normal focus order and semantics.

### DIFF
--- a/android/desktop-app/src/main/kotlin/dev/jasonpearson/automobile/desktop/AutoMobileDesktopApp.kt
+++ b/android/desktop-app/src/main/kotlin/dev/jasonpearson/automobile/desktop/AutoMobileDesktopApp.kt
@@ -61,6 +61,7 @@ import dev.jasonpearson.automobile.desktop.core.workspace.WorkspaceUiState
 import dev.jasonpearson.automobile.desktop.core.workspace.WorkspaceViewModel
 import dev.jasonpearson.automobile.desktop.core.workspace.buildWorkspaceCommands
 import dev.jasonpearson.automobile.desktop.core.workspace.deriveWorkspaceStatus
+import dev.jasonpearson.automobile.desktop.core.workspace.isolatedBehindOverlay
 import dev.jasonpearson.automobile.desktop.core.workspace.parseBootedDeviceSessionUuids
 import dev.jasonpearson.automobile.desktop.core.workspace.parseBootedLockStates
 import dev.jasonpearson.automobile.desktop.core.workspace.parseDeviceLockStates
@@ -471,6 +472,10 @@ fun AutoMobileDesktopApp(
                 onAction = workspaceViewModel::onAction,
                 onOpenPicker = workspaceViewModel::openPicker,
                 onOpenPalette = { paletteOpen = true },
+                // The ⌘K command palette is a sibling overlay hosted here (not inside the shell),
+                // so isolate the whole workspace behind it — matching how the shell isolates its
+                // own scrimmed panels — to keep the palette modal to keyboard/a11y focus (#4846).
+                modifier = Modifier.isolatedBehindOverlay(paletteOpen),
                 status = workspaceStatus.status,
                 statusDetail = workspaceStatus.detail,
                 updateStatus = updateStatus,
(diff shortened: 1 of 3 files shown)


kaeawc:
Follow-up: [#6022](https://github.com/kaeawc/auto-mobile/pull/6022) seals the keyboard-focus gap Codex flagged here. `canFocus = false` on the `focusGroup()` deactivated the group node but not its focusable descendants, so Tab could still reach a control behind the scrim (the semantics-only tests passed because `clearAndSetSemantics {}` did remove the a11y subtree). #6022 adds `onEnter = { cancelFocusChange() }` to trap Tab at the group boundary plus a focus-clear on open for the already-focused case, with two focus-observing regression tests that were red on this impl.

Issue #6078, opened 2026-09-04
observe.activeWindow names the app behind an open notification shade while returning the shade's hierarchy (API 29/31/34/35)


kaeawc:
While the Android notification shade is open, `observe` returns the **shade's**
hierarchy but reports an `activeWindow` describing the app *behind* the shade.
The two halves of a single observation disagree with each other and with
`dumpsys window` focus, so a caller has no field that says "the shade is up".

Reproduced on a clean daemon built from `979e2fe8a` (TS-only rebuild; natives at
the 2026-09-04 nightly re-cut `92dd3186e`, no `android/`/`ios/` changes in range).

## Repro

```bash
adb -s emulator-5554 shell cmd statusbar expand-notifications
# then, over one MCP session:
#   getAndroid {deviceId: emulator-5554} -> observe {platform: "android"}
```

## Observed — API 34 (`am-api34-ga-arm64`, emulator-5554)

adb ground truth:

```
mCurrentFocus=Window{8ddaeb2 u0 NotificationShade}
```

`observe` in the same moment:

```
activeWindow: {"appId":"com.google.android.settings.intelligence",
               "activityName":"com.google.android.settings.intelligence.modules.search.SearchActivity",
               "layoutSeqSum":13205}
skeleton labels: Internet, AndroidWifi | Bluetooth, On | Flashlight, Off |
                 Do Not Disturb, Off | Security & privacy | Collapse |
                 Silent | Expand | Manage | Clear all
```

The hierarchy is unambiguously the shade (quick-settings tiles, `Collapse`,
`Clear all`). `activeWindow` names a Settings-intelligence search activity that
is fully occluded and not focused.

## API-level matrix (shade open, adb confirms `mCurrentFocus=NotificationShade` in every row)

| AVD | API | `activeWindow.appId` | `activeWindow.activityName` | coherent? |
|---|---|---|---|---|
| `am-api29-ga-arm64` | 29 | `com.android.systemui` | `com.android.settings.homepage.SettingsHomepageActivity` | no — the two fields name **different apps** |
| `am-api31-ga-arm64` | 31 | `com.google.android.settings.intelligence` | `...modules.search.SearchActivity` | no — shade invisible |
| `am-api34-ga-arm64` | 34 | `com.google.android.settings.intelligence` | `...modules.search.SearchActivity` | no — shade invisible |
[...]

Issue #6070, opened 2026-09-03
API 35: activeWindow.activityName still intermittently empty/stale while backStack and adb agree (#5992 residual)


kaeawc:
## Summary

[#5992](https://github.com/kaeawc/auto-mobile/issues/5992) is materially improved by [#6043](https://github.com/kaeawc/auto-mobile/pull/6043) — the "empty diff for a full screen change" symptom no longer reproduces — but on **API 35** `activeWindow.activityName` is still **intermittently wrong** (empty or stale) while `backStack.currentActivity` and adb ground truth agree with each other in the same observation. That is the internal contradiction #5992 named.

Verified on `main` @ `16fd70697` (nightly manual-test run 2026-09-03), `am-api35-ga-arm64`, CtrlProxy APK freshly built from this commit and confirmed installed on-device.

## What is fixed (confirmed)

- Taps that change screen now emit a real diff. `tapOn` "Network & internet" produced a populated `added`/`removed`/`changed` diff with `effect.screenChanged: true` — no empty diff.
- A standalone `observe` after that tap reported `activeWindow.activityName = com.android.settings.SubSettings`, matching `backStack.currentActivity.name` and `dumpsys activity activities` (`topResumedActivity=... com.android.settings/.SubSettings`).
- `launchApp` foreground verification works and correctly *refuses* a mis-attributed capture (see [#6048](https://github.com/kaeawc/auto-mobile/pull/6048) note below).

## Residual defect — intermittent, `activeWindow`-only

Two distinct wrong values observed on API 35 within a single run:

**(a) empty `activityName`, frozen `layoutSeqSum`.** `launchApp { appId: "com.android.settings" }` returned `verified: true` with

```json
"activeWindow": { "appId": "com.android.settings", "activityName": "", "layoutSeqSum": 3478 }
"backStack": { "currentActivity": { "name": "com.android.settings.Settings", "taskId": 14 } }
```

adb agreed with `backStack` (`com.android.settings/.Settings`). A subsequent `tapOn "Apps"` navigated for real (large diff, `effect.screenChanged: true`) but **both** sides of `observationDiff` still carried `activityName: ""` and the same `layoutSeqSum: 3478`. `effect.basis` fell back to `"viewHierarchy changed"`, which is why effect detection still worked.

**(b) stale `toScreen` on a genuine activity change.** In an earlier session on the same device, `tapOn "Network & internet"` reported:

```json
[...]

Issue #5992, opened 2026-08-31
API 35: post-tap activeWindow.activityName contradicts backStack.currentActivity and adb ground truth in the same observation, with one tap emitting an empty diff for a full screen change


kaeawc:
On API 35, the `activeWindow.activityName` a post-tap observation reports disagrees with `backStack.currentActivity.name` **in the same payload**, and with the hierarchy content that payload carries. `freshness.verified` is `true` for all of them. Three consecutive taps on the stock Settings app, three distinct wrong answers — one of which emits an empty diff for a full screen change, which is the shape that actually breaks a client.

Not the same as #5981: there the tree collapses to status-bar-only and stays broken. Here the *hierarchy* is usually right and it is the *attribution* that lags, so nothing looks obviously wrong — which is worse for a client that steers on `activeWindow`. #5975 was supposed to reconcile stale active-window attribution; on API 35 it does not.

## Environment

- HEAD `fc89f5422`, TS dist rebuilt from that commit; released CtrlProxy APK (re-cut at that same commit)
- `emulator-5556`, AVD `am-api35-ga-arm64`, API 35
- Single daemon, `--daemon-mode --debug --debug-perf`
- API 34 (`emulator-5554`) shows a related but milder form; the reproductions below are all API 35

## Repro — three consecutive taps in `com.android.settings`

**Tap 1 — `{ text: "Connected devices" }` from the Settings home.** An empty diff for a full screen change:

```json
"effect":            { "screenChanged": true, "basis": "viewHierarchy changed" },
"observation":       { "isDiff": true, "added": [], "removed": [], "changed": [] },
"observationDiff":   { "fromScreen": { "activeWindow": { "activityName": "", "layoutSeqSum": 546 } },
                       "toScreen":   { "activeWindow": { "activityName": "", "layoutSeqSum": 546 } } }
```

`effect` says the hierarchy changed; the diff computed from that same capture says nothing changed; `from` and `to` carry an identical `layoutSeqSum`. A client running in the default diff mode is told the tap did nothing. Ground truth at that moment:

```
ResumedActivity=ActivityRecord{f4ee953 u0 com.android.settings/.SubSettings t7}
mCurrentFocus=Window{f895511 u0 com.android.settings/com.android.settings.SubSettings
```

A follow-up bare `observe` returns the correct Connected-devices tree, so the destination was reachable — only the post-tap capture was pre-navigation.

[...]

Issue #5867, opened 2026-08-29
observe returns a stale wrong-window hierarchy while reporting freshness.verified=true, and a client cannot recover from it


kaeawc:
Filed by the MCP dogfood loop: I drove AutoMobile as an MCP client (Claude) to do
real end-user tasks on `emulator-5554` (AVD `am-api34-ga-arm64`, Android 14),
local build `0.0.66+ga99f379ce400`.

## What happened

While doing "toggle Wi-Fi from Settings", `observe` began returning a view of a
**different app than the one in the foreground**, and stayed wrong indefinitely.

```
try1  observe.activeWindow: com.google.android.calendar | skeleton nodes=6 | freshness.verified=true   ||  adb: com.android.settings/.SubSettings
try2  observe.activeWindow: com.google.android.calendar | skeleton nodes=6 | freshness.verified=true   ||  adb: com.android.settings/.SubSettings
try3  observe.activeWindow: com.google.android.calendar | skeleton nodes=6 | freshness.verified=true   ||  adb: com.android.settings/.SubSettings
```

`adb shell dumpsys activity activities | grep topResumedActivity` is the ground
truth and says Settings every time. The 6 "nodes" `observe` returns are only the
status bar (clock, two notification icons, wifi/signal/battery) — no app content
at all. `layoutSeqSum` was frozen at `7092` across every one of these calls.

## Repro

1. `launchApp` `com.android.settings`
2. `tapOn` text `Network & internet`
3. `tapOn` the Internet row (`text: "AndroidWifi"`)
4. `observe` → `activeWindow.appId = com.google.android.calendar`, 6 nodes

## Why this is the worst possible shape for a client

- **It is silent.** `freshness.verified: true` and `freshness.isFresh: true` on
  every bogus response. The only in-band hint is a suspiciously small node count,
  which is not something a client can threshold on.
- **Every downstream call is then wrong.** My next `tapOn text:"Internet"` matched
  something in the stale Calendar view and reported `"Tapped on element"`. I only
  found out by observing afterwards and seeing a different app.
- **A client cannot recover through the MCP surface.** I tried
  `terminateApp com.google.android.calendar` — it returned
  `"Terminated app com.google.android.calendar"` and the very next `observe`
  *still* reported Calendar as `activeWindow`, with the same frozen
  `layoutSeqSum: 7092`, for a process that no longer existed. Only
  `homeScreen` cleared it, and re-launching Settings re-entered the stuck state
[...]

PR #6098, opened 2026-09-04
fix(observe): always recapture on the bootstrap active-window path when backStack is present

### LINKED ISSUE #6088: Same-app A→B activeWindow attribution skew on the bootstrap Window path
## Summary

`observe`'s bootstrap active-window path (used during the lazy CtrlProxy-install interval, when the accessibility service supplies no usable `foregroundActivity`) can publish a **same-app** activity from a *later* navigation against an *earlier* hierarchy, with `freshness.verified: true`. Surfaced by Codex review on [#6087](https://github.com/kaeawc/auto-mobile/pull/6087).

## Mechanism (verified)

During the CtrlProxy-install bootstrap interval:

1. The hierarchy is captured for activity **A**.
2. A same-app navigation **A → B** happens.
3. `backStack` is read → reports **B**.
4. `collectActiveWindow` reads the window → reports **B**.

`reconcileActiveWindowAttribution` (`src/features/observe/ObserveScreen.ts`) only runs its recapture / temporal-confirmation when `activeWindow.activityName !== backStack.currentActivity.name`. Here both are **B**, so it **skips the recapture** and takes the "agrees" path, publishing **B**'s activity name onto **A**'s hierarchy and leaving `freshness.verified: true`.

The window-identity mismatch guard (`resolveWindowIdentityMismatch`, [#5867](https://github.com/kaeawc/auto-mobile/issues/5867)) does not catch this because it is **package-level** — A and B are the same app, so the packages match and no mismatch is reported.

Consequence: `waitFor(activeWindow.activityName == B)` can succeed while the returned `elements` still describe screen **A**.

## Pre-existing, not introduced by [#6070](https://github.com/kaeawc/auto-mobile/issues/6070)

This is a **frequency-widening** of a pre-existing limitation, not a new bug:

- `reconcile` already skips recapture when `activeWindow.activityName == backStack.activityName` on `main`.
- On `main`, `collectActiveWindow` calls `Window.getActive(false)`, which on a **cache miss** does a fresh `dumpsys` and returns **B** — hitting the same skew. The stale cache only *accidentally* protected the cache-*hit* case by holding an older activity (A / empty), which then tripped the recapture guard.
- [#6070](https://github.com/kaeawc/auto-mobile/issues/6070) / PR [#6087](https://github.com/kaeawc/auto-mobile/pull/6087) changed that read to `Window.getActive(true)` (a deterministic fix for the frozen stale-cache: empty `activityName` + frozen `layoutSeqSum` served across a navigation). Force-refresh removes the cache-hit accidental protection, so the skew now reproduces more often on the bootstrap path.

The force-refresh's deterministic benefit (killing the reproducible frozen-cache #6070 symptom) was judged to outweigh this narrow race, so it landed in [#6087](https://github.com/kaeawc/auto-mobile/pull/6087) with this skew tracked here.

## Fix direction

Correlate the bootstrap window read with the hierarchy, or **always recapture on the bootstrap path when `backStack` is present** — do not treat a `window == backStack` agreement as proof the hierarchy is aligned when the hierarchy carries no activity signal (the bootstrap case).

Risk to weigh: a forced recapture during the CtrlProxy-install bootstrap interval can fail or return an unusable hierarchy (CtrlProxy may not yet be available), which the current guard treats as a mismatch and **retracts freshness**. Applied unconditionally on the bootstrap path this could **over-retract** `freshness.verified` on otherwise-legitimate observations. The fix needs fixture tests that cover both the genuine A→B skew (must retract / resolve to unknown) and a stable bootstrap observation (must NOT over-retract).

## References

- Residual of [#6070](https://github.com/kaeawc/auto-mobile/issues/6070) (fixed the deterministic frozen-cache half in PR [#6087](https://github.com/kaeawc/auto-mobile/pull/6087)).
- Related package-level guard: [#5867](https://github.com/kaeawc/auto-mobile/issues/5867); activity-level recapture: [#5992](https://github.com/kaeawc/auto-mobile/issues/5992).

kaeawc:
## Summary

On `observe`'s bootstrap active-window path (lazy CtrlProxy install interval, no usable `foregroundActivity` from the accessibility service), `reconcileActiveWindowAttribution` only ran its recapture/temporal confirmation when `activeWindow.activityName` disagreed with `backStack.currentActivity.name`. But on that path both values are post-capture adb reads, so a same-app A→B navigation between the hierarchy capture and those reads made both report B — the "agrees" path then published B's activity name onto A's hierarchy with `freshness.verified: true`. The package-level guard from #5867 cannot distinguish A from B.

The fix follows the issue's second suggested direction: **always recapture on the bootstrap path when a back-stack attribution is present**. The back-stack branch of `reconcileActiveWindowAttribution` is extracted into `reconcileAgainstBackStack` (also keeps the method under the complexity ratchet), with a `hierarchyCarriesActivitySignal` helper identifying the bootstrap case. When the recapture cannot confirm, the bootstrap window's activity name resolves to `""` (unknown) instead of staying published against an uncorrelated tree, and freshness is retracted through the existing `activityAttributionMismatch` path.

## Linked Issue

Closes #6088

## Bug / Regression Context

- **Prior attempts:** #6087 fixed the deterministic frozen-cache half of #6070 via `Window.getActive(true)`; that force-refresh widened the frequency of this pre-existing skew, tracked as #6088.
- **Observed symptom:** `waitFor(activeWindow.activityName == B)` could succeed while `elements` still described screen A, with `freshness.verified: true`.
- **Repro steps / conditions:** bootstrap interval; hierarchy captured for A; same-app navigation A→B; `backStack` and `Window.getActive` both read B.

## Acceptance criteria → tests (all in `test/features/observe/ObserveScreenWindowIdentity.test.ts`)

- **Genuine A→B skew, recapture unusable → must resolve to unknown / retract:** "does not publish an agreeing later same-app window/backStack activity onto an earlier hierarchy when the recapture cannot confirm it (#6088)" — recapture runs (2 hierarchy reads), `activityName` is `""`, `verified: false`, A's tree preserved.
- **Stable bootstrap observation → must NOT over-retract:** "does not over-retract a stable bootstrap observation whose window and backStack agree (#6088)" — recapture confirms, name backfilled, `layoutSeqSum` kept, `verified: true`, no warning.
- **Skew where the recapture does confirm B (inferred):** "resolves a genuine bootstrap A->B skew to B's fresh hierarchy when the recapture confirms it (#6088)" — B's fresh tree is published with B's name, `verified: true`.

Existing #5992 / #6070 / #6078 tests are unchanged and green.

## Validation

- [x] `bun run lint` + `bun test` pass (two pre-existing worktree-only failures: `oxlintConfigScoping.integration` needs `node_modules/.bin/oxlint` absent in a `.claude` worktree; `webrtcDeviceCaptureLoad.integration` asserts on bun skip-line output — both unrelated to this diff)
- [x] `bun run format:check` clean
- [x] New code uses interfaces + fakes + FakeTimer; new tests run in <1ms each
- [x] No new dependencies or helpers beyond a one-line private predicate
- [x] Branched from latest `main`

## Trade-off noted in the issue

A recapture that fails during the bootstrap interval now retracts freshness on the agreeing path too (previously it was skipped). This is deliberate: on that path the window name cannot be correlated with the tree without the recapture, so "unknown + retracted" is the honest verdict, and the caller re-observes. The stable-observation test pins that a successful recapture never over-retracts.

### DIFF
--- a/src/features/observe/ObserveScreen.ts
+++ b/src/features/observe/ObserveScreen.ts
@@ -1205,6 +1205,12 @@ export class RealObserveScreen implements ObserveScreen {
     );
     const confirmed = resolveBackStackActivityAttribution(recapturedState);
     if (!this.matchesExpectedBackStackAttribution(confirmed, expected)) {
+      // A failed confirming back-stack read retracts freshness; surface its
+      // cause on the published result so the retraction is diagnosable rather
+      // than an unexplained "could not reconcile" (#6088).
+      for (const error of recapturedState.errors ?? []) {
+        appendObserveError(result, error);
+      }
       return false;
     }
 
--- a/src/features/observe/ObserveScreen.ts
+++ b/src/features/observe/ObserveScreen.ts
@@ -226,6 +226,18 @@ function isAccessibilityViewClass(foregroundActivity: string): boolean {
   );
 }
 
+/**
+ * Whether the captured hierarchy itself names the foreground activity (the
+ * accessibility-service path). When it does not — the lazy-CtrlProxy bootstrap
+ * interval, where `activeWindow` comes from the legacy `Window.getActive()`
+ * read — the tree carries no activity signal that could be correlated with the
+ * adb reads that follow it (#6088).
+ */
+function hierarchyCarriesActivitySignal(hierarchy: ObserveResult["viewHierarchy"]): boolean {
+  const foregroundActivity = hierarchy?.foregroundActivity;
+  return typeof foregroundActivity === "string" && !isAccessibilityViewClass(foregroundActivity);
+}
+
 function isActivityInPackage(activityName: string, packageName: string): boolean {
   return activityName === packageName || activityName.startsWith(`${packageName}.`);
 }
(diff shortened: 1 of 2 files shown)


kaeawc:
Follow-up filed: #6099 tracks the recapture fresh-wait latency this PR makes more frequent on the bootstrap path (about one second per static-screen bootstrap observe), plus the stale "rare" comment in `DeviceStateCollector.collectActiveWindow`. The second commit here surfaces a failed confirming back-stack read in `result.errors` so the retraction is diagnosable.

Issue #6123, opened 2026-09-04
DefaultUIStateSetup.tapCloseButton returns true after the first candidate regardless of outcome — "Cancel"/"Dismiss"/"×"/"✕" are never tried


kaeawc:
## The candidate loop assumes `callInternal` throws on a missing element; it does not

`src/features/navigation/DefaultUIStateSetup.ts:520-549`:

```ts
for (const text of closeTexts) {
  try {
    await ToolRegistry.callInternal(tapTool, { selector: { text }, action: "tap", ... });
    return true;          // <- reached even when the tap failed
  } catch (error) {
    continue;
  }
}
```

`TapOnElement.execute` catches "Element not found" and returns `{success: false}` (`TapOnElement.ts:1776-1798`); `tapOnHandler` spreads that into the envelope (`interactionTools.ts:1006-1016`); the registry wrapper only rethrows `ActionableError`s that escape. So `callInternal` resolves, `tapCloseButton` returns `true` after the "Close" attempt, and the remaining four candidates are dead code. `dismissWithCloseButton` then fails its `isModalDismissed` check and gives up.

The sibling `tapOnElement` in the same file (`:313-314`) already does this correctly with `throwIfInternalToolFailed(response, "tapOn", platform)`.

### Trigger

`setUIState` / navigation setup on a dialog or bottom sheet whose only dismiss affordance is "Cancel" or "Dismiss" → the modal is not dismissed via the close-button strategy; it falls through to the generic back-press fallback after paying a full observe + search for a "Close" selector that does not exist.

### Evidence

Repro (`bun test`, `tapOn` stubbed to return `success: false`):

```
{ tapped: true, attempted: [ "Close" ] }
```

### Fix

Call `throwIfInternalToolFailed(response, "tapOn", platform)` inside the loop (as `tapOnElement` does) so a failed candidate falls through to the next text. Add a test where only "Cancel" exists and assert `attempted` is `["Close", "Cancel"]` and `tapped` is true.

### Acceptance criteria

- [ ] A failed candidate does not short-circuit the loop.
- [ ] All five candidates are tried when none match; the method returns `false`.
- [ ] Red before / green after.

### Relationships

- Related: [#3106](https://github.com/kaeawc/auto-mobile/issues/3106) (previous dead-code finding in `dismissTopModal` of the same file).
- Related: [#3108](https://github.com/kaeawc/auto-mobile/issues/3108) (the `callInternal` seam this loop was migrated to).

[...]

Issue #6056, opened 2026-09-03
pinchOn reports "Pinched <direction>" on failed gestures (success:false + error), unlike swipeOn after #4183


kaeawc:
## Summary

`pinchOn` reports a success-shaped `message` ("Pinched in" / "Pinched out") even when the gesture never ran and the payload carries `success: false` plus an `error`. A client that reads `message` — the field most naturally surfaced to a user or an LLM — is told the pinch happened when it did not.

This is the same defect class already fixed for `swipeOn` in #4183 (P4). `swipeOn` routes its message through `formatSwipeOnMessage()`, which returns the error on failure; `pinchOn` never got the equivalent.

## Repro

Deterministic, no app state required. Against a booted Android emulator with an acquired session:

```
pinchOn { platform: "android", deviceId: "emulator-5554", sessionUuid: "<uuid>", direction: "in", scale: 0 }
```

Observed result (observation stripped):

```json
{
 "message": "Pinched in",
 "success": false,
 "direction": "in",
 "distanceStart": 0,
 "distanceEnd": 0,
 "duration": 0,
 "scale": 0,
 "centerX": 0,
 "centerY": 0,
 "targetType": "screen",
 "error": "scale must be greater than 0"
}
```

`isError` is also unset on the MCP envelope.

Same shape for the other validation rejections:

- `scale: -5` → `error: "scale must be greater than 0"`, `message: "Pinched in"`
- `distanceStart: -999` → `error: "distanceStart must be greater than 0"`, `message: "Pinched out"`

## Observed vs expected

**Observed:** `message` is unconditionally `Pinched ${direction}` regardless of outcome.

**Expected:** on failure, `message` should carry the error, mirroring `swipeOn` — e.g. `"scale must be greater than 0"`, with a non-empty fallback such as `Pinch ${direction} failed` when `error` is empty (the `||`-not-`??` detail #4183 called out).

## Cause

`src/server/interactionTools.ts:1356` — `pinchOnHandler` builds the response with a hardcoded message:

```ts
return createJSONToolResponse({
  message: `Pinched ${args.direction}`,
  observation: result.observation,
  ...result,
});
```

Compare `swipeOnHandler` 30 lines above (`src/server/interactionTools.ts:1326`):

```ts
message: formatSwipeOnMessage(result, args.direction),
```

and `formatSwipeOnMessage` at `src/server/interactionTools.ts:813`, which already encodes the intended contract.

[...]

Issue #4183, opened 2026-07-22
test(server-runtime): burn down 36 verified test-quality findings


kaeawc:
Part of the unit-test-quality audit. This slice: **server-runtime**.

## How to work this issue

For **every** item below, in rank order:

1. **Write/change the test and run it.** `bun test <file>` — it must pass.
2. **Break the behavior it claims to guard** (edit `src/`, invert the condition, drop the call).
3. **Confirm it goes red.** If it still passes, the test is worthless — fix the test, not the assertion count. This is the whole point: a test that cannot fail is worse than no test.
4. **Restore `src/`** and confirm green again.
5. Ship.

Do not skip step 2. An earlier pass of this audit produced proposals that passed against deliberately broken code, and one that "passed" only by hanging to bun's 5s timeout.

## Hard rules

- Interface + Fake DI; `FakeTimer` for all time. No real timers, sleeps, fs, network, or sockets.
- Never resolve the real file-backed `getDatabase()` — use `createTestDatabase()` or a harness from `test/db/`.
- Each test under 100ms.
- Assert the **observable outcome**, not that a mock was called. Table rows are the specification — include the boundary rows (empty, zero, negative, non-finite, unicode, very long, duplicate, out-of-order).
- Behavior-first test names: `<subject> <verb-phrase> when <condition>`.

## Verified action list

Priority = (blast radius of the untested behavior) x (silence of the failure) / effort.

| Rank | Action | Category | Target (file:line) | Effort | Regression prevented |
|---|---|---|---|---|---|
| 1 | Assert the expand-button centre, not any tap | REWRITE | `test/server/systemTray.test.ts:897` | S | Tapping the notification row instead of the expand button — the exact documented failure that opens the app generically |
| 2 | Actually call `expandNotificationGroup` between the two matches | REWRITE | `test/server/systemTray.test.ts:935` | M | Deleting `expandNotificationGroup` from `src/` leaves the suite green today |
| 3 | `compareVersions` boundary table (finds a live bug) | PARAM+GAP | `test/server/deviceMatcher.test.ts:24` / `src/server/deviceMatcher.ts:26-40` | S | Non-numeric `osVersion` returns `NaN` → device silently dropped by *both* min and max filters |
| 4 | `artifactMode: "oversized"` + 64 KB threshold | ADD | `src/server/finalizeToolResponse.ts:352,363` | M | Auto-spill becomes always-on or never-on with no test failing |
[...]

PR #6162, opened 2026-09-05
fix(mcp): tapOn returns isError on selector miss

### LINKED ISSUE #6152: tapOn still returns a success-shaped envelope ("Tapped on element", no isError) when its selector matches nothing — the #5902 fix landed on inputText only
Filed by the MCP dogfood routine (2026-09-05, daemon v0.0.67). Follow-up to #5902, which fixed this exact failure shape for `inputText` but left `tapOn` untouched.

## Side-by-side, same session, same daemon, same miss condition

`inputText` with a selector that matches nothing — **correct**:

```jsonc
// result envelope
{ "content": [...], "isError": true }
// body
{ "message": "Failed to input text: Failed to perform tap on element: Element not found with provided text 'ZZZ_NO_FIELD_ZZZ'",
  "success": false }
```

`tapOn` with a selector that matches nothing — **wrong**:

```jsonc
// result envelope — note: no isError at all
{ "content": [...], "structuredContent": {...}, "success": false, "error": "..." }
// body
{ "message": "Tapped on element (0 view hierarchy changes over 29 requests within 1521ms)",
  "success": false,
  "error": "Failed to perform tap on element: Element not found with provided text 'ZZZ_NO_SUCH_TEXT_ZZZ'",
  "element": { "bounds": { "left": 0, "top": 0, "right": 0, "bottom": 0 } },
  "debug": { "elementSearch": { "totalElementsChecked": 107 } } }
```

## Why this costs me as a client

- `message` is the field I read first and the one surfaced in most transcripts. It says **"Tapped on element"** for a tap that never happened. The correct information (`success: false`, `error`) is buried two levels down in a body I only parse when something already looks wrong.
- **No `isError` on the envelope.** MCP defines `isError` precisely so a client can branch without understanding the payload schema. `inputText` sets it; `tapOn` does not. A conforming client sees an unremarkable successful tool result.
- The envelope does carry non-standard top-level `success`/`error` keys, but those are not part of the MCP contract and a generic client will not look at them.

This bit me for real in this run: with a nonexistent resource id (`com.google.android.deskclock:id/definitely_not_here`) I got `"Tapped on element"` and moved on believing the tap landed.

## Smallest change

In `tapOn`'s result construction, mirror what `inputText` already does: when `success === false`, set the envelope's `isError: true` and make `message` the failure text (`"Failed to tap: <error>"`) instead of `"Tapped on element …"`. Worth auditing the other action tools for the same gap while in there — `inputText` was fixed in isolation and `tapOn` is the most-used tool of the family.

kaeawc:
## Summary

`tapOn` with a selector that matches nothing still returned a success-shaped result: the `message` read `"Tapped on element (0 view hierarchy changes over 29 requests within 1521ms)"` and the MCP envelope carried no `isError`, so a conforming client saw an ordinary successful tool result and moved on. [#5902](https://github.com/kaeawc/auto-mobile/issues/5902) fixed this exact shape for `inputText` (commit `6b4b96637`) but left `tapOn` untouched.

This mirrors that fix for `tapOn`: the message is gated on `result.success` (`"Failed to tap: <error>"`, where `<error>` already names the selector that matched nothing, followed by the existing search summary when the tap polled — e.g. `"Failed to tap: … Element not found with provided text 'ZZZ' (0 view hierarchy changes over 29 requests within 1521ms)"`) and the envelope gets `isError: true` on failure. Successful taps are byte-for-byte unchanged.

## Linked Issue

Closes #6152

## Bug / Regression Context

- **Prior attempts:** [#5941](https://github.com/kaeawc/auto-mobile/pull/5941) fixed the same failure shape for `inputText` only.
- **Observed symptom:** selector miss → `message: "Tapped on element (…)"`, `success: false`, `error: "… Element not found …"`, and no `isError` on the envelope.
- **Repro steps / conditions:** any `tapOn` whose selector resolves nothing (e.g. `selector: { text: "ZZZ_NO_SUCH_TEXT_ZZZ" }` or a nonexistent resource id).

## Root cause

`tapOnHandler` in `src/server/interactionTools.ts` called `buildTapOnResultMessage(...)` unconditionally (that helper only knows how to describe a match, so it always emits "Tapped on element") and returned `createStructuredToolResponse(...)` as-is, never setting `isError`. The non-standard top-level `success`/`error` keys that `createStructuredToolResponse` hoists are not part of the MCP contract, so a generic client never saw the failure.

## Changes

- `src/server/interactionTools.ts`
  - `tapOnHandler`: message is `result.success ? buildTapOnResultMessage(...) : "Failed to tap: <error>"` (`||` not `??` so an empty-string error still yields `"unknown error"`, per the #4183 P4 precedent), and the response is `result.success ? response : { ...response, isError: true }` — the same one-liner `inputTextHandler` uses.
  - Hoisted `tapOnHandler` to module scope with an injectable `TapOnElement` factory (`setTapOnElementFactory` / `resetTapOnElementFactory`), the same seam `pinchOnHandler` gained in #6056, so the registered handler wiring — not just a formatter — is pinned by a unit test.
  - Extracted the search-summary computation into `buildTapOnSearchSummary` (pure move; keeps the handler under the `complexity` 12 threshold). Its early `return` narrows `searchStats`, retiring two baselined `TS18048` errors.
- `scripts/typecheck-baseline.txt`: ratcheted down 165 → 163 (`bun run typecheck:update`).
- `test/server/interactionTools.tapOnMessage.test.ts`: new `tapOnHandler (registered handler wiring)` suite.

No tool schema or description text changed; `schemas/tool-definitions.json` is unchanged (pre-commit regen confirmed).

## Tests

- registration pin: `ToolRegistry.getTool("tapOn")?.deviceAwareHandler` is the module-scope `tapOnHandler` the direct-call tests exercise, so the two cannot drift.
- selector miss with search stats → `isError: true`, message `"Failed to tap: Failed to perform tap on element: Element not found with provided text 'ZZZ_NO_SUCH_TEXT_ZZZ' (0 view hierarchy changes over 29 requests within 1521ms)"`, message does not contain "Tapped on element", and `structuredContent.message`/`success` agree with the wire text.
- selector miss without search stats → `"Failed to tap: Element not found with provided text 'Missing'"` (no empty parenthetical).
- `error: undefined` and `error: ""` → `isError: true` with `"Failed to tap: unknown error"` (never blank).
- successful tap → no `isError`, message unchanged (`Tapped on element (matched text="Internet"; 1 match; 2 view hierarchy changes over 5 requests within 300ms)`).

All of these failed before the change (the seam did not exist; with the old wiring the miss case produced "Tapped on element (…)" and no `isError`).

## Review notes

Reviewed with the repo's three-lens code review (runtime behavior, delivery/enforcement, and a generated lens over downstream consumers of the `isError` bit — `src/server/index.ts` session binding, `daemonMcpProxy` replay lease, plan executor, CLI). No blocking findings; every downstream consumer already treats an `inputText` miss the same way. One adjacent gap the downstream lens surfaced — `DefaultUIStateSetup.tapCloseButton` returning `true` after the first candidate regardless of outcome — is already tracked as [#6123](https://github.com/kaeawc/auto-mobile/issues/6123) and is deliberately not touched here.

## Audit of the other action handlers (not changed here)

The issue asks for an audit of the rest of the family. None of the others is the identical one-line pattern, so they are deliberately left out of this PR to keep it minimal:

- `swipeOn`, `pinchOn`: message is already outcome-gated, but the envelope never sets `isError` on `success: false`.
- `tapAny`, `dragAndDrop`, `clearText`, `selectAllText`, `pressButton`, `imeAction`: hard-coded success message **and** no `isError` — the same two-part gap `tapOn` had.

Tracked as a follow-up (linked in the comments below).

## Validation

- [x] `turbo run lint build test` passes
- [x] `bun run format:check` passes
- [x] `bun run typecheck` reports no new errors (baseline ratcheted down to 163)
- [x] New code uses interfaces + fakes; the new suite runs in well under 100ms with no device or DB
- [x] No new dependencies or helpers beyond the two module-scope functions above
- [x] Based on latest `main` (`fb68db14f`)

### DIFF
--- a/scripts/typecheck-baseline.txt
+++ b/scripts/typecheck-baseline.txt
@@ -138,8 +138,6 @@ src/server/compactBoundsAdvertisement.ts: error TS2698: Spread types may only be
 src/server/formTools.ts: error TS2345: Argument of type 'AdbExecutor | null' is not assignable to parameter of type 'AdbClient | null | undefined'.
 src/server/interactionTools.ts: error TS18048: 'freshness.actualTimestamp' is possibly 'undefined'.
 src/server/interactionTools.ts: error TS18048: 'freshness.requestedAfter' is possibly 'undefined'.
-src/server/interactionTools.ts: error TS18048: 'searchStats' is possibly 'undefined'.
-src/server/interactionTools.ts: error TS18048: 'searchStats' is possibly 'undefined'.
 src/server/interactionTools.ts: error TS2322: Type 'SwipeDirection | undefined' is not assignable to type 'SwipeDirection'.
 src/server/navigationTools.ts: error TS2339: Property 'coverage' does not exist on type 'ExploreExecutionResult'.
 src/server/navigationTools.ts: error TS2339: Property 'interactionsPerformed' does not exist on type 'ExploreExecutionResult'.
@@ -206,10 +204,10 @@ src/utils/ios-cmdline-tools/XcodeSigning.ts: error TS2322: Type 'Promise<string 
 # swap-fp: 3 :: src/utils/Backoff.ts: error TS2322: Type 'BackoffPolicy | readonly number[]' is not assignable to type 'BackoffPolicy'.
 # swap-fp: 30 :: src/features/action/IosSimulatorPermissions.ts: error TS2352: Conversion of type 'TccPermissionRow' to type 'Record<string, null | number | string>' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
 # swap-fp: 30 :: src/features/observe/DeviceServiceUtils.ts: error TS2694: Namespace '"src/features/observe/DeviceService"' has no exported member 'ImeActionResult'.
+# swap-fp: 30 :: src/server/interactionTools.ts: error TS18048: 'freshness.actualTimestamp' is possibly 'undefined'.
 # swap-fp: 30 :: src/server/observeTools.ts: error TS2322: Type 'TimingData' is not assignable to type 'TimingEntry'.
 # swap-fp: 32 :: src/features/accessibility/WcagAudit.ts: error TS2339: Property 'class' does not exist on type 'ViewHierarchyNode'.
 # swap-fp: 32 :: src/features/observe/ios/CtrlProxyGestures.ts: error TS2459: Module '"./types"' declares 'GestureTimingResult' locally, but it is not exported.
-# swap-fp: 32 :: src/server/interactionTools.ts: error TS18048: 'freshness.actualTimestamp' is possibly 'undefined'.
 # swap-fp: 32,40 :: src/db/performanceAuditRepository.ts: error TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'OperandValueExpressionOrList<Database, "performance_audit_results", "timestamp">'.
 # swap-fp: 33 :: src/features/utility/DefaultElementSelector.ts: error TS2339: Property 'findClickableElementsInContainer' does not exist on type 'ElementFinder'.
 # swap-fp: 33 :: src/features/utility/DefaultElementSelector.ts: error TS2339: Property 'findClickableParentsContainingText' does not exist on type 'ElementFinder'.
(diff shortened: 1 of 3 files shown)


kaeawc:
Follow-up for the other action handlers audited in the PR body (swipeOn/pinchOn missing `isError`; tapAny/dragAndDrop/clearText/selectAllText/pressButton/imeAction with hard-coded success messages): [#6163](https://github.com/kaeawc/auto-mobile/issues/6163).

Issue #2724, opened 2026-06-30
Daemon log file is unlinked by bunx temp-dir cleanup; on-disk logs end up empty


kaeawc:
## Summary

When the daemon is launched via `bunx`, it writes its log under the bunx temp directory. That directory is later cleaned up, but the daemon keeps the file descriptor open and continues writing to the now-unlinked inode. Result:

- The on-disk `…/auto-mobile/logs/server-*.log` files are **0 bytes**.
- The live log is only recoverable via the open fd, e.g. `tail /proc/<daemon-pid>/fd/1`.

This makes post-hoc debugging hard: by the time you look, the visible log files are empty, and you have to know to read the process's fd while the daemon is still alive.

Observed:
```
$ ls -la .../auto-mobile/logs/
-rw-r--r-- ... 0 server-26926.log
-rw-r--r-- ... 0 server-35452.log
...
$ ls -l /proc/<pid>/fd/1
... -> .../auto-mobile-daemon-XXXX/daemon.log (deleted)
```

A related symptom: cache writes also fail with `ENOENT` because the temp dir is gone, e.g.:
```
[OBSERVE_CACHE] Failed to save observe result to disk: ENOENT ... /auto-mobile/observe_results/observe_...json
```

## Ask

Write the daemon log (and persistent caches) to a stable, non-ephemeral path that outlives bunx temp-dir cleanup — e.g. a configurable log dir, or the user data dir rather than the bunx extraction dir.

PR #6193, opened 2026-09-05
fix(logPruner): stop daemon own-file cap from claiming daemon-launch-<pid>.log

### LINKED ISSUE #6120: logPruner: daemon's own-file cap claims peer `daemon-launch-<pid>.log` files, deleting the daemon's rotated backups while launch logs are never swept
## `isOwnedBy(file, "daemon")` matches `daemon-launch-<pid>.log`

`src/utils/logPruner.ts:34-36`:

```ts
function isOwnedBy(file: string, ownPrefix: string): boolean {
  return file === `${ownPrefix}.log` || file.startsWith(`${ownPrefix}-`);
}
```

The daemon's log prefix is the bare string `"daemon"` (`src/utils/logger.ts:128-130`), and the manager writes launch-capture logs named `daemon-launch-<pid>.log` (`src/daemon/manager.ts:1221`). `"daemon-launch-123.log".startsWith("daemon-")` is `true`, so in the daemon process:

1. **Own-file cap (a)** (`:79-85`): every manager's launch log counts toward the daemon's 10-file cap. Sorted lexically, `daemon-<ISO timestamp>-...log` (the daemon's rotated backups) sort before `daemon-launch-...`, so the backups are the files that get unlinked while the launch logs survive.
2. **Sweep (b)** (`:88-91`): `continue`s on anything "owned", so the daemon never reaps stale `daemon-launch-*.log` files even after `ownerPid()` (which explicitly recognizes them, per the comment for [#2724](https://github.com/kaeawc/auto-mobile/issues/2724)) says the owner exited.

The "exact PID boundary" comment describes the intent for `stdio-<pid>`, but the `daemon` prefix has no PID and collides with `daemon-launch-`.

### Trigger

≥ 7 agent launches leave `daemon-launch-*.log` in the logs dir; the daemon rotates a 10 MB log → all its rotated backups are unlinked, launch logs stay.

### Evidence

Repro (`bun test`, real `pruneLogFiles`, 3 rotated + 9 launch + active, `maxOwnFiles: 10`):

```
before: daemon-2026-01-01.log daemon-2026-01-02.log daemon-2026-01-03.log daemon-launch-100..900.log daemon.log
after:  daemon-launch-100..900.log daemon.log
```

### Fix

Make `isOwnedBy` match only `<prefix>.log` or `<prefix>-<rotation timestamp>.log` (e.g. `^${prefix}-\d{4}-\d{2}-\d{2}T`), or exclude any name `ownerPid()` recognizes. Add the repro above as a unit test.

### Acceptance criteria

- [ ] Daemon rotated backups are retained up to `maxOwnFiles`; launch logs are not counted.
- [ ] `daemon-launch-<dead pid>.log` older than the stale threshold is removed by the daemon's sweep.
- [ ] Red before / green after.

### Relationships

- Follow-up to: [#2724](https://github.com/kaeawc/auto-mobile/issues/2724) (added launch-log recognition to `ownerPid`, but `isOwnedBy` was not adjusted).

Confidence: high. Found in the 2026-09-04 TypeScript bug hunt.

kaeawc:
## Summary

Fixes `isOwnedBy("daemon", ...)` in `src/utils/logPruner.ts` wrongly matching
peer `daemon-launch-<pid>.log` files as owned by the `daemon` process.

## Root cause

`isOwnedBy` matched any file that merely `startsWith(\`${ownPrefix}-\`)`. The
daemon's log prefix is the bare string `daemon`, and the manager writes
launch-capture logs named `daemon-launch-<pid>.log` (owned by the spawning
manager, not the daemon). `"daemon-launch-123.log".startsWith("daemon-")` is
`true`, so:

1. **Own-file cap**: every manager's launch log counted toward the daemon's
   `maxOwnFiles` cap. Sorted lexically, the daemon's genuinely-owned rotated
   backups (`daemon-<timestamp>.log`) sort before `daemon-launch-*.log`, so
   the backups were the files evicted while launch logs survived.
2. **Sweep**: the sweep step `continue`s on anything "owned", so the daemon
   never reaped stale `daemon-launch-*.log` files via its own pruning pass
   even after `ownerPid()` recognized their owning PID had exited.

## Fix

`isOwnedBy` now excludes the one named sub-role that collides with the
prefix boundary — `<prefix>-launch-<pid>.log` — while keeping the existing
exact-PID-boundary match (`stdio-12` never claims `stdio-123`) for every
other rotated-file shape used elsewhere in the suite.

## Testing

- `test/utils/logPruner.test.ts` (new): reproduces the issue's exact repro
  shape (3 rotated daemon backups + 9 `daemon-launch-*.log` + active,
  `maxOwnFiles: 10`) and asserts the rotated backups survive the cap and the
  launch logs are untouched by it; also covers the daemon's own-file cap
  still pruning its own excess backups, and the sweep both removing a dead
  launch log past the stale threshold and never removing a live peer's.
- `test/daemon/loggerPrune.test.ts` (pre-existing, unmodified): still green,
  confirming the fix preserves the exact-PID-boundary behavior for
  `stdio-<pid>` and the existing `daemon-launch-<pid>.log` sweep coverage
  from #2724.
- Full suite: `turbo run lint build test`, `bun run typecheck`,
  `bun run format:check` all green locally.

## Acceptance criteria

- [x] Daemon rotated backups are retained up to `maxOwnFiles`; launch logs
      are not counted.
- [x] `daemon-launch-<dead pid>.log` older than the stale threshold is
      removed by the daemon's sweep.
- [x] Red before / green after.

Closes #6120

### DIFF
--- a/src/utils/logPruner.ts
+++ b/src/utils/logPruner.ts
@@ -30,11 +30,24 @@ function defaultIsProcessAlive(pid: number): boolean {
 
 /**
  * Whether `file` belongs to `ownPrefix`, matched on an exact PID boundary so
- * `stdio-12` never claims `stdio-123`'s files. Filenames are
- * `<prefix>.log` (active) and `<prefix>-<ts>.log` (rotated).
+ * `stdio-12` never claims `stdio-123`'s files. Filenames are `<prefix>.log`
+ * (active) and `<prefix>-<rotation suffix>.log` (rotated).
+ *
+ * `daemon-launch-<pid>.log` is the manager's launch-capture log — owned by
+ * whichever process spawned it, not the daemon — but it textually starts with
+ * `daemon-` too, so a bare `startsWith(\`${ownPrefix}-\`)` wrongly claims it for
+ * the `daemon` owner. Excluding that one named sub-role keeps the exact-PID-
+ * boundary match for everything else (issue #6120).
  */
 function isOwnedBy(file: string, ownPrefix: string): boolean {
-  return file === `${ownPrefix}.log` || file.startsWith(`${ownPrefix}-`);
+  if (file === `${ownPrefix}.log`) {
+    return true;
+  }
+  if (!file.startsWith(`${ownPrefix}-`)) {
+    return false;
+  }
+  const rest = file.slice(ownPrefix.length + 1);
+  return !/^launch-\d+\.log$/.test(rest);
 }
 
 /**
--- a/test/utils/logPruner.test.ts
+++ b/test/utils/logPruner.test.ts
@@ -1,0 +1,123 @@
+import { describe, expect, test } from "bun:test";
+import { mkdtemp, readdir, rm, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import { pruneLogFiles } from "../../src/utils/logPruner";
+
+async function withTempLogDir(fn: (dir: string) => Promise<void>): Promise<void> {
+  const dir = await mkdtemp(path.join(tmpdir(), "logpruner-test-"));
+  try {
+    await fn(dir);
+  } finally {
+    await rm(dir, { recursive: true, force: true });
+  }
+}
+
+describe("logPruner isOwnedBy (via pruneLogFiles)", () => {
+  test("daemon own-file cap does not count daemon-launch-<pid>.log and does not delete them", async () => {
+    await withTempLogDir(async (dir) => {
+      // 3 rotated daemon backups + active + 9 daemon-launch capture logs, matching
+      // the issue's repro shape (issue #6120).
+      const rotated = [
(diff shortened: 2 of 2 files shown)


kaeawc:
The daemon-launch-log fd-lifecycle hazard (a still-running daemon holding an inherited stdio fd to a launch log the sweep may unlink) is real and pre-existing (#2724) — this PR only widens exposure because isOwnedBy now correctly includes these files. The suggested fix (never sweep them for the daemon) contradicts #6120's acceptance criterion that a stale *dead-pid* launch log must be swept; the correct resolution is to correlate launch-log liveness to the daemon's own pid rather than the manager's, which is a separate redesign. Deferred to #6194. This PR fixes #6120's stated defect (rotated backups wrongly evicted) with both acceptance criteria under test. Refs #6194

PR #6206, opened 2026-09-05
test(record): property-test GestureClassifier classification geometry

### LINKED ISSUE #6717: Complete GestureClassifier property coverage, including zero-scale pinches
Track the property-coverage change in [PR 6206](https://github.com/kaeawc/auto-mobile/pull/6206).

The example-based GestureClassifier suite does not systematically cover the single-finger partition, swipe geometry, finite pinch scaling, and finite numeric outputs. The draft adds deterministic fast-check properties, but its positive-scale assertion incorrectly rejects a valid converged pinch: two fingers start at (0,0)/(100,0), both finish at (50,0), and the classifier correctly emits scale 0 with inward direction.

Acceptance criteria:

- Preserve deterministic property coverage for tap/long-press/swipe partitioning, swipe geometry and velocity, pinch scaling/direction, and finite emitted numeric fields.
- Accept zero-scale convergence as valid, with an explicit regression case. Do not change production gesture semantics to satisfy a faulty test assertion.
- Keep each added unit test within the repository's 100 ms budget and pass relevant record tests, formatting, lint, build, and type checking.

The zero-scale counterexample was reproduced during the authored-PR burndown; production behavior is correct. This issue tracks completion and delivery of the test change.

kaeawc:
Adds deterministic property coverage for GestureClassifier's single-finger classification, swipe geometry, pinch scale and direction, and finite numeric output. A converged two-finger gesture is explicitly covered: a valid inward pinch can end at scale zero.

Closes [issue 6717](https://github.com/kaeawc/auto-mobile/issues/6717).

Validation: 126 recording tests passed, including all eight property/regression tests below the timing limit. Lint, build, typecheck baseline gate, and formatting passed. Production code is unchanged.

### DIFF
--- a/test/features/record/android/GestureClassifier.property.test.ts
+++ b/test/features/record/android/GestureClassifier.property.test.ts
@@ -1,0 +1,267 @@
+import { describe, test, expect } from "bun:test";
+import fc from "fast-check";
+import { GestureClassifier } from "../../../../src/features/record/android/GestureClassifier";
+import { GESTURE_THRESHOLDS } from "../../../../src/features/record/android/types";
+import type { RawTouchFrame } from "../../../../src/features/record/android/types";
+import type { CoordScaler } from "../../../../src/features/record/android/AxisRanges";
+
+// Property-based tests for GestureClassifier's pure classification geometry.
+// The example-based suite (GestureClassifier.test.ts) pins specific gestures;
+// these assert the invariants the classifier *promises* over the whole input
+// space: the tap/longPress/swipe partition, direction derivation, pinch
+// scale/direction agreement, and the guard that a completed gesture never
+// carries a non-finite number. See Backoff.property.test.ts / ElementGeometry
+// .property.test.ts for the pinned-seed rationale — the same generated cases
+// run every CI invocation, so any counterexample is reproducible.
+const RUN_OPTIONS = { seed: 0x6e_57_1a_2f, numRuns: 400 } as const;
+
+// Identity scaler: raw sensor coordinates == logical screen pixels, so the
+// test can reason about geometry in a single coordinate space.
+const identityScaler: CoordScaler = {
+  toScreenX: (x: number) => x,
+  toScreenY: (y: number) => y,
+};
+
+function makeFrame(
+  arrivedAt: number,
+  activeSlots: Array<{ slotId: number; trackingId: number; x: number; y: number }>,
+  releasedSlots: number[] = [],
+): RawTouchFrame {
+  return {
+    arrivedAt,
+    activeSlots: activeSlots.map((s) => ({ ...s, pressure: 0 })),
+    releasedSlots,
+  };
+}
+
+const dist = (x1: number, y1: number, x2: number, y2: number): number =>
+  Math.sqrt((x2 - x1) ** 2 + (y2 - y1) ** 2);
+
+// Density in the range real Android screens report (mdpi..xxxhdpi ≈ 1..4).
+const density = fc.integer({ min: 10, max: 40 }).map((d) => d / 10);
+const coord = fc.integer({ min: 0, max: 2000 });
+
+/**
+ * Drive a single finger through DOWN → (optional MOVE) → UP against a fresh
+ * classifier and return the completed gesture. A fresh classifier means
+ * `lastTap` is null, so a low-displacement short contact can only be a `tap`,
+ * never a `doubleTap`.
+ */
+function singleFinger(
+  d: number,
+  down: { x: number; y: number },
+  up: { x: number; y: number },
+  durationMs: number,
+) {
+  const c = new GestureClassifier(identityScaler, d);
+  c.feedFrame(makeFrame(0, [{ slotId: 0, trackingId: 1, x: down.x, y: down.y }]));
+  c.feedFrame(makeFrame(1, [{ slotId: 0, trackingId: 1, x: up.x, y: up.y }]));
+  return c.feedFrame(makeFrame(durationMs, [], [0]));
+}
+
+/** Drive two fingers through DOWN → MOVE → simultaneous UP against a fresh classifier. */
+function twoFinger(
+  d: number,
+  a0: { x: number; y: number },
+  b0: { x: number; y: number },
+  a1: { x: number; y: number },
+  b1: { x: number; y: number },
+) {
+  const c = new GestureClassifier(identityScaler, d);
+  c.feedFrame(
+    makeFrame(0, [
+      { slotId: 0, trackingId: 1, x: a0.x, y: a0.y },
+      { slotId: 1, trackingId: 2, x: b0.x, y: b0.y },
+    ]),
+  );
+  c.feedFrame(
+    makeFrame(100, [
+      { slotId: 0, trackingId: 1, x: a1.x, y: a1.y },
+      { slotId: 1, trackingId: 2, x: b1.x, y: b1.y },
+    ]),
+  );
+  return c.feedFrame(makeFrame(200, [], [0, 1]));
+}
+
+describe("GestureClassifier (property-based)", () => {
+  // -------------------------------------------------------------------------
+  // Single-finger classification partition
+  // -------------------------------------------------------------------------
+
+  test("a zero-displacement contact is a tap iff short, a longPress iff long", () => {
+    fc.assert(
+      fc.property(density, coord, coord, fc.integer({ min: 0, max: 3000 }), (d, x, y, durationMs) => {
+        // start == end ⇒ displacement 0 < slop for every positive density,
+        // so the swipe branch is never taken and classification hinges only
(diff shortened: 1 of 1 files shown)


kaeawc:
## CI status — two ubuntu failures are pre-existing environment flakes, not this PR

This PR is **test-only** (one new file, `test/features/record/android/GestureClassifier.property.test.ts`; no `src/` changes). Two checks failed on the current head `f4897eb`, both on `ubuntu-latest`, and neither is caused by this diff:

1. **Node Unit Tests (ubuntu-latest)** — the unrelated test `shared-storage tools › advertises defaulted fields as optional` exceeded the repo's 100 ms per-test budget at **112 ms**, which the job enforces as a hard failure. The Node unit job runs each test file in an isolated `bun test` invocation, so a new gesture-classifier test file cannot slow a shared-storage test — this is timing variance on a loaded runner (this run coincided with a burst of other PR CI).

2. **Node Host Integration Tests (ubuntu-latest)** — `1862 pass, 1 fail, 2 errors`, all from `test/server/deviceLabelSessionReleaseOrdering.integration.test.ts`. Root cause is a Bun runtime/fd error, not an assertion about this change:
   ```
   error: EEXIST: file already exists, epoll_ctl
       at new WriteStream (internal:fs/streams:244:58)
   ```
   That aborted the run, which cascaded into the second error (`Cannot call describe() after the test run has completed`).

Locally on this branch `turbo run lint build test`, `bun run typecheck`, `bun run format:check`, and the new file (7 properties, 7841 assertions) all pass; the two failing tests live in unrelated areas (`shared-storage`, `deviceLabelSessionReleaseOrdering`) and fixing them would widen a test-only PR, so I have not touched them.

I attempted to re-run the failed jobs to confirm the flake but got `403 Resource not accessible by integration` — I don't have permission to re-run CI here. A maintainer re-run of the two failed ubuntu jobs (or the next push) should clear them. I'm keeping this PR watched until CI is green.

---

kaeawc:
## Steward: flagged `needs-human` — red >12h, owning routine can't self-heal

PR-queue steward pass. This PR's head `f4897eb` has been red for ~12.4h (Node Tests failed 2026-09-05T23:15Z) and the owning `property-test-expander` routine last acted at 2026-09-05T22:51Z — it diagnosed the two ubuntu failures as pre-existing environment flakes, [tried to re-run the jobs but got `403 Resource not accessible by integration`](https://github.com/kaeawc/auto-mobile/pull/6206#issuecomment-5555325769), and has not refired since.

Per the steward rules I don't re-run CI or otherwise fix this myself. A human is needed to either re-run the two failed ubuntu jobs to confirm the flake, or sequence this draft. No unresolved review threads and no cross-PR file collisions on this branch.

---

kaeawc:
## Re-run update — ubuntu integration suite is flaky independent of this diff

A fresh full run (`34031000962`) went through on head `f4897eb`. Outcome:

- **`Node Unit Tests (ubuntu-latest)` now passes** — the earlier 100 ms-budget overshoot was a one-off timing flake, as expected.
- **`Node Host Integration Tests (ubuntu-latest)` failed again, but with a _different_ failure**: this time the suite **hung** on `test/server/androidNavigationWorkflow.integration.test.ts` and was killed by the 900 s wall-timeout (`##[error]Process completed with exit code 124`). Last run it was an `EEXIST: file already exists, epoll_ctl` crash in a _different_ test (`deviceLabelSessionReleaseOrdering.integration.test.ts`). A different failure point each run, on ubuntu only, is the signature of a pre-existing flaky/hang-prone integration suite — not something a downstream diff introduced.

Decisive point on causality: the `Node Host Integration Tests` job runs `scripts/test-ts.sh integration`, which loads **only `*.integration.test.ts` files**. This PR adds `test/features/record/android/GestureClassifier.property.test.ts` — a **unit** test that this integration job never loads or executes. So these ubuntu integration failures cannot originate from this diff. Consistent with that, the **macos-latest and windows-latest** variants of both Node jobs pass.

I still can't re-run CI myself (`403 Resource not accessible by integration`). Correcting my earlier note: a plain re-run did not clear it, because the ubuntu integration suite flakes at a new spot each time. Getting this PR green will need either that suite stabilized on ubuntu (separate, pre-existing issue) or the `Node Host Integration Tests (ubuntu-latest)` job re-run until it passes. The change itself is sound — happy to keep this watched until it's green.

---

kaeawc:
Addressed the Codex P2 (density-scaled touch-slop boundary) in `7a2d4bc`, on top of your converged-pinch fix.

Added a property that straddles the threshold with a single-axis move, so the Euclidean displacement equals the integer offset exactly (identity scaler): strictly below `TOUCH_SLOP_DP * d` must classify as `tap`, at-or-above as `swipe`. The exact-boundary case (`displacement === slopPx`) is exercised whenever `8*d` is integral (d = 1.0, 1.5, 2.0, …), which pins the classifier's `<`-vs-`<=` comparison; the below-slop assertion pins the density scaling. Both regressions Codex named now fail the suite.

Validated locally: `bun test` on this file (9 properties pass, ~42 ms each), `bun run format:check`, `bun run lint` (no new ratchet violations), `bun run typecheck` (no new errors). Test-only; happy to adjust if you'd prefer a different shape.

---

kaeawc:
Addressed the zero-duration speed finding in `e13d565`.

Added a property that drives a swipe whose DOWN and UP frames share a timestamp (`durationMs === 0`, reachable since `arrivedAt` is `Date.now()`) and asserts `"normal"`, with a contrasting 1 ms swipe asserting `"fast"` — so the classifier's `durationMs > 0 ? … : 0` guard is shown to be what produces `"normal"`, not a blanket result. Dropping that guard (Infinity velocity → `"fast"`) now fails the suite.

On the fling-threshold boundary half: I deliberately did **not** add an exact-threshold assertion. `FLING_MIN_DP_PER_S * d` isn't exactly representable for most `d = k/10`, so a `velocity === threshold` check is float-fragile (my first attempt flaked on the roundtrip). More importantly, the existing "…clears touch-slop is always a swipe" property already pins the threshold: it recomputes `velocity >= flingThreshPx` with the same density-scaled formula and asserts the classifier agrees across random durations, which catches a dropped `* d` or an altered comparison. Happy to add an explicit near-threshold pair if you'd prefer it spelled out.

Validated locally: 10 properties pass (~34 ms each), `format:check`, `lint` (no new ratchet violations), `typecheck` (no new errors).

---

kaeawc:
Addressed the pinch-boundary finding in `a2dc8e5`. Added a deterministic property that brackets `PINCH_MIN_SCALE_DELTA` with axis-aligned fingers (exact-integer distances → clean `finalDist/initialDist` ratio): `|scale − 1| = 0.11` (finalDist 111 / 89) must emit with the matching direction, `0.09` (109 / 91) must be suppressed — tight to the 0.1 delta but clear of float-equality fragility.

That completes deterministic boundary coverage for the classifier's three thresholds — touch-slop (`a2dc8e5`… actually `7a2d4bc`), fling velocity (existing property mirrors the density-scaled formula), and now the pinch min-scale-delta. Validated locally: 11 properties pass (~59 ms each), `format:check`, `lint` (no new ratchet violations), `typecheck` (no new errors).

---

kaeawc:
Codex's latest pass on `a2dc8e5` raised four more suggestions. Since you're driving this PR and a couple of these are non-trivial test-design changes (not small nits), I'm holding off pushing further and leaving the scope call to you rather than churning more CI runs. My read:

**Worth doing:**
- **Coordinate scaling (#1)** — every helper uses the identity scaler, so `screenDist`'s scaler calls are effectively untested. An anisotropic fake scaler (different X/Y factors) with expected gestures derived from *scaled* coordinates would close a real gap. This is a meaningful new test, not a one-liner.
- **DPI-derived density (#2)** — the `density` arbitrary emits `k/10` values, but production `queryDensity` returns `DPI/160` (e.g. `420/160 = 2.625`, fallback `2.75`). A regression rounding density to one decimal would pass today. Switching the generator to realistic integer DPIs `/160` and pinning `2.625`/`2.75` is a good change — but it touches the shared arbitrary every property uses, so I'd rather you okay it than silently reshape the whole file.

**Lower value / float-fragile:**
- **Exact fling equality (#3)** and **exact pinch delta (#4)** — these want assertions *exactly at* the threshold. In double precision the constructed values land just off the mathematical boundary (Codex's own `40px/800ms` computes to ≈`50.0000000000000071 > 50`, so `>=` vs `>` isn't actually distinguished; `100→90` suppresses while `100→110` emits, so `0.10` isn't symmetric). An earlier exact-threshold attempt of mine flaked for exactly this reason. If you want these pinned, the robust form is an explicit documented tolerance rather than literal equality.

Happy to implement #1 and #2 (and #3/#4 with an explicit tolerance) — just say the word and I'll push on top of your branch. Otherwise the change as it stands has all three classifier thresholds bracketed and the zero-duration guard covered.

---

Issue #6258, opened 2026-09-06
effect.screenChanged reports false when a tap opens a dialog (activeWindow basis never falls through to viewHierarchy), and diff-mode observations omit activeWindow/freshness


kaeawc:
## Context

Dogfood loop, 2026-09-06, Android emulator API 34, local build of main `b7f61eb1b`. This is a follow-up to https://github.com/kaeawc/auto-mobile/issues/6219, whose original defect **is fixed** — the observation returned by an action tool is now the post-transition capture. Verified: `tapOn {text:"Add alarm"}` now returns the time-picker skeleton, not the alarm list.

But the pair is still inconsistent, in the opposite direction, and the diff-mode observation now drops the two fields a client uses to orient itself.

## 1. `effect.screenChanged` is false for a tap that opens a dialog

```jsonc
// tapOn { selector: { text: "Add alarm" } } — opens the Material time picker
{
  "message": "Tapped on element (matched id=…:id/fab text=\"Add alarm\"; 1 match)",
  "effect": { "screenChanged": false, "basis": "activeWindow unchanged" },   // <- wrong
  "observation": { "isDiff": true, "skeleton": [
      { "elementId": "s-4c7c9a06373fb435", "label": "12" },
      { "elementId": "…:id/material_clock_period_am_button", "label": "AM" },
      { "elementId": "…:id/material_timepicker_mode_button", … } ] },        // <- the dialog, correct
  "observationDiff": { "mode": "diff", "reason": "diff_emitted",
      "fromScreen": { "activeWindow": { "activityName": "com.android.deskclock.DeskClock" } },
      "toScreen":   { "activeWindow": { "activityName": "com.android.deskclock.DeskClock" } } }  // <- also wrong
}
```

The skeleton in that very response is the dialog. `effect` says nothing changed, because `basis` is `activeWindow` and the dialog window is not reflected there (the known dialog-window gap, https://github.com/kaeawc/auto-mobile/issues/6151). A client that branches on `effect.screenChanged` — which is what the field is for — concludes the tap did nothing and retries, double-opening or dismissing the dialog.

`effect` clearly *can* do better: the same session's `tapOn` on a notification `Reply` button returned `{"screenChanged":true,"basis":"viewHierarchy changed"}`. So the hierarchy-based basis exists and works; the dialog case just resolves to the activeWindow basis first and stops there.

[...]

Issue #5377, opened 2026-08-18
observe freshness: Android ageMs/warning corrupted by device↔host clock skew (isFresh protected, age math is not)


kaeawc:
## Summary

On Android, `observe`'s `freshness.ageMs` and `freshness.warning` are computed
by subtracting a **device-authored** capture timestamp from a **host-authored**
`now`, so any clock skew between the emulator and the host is misreported as
observation age. `freshness.isFresh` is correct (it is protected by the
delegate's `fresh` verdict — see below), but the reported *age* and *warning*
are wrong whenever the device clock differs from the host clock, which is common
on emulators.

Discovered during a post-`0.0.58` manual-test pass verifying #5272
(https://github.com/kaeawc/auto-mobile/pull/5272). iOS is unaffected (the
simulator shares the host clock).

## Reproduction (emulator-5554, API 34, current `main` @ `7f6d641`)

Emulator clock measured ~24s behind the host:

```
host epoch ms: 1787057412905 .. 1787057414918
emu  epoch ms:  1787057388878        # ~24s behind host
```

A single `observe` immediately after navigating (fresh screen content, e.g.
skeleton contains "Wifi") returns:

```json
"freshness": {
  "actualTimestamp": 1787057389299,   // == emu clock ~now (+421ms), i.e. a fresh capture
  "ageMs": 25597,                      // == hostNow - deviceTimestamp  (WRONG)
  "verified": true,
  "isFresh": true,
  "warning": "Hierarchy was verified against the device on this call but is already 25597ms old (budget 5000ms) — the rest of the observation was slow."
}
```

Correlation proves the timestamp is device-domain, not stale:
`actualTimestamp` is only **+421ms** past the emulator's current clock but
**−25619ms** relative to the host clock — i.e. the tree was captured ~now on the
device, and the ~25s "age" is pure clock skew. `ageMs` (25597) ≈ `hostNow −
deviceTimestamp`. The warning's "the rest of the observation was slow" is a false
diagnosis — the observation was not slow.

## Root cause

- `ObserveScreen.resolveObservationTimestampMs` returns the device-authored
  `viewHierarchy.updatedAt` (`src/features/observe/ObserveScreen.ts:640-652`),
  set from the Android runner payload at
  `src/features/observe/android/CtrlProxyHierarchy.ts:376-377`.
- `computeFreshness` is called with `now: this.timer.now()` — the **host** clock
  (`src/features/observe/ObserveScreen.ts:346-355`) — and computes
[...]

Issue #10, opened 2025-12-23
Retain and Reuse Navigation


kaeawc:
Extending #9 to keep reusing the retained navigation across runs and different machines.

PR #6309, opened 2026-09-07
fix(tap): coherent post-tap hierarchy-only settle under device-clock skew (#6284)

### LINKED ISSUE #6284: Post-tap hierarchy-only transient settling under device-clock skew (coherent design)
Split out of PR #6266 (issue #6258). The core #6258 fix — `deriveTapEffect`'s activeWindow→viewHierarchy fall-through (so a dialog-opening tap reports `screenChanged: true`) plus diff-mode observations carrying `activeWindow`/`freshness` — lands in #6266 on its own and delivers the actual repro without any settle-poll machinery.

This issue tracks the **settle-poll / freshness-consistency hardening** that was attempted incrementally in #6266 and produced 5 rounds of interacting-guard findings. It needs a coherent design pass rather than per-call-site patching. Root cause of the chain (per an architecture review): the `bothFresh` gate added to `compareViewHierarchy` — once a stale baseline can block all future diffs, everything downstream (freshness realignment at every hierarchy-replace site, device-clock-domain poll-floor seeding, a monotonic floor, screen-off terminal preservation, `enforceFreshnessConsistencyWithEffect`) becomes load-bearing and mutually entangled.

## Scope
Design robust settling of **hierarchy-only post-tap transients** such that:
- The observation timestamp floor lives in ONE clock domain end-to-end (device-authored `updatedAt`), so an Android device clock trailing the daemon does not make every fresh capture look stale. No per-call-site host-vs-device floor patching; `waitForPostTapChange` and all default callers must use the same domain.
- A stale cached observation cannot lower a monotonic floor, yet the default path must still be able to seed the floor to the first observation's device timestamp (the round-4/round-5 tension).
- A definitive screen-off (`wakefulness: "Asleep"`) terminal is preserved as `screenChanged: true` through BOTH re-derivation and any freshness-consistency guard.
- Every site that replaces an observation's hierarchy also refreshes its freshness (structurally enforced — a single sanctioned helper).
- A real stability interval is required before declaring a hierarchy-only transient settled, without missing a later destination.

## Findings captured from #6266 review (for the redesign)
- `bothFresh` gate in `compareViewHierarchy` is the entanglement root.
- Device-clock-domain floor: `initialMinTimestampMs` seeding + `Math.max` monotonic floor interact badly for callers without `initialMinTimestampMs` (host floor never decreases to device `updatedAt`).
- `enforceFreshnessConsistencyWithEffect` (#6219) overwrites the screen-off terminal with `isFresh:false` + wrong warning.
- Freshness realignment needed at the pre-tap-stability path AND the `searchForElement` replace path.

No `bothFresh`/settle machinery should be reintroduced piecemeal; design it whole here.

kaeawc:
Split out of [#6266](https://github.com/kaeawc/auto-mobile/pull/6266) (issue [#6258](https://github.com/kaeawc/auto-mobile/issues/6258)) as a coherent design pass, after the settle/freshness hardening was attempted incrementally there and produced five rounds of interacting-guard findings. The core [#6258](https://github.com/kaeawc/auto-mobile/issues/6258) fix (`deriveTapEffect` fall-through) already merged; this PR is only the settle-poll / freshness-consistency hardening, designed whole.

## Root cause of the earlier entanglement
The earlier attempt added a `bothFresh` gate to `compareViewHierarchy`: once a stale baseline could block *all* future diffs, every downstream guard (freshness realignment at each replace site, device-clock-domain floor seeding, a monotonic floor, screen-off terminal preservation) became load-bearing and mutually entangled — and a same-window dialog could never be reported changed.

This redesign moves staleness robustness **out of the comparison** and **into a single device-clock-domain monotonic poll floor**, so `compareViewHierarchy` stays a pure hash diff and legitimate diffs are never blocked.

## What changed
- **One clock domain end-to-end.** `pollObserveUntil`'s `minTimestamp` floor is now purely device-authored `updatedAt`. It is seeded from the caller's device-domain `initialMinTimestampMs` (or from the first observation when unseeded), **never** from the host clock — so an Android device clock trailing the daemon no longer makes fresh captures look stale ([#5377](https://github.com/kaeawc/auto-mobile/issues/5377) family). `waitForPostTapChange` and the settle both seed from the entering capture's `updatedAt`.
- **Monotonic floor.** The floor is only ever raised (`Math.max`), so a stale/cached capture returned by a timed-out sub-poll cannot lower it and re-serve the same stale hash. The default path still seeds the floor to the first observation's device timestamp.
- **Real stability interval, without missing a later destination.** Hierarchy-only post-tap changes are settled (hash unchanged across two consecutive poll comparisons, or promoted to an authoritative `activeWindow` change) before being trusted as terminal, so `baseline → A → A → B` reaches **B** rather than returning the transient A.
- **Screen-off terminal preserved.** A definitive `wakefulness: "Asleep"` terminal keeps the already-established `screenChanged: true` through both settle re-derivation **and** `enforceFreshnessConsistencyWithEffect`.
- **Freshness realigned at every hierarchy-replace site.** Both mutation sites route through one sanctioned helper (`replaceObservationHierarchy`) that also refreshes freshness on a live device re-capture.

## Tests (deterministic, <100ms, no device/DB)
`test/features/observe/ObservePoll.test.ts` and `test/features/action/TapOnElement.hierarchySettle.test.ts`: device-clock-trails-host seeds a device-domain floor; monotonic floor not lowered by a stale poll; `baseline → A → A → B` reaches B; same-activity dialog reported changed; asleep terminal preserves `screenChanged: true` (asserts the effect); freshness realigned at the replace site; screen-off guard in the consistency check.

## Validation
`bun test test/features/action/TapOnElement*.test.ts test/features/observe/*.test.ts` (1119 pass) · `bun run lint` (ratchet: no new violations) · `bun run typecheck` (no new errors) · `bun run format:check` · `bunx turbo run build`.

Closes #6284

### DIFF
--- a/schemas/tool-definitions.json
+++ b/schemas/tool-definitions.json
@@ -6412,6 +6412,16 @@
             },
             "warning": {
               "type": "string"
+            },
+            "category": {
+              "type": "string",
+              "enum": [
+                "cache_age",
+                "window_identity",
+                "requested_min",
+                "no_timestamp",
+                "unavailable"
+              ]
             }
           },
           "required": ["isFresh"],
@@ -10048,6 +10058,16 @@
                     },
                     "warning": {
                       "type": "string"
+                    },
+                    "category": {
+                      "type": "string",
+                      "enum": [
+                        "cache_age",
+                        "window_identity",
+                        "requested_min",
+                        "no_timestamp",
+                        "unavailable"
+                      ]
                     }
                   },
                   "required": ["isFresh"],
--- a/schemas/tool-definitions.json
+++ b/schemas/tool-definitions.json
@@ -6420,7 +6420,8 @@
                 "window_identity",
                 "requested_min",
                 "no_timestamp",
-                "unavailable"
+                "unavailable",
+                "effect_inconsistent"
               ]
             }
           },
@@ -10066,7 +10067,8 @@
                         "window_identity",
                         "requested_min",
                         "no_timestamp",
-                        "unavailable"
+                        "unavailable",
+                        "effect_inconsistent"
                       ]
                     }
                   },
--- a/scripts/oxlint-baseline.txt
+++ b/scripts/oxlint-baseline.txt
@@ -56,6 +56,7 @@
 1	src/features/observe/GetBackStack.ts	eslint(complexity)
 1	src/features/observe/Idle.ts	eslint(complexity)
 1	src/features/observe/ListInstalledApps.ts	eslint(complexity)
+1	src/features/observe/ObservePoll.ts	eslint(complexity)
 1	src/features/observe/ObserveScreen.ts	eslint(max-depth)
 1	src/features/observe/TakeScreenshot.ts	auto-mobile(catch-convention)
 1	src/features/observe/TakeScreenshot.ts	typescript(no-floating-promises)
--- a/src/features/action/TapOnElement.ts
+++ b/src/features/action/TapOnElement.ts
@@ -116,15 +116,13 @@ const POST_TAP_EFFECT_POLL_MS = 150;
 
 /**
  * Minimum wall-clock time a hierarchy-only post-tap frame must hold UNCHANGED
- * before it is accepted as the settled destination (issue #6284 P1). Set to two
- * full poll intervals so a settle can never be declared on the first, zero-
- * elapsed comparison, nor after a single interval — a real quiet period has to
- * elapse. A delayed dialog/navigation frame `B` that arrives more than one poll
- * interval after a transient `A` (the normal case for the delayed transitions
- * this settle targets) therefore resets the quiet period and is reached, rather
- * than being pre-empted by settling on `A`.
+ * before it is accepted as the settled destination (issue #6284 P1). Activity
+ * transitions commonly take 1–2 seconds, so a transient hierarchy must stay
+ * quiet for a full second before it is terminal evidence. This prevents a
+ * routine delayed destination from being pre-empted by an intermediate frame
+ * that merely survives the first few 150ms polls.
  */
-const POST_TAP_SETTLE_QUIET_PERIOD_MS = 2 * POST_TAP_EFFECT_POLL_MS;
+const POST_TAP_SETTLE_QUIET_PERIOD_MS = 1000;
 
 /** Brief debounce between the original tap and the retry tap when a ghost tap
  *  was detected. Just enough to let any inflight gesture queue drain. */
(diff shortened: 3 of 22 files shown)


kaeawc (commented):
Round-2 P1 fixes pushed (5453d7cc9):

- **Force the first wait poll past the existing cache** — \`ObservePoll\` now forces the first poll STRICTLY past the entering reference (seeded: \`reference+1\`; unseeded: the first observation is a non-terminal baseline and later polls are forced past it), so a still-fresh pre-call cache entry can neither be re-read nor accepted as terminal evidence. Once a genuine post-invocation capture arrives the floor relaxes to the inclusive monotonic device floor so a static screen still settles.
- **Require a meaningful quiet period** — \`settleHierarchyOnlyChange\` settles on a wall-clock quiet-period deadline (two poll intervals) rather than a count of equal comparisons, so a transient A that persists beyond one interval before a delayed B no longer pre-empts B. New FakeTimer regression covers A-holds-two-frames-then-B.
- **Preserve freshness failures unrelated to hierarchy age** — added a stable \`freshness.category\` discriminant; the post-refresh realignment promotes ONLY a \`cache_age\` failure, never a \`window_identity\`/attribution/incomplete-capture failure the hierarchy swap does not resolve. New test asserts a wrong-window verdict survives the swap.

No \`compareViewHierarchy\` bothFresh gate reintroduced; staleness robustness stays in the poll floor.

kaeawc:
## Systemic review — handoff for fix agents

Umbrella plan: #6333. This PR is the **reference for W6 (settle/confirm under device-clock skew)** and W7 (warning-scope on refresh).

- **W6:** the whole PR is one thesis — one clock domain end-to-end. Seed the `ObservePoll` monotonic floor **only from device-authored timestamps** (not host, #4/#1); gate **both** termination and predicate-`previous` state on the raised floor (#5/#9); require a post-invocation capture before any terminal verdict (#1); use a wall-clock quiet period, not count-of-equal (#3); exclude explicitly-stale observations as terminal evidence (#10). These belong in the shared `settleUntilStable`/`RealSettleObserve` primitive so Rotate (6301), stream-reopen (6310), and readiness-upgrade (6280) stop re-rolling them. Age via `observationFreshness.hostAgeBasisMs` (#5377/#6284).
- **W7:** #2/#7 — a hierarchy-only refresh promotes only `cache_age` and only when `hasCompleteLiveHierarchy`; never clear `window_identity`/incomplete-capture warnings it didn't reconcile.
- **Status:** all threads resolved. **Safe to land.**

PR #6315, opened 2026-09-07
ci: route kaeawc's non-simulator macOS jobs to a self-hosted runner

kaeawc:
## What

Routes the **non-simulator** macOS jobs to a self-hosted Apple Silicon runner (label `automobile-mac`) for **kaeawc-authored runs**, falling back to the GitHub-hosted `macos-26` runner for every other author.

Stacked on [#6312](https://github.com/kaeawc/auto-mobile/pull/6312) (base branch = `work/macos-github-actions-capacity-5a738a`); merge that first.

### Jobs routed

`pull_request.yml` (author = `github.event.pull_request.user.login`):
- `swiftlint`
- `swift-code-coverage`
- `ios-swift-packages` (matrix)
- `ios-spm-root-package-build`

`merge.yml` (author = `github.actor`, since it triggers on `push`):
- `swift-code-coverage` — the other macOS jobs (`ios-swift-build`, `ios-swift-test`, `ios-xcodegen`) are `if: ${{ false }}` / disabled, so wiring them would be dead config.

### Routing mechanism

```yaml
runs-on: ${{ github.event.pull_request.user.login == 'kaeawc'
  && fromJSON('["self-hosted", "automobile-mac"]') || 'macos-26' }}
```

`fromJSON` supplies the label array on the kaeawc branch; the hosted fallback is a plain string (or `matrix.config.runner` for matrix jobs).

### Xcode handling

Jobs with a pinned `maxim-lobanov/setup-xcode` step (`ios-swift-packages`, `ios-spm-root-package-build`) gate that step with `if: runner.environment == 'github-hosted'`, so the self-hosted Mac uses its **ambient** Xcode instead of trying to select a version it doesn't have. The self-hosted runner must therefore carry an Xcode whose Swift toolchain is **>= 6.3** (the root SPM manifest's floor).

## Excluded (simulator jobs — stay on GitHub-hosted)

`ios-xcode-build`, `ios-playground-tests`, `ios-xctest-runner-simulator-tests`, `ios-device-webrtc` — all pull in `ensure-ios-simulator-runtime` or boot a simulator, which is stateful/flaky on a laptop.

## Security

The runner is registered on a **public** repo. Alongside this, the repo's fork-PR approval policy was hardened to `all_external_contributors`, so every fork/outside PR needs manual approval before any workflow runs. The `runs-on` author guard is defense-in-depth: a non-kaeawc PR never lands on the self-hosted Mac at all.

## Validation

- `actionlint` clean on the added `runs-on` expressions and `runner.environment` gates (pre-existing `background:`/`wait:`/`if: false` warnings are unrelated).
- Both workflows parse as valid YAML.
- `test/bats/pr-gate-wiring.bats` + `workflow-artifact-name-uniqueness.bats` pass (24 tests); `validate_workflow_assertion_triggers.sh` passes. No test asserts job `runs-on`, so nothing is coupled to this change.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -1503,7 +1503,12 @@ jobs:
 
   swift-code-coverage:
     name: "Swift Code Coverage"
-    runs-on: macos-26
+    # Self-hosted routing: merge runs triggered by kaeawc (push/merge to main) run
+    # on the self-hosted Apple Silicon runner (label automobile-mac); anyone else's
+    # merge falls back to the GitHub-hosted macos-26 runner. This workflow triggers
+    # on push, so github.actor (not pull_request.user) identifies the actor. No
+    # setup-xcode step, so it uses the runner's ambient Xcode.
+    runs-on: ${{ github.actor == 'kaeawc' && fromJSON('["self-hosted", "automobile-mac"]') || 'macos-26' }}
     steps:
       - name: "Git Checkout"
         uses: actions/checkout@v6
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1648,13 +1648,10 @@ jobs:
   swiftlint:
     timeout-minutes: 20
     name: "SwiftLint"
-    # Self-hosted routing: PRs authored by kaeawc run on the self-hosted Apple
-    # Silicon runner (label automobile-mac); every other author falls back to the
-    # GitHub-hosted macos-26 runner. runs-on wants a label array for self-hosted
-    # and a plain string for hosted, so fromJSON supplies the array on the kaeawc
-    # branch. This workflow is pull_request-only, so pull_request.user.login is
-    # always populated. Non-simulator job (no ensure-ios-simulator-runtime).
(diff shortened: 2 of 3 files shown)


kaeawc:
## Systemic review — handoff for fix agents

Umbrella plan: #6333. This PR is a **genuine one-off** (CI supply-chain), not part of the mobile-automation clusters.

- **Resolved (correct):** self-hosted routing removed from `pull_request.yml` entirely; only trusted post-merge `merge.yml` retains the `automobile-mac` route, and PR jobs pin `macos-26` + explicit `Select Xcode 26.5`. This closes the untrusted-PR-YAML-on-self-hosted-runner risk and the ambient-Xcode toolchain-floor drift.
- **Residual (open, follow-up — not a blocker):** F4 — `test/scripts/spmRootToolchainWorkflow.test.ts` rejects only the literal substring `self-hosted`; a job could target the runner directly via `runs-on: automobile-mac`, which contains no such substring. Tighten the guard test to also reject the bare label. No such config exists today, so this is a latent coverage gap.
- **Land-readiness:** **Safe to land**; F4 as a tracked follow-up.

Issue #6649, opened 2026-09-08
db: cached-statement hit re-reads PRAGMA schema_version on every query


plx:
Child of [#6379](https://github.com/kaeawc/auto-mobile/issues/6379).

### Context

`BunSqliteDialect#getStatement` (`src/db/bunSqliteDialect.ts:459-477`) is the lookup path used for every non-DDL query:

```ts
#getStatement(db: BunDatabase, sql: string): BunStatement {
  if (this.#statementCache.size === 0) {
    this.#observedSchemaVersion = this.#readSchemaVersion(db);
  }

  const cached = this.#statementCache.get(sql);
  if (cached) {
    this.#invalidateCacheIfSchemaVersionChanged(db);
    const current = this.#statementCache.get(sql);
    if (!current) {
      return this.#prepareAndCacheStatement(db, sql);
    }
    this.#statementCache.delete(sql);
    this.#statementCache.set(sql, current);
    return current;
  }

  return this.#prepareAndCacheStatement(db, sql);
}
```

`#invalidateCacheIfSchemaVersionChanged` (`:493-503`) always calls `#readSchemaVersion(db)` (`:505-517`), which does its own `db.prepare("PRAGMA schema_version")`, `.get()`, and `.finalize()` — a full extra statement compile/execute/finalize cycle — on every single cache hit.

The cache is already invalidated on the one path that actually changes schema: `#isSchemaChangingSql` (`:519-521`) detects `create`/`alter`/`drop` statements, and after executing one the query path calls `this.#clearStatementCache()` (`:369` for SELECT/RETURNING, `:384` for writes) before returning. All DDL on this connection goes through `executeQuery`, so schema changes made by this dialect are already reflected by the unconditional cache clear.

### What goes wrong

1. A warm daemon's dominant workload is repeated SELECTs against a small set of prepared statements (observe, nav, telemetry) — the cache-hit branch at `:464-472`.
2. Every one of those hits pays a second SQLite round trip (`PRAGMA schema_version`) purely to detect a schema change that, on this connection, can only originate from this same dialect's own DDL execution — which already clears the cache directly.
3. The check also has a gap in the opposite direction: a cache MISS while the cache is non-empty (`:476`, reached when `cached` is falsy) does not call `#invalidateCacheIfSchemaVersionChanged` at all, so the schema-version bookkeeping is inconsistently applied — the hit path pays a cost the miss path doesn't, while neither path is actually needed for correctness against this connection's own DDL.

### Impact

[...]

Issue #6379, opened 2026-09-08
epic(db): persistence, retention, and shared utility hygiene


plx:
Child of [#6367](https://github.com/kaeawc/auto-mobile/issues/6367).

## Context

SQLite/Kysely layer and shared utilities (`src/db/`, `src/utils/cache/`, `logger.ts`, retry primitives): unique-index violations, unbounded tables, per-query schema checks, lock fairness and long transactions, migration lifecycle on close, health-probe timeouts, and duplicated primitives.

Findings were verified against `main@02c1188db`; file and line references are current as of that commit.

## Children

**Severity high**
- [x] [#6491](https://github.com/kaeawc/auto-mobile/issues/6491) snapshot: concurrent captures run overlapping archive-limit evictions and over-evict

**Severity medium**
- [x] [#6492](https://github.com/kaeawc/auto-mobile/issues/6492) snapshot: on-disk settings.json/metadata.json are never read back; legacy scan only sees flat manifest.json
- [x] [#6493](https://github.com/kaeawc/auto-mobile/issues/6493) snapshot: no property test pins that resolved snapshot paths stay inside the snapshots base directory
- [x] [#6463](https://github.com/kaeawc/auto-mobile/issues/6463) db: linkUIElementsToEdge has no onConflict/delete-before-insert and can violate the edge_ui_elements unique index
- [x] [#6464](https://github.com/kaeawc/auto-mobile/issues/6464) db: crashes, anrs, tool_calls, and device_sessions grow without bound; the one pruning method is unreachable and its tool_calls filter never matches
- [x] [#6649](https://github.com/kaeawc/auto-mobile/issues/6649) db: cached-statement hit re-reads PRAGMA schema_version on every query
- [x] [#6650](https://github.com/kaeawc/auto-mobile/issues/6650) db: NavigationRetention.prune holds the daemon's single connection in one transaction for the whole pass
- [x] [#6651](https://github.com/kaeawc/auto-mobile/issues/6651) logging: checkAndRotateLog performs a stat syscall on every serialized log write instead of on an interval
- [x] [#6652](https://github.com/kaeawc/auto-mobile/issues/6652) db: provisionDevice operation store restarts provisioning for a 'running' row instead of reporting in-progress, and never prunes its table
- [x] [#6653](https://github.com/kaeawc/auto-mobile/issues/6653) cache: TTLCache.cleanup() is documented as automatic but never invoked, and size-capped eviction is O(n) per insert
[...]

PR #6763, opened 2026-09-09
perf(db): FIFO admission for the SQLite connection, no broadcast, bounded fairness (#6699)

### LINKED ISSUE #6699: db: BunSqliteDialect broadcasts every query completion to all waiters and can starve queries behind transaction bursts
Child of [#6379](https://github.com/kaeawc/auto-mobile/issues/6379).

### Context

`BunSqliteConnectionState` coordinates all leases over the daemon's singleton SQLite connection with three counters and one shared waiter array:

```ts
// src/db/bunSqliteDialect.ts:172-175
#transactionOwner: symbol | null = null;
#pendingTransactions = 0;
#activeQueries = 0;
#waiters: Array<() => void> = [];
```

Every query completion calls `#notifyWaiters()` (`src/db/bunSqliteDialect.ts:294-297`), and `beginTransaction()` calls it again after a reservation attempt (`:231-239`). `#notifyWaiters()` drains and resolves the entire array (`:581-587`), even though normally only one transaction can acquire the connection and most woken operations immediately park again.

The admission rule also gives every pending transaction priority over every non-transaction query:

```ts
// src/db/bunSqliteDialect.ts:554-564
while (
  (this.#transactionOwner !== null && this.#transactionOwner !== owner) ||
  (this.#transactionOwner === null && this.#pendingTransactions > 0)
) {
  this.#assertOpen();
  await this.#waitForStateChange();
}
```

The existing test at `test/db/database.integration.test.ts:352-422` intentionally pins that one pending transaction runs before a later query, but there is no bound on how many consecutive pending transactions may retain that priority.

### What goes wrong

1. With `M` parked queries and `N` pending transactions, each statement completion resolves all `M + N` waiter promises.
2. Those continuations all re-run their admission predicates in the same microtask turn. One eligible operation progresses; the rest allocate and enqueue fresh promises.
3. The pattern repeats after the next statement, producing a thundering herd whose scheduler work scales with the number of parked operations rather than the number that can proceed.
4. A non-transaction query also remains blocked as long as `#pendingTransactions > 0`. If transactions continue arriving before that counter reaches zero, the query has no fairness bound and can wait behind the entire burst even when it was queued before most of those transactions.

### Impact

The SQLite statements themselves are synchronous and usually short, so waking every waiter can become a material share of the work during a multi-device burst. More importantly, reads and telemetry writes can acquire unbounded queue latency behind a sustained stream of short transactions. This does not corrupt data, but it makes the daemon's single database connection a noisier and less predictable chokepoint.

### How to reproduce or observe

Extend the connection-state tests in `test/db/database.integration.test.ts` or `test/db/bunSqliteDialect.integration.test.ts` with controlled deferred queries:

1. Hold one ordinary query in `beforeQuery`.
2. Queue several transactions and several ordinary queries in a known order.
3. Release one operation at a time and record which queued operation enters `beforeQuery`.
4. Show that every pending transaction runs before the already-queued queries.

For the broadcast behavior, add an injected/test-only waiter scheduler or admission trace and assert that one statement completion resumes only the next eligible waiter. The current implementation resumes every parked waiter on each completion.

### Suggested direction

Replace the shared broadcast array with an explicit FIFO admission queue (or separate transaction/query queues with a documented fairness policy). Wake only the operation or cohort that can make progress. Preserve bounded transaction preference so a transaction cannot be starved by a constant read stream, but add the symmetric bound so a constant transaction stream cannot starve earlier queries.

Avoid a polling delay or larger SQLite `busy_timeout`: this contention is entirely in the JavaScript lease scheduler, before SQLite sees a statement.

### How to verify the fix

- A deterministic ordering test proves both operation classes make progress under a mixed sustained queue.
- An admission trace/counter proves one state transition does not resume every parked waiter.
- Existing transaction atomicity and "pending transaction before later query" coverage continues to pass with the documented bounded policy.

### Scope

- `src/db/bunSqliteDialect.ts`: `beginTransaction`, `executeQuery` cleanup, `#reserveTransaction`, `#enterQuery`, `#waitForStateChange`, `#notifyWaiters`
- `test/db/database.integration.test.ts`
- `test/db/bunSqliteDialect.integration.test.ts`

Related: [#6650](https://github.com/kaeawc/auto-mobile/issues/6650)

plx:
## Summary

`BunSqliteConnectionState` coordinated all leases over the daemon's singleton SQLite
connection with a shared `#waiters` array that `#notifyWaiters` **drained and resolved in
full on every statement completion**. With M parked queries and N pending transactions,
each completion woke all M+N — one progressed, the rest re-parked (a thundering herd
scaling with parked operations, not admittable ones). The admission rule also gave every
pending transaction priority over every query (`#enterQuery` waited while
`#pendingTransactions > 0`), so a sustained transaction stream could **starve** an
earlier-queued read with no fairness bound.

Replaces the broadcast array with an explicit **FIFO admission queue**; `#pump` is the
sole authority that sets `#transactionOwner`/`#activeQueries` on admission:
- `#pump` resolves **only** the waiters that can proceed now — one transaction
  (exclusively) or a run of consecutive head queries (concurrent reads) — and never wakes
  a waiter that would immediately re-park.
- Strict FIFO gives **symmetric bounded fairness**: a head transaction that can't run yet
  blocks what's behind it (never starved by a later query stream); a query queued ahead
  of a transaction admits first (never starved by a later transaction stream). The
  transaction owner's own statements still bypass the queue.
- On close, `#pump` drains the queue by rejecting each parked waiter with the
  closed-database error (preserves the `#2792` shutdown safety).

`#pendingTransactions` is removed (the queue subsumes it). Part of epic #6379.
**Stacked on #6649** (`work/issue-6649-cached-stmt-pragma`); retargets to `main` when it merges.

## Linked Issue

Closes #6699

## Validation

- [x] New tests (`test/db/bunSqliteDialect.integration.test.ts`) via `resumeCount`/`queuedWaiterCount` seams: one completion resumes only the eligible waiter(s) (old broadcast woke all); FIFO across mixed query/transaction waiters (`[q1, t2, q3]`); an earlier query isn't starved by a sustained later transaction stream; a transaction queued before a query still runs first (bounded preference preserved). These fail against the old global-transaction-priority behavior.
- [x] Existing coverage green: bunSqliteDialect + database integration (37), incl. the "pending transaction before later query" atomicity test. Broad `test/db` — 1062 pass, 0 fail.
- [x] `bun run format:check`, `bun run lint`, `bun run typecheck` — all clean.

## Follow-ups
- None.

### DIFF
--- a/src/db/bunSqliteDialect.ts
+++ b/src/db/bunSqliteDialect.ts
@@ -164,14 +164,34 @@ class BunSqliteDriver implements Driver {
   }
 }
 
+// One parked operation in the FIFO admission queue (#6699). `kind` decides the
+// admission rule: a "txn" needs exclusive access (no owner, no active queries)
+// and blocks everything behind it until it can run; a "query" runs as soon as no
+// transaction holds the connection, concurrently with other queued queries.
+interface AdmissionWaiter {
+  kind: "txn" | "query";
+  owner: symbol;
+  resolve: () => void;
+  reject: (error: Error) => void;
+}
+
 export class BunSqliteConnectionState {
   readonly #databaseSource: BunDatabase | (() => BunDatabase);
   readonly #beforeQuery?: () => Promise<void>;
   #db: BunDatabase | null = null;
   #transactionOwner: symbol | null = null;
-  #pendingTransactions = 0;
   #activeQueries = 0;
-  #waiters: Array<() => void> = [];
+  // FIFO admission queue over the singleton connection (#6699). Replaces the old
+  // broadcast waiter array: each state change wakes only the next eligible
+  // waiter — one transaction (exclusively) or a run of consecutive queries
+  // (concurrent reads) — via #pump, never the whole queue. Admission is strictly
+  // first-come, so a sustained transaction stream cannot starve an earlier-queued
+  // query (it queues ahead of later transactions), and a sustained query stream
+  // cannot starve an earlier-queued transaction (later queries queue behind it).
+  #queue: Array<AdmissionWaiter> = [];
+  // Test seam: cumulative waiter resolutions. Pins that one completion resumes
+  // only the eligible waiter(s), not every parked operation (the old herd).
+  #resumeCount = 0;
   #statementCache = new Map<string, BunStatement>();
   #closed = false;
   readonly #maxRetryAttempts: number;
@@ -213,7 +233,7 @@ export class BunSqliteConnectionState {
       this.#closed ||
       !this.#db ||
       this.#transactionOwner !== null ||
-      this.#pendingTransactions > 0 ||
+      this.#queue.length > 0 ||
       this.#activeQueries > 0
     ) {
       return;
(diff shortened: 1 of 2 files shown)


kaeawc (approved):
Approved: exact head reviewed, all checks green, no unresolved review feedback.

Issue #6372, opened 2026-09-08
epic(ios): simctl, xcrun and xcodebuild assumptions and CtrlProxy runner lifecycle


plx:
Child of [#6367](https://github.com/kaeawc/auto-mobile/issues/6367).

## Context

How the server models CoreSimulator and the XCTest-hosted CtrlProxy runner versus their real semantics: runner process ownership and abort signals, boot/shutdown state verification, readiness probes and availability caching, builder and manager lifecycle, xctestrun handling, privacy grants, and free-text parsing of simctl output. Most items are in `src/utils/ios-cmdline-tools/`, `src/utils/IOSCtrlProxyManager.ts`, and `src/utils/IOSCtrlProxyBuilder.ts`.

Findings were verified against `main@02c1188db`; file and line references are current as of that commit.

## Children

**Severity high**
- [x] [#6496](https://github.com/kaeawc/auto-mobile/issues/6496) accessibility(ios): VoiceOverToggle reports a disable as applied when the confirmation probe never succeeded
- [x] [#6410](https://github.com/kaeawc/auto-mobile/issues/6410) ctrlproxy(ios): resident xcodebuild runner inherits the per-request abort signal and is SIGTERMed on cancel
- [x] [#6411](https://github.com/kaeawc/auto-mobile/issues/6411) simctl: boot retry treats an unreadable device state as a failed boot and re-boots into an in-flight shutdown
- [x] [#6415](https://github.com/kaeawc/auto-mobile/issues/6415) ctrlproxy(ios): runner readiness gate accepts any /health responder — isRunning/waitForHealthEndpoint skip the device-identity check
- [x] [#6416](https://github.com/kaeawc/auto-mobile/issues/6416) ctrlproxy(ios): setup() answers "CtrlProxy was already running" from a 1-hour positive availability cache the iOS readiness path never reopens
- [x] [#6417](https://github.com/kaeawc/auto-mobile/issues/6417) ctrlproxy(ios): concurrent device setups share one unsynchronized builder that rm -rf's the derived-data root

**Severity medium**
- [x] [#6412](https://github.com/kaeawc/auto-mobile/issues/6412) simctl: the two boot-completion paths disagree on isAvailable, so one resolves the boot the other rejects
- [x] [#6418](https://github.com/kaeawc/auto-mobile/issues/6418) ctrlproxy(ios): xctestrun runner-environment injection only understands the FormatVersion 1 layout
[...]

PR #6787, opened 2026-09-09
fix(ios): escalate straight to SIGKILL on shutdown-time force-stop (#6578)

### LINKED ISSUE #6578: ctrlproxy(ios): shutdown's 1.2s/250ms budget cannot cover terminateProcessTree's TERM+KILL ladder
Child of [#6372](https://github.com/kaeawc/auto-mobile/issues/6372).

### Context

Daemon shutdown gives the iOS proxies one bounded stage: `cleanupDaemonChildProcesses` wraps `IOSCtrlProxyManager.shutdownAll()` in `settleWithin(..., 1_500)` (`src/daemon/childProcessCleanup.ts:17`, `:93`). Inside it, `shutdownAll` (`src/utils/IOSCtrlProxyManager.ts:450-461`) runs `stopWithinShutdownDeadline` per instance (`:463-493`), which races `stop()` against `SHUTDOWN_STOP_TIMEOUT_MS = 1_200` and then `forceStopForShutdown()` against `SHUTDOWN_FORCE_STOP_TIMEOUT_MS = 250` (`:38-39`, `:496-511`), and finally clears the instance map (`:455`).

Both kill paths funnel into `terminateProcessTree`, which is a two-phase ladder:

```ts
async terminateProcessTree(pid: number, deadline?: number): Promise<void> {
  const descendants = await this.findDescendantProcessIds(pid, deadline);
  const targets = [...descendants].reverse().concat(pid);
  await this.signalGroup(pid, "TERM", deadline);
  await this.signalPids(targets, "TERM", deadline);
  if (await this.waitForExit([pid, ...descendants], deadline)) {
    return;
  }
  await this.signalGroup(pid, "KILL", deadline);
  await this.signalPids(targets, "KILL", deadline);
  if (!(await this.waitForExit([pid, ...descendants], deadline))) {
    throw new Error(`CtrlProxy process tree rooted at PID ${pid} remained alive after SIGKILL`);
  }
}
```
(`src/utils/ios/IOSCtrlProxyProcessClient.ts:196-209`)

`waitForExit` loops `releaseAttempts = 4` times sleeping `releaseGraceMs = 250` between `kill -0` probes (`:45-46`, `:331-341`), so each phase can consume ~1s of sleep plus one exec per pid per attempt. Neither call site passes a `deadline`: `src/utils/IOSCtrlProxyManager.ts:1272` (`stop()`) and `:550` (`forceStopForShutdown`) both call it with a bare pid, so `remainingTimeoutMs(undefined)` leaves every sleep and every exec at full length.

### What goes wrong

1. A local simulator `stop()` reaches `terminateProcessTree` at `:1272`. If the runner tree exits on the first `SIGTERM`, `waitForExit` short-circuits on its first `haveExited` check and the whole call is fast — the common case.
2. If the tree does **not** exit promptly (an `xcodebuild test` runner wedged mid-teardown), the TERM phase alone sleeps up to 4 x 250 ms before escalation, and `stop()` blows past the 1_200 ms race in `stopWithinShutdownDeadline` (`:479-482`).
3. The timeout branch starts `forceStopForShutdown` (`:513-554`), which calls `terminateProcessTree(runnerPid)` again at `:550` — a *fresh* ladder that starts over at TERM. Its budget is 250 ms (`:39`, `:503-505`), less than a single grace sleep, so the `SIGKILL` escalation at `IOSCtrlProxyProcessClient.ts:204-206` cannot be reached before the race resolves.
4. `shutdownAll` clears the map (`:455`) and resolves. The original `stop()` promise is only raced — never awaited and never cancelled — and the force-stop promise is likewise abandoned; both keep executing detached.
5. The daemon proceeds to exit while the tree is still being signalled, leaving exactly the orphaned `xcodebuild`/runner processes that `startOrphanRunnerReapOnStartup` (`:560`) exists to sweep on the next launch.

### Impact

In a long multi-device session that ends with a wedged iOS runner, daemon shutdown reports success while a runner tree survives. The orphan holds the simulator's CtrlProxy port and its in-simulator runner, so the next daemon start pays the startup reap (`STARTUP_ORPHAN_RUNNER_REAP_DEADLINE_MS = 5_000`, `:34`) and, if the reap does not finish inside its deadline, the first `startDevice` for that simulator would contend with a stale runner it can re-adopt. The healthy path is unaffected — this only bites when the runner ignores `SIGTERM`, which is precisely the case the force path exists for.

### How to reproduce or observe

Deterministic unit test alongside the existing `describe("shutdownAll")` block in `test/utils/IOSCtrlProxyManager.test.ts:288-333` (which already drives `shutdownAll(timer)` with `FakeTimer` and `advanceTime`):

1. Build a manager with `IOSCtrlProxyManager.createForTestingWithDeps(..., processClient)` (`src/utils/IOSCtrlProxyManager.ts:408-419`), injecting an `IOSCtrlProxyProcessClient` over a fake `HostCommandExecutor` whose `kill -0` always reports the pid alive and whose `ps` reports one descendant.
2. Track the tracked runner pid, then call `shutdownAll(fakeTimer)` and advance the fake clock by 1_500 ms (the daemon's own stage budget from `childProcessCleanup.ts:17`).
3. Assert on the recorded exec calls: with today's code the fake sees `kill -TERM` but never `kill -KILL`, and `shutdownAll` has already resolved.

`test/utils/ios/IOSCtrlProxyProcessClient.test.ts` is the closest pattern for driving the ladder itself with a fake executor plus `FakeTimer`.

### Suggested direction

One or both, keeping the existing budgets intact:

- Thread a deadline through both call sites — `terminateProcessTree` already accepts one (`IOSCtrlProxyProcessClient.ts:196`) and clamps its sleeps and execs against it (`:331-341`) — so a shutdown-time kill degrades predictably instead of overrunning.
- Make the force path an actual force: have `forceStopForShutdown` (`:549-553`) go straight to `signalGroup`/`signalPids` with `"KILL"` (or a `terminateProcessTree(pid, deadline, { skipGraceful: true })` variant) rather than replaying a TERM grace ladder that has already demonstrably timed out.

Inflating `SHUTDOWN_STOP_TIMEOUT_MS` is not the fix — the comment at `:35-37` records that this stage shares the 10s force-exit budget with capture cleanup, and a larger window would starve the other owners.

### How to verify the fix

The unit test above asserts the invariant: within the daemon's own 1_500 ms shutdown stage, a runner tree that ignores `SIGTERM` receives `SIGKILL` before `shutdownAll` resolves. Add a companion case asserting that a promptly-exiting tree still resolves without any `KILL`, so the fix does not turn every shutdown into a hard kill.

### Scope

- `src/utils/IOSCtrlProxyManager.ts` — `SHUTDOWN_STOP_TIMEOUT_MS`/`SHUTDOWN_FORCE_STOP_TIMEOUT_MS` (`:38-39`), `shutdownAll` (`:450`), `stopWithinShutdownDeadline` (`:463`), `forceStopWithinShutdownDeadline` (`:496`), `forceStopForShutdown` (`:513`), `stop()` (`:1272`)
- `src/utils/ios/IOSCtrlProxyProcessClient.ts` — `terminateProcessTree` (`:196`), `waitForExit` (`:331`)
- `src/daemon/childProcessCleanup.ts` — shutdown stage budget (`:17`, `:93`)

Related: [#5818](https://github.com/kaeawc/auto-mobile/issues/5818), [#6260](https://github.com/kaeawc/auto-mobile/issues/6260)

plx:
## Summary

The force-stop path that runs after stop()'s graceful attempt times out replayed a fresh, unbounded SIGTERM ladder before ever reaching SIGKILL, so a wedged runner tree could outlive the daemon's bounded shutdown stage entirely and be orphaned. terminateProcessTree in IOSCtrlProxyProcessClient now accepts a skipGraceful option that jumps straight to SIGKILL, and forceStopForShutdown uses it since it only runs after a prior timeout. Deadlines are threaded through both stop() and forceStopForShutdown from the shared shutdown Timer (computed at each stage's start) so terminateProcessTree's sleeps and execs clamp to the remaining budget instead of running unbounded.

Part of epic #6372.

Stacked on `work/issue-6416-ios-availability-cache-reopen`; GitHub retargets to `main` once it merges.

## Linked Issue

Closes #6578

## Validation

Two FakeTimer tests in the shutdownAll block: a SIGTERM-ignoring runner now receives SIGKILL within the shutdown stage (red before — only SIGTERM was seen); a promptly-exiting runner still resolves without ever reaching force-stop/SIGKILL. Targeted IOSCtrlProxyManager 111 + IOSCtrlProxyProcessClient 10 + broader test/utils 3107 pass.

- [x] `bun run format:check`, `bun run lint`, `bun run typecheck` — clean

### DIFF
--- a/src/utils/IOSCtrlProxyManager.ts
+++ b/src/utils/IOSCtrlProxyManager.ts
@@ -462,7 +462,7 @@ export class IOSCtrlProxyManager implements CtrlProxyIosManager {
     // Threaded into stop()'s own terminateProcessTree call so its sleeps/execs
     // degrade predictably against this stage's budget instead of running
     // unbounded (#6578).
-    const stopDeadline = timer.now() + SHUTDOWN_STOP_TIMEOUT_MS;
+    const stopDeadline = instance.timer.now() + SHUTDOWN_STOP_TIMEOUT_MS;
     const settled = instance.stop(stopDeadline).then(
       () => null,
       (error) => error,
@@ -498,7 +498,7 @@ export class IOSCtrlProxyManager implements CtrlProxyIosManager {
     let timeout: NodeJS.Timeout | undefined;
     // Computed here (not at the start of the overall shutdown stage) since
     // this stage's budget only starts once stop()'s own attempt has timed out.
-    const forceStopDeadline = timer.now() + SHUTDOWN_FORCE_STOP_TIMEOUT_MS;
+    const forceStopDeadline = instance.timer.now() + SHUTDOWN_FORCE_STOP_TIMEOUT_MS;
     const deadline = new Promise<void>((resolve) => {
       timeout = timer.setTimeout(resolve, SHUTDOWN_FORCE_STOP_TIMEOUT_MS);
     });
--- a/src/utils/IOSCtrlProxyManager.ts
+++ b/src/utils/IOSCtrlProxyManager.ts
@@ -304,7 +304,7 @@ export class IOSCtrlProxyManager implements CtrlProxyIosManager {
     this.builder = builder || IOSCtrlProxyBuilder.getInstance();
     this.processExecutor = processExecutor;
     this.xcodebuild = xcodebuild;
-    this.processClient = processClient ?? new IOSCtrlProxyProcessClient();
+    this.processClient = processClient ?? new IOSCtrlProxyProcessClient(processExecutor, timer);
     this.signingManager = signingManager;
     this.deviceAppManager = deviceAppManager;
     this.hostPortAvailabilityChecker = hostPortAvailabilityChecker;
--- a/src/utils/IOSCtrlProxyManager.ts
+++ b/src/utils/IOSCtrlProxyManager.ts
@@ -458,7 +458,7 @@ export class IOSCtrlProxyManager implements CtrlProxyIosManager {
     timer: Timer,
   ): Promise<unknown | null> {
     let timeout: NodeJS.Timeout | undefined;
-    let forceStopStarted = false;
+    let forceStop: Promise<void> | undefined;
     // Threaded into stop()'s own terminateProcessTree call so its sleeps/execs
     // degrade predictably against this stage's budget instead of running
     // unbounded (#6578).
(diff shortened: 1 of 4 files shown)


plx:
Superseded by #6808 ("fix(ios): force runner termination within shutdown budget"), which the maintainer merged to `main` and which closed the underlying issue #6578. Closing this redundant PR from the epic #6372 burndown; the fix is already on `main`.

PR #6804, opened 2026-09-10
fix: harden device provisioning readiness

kaeawc:
## Summary

Booted `provisionDevice` results expose `sessionUuid` alongside the equal `sessionId` compatibility alias. Both direct MCP connections and the daemon proxy bind these acquired sessions; the proxy starts ownership heartbeats even when optional resource configuration returns an error with a retained session. Replay storage continues to use only `sessionId`.

Android runner setup retries explicit missing package/settings service failures within the original absolute deadline. Manager status reads preserve these transient failures, including when downloads are disabled. Permanent setup failures remain terminal. The existing retry executor provides cancellation-aware backoff, and status reads finish before another attempt begins.

Optional framework inspection stays diagnostic: synchronous throws and rejected promises cannot prevent readiness. Inspection has its own bounded lifetime, aborts when readiness settles, and observes bounded cleanup. Logs include redacted diagnostics.

## Validation

- 380 tests passed across provisioning, direct session binding, daemon proxy, Android manager, framework failure classification, and runner readiness suites.
- Regression coverage includes retained-session errors, Android/iOS aliases and persistence, replay/rebound, disabled boot/downloads, pending/rejecting/throwing probes, transient recovery, absolute deadlines, cancellation, and overlapping status reads.
- `bun run lint`: no new violations; all 13 boundary checks passed.
- `bun run build`, `bun run format:check`, and `git diff --check` passed.
- `bun run typecheck`: no new errors against the existing baseline.

The narrow framework classifier uses the standard library and existing error-message utility; no dependency was added.

### DIFF
--- a/src/server/deviceTools.ts
+++ b/src/server/deviceTools.ts
@@ -3247,10 +3247,15 @@ function createProvisionDeviceResponse(result: Record<string, unknown>) {
   const device = result.device as { name: string; platform: string };
   const resources = result.resources as DeviceResourceConfigurationResult | undefined;
   const resourceFailure = resources?.success === false;
+  const sessionId = typeof result.sessionId === "string" ? result.sessionId : undefined;
   return {
     ...createJSONToolResponse({
       message: `${device.platform} '${device.name}' provisioned (${result.lifecycleState})${resourceFailure ? "; requested resource configuration was not fully applied" : ""}`,
       ...result,
+      // `sessionId` remains the persisted daemon-internal handle for replay and
+      // recovery. Public preparation APIs use `sessionUuid`; retain sessionId
+      // as a compatibility alias for provisionDevice's existing consumers.
+      ...(sessionId ? { sessionUuid: sessionId, sessionId } : {}),
       ...(resources ? { success: resources.success } : {}),
     }),
     ...(resourceFailure ? { isError: true } : {}),
--- a/src/utils/RunnerReadinessService.ts
+++ b/src/utils/RunnerReadinessService.ts
@@ -142,6 +142,12 @@ export interface AndroidRunnerConnectDiagnostic {
   primaryUserStartState?: string;
 }
 
+export interface AndroidFrameworkReadinessResult {
+  ready: boolean;
+  unavailableResources?: string[];
+  diagnostic?: string;
+}
+
 interface SystemUiAnrReadinessClient {
   getAccessibilityHierarchy: NonNullable<ReadinessClient["getAccessibilityHierarchy"]>;
   requestTapCoordinates: NonNullable<ReadinessClient["requestTapCoordinates"]>;
@@ -149,6 +155,11 @@ interface SystemUiAnrReadinessClient {
 
 export interface RunnerReadinessDependencies {
   timer: Timer;
+  getAndroidFrameworkReadiness?(
+    device: BootedDevice,
+    signal: AbortSignal,
+    timeoutMs: number,
+  ): Promise<AndroidFrameworkReadinessResult>;
   getAndroidManager(device: BootedDevice): ReadinessAndroidManager;
   getAndroidClient(device: BootedDevice): ReadinessClient;
   getIosManager(device: BootedDevice): ReadinessIosManager;
@@ -234,6 +245,7 @@ export class RunnerReadinessService {
   }
 
   private async ensureAndroidReady(context: ReadinessAttemptContext): Promise<void> {
+    this.inspectAndroidFrameworkBestEffort(context);
     const manager = this.dependencies.getAndroidManager(context.device);
     const client = this.dependencies.getAndroidClient(context.device);
     if (context.skipCtrlProxyDownload) {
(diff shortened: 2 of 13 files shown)


src/server/deviceTools.ts
```diff
@@ -3247,10 +3247,15 @@ function createProvisionDeviceResponse(result: Record<string, unknown>) {
...
   const resources = result.resources as DeviceResourceConfigurationResult | undefined;
   const resourceFailure = resources?.success === false;
+  const sessionId = typeof result.sessionId === "string" ? result.sessionId : undefined;
   return {
     ...createJSONToolResponse({
       message: `${device.platform} '${device.name}' provisioned (${result.lifecycleState})${resourceFailure ? "; requested resource configuration was not fully applied" : ""}`,
       ...result,
+      // `sessionId` remains the persisted daemon-internal handle for replay and
+      // recovery. Public preparation APIs use `sessionUuid`; retain sessionId
+      // as a compatibility alias for provisionDevice's existing consumers.
```
kaeawc, line 3257: `provisionDevice` now emits `sessionUuid`, but the result-minted session is not treated as an acquisition. `DEVICE_SESSION_ACQUISITION_TOOLS` still contains only `getAndroid`, `getApple`, and `startDevice`, so the direct MCP path skips connection binding and the daemon proxy skips ownership-heartbeat establishment. The returned session can therefore be reaped before the caller uses it.

Reproduction:

```bash
bun -e 'import {isDeviceSessionAcquisitionTool} from "./src/server/deviceSessionResult"; console.log(isDeviceSessionAcquisitionTool("provisionDevice"))'
# false
```

Please add `provisionDevice` to the shared acquisition set and cover both direct-server binding and daemon-proxy heartbeat establishment. This should block merge.

src/utils/RunnerReadinessService.ts
```diff
@@ -268,6 +280,38 @@ export class RunnerReadinessService {
...
+  private inspectAndroidFrameworkBestEffort(context: ReadinessAttemptContext): void {
+    const probe = this.dependencies.getAndroidFrameworkReadiness;
+    if (!probe) {
+      return;
+    }
+    const remainingMs = Math.max(0, context.totalDeadlineMs - this.dependencies.timer.now());
+    if (remainingMs <= 0) {
+      return;
+    }
+    const inspectionSignal = context.signal ?? new AbortController().signal;
```
kaeawc: This fire-and-forget probe can outlive a completed readiness request and stack with subsequent calls. When the caller supplied no signal, the new `AbortController` is never aborted; even with a caller signal, successful request completion does not abort it. Each production probe launches two ADB commands, so repeated warm readiness calls can leave overlapping package/settings processes that race teardown or recovery.

I reproduced this with two completed `ensureReady` calls whose injected probes remained pending: two probes were active and both signals were still un-aborted.

Please give diagnostic work an owned lifetime: either await a strictly bounded inspection at a point where its result cannot gate readiness, or create a child controller that is aborted when `ensureAndroidReady` settles and await/observe process settlement. Add a repeated-call test proving no probe remains active after readiness returns. This should block merge.

src/utils/RunnerReadinessService.ts
```diff
@@ -234,6 +245,7 @@ export class RunnerReadinessService {
   }
 
   private async ensureAndroidReady(context: ReadinessAttemptContext): Promise<void> {
```
kaeawc, line 256: The diagnostic probe does not delay or retry the required framework-dependent work, so the reported boot failure remains. Immediately after launching this detached inspection, `ensureCompatibleVersion`, `isInstalled`, `isEnabled`, and `setup` run once. `AndroidCtrlProxyManager.isInstalled`/`isEnabled` turn missing `package`/`settings` services into `false`, and setup/install failures become a terminal `RunnerReadinessError`; none of these paths retries when Android services appear later.

The new permanent-inspection test passes only because its fake manager reports compatibility, installation, and enablement as ready despite the probe claiming both services are absent. It does not model the observed failure.

Please retry the *required* CtrlProxy setup sequence when its failure is classified as transient framework boot readiness, under the existing absolute deadline and cancellation signal. Keep optional inspection non-gating, but add deterministic tests where manager operations fail with `Can't find service: package/settings` and later succeed, plus deadline and cancellation cases. This should block merge.

test/utils/RunnerReadinessService.test.ts
```diff
@@ -294,6 +301,52 @@ function appScreenMimickingEnglishSystemUiAnr(): ViewHierarchyResult {
 }
 
 describe("RunnerReadinessService", () => {
+  test("does not let Android framework inspection block runner setup", async () => {
+    const inspectionStarted = Promise.withResolvers<void>();
+    const androidClient = new FakeReadinessClient();
+    androidClient.connected = false;
+    const { service, androidManager } = createService({
+      androidClient,
+      getAndroidFrameworkReadiness: async () => {
```
kaeawc, line 311: This test does not prove that inspection is non-blocking because the injected probe resolves immediately. An implementation that `await`s the probe still passes unchanged. The sibling “permanent failure” test also resolves `{ ready: false }`, so the rejection handler is not exercised.

Use a pending probe and assert `ensureReady` completes before releasing it, then add a rejecting probe and assert required readiness still succeeds. That will discriminate detached/non-gating behavior from accidentally awaited behavior (and should be paired with the lifecycle test from the implementation comment proving no work remains after return). This should block merge because the central behavior claimed by the PR is currently untested.

test/server/deviceTools.provisionDevice.test.ts
```diff
@@ -706,10 +706,51 @@ describe("provisionDevice handler", () => {
...
+      const response = JSON.parse(
+        (
+          (await ToolRegistry.getTool("provisionDevice")!.handler({
+            ...provisionTestArgs(platform, `operation-public-session-${platform}`),
+            readiness: "none",
+          })) as any
+        ).content[0].text,
+      );
+
+      expect(response.sessionUuid).toEqual(expect.any(String));
```
kaeawc, line 742: The requested compatibility contract is not asserted: this only proves `sessionUuid` is some string. It would pass if `sessionUuid` and retained `sessionId` referred to different sessions, and existing replay/rebound tests still assert only `sessionId`.

Please assert exact public key presence and `sessionUuid === sessionId` for fresh Android/iOS results, completed-operation replay, and daemon rebound/recovery. Also assert the persisted operation result remains internal (`sessionId` only, no `sessionUuid`) so response shaping cannot accidentally change replay storage semantics. The resource-failure response should receive the same alias coverage. This is required contract coverage and should block merge.

PR #7036, opened 2026-09-14
fix(observe): update the terminal capture's cache entry in place and make settle polls side-effect-free

### LINKED ISSUE #6932: Embedded observation settle poll should be a side-effect-free hierarchy read
Follow-up from the Codex review of [#6890](https://github.com/kaeawc/auto-mobile/pull/6890). After several review rounds on the same mechanism the PR is being merged as-is (anti-thrash decision) and the remaining findings are tracked here so they can be addressed structurally rather than one fence at a time.

### `src/server/embeddedObservationSettle.ts:81`
**Fence recomposition setup with the settle deadline**

When Android recomposition tracking was not configured before this gate—for example, after CtrlProxy disconnects—each poll enters `HierarchyCollector.collect` and calls `configureRecompositionTracking(true)` before reading the hierarchy. That reaches `CtrlProxyHierarchy.setRecompositionTrackingEnabled`, whose bare `ensureConnected(perf)` can consume the five-second handshake budget without observing this combined signal, so the advertised one-second gate can still block for several seconds. Fresh evidence beyond the earlier `getLatestHierarchy` handshake fix is this preceding configuration path; either forward the deadline into it or skip recomposition setup for settle polls.

### `src/server/embeddedObservationSettle.ts:78`
**Skip recomposition accounting in hidden settle polls**

On Compose hierarchies containing recomposition annotations, every call to this full `RealObserveScreen` settle delegate runs `RecompositionTracker.processObservation`, which advances the singleton baseline, inserts `recomposition_metrics` rows, and emits telemetry for each hidden poll. The adopted observation's `recompositionSummary.sinceLastObservation` consequently measures only since the immediately preceding hidden poll—often reporting zero—while persisted telemetry is duplicated, corrupting performance results. Add an intermediate-poll option that skips recomposition processing and process the adopted terminal capture once.

### `src/server/embeddedObservationSettle.ts:97`
**Refresh screenshot evidence after adopting a later capture**

When automatic post-action screenshots or accessibility auditing are enabled, the handler captures its terminal screenshot before this after-call gate runs, while settle polls explicitly skip screenshots. If these lines adopt a later settled hierarchy, the response and session hierarchy cache describe that later frame but the latest-screenshot resource still exposes the handler's earlier, potentially half-inflated frame, so visual evidence and element bounds can disagree. Recapture once for an adopted terminal observation when screenshot evidence was requested, or invalidate the older screenshot.

### LINKED ISSUE #7020: Follow-up (#6999 round 6): <sub><sub>![P2 Badge](https://img.shields.io/badge/P2-yellow?style=flat)</sub></sub>  Update the terminal capture's existing cache entry
Round-6 review finding on the observe settle-poll change merged via #6999 (issue #6932). Rounds 1-5 cover admissible-evidence gating, deferred metric persistence, abort checks around the handshake, the generation fence, deferred-write ordering, and host-clock recency. Deferred here per the release cut to stop review churn.

Thread quoted verbatim from [#6999](https://github.com/kaeawc/auto-mobile/pull/6999):

### `src/features/observe/ObserveScreen.ts:1030`

> **<sub><sub>P2 Badge</sub></sub>  Update the terminal capture's existing cache entry**
> 
> When observation collection takes at least one host-clock tick, `execute()` has already cached this same object at completion, while the deferred write receives its earlier poll-start timestamp; this condition therefore mistakes the capture's own entry for a newer capture and skips reserialization. Memory appears updated because it shares the mutated object, but disk still lacks `recompositionSummary` and per-node metrics, so a daemon restart within the cache TTL reloads incomplete data. Fresh evidence beyond the earlier persistence comment is that the newly added host-time recency guard now rejects the same capture's update; compare capture identity separately and update its existing entry while still rejecting genuinely newer captures.
> 
> Useful? React with 👍 / 👎.

kaeawc:
Two observe-pipeline fixes under one rule: observation identity `(deviceId, observationId, hostCachedAt, generation)` keys every cache write, and settle polls are read-only.

**[#7020](https://github.com/kaeawc/auto-mobile/issues/7020)**: `ObserveScreen.cacheObserveResult`'s deferred write compared its poll-start `cachedAt` against the store's newest entry, which, once collection spanned a host-clock tick, was the same capture's own completion-time entry written by `execute()`; the guard read its own capture as "a newer one exists" and dropped the deferred write, and even when it did not, the write keyed by the earlier timestamp created a second entry that could never become newest. Fixed by minting a stable `observationId` per capture (through the injected `IdGenerator`) and having `FileSystemObserveCacheStore.put()` update an entry with a matching `observationId` in place (same key, same file, unchanged recency); the skip guard fires only for a genuinely different, strictly newer observation. Disk-cache recency now reads the timestamp embedded in the filename rather than mtime, since in-place updates rewrite mtime.

**[#6932](https://github.com/kaeawc/auto-mobile/issues/6932)**: the embedded settle poll still called `configureRecompositionTracking(true)` (a live CtrlProxy handshake that can burn the settle budget) and wrote directly to the observe cache on every intermediate poll. `HierarchyCollector.collect` gains a `readOnly` flag derived from the existing `skipRecompositionTracking` option, which now also gates `execute()`'s own cache write; intermediate polls touch neither tracking configuration, telemetry, nor the cache, and the deferred write from `ObservePoll.finalize()` is the only write for a settle-gated capture. The screenshot-refresh item on #6932 was out of scope.

Overlap note: [#7000](https://github.com/kaeawc/auto-mobile/pull/7000) also mints `observationId` in `createBaseResult` and makes it required; this branch adds it as optional. #7000 merges first and this branch reconciles onto it before merge.

Tests: stable id per capture; skip versus update-in-place; read-only versus ordinary execute side effects; disk-backed update-in-place across a store reload. Observe, settle and lint suites: 3037 pass.

Closes #7020
Closes #6932


kaeawc:
Reconciled onto main after #7000 merged: one `observationId` mint in `createBaseResult` using main's injected `IdGenerator`, the field stays required, and this branch's update-in-place cache write, stale-write skip and read-only settle gating are intact. Observe, resource, settle and lint suites: 3097 pass.

PR #7100, opened 2026-09-14
fix(daemon): preserve provisioning across restart handoffs

### LINKED ISSUE #7088: Concurrent daemon restart interrupts provisionDevice and triggers AVD rollback
## Summary

Overlapping AutoMobile clients sharing one daemon namespace can terminate a daemon while `provisionDevice` is starting an Android AVD. The replacement daemon can then fail strict startup because the previous process still owns the port. This turns a transient daemon-lifecycle race into a failed provision and deletes the newly created AVD.

## Actual behavior

During Android provisioning, the daemon receives `SIGTERM` shortly after the emulator is launched. A replacement daemon starts while the prior process still holds its port, then exits before becoming ready because the port is occupied. The in-flight `provisionDevice` call receives a terminal daemon-startup error, and cleanup deletes the new AVD.

This is related to the concurrent-startup race in #6103, but has a distinct destructive consequence for an active provisioning request. It does not involve an ADB-offline timeout.

## Expected behavior

A daemon with an active device operation must not be terminated by an overlapping client startup or probe. Concurrent clients should join the existing daemon or wait for a verified handoff. A transient port collision must be retried or correlated with the in-flight owner rather than surfaced as a terminal provisioning failure.

## Reproduction

1. Start `provisionDevice` for an Android AVD.
2. While provisioning is in progress, create overlapping AutoMobile clients using the same daemon namespace.
3. Observe a daemon termination/restart during provisioning and a replacement startup that collides with the previously held strict port.
4. Observe `provisionDevice` fail and the new AVD be deleted during cleanup.

## Acceptance criteria

- Active provisioning prevents conflicting daemon termination or hands off without interrupting the operation.
- Concurrent clients coalesce around one healthy daemon generation.
- A replacement waits/retries when a prior daemon still owns its port.
- `provisionDevice` receives a retryable, actionable error for a genuine daemon handoff failure and does not delete an otherwise viable AVD solely because of that race.
- Add coverage for concurrent client startup during Android provisioning.

kaeawc:
## Summary

Make automatic daemon compatibility restarts safe while Android provisioning is active. The daemon now atomically admits a restart only for its verified generation, initiates the accepted handoff itself, fences new provisioning until shutdown, and persists a retryable handoff outcome before later shutdown cleanup. `provisionDevice` preserves an otherwise viable AVD when the handoff genuinely interrupts it.

## Linked Issue

Closes [#7088](https://github.com/kaeawc/auto-mobile/issues/7088)

## Bug / Regression Context

- **Prior attempts:** Status-only restart checks could race a provisioning start.
- **Observed symptom:** An overlapping client could terminate the daemon during AVD creation, then a failed replacement startup led to destructive rollback.
- **Repro steps / conditions:** Start Android `provisionDevice`, then connect a newer or mismatched client in the same daemon namespace.

## Validation

- [x] `scripts/all_fast_validate_checks.sh` equivalent Node gate passes via `bash scripts/prepush-node.sh`
- [x] New code uses interfaces + fakes + FakeTimer; focused unit/integration tests run in <100ms
- [x] No new direct dependencies

## Follow-ups

- [x] No follow-up issues needed


kaeawc:
**Cross-PR file collision — needs human sequencing**

This PR overlaps with [#7099](https://github.com/kaeawc/auto-mobile/pull/7099) (`fix: retain stopped provision on notification failure`). Both open bot PRs modify the same files:

- `src/server/deviceTools.ts`
- `test/server/deviceTools.provisionDevice.test.ts`

#7099 rewrites the stopped-device notification path in `provisionDevice` to a best-effort background notification; this PR (the larger daemon restart-handoff change) also edits that same handler and its test. Whichever merges second will hit a conflict or silently drop the other's change to `deviceTools.ts`.

The per-PR create-time overlap check can't catch this once both are already open, so labeling `needs-human` for a human to sequence the merges (likely land the small #7099 first, then rebase this one). No code changed here — steward flag only.

---

PR #7198, opened 2026-09-17
fix(systemTray): expand collapsed groups before dismiss and never swipe the group

### LINKED ISSUE #7185: systemTray dismiss wipes the whole collapsed notification group (and further collateral) while reporting a single "Dismissed notification"
## Summary

`systemTray { action: "dismiss" }` with a single-notification criterion dismisses the **entire
collapsed group** (and observed collateral beyond it), while reporting
`{"message":"Dismissed notification","success":true}` as if exactly one row was removed.

With only 2 notifications posted (no group formed) dismissal is correct. The bug appears as
soon as Android collapses same-app notifications into a group.

## Repro (deterministic, API 36)

Cold-boot `am-api36-ga-arm64`, `getAndroid`, `setToolEnabled systemTray`.

**Control — 2 notifications, no group → correct:**

```
adb shell cmd notification post -S bigtext -t "Del1" dt1 "DB1"
adb shell cmd notification post -S bigtext -t "Del2" dt2 "DB2"
systemTray { action: "open" }
systemTray { action: "dismiss", notification: { title: "Del1" } }
```
`dumpsys notification --noredact` after → `Del2` survives. Correct.

**Bug — 3+ notifications, group formed → all destroyed:**

```
adb shell cmd notification post -S bigtext -t "Gam1" gt1 "GB1"
adb shell cmd notification post -S bigtext -t "Gam2" gt2 "GB2"
adb shell cmd notification post -S bigtext -t "Gam3" gt3 "GB3"
systemTray { action: "open" }
systemTray { action: "dismiss", notification: { title: "Gam2" } }
```

Before: `Gam1 Gam2 Gam3`. After: **nothing**. Tool returns `"Dismissed notification"`, `success: true`.

## Observed collateral outside the group

A second pass with 4 `Zeta*` notifications, counting `pkg=` lines in
`dumpsys notification --noredact`:

```
BEFORE:  15 pkg=android   6 pkg=com.android.shell   8 pkg=...wellbeing   1 pkg=...safetycenter
AFTER:    9 pkg=android   0 pkg=com.android.shell   8 pkg=...wellbeing   1 pkg=...safetycenter
```

All 6 `com.android.shell` records gone (unambiguous), and `pkg=android` dropped 15 → 6 fewer.
`wellbeing` and `safetycenter` were untouched. I am confident about the group-wide wipe;
the `pkg=android` delta is a raw `grep -c` over dumpsys (which prints `pkg=` in more than one
section) so treat that line as a lead to confirm, not as established.

## Cause

[`src/server/interactionTools.ts:1893`](../blob/main/src/server/interactionTools.ts#L1893) —
the dismiss branch resolves `resolveNotificationSwipeElement(match, ...)` and calls
`swipeElement(device, swipeTarget)`. Inside a collapsed group the resolved element is the group
container/header rather than the individual row, so the swipe clears the whole bundle.

`list` corroborates this: every row in a group comes back with `"ownership":"header"` and
`"inGroup":true` —

```json
{"message":"Listed 5 notifications for com.android.shell","notifications":[
 {"texts":["ExplTitle5","Body"],"inGroup":true,"ownership":"header", ...}, ... ]}
```

so the per-row element the swipe needs is never distinguished from the header.

This is the same hazard `CLAUDE.md` documents for `tap` ("you must expand the group before
tapping a specific notification") — `dismiss` never got the matching expand-first treatment,
and unlike `tap` it is destructive.

## Expected

Either expand the group first and swipe the specific row (the documented `tap` flow), or —
if only the header is resolvable — refuse with an `ActionableError` rather than silently
clearing the group. In no case should a group-wide clear report `"Dismissed notification"`
with `success: true`.

Related: `list` returning `title: null` / `body: null` for every grouped row while the real
text sits in `texts` is likely the same header-vs-row conflation.

---
Filed by the exploratory-tester routine. Daemon + client `@kaeawc/auto-mobile@latest`,
device `emulator-5554` / `am-api36-ga-arm64` API 36.

kaeawc:
## Summary
`systemTray { action: "dismiss" }` on a notification inside a collapsed group swiped the group container, clearing the whole bundle while reporting `Dismissed notification`.

Dismiss now uses the same expand-first flow as `tap`, via a shared `expandAndRematchIfCollapsed` helper:
- **Expansion state is locale-independent and validated against real CtrlProxy captures.** CtrlProxy's systemui visibility bypass keeps one `expandableNotificationRow` per child in *both* states (unlike `uiautomator dump`), and `groupNode` ancestry persists after expansion, so neither "has child rows" nor ancestry is a usable signal. `resolveNotificationGroupExpansionState` returns `expanded` when a child row contains the inflated `status_bar_latest_event_content` template, corroborates with geometry (collapsed children are ~51px stubs overlapping the header; expanded children are full rows below it), falls back to the header button's content-desc, and otherwise reports `unknown`. On `unknown`, `tap` still expands (main's behavior) and `dismiss` refuses. The header expand button is found by resource-id, with a content-desc fallback.
- An already-expanded group is not tapped again (that would collapse it) — this also hardens `tap`.
- The post-expansion re-match must be the **same notification row** (its own title/body text) in the **same group**; otherwise the action aborts instead of acting on an unrelated notification.
- The re-match uses the **remaining** budget of one absolute deadline, so the action never exceeds `awaitTimeoutMs`.
- Composite criteria (`appId` + title/body) keep their group association and resolve to the child row; dismiss never swipes an element that is or contains the children container, or whose bounds equal the group row. If a row cannot be isolated, dismiss throws an `ActionableError` and issues no swipe.
- `clearAll` and the success shape are unchanged. `docs/design-docs/plat/android/system-tray-lookfor.md` documents the new dismiss contract.

## Tests
`bun test test/server test/lint` (4028 pass). Realistic (non-flattened) group fixtures mirroring the device dumps, incl. localized headers: collapsed → expand → swipe child-row bounds; already-expanded → no expand tap; failed expansion → error, zero swipes; cross-group re-match rejected / shifted-top same group accepted; remaining-budget timeout; composite match swipes the child row, never the container. typecheck, lint, format:check clean.

The expansion predicate was run against raw CtrlProxy hierarchies captured from an API 36 emulator in both states (collapsed → `collapsed`, expanded → `expanded`), and the test fixtures are built from those captures. The full tool flow has not yet been run end-to-end against a daemon built from this branch.

Closes #7185


kaeawc:
Device-verified on Android API 36 (0.0.74): posted a shell notification group, then `systemTray dismiss {appId, title}` removed **only** the targeted row — dumpsys confirms `ManualTestFind` 1→0 while sibling `MTX` remained (1). Expand-then-dismiss, no whole-group swipe. `systemTray list` also returned the grouped rows with `inGroup:true / ownership:header` (exercises #7208/#7210 expansion) with no timeout. ✓

PR #7199, opened 2026-09-17
fix(postNotification): probe for the SDK receiver before broadcasting

### LINKED ISSUE #7184: postNotification misreports an absent SDK receiver as a receiver-side failure (supported:true) — am broadcast to a nonexistent explicit component prints result=0, not "no receiver"
## Summary

`postNotification` (Android SDK path) cannot tell "the target app has no AutoMobile
notification receiver" from "the receiver ran and returned failure". It misreports the
first case as the second, with `supported: true` and the non-actionable message
`SDK notification receiver reported a failure.`

The detection at [`src/features/utility/PostNotification.ts:358`](../blob/main/src/features/utility/PostNotification.ts#L358)
string-sniffs `am broadcast` stdout for `no receiver` / `not found` / `does not exist` /
`securityexception`. But `am broadcast` with an **explicit `-n <component>`** that does not
exist prints none of those — it prints `Broadcast completed: result=0`, byte-identical to a
real receiver returning a failure result. So the absent-receiver case falls past the
`supported: false` branch at line 267 and lands on line 299.

The receiver only exists in the SDK's `debug` source set
(`android/auto-mobile-sdk/src/debug/kotlin/.../AutoMobileNotificationReceiver.kt`), so **any**
release build of a target app hits this — including the CtrlProxy APK AutoMobile itself
installs.

## Repro (deterministic)

Cold-boot `am-api36-ga-arm64` (API 36), `getAndroid`, `setToolEnabled postNotification`, then:

```
postNotification {
  sessionUuid: "<uuid>",
  appId: "dev.jasonpearson.automobile.ctrlproxy",
  title: "Expl 1",
  body: "payload-1"
}
```

Observed (identical on all 5 attempts, and still identical after granting
`POST_NOTIFICATIONS` via `setAppPermissions` — the permission is **not** the cause):

```json
{"message":"Failed to post notification: SDK notification receiver reported a failure.",
 "success":false,"supported":true,"method":"sdk",
 "error":"SDK notification receiver reported a failure."}
```

Expected: `supported: false` with the message that already exists at line 273 —
`AutoMobile notification receiver not found in the target app.`

## Evidence

The receiver is genuinely absent from the installed package:

```
$ adb shell cmd package query-receivers --brief -a dev.jasonpearson.automobile.sdk.NOTIFICATION_POST
No receivers found
```

`dumpsys package dev.jasonpearson.automobile.ctrlproxy` (v0.0.74-SNAPSHOT) Receiver Resolver
Table lists only `ProfileInstallReceiver` and `AutoMobileDeviceAdminReceiver` — no
`AutoMobileNotificationReceiver`.

The broadcast is enqueued and completes with no receiver ever running:

```
$ adb shell am broadcast -a dev.jasonpearson.automobile.sdk.NOTIFICATION_POST \
    -n dev.jasonpearson.automobile.ctrlproxy/dev.jasonpearson.automobile.sdk.notifications.AutoMobileNotificationReceiver
Broadcasting: Intent { act=...NOTIFICATION_POST flg=0x400000 cmp=... }
Broadcast completed: result=0
```

logcat shows `ActivityManager: Enqueued broadcast ...: 0` and **zero** receiver-side lines.

## Suggested fix

Replace the stdout sniff with a real capability probe before broadcasting — e.g.
`cmd package query-receivers --brief -a dev.jasonpearson.automobile.sdk.NOTIFICATION_POST`
scoped to `appId` (or a `pm` resolve) — and return the existing `supported: false` shape when
it resolves nothing. Keep the `result=0` branch for the case where a receiver really did run.
`isReceiverUnavailable`'s substring list is unreliable regardless: `not found` would also
match unrelated stderr text.

## Impact

This has blocked the exploratory routine's notification-shade / collapsed-group coverage for
three consecutive runs (2026-09-01, 2026-09-03, 2026-09-17), each time costing time to a
message that points at the wrong layer. `supported: true` actively tells callers the app
supports SDK notifications when it does not.

---
Filed by the exploratory-tester routine. Daemon + client `@kaeawc/auto-mobile@latest`,
device `emulator-5554` / `am-api36-ga-arm64` API 36.

kaeawc:
## Summary
`am broadcast -n <component>` to a nonexistent explicit receiver prints `Broadcast completed: result=0`, byte-identical to a receiver that ran and failed — so an app without the SDK notification receiver was reported as `supported: true, "SDK notification receiver reported a failure."`.

`postNotification` now probes before broadcasting with `cmd package query-receivers --brief -a <action> -p <appId>`:
- `No receivers found` (or an `N receivers found` list without `<appId>/`) → the existing `supported: false` "receiver not found" result, and no broadcast is sent.
- `<appId>/…` listed → broadcast as before; `result=0` still means a real receiver-side failure.
- Anything else (`Unknown command`, empty output, probe throws) → debug log and fall through to the broadcast path.

Also drops the over-broad `not found` substring from `isReceiverUnavailable`. Probe output shapes were checked against a live API 36 emulator.

## Tests
`bun test test/features/utility/PostNotification.test.ts test/features/utility/PostNotificationIos.test.ts` (22 pass), using the real device output strings. typecheck, lint, format:check clean.

Closes #7184


kaeawc:
Device-verified on Android API 36 (0.0.74, HEAD c2e713c8f): `postNotification` now probes for the AutoMobile receiver first — returns `success:false, supported:false, "AutoMobile notification receiver not found in the target app."` for both the default foreground target and an explicit `--appId`, instead of the prior silent false-success broadcast. ✓

PR #7200, opened 2026-09-17
fix(devices): keep configured AVD inventory when the running-state overlay fails

### LINKED ISSUE #7169: Preserve configured AVD inventory when booted-device discovery fails
## Current behavior

Android device-image discovery combines configured AVD enumeration with an optional booted-device scan used only to overlay `isRunning`. If `emulator -list-avds` succeeds but ADB discovery or AVD-name enrichment fails, the rejected booted scan causes the whole `Promise.all` to fail. Callers then receive no configured images and an incomplete Android inventory, even though configured AVD discovery completed successfully.

This is present on current `main` in `MultiPlatformDeviceManager.listAndroidDeviceImages`.

## Expected behavior

Configured AVDs remain visible when the optional live-state overlay cannot be collected. Their `isRunning` values may conservatively remain `false`. Caller cancellation must still propagate and stop the booted-device scan rather than being swallowed as an ordinary discovery failure.

## Reproduction

1. Make `listAvds({ signal })` return at least one configured Android image.
2. Make `getBootedDevicesChecked(...)` reject with a non-cancellation ADB discovery error.
3. Call `getDeviceImagesDetailed("android")` or the Android device-images resource.
4. Observe that the configured image is discarded and Android inventory is marked incomplete.

## Acceptance criteria

- A successful configured-AVD result is retained when only the booted-device overlay fails.
- Retained images conservatively report `isRunning: false` when live state is unavailable.
- An aborted caller signal still rejects promptly and cancels the booted-device scan.
- Tests distinguish a normal ADB discovery failure from cancellation and cover both behaviors.

kaeawc:
## Summary
`listAndroidDeviceImages` overlays `isRunning` from a booted-device scan that is optional, but only an *async rejection* of that scan was isolated. A synchronous throw from `getBootedDevicesChecked`, or a failure while deriving the overlay from its result, escaped the `Promise.all` and discarded the successfully-enumerated configured AVDs (marking Android inventory incomplete).

The fetch + derivation now live in `getAndroidRunningStateOverlay`, which isolates every non-cancellation failure (`logger.warn`, overlay `undefined`) so images are returned with `isRunning: false` / `isRunningStateKnown: false`. Cancellation is still checked first and propagates; a `listAvds` failure still fails the listing.

## Tests
`bun test test/utils/deviceUtils.test.ts` (41 pass): sync-throw overlay failure, overlay-mapping failure, `getDeviceImagesDetailed("android")` staying complete; the existing async-rejection and abort cases remain. typecheck, lint, format:check clean.

Closes #7169


kaeawc:
Device-verified on Android API 36 (0.0.74): `listDeviceImages` returns `configuredInventory.complete:true` with the full `capabilityInventory` for `am-api36-ga-arm64` — configured inventory preserved. ✓

PR #7197, opened 2026-09-17
fix(boot): fail closed on incomplete Android discovery and re-discover before booting

kaeawc:
## Summary
Two coupled fixes in `DeviceBootService`:

- **#7179** — `discoverBootedDevices` dropped `succeededPlatforms`, so a transient ADB failure looked like an authoritative empty result. Android discovery now always goes through `getBootedDevicesDetailed` (still a single call site for the reconcile funnel) and throws a typed `AndroidBootedDeviceDiscoveryIncompleteError` when `android` is missing from `succeededPlatforms`. `provisionDevice` surfaces it as `discovery_incomplete` with `retryable: true` instead of the terminal `platform_command_failed`.
- **#7178** — `bootMatchedImage` short-circuited to boot on a cached `isRunning: false`. Android image-name boots now always run a fresh cache-bypassing discovery first: a unique live match is adopted, multiple same-name transports raise `identity_conflict`, none boots as before. iOS is unchanged.

Concurrency: fresh discovery can see an emulator that *this process* is still cold-booting. Those launches are tracked (per lifecycle coordinator + AVD name, across the whole boot-recovery retry loop) and routed back through `bootImage`, so the existing duplicate-launch coordination applies and a second concurrent `startDevice` does not adopt a mid-boot device. `preferRunning: false` keeps the uniqueness/completeness checks and only forces a boot when the snapshot said the image was not running; an image already reported running is still reused (#3334).

`FakeDeviceUtils` gains `setAndroidDiscoveryIncomplete()`.

## Tests
`bun test test/server test/utils test/lint` (7649 pass): incomplete discovery → typed failure, nothing booted; stale-false overlay with one / two / zero live transports; owned in-flight launch not adopted (incl. across a recovery retry); `preferRunning:false` conflict / incomplete / stale-snapshot / running-image cases; `provisionDevice` → `discovery_incomplete`, retryable. The shared-cold-boot reservation-race tests are unmodified. Two `startDevice` cold-boot fixtures now publish the transport after `startDevice` (the old fixture had a live transport alongside `isRunning:false`, which #7178 now correctly adopts). typecheck, lint, format:check clean.

Closes #7179
Closes #7178

PR #7202, opened 2026-09-17
fix(provisionDevice): settle session release before persisting replay-visible failure

### LINKED ISSUE #7180: Keep provision replay non-successful until failed-session cleanup settles
## Current behavior

Finalization of a failed `provisionDevice` completion can persist the replay-visible terminal status concurrently with releasing the associated device session. A same-operation retry may observe a successful/replayable row while that session still appears live, return it, and then lose it when detached cleanup finishes.

This exact ordering race remains after #7135 / #7146; #7147 covers adjacent lease cleanup only.

## Expected behavior

A failed operation is not published as replayable success until session cleanup has settled or the persisted state explicitly represents cleanup still in progress.

## Acceptance criteria

- Same-operation retries cannot return a session whose release is pending.
- Cleanup completion and replay-visible status have an explicit safe ordering/state transition.
- A deferred-cleanup regression test allows persistence to finish first and proves retry remains blocked or non-successful.

kaeawc:
## Summary
`finalizeFailedProvisionDeviceCompletion` persisted the replay-visible terminal status (`store.fail`, which returns a `replaying` row to `succeeded`) concurrently with releasing the bound device session. A same-operation retry could observe the replayable row while the session still looked live, return it, and then lose it when the detached cleanup finished.

Finalization is now ordered: session release settles first, then the failure is persisted. While release is pending the row stays non-terminal, so a retry gets `operation_in_progress`. A release failure is logged (`logger.warn`) and the failure is still persisted, so the operation cannot stay in-progress because cleanup threw. The superseded path still only releases. `persistFailedProvisionDeviceOperation` and `retireLateProvisionDeviceOperationBegin` were audited and have no concurrent release; no new persisted state was needed.

Trade-off: a *stalled* (never-settling) release now also delays failure persistence; finalization remains detached from the request deadline as before.

## Tests
`bun test test/server/deviceTools.provisionDevice.test.ts test/server/provisionDevice.property.test.ts test/db/provisionDeviceOperationRepository.integration.test.ts` (136 pass): deferred-cleanup regression (retry blocked, `fail` not called before release settles, then failure persisted), release-rejects case, and two existing ordering tests updated to the new contract. typecheck, lint, format:check clean.

Closes #7180


kaeawc:
🧭 **PR-queue steward: cross-PR file collision**

This PR changes files that other currently-open PRs also change. The per-PR create-time overlap check can't catch collisions that only form once both PRs are open, so I'm flagging this for a human to sequence the merges / plan rebase order:

- `src/server/deviceTools.ts` — also changed by #7188, #7197, #7203
- `test/server/deviceTools.provisionDevice.test.ts` — also changed by #7197, #7203

These are the device-provisioning/boot PRs and they overlap on the same hot files, so merging them in an arbitrary order will produce conflicts. Labeling `needs-human` so a person can decide the merge order. No code change is being requested here.

---

Issue #6871, opened 2026-09-13
observe(android): the soft keyboard floods the skeleton with ~40 per-keycap tap nodes on every call while the IME is open — collapse the IME window into one node


kaeawc:
## Context

MCP dogfood loop, 2026-09-12, Android emulator API 35, local build of `f888f1914`. Task: *reply to the most recent notification*. Completed — the reply is in the SMS provider — but the response that told me I had succeeded was almost entirely soft-keyboard keys.

## The problem

Whenever the IME is up, the skeleton lists every individual key as its own actionable node. One `inputText` response, trimmed:

```
  [skeleton] com.google.android.inputmethod.latin:id/key_pos_1_2 | d | tap
  [skeleton] com.google.android.inputmethod.latin:id/key_pos_1_3 | f | tap
  [skeleton] com.google.android.inputmethod.latin:id/key_pos_1_4 | g | tap
  ... (~40 entries: every letter, Shift, Delete, Space, Symbol, Emoji, Enter)
  [skeleton] com.google.android.inputmethod.latin:id/key_pos_ime_action | Enter | tap
  [skeleton] android:id/input_method_nav_back | Back | tap
```

About 40 of the response's nodes were keycaps. The app's own actionable elements — the ones I needed to find `remote_input_send` among — were pushed past them. The skeleton exists to be the compact actionable list; here it is dominated by a control surface no client should ever drive by tapping, because `inputText` and `sendKeys` are the supported path and per-key tapping is explicitly not (`mode: "eventOnly"` exists precisely so nobody has to).

This repeats on **every** call while the keyboard is up, so a form with four fields pays it four or five times.

## What I expected as a client

The keyboard is a mode, not a set of targets. I want to know it is open — `keyboard` already reports that — not to be handed 40 tap targets.

## Smallest change

Collapse the IME window into a single skeleton node rather than enumerating its keys:

```
[skeleton] <ime> | Keyboard (com.google.android.inputmethod.latin) | input
```

Keys are identifiable as descendants of the IME window (`android.inputmethodservice` / the input-method window type), or by the `key_pos_*` resource-id family as a narrower first cut. Anything genuinely non-keycap in that window (the `input_method_nav_back` affordance) can stay.

`project: "full"` / `raw: true` should keep emitting the keys, so the rare client that wants them is not blocked.

## Smaller thing, same response shape

Container nodes come through with a literal `undefined` label:

```
[...]

PR #7215, opened 2026-09-18
fix(observe): collapse the iOS soft keyboard into one <ime> skeleton row

### LINKED ISSUE #7214: observe(ios): soft keyboard still floods the skeleton with ~33 per-keycap nodes and emits no <ime> row — #6871's collapse landed on Android only
## Context

MCP dogfood loop, 2026-09-17, **iOS Simulator** (iPhone 16, iOS 18.6), published release **0.0.75**, driving AutoMobile as the desktop app's directly-wired MCP client. This is the iOS half of the cross-platform run; the Android half re-confirmed all seven 2026-09-12 fixes, including #6871 (Android keyboard now collapses to a single `<ime>` skeleton row).

## The problem

**#6871's keyboard-collapse fix is Android-only — iOS `observe` still enumerates every keycap in the skeleton, with no `<ime>` node.** #6871 was filed and titled `observe(android)`; the iOS side was never covered, so the same flood I reported for Android is still live on iOS.

Task: add a contact (Contacts → New Contact → type a Last name). The `inputText` response's skeleton contained a node per key:

```
s2-f4f27f88982be7ea | q | [tap]
s2-41fd77991aa6ff1b | w | [tap]
s2-9b56ed0226204255 | e | [tap]
... every letter q…p, a…l, z…m ...
shift  | shift  | [tap]
delete | delete | [tap]
more   | numbers| [tap]
space  | space  | [tap]
Return | return | [tap]
s2-e1ef42a66824348f | Emoji   | [tap]
dictation | Dictate | [tap]
```

~33 `UIKeyboardKey` / keyboard-button nodes as individual `tap` targets, pushing the screen's real actionable elements (the field, Done) down the list. On Android the same moment now yields exactly one row:

```
<ime> | Keyboard (com.google.android.inputmethod.latin) | [input]
```

There is **no `<ime>` node on iOS at all** — I checked the skeleton; the keys are the only representation of the keyboard.

It cannot be for lack of signal: the iOS `observe` already computes `screenIdentity` with `keyboardVisible: true` on the very same response, and the raw tree tags the container `className: "UIKeyboard"` (`resource-id: "UIKeyboardLayoutStar Preview"`) with each key `className: "UIKeyboardKey"`. So the server knows the keyboard is up and which nodes are keys — it just does not collapse them the way the Android path now does.

## What I expected as a client

The same thing #6871 delivered on Android: the keyboard is a mode, driven by `inputText`/`sendKeys`, not 33 tap targets. One `<ime>` skeleton row, keys kept only under `raw`/`project:"full"`.

## Smallest change

Extend the #6871 collapse to the iOS skeleton builder: fold the `UIKeyboard` subtree (or the `UIKeyboardKey` class family, plus the `shift`/`delete`/`more`/`space`/`Return`/`dictation` buttons and the Emoji button inside it) into a single `<ime>` node —

```
<ime> | Keyboard (com.apple.keyboard) | [input]
```

— and keep the individual keys in the raw/full projection only, mirroring the Android behavior exactly so the two platforms present a typing screen identically.

## Also verified on iOS in this run (0.0.75), for the record

- **Add-contact task completed** (Priya Nair) and a **Settings toggle task completed** (Accessibility → Display & Text Size → Reduce Transparency ON; the `changed` diff reported `checked: {to:"true"}`).
- iOS **switch rows expose `toggle` + `checked` correctly** — the Android-only loss in #6257 does not exist here.
- `getDeviceState` on iOS returns `success:false` with genuinely excellent, actionable messaging (explains the private `donotdisturbd` Focus daemon, that the simulator shares the host network stack, that `simctl status_bar override` only paints the bar, and to read connectivity from Settings) — no change wanted there.
- `getApple.enableTools[]` and the batch session mint worked the same as Android.

Minor, not filed (likely image-specific, not a tool bug): `launchApp com.apple.mobiletimer` (Clock) failed with a raw `FBSOpenApplicationServiceErrorDomain code=4` — the app appears simply absent from this simulator image. If the intent is for such launches to be client-actionable, the raw FBS error is opaque; but that is a separate, lower-value observation.

Filed by the MCP dogfood loop. Parity follow-up to #6871.

kaeawc:
## Summary
The iOS observe skeleton still flooded with ~33 per-keycap `UIKeyboardKey` rows and never emitted an `<ime>` row: the #6871 keyboard collapse is driven by `ElementProvenance.keyboardPackage`, which only the Android collector set (from the `automobile:imePackage` window extra, gated on `platform === "android"`). The skeleton projection itself is shared.

The iOS collector now opens a keyboard root when a node's class is `UIKeyboard` or `XCUIElementTypeKeyboard` (the same class set `keyboardVisible` already keys on, now exported as `IOS_KEYBOARD_CONTAINER_CLASSES`), labeled `com.apple.keyboard` since iOS has no IME package. Everything downstream is unchanged: the keys fold into one `<ime>` row (`Keyboard (com.apple.keyboard)`, bounds = union of the keys, `affordances: ["input"]`) and the `keyboard: { visible: true, package }` summary is emitted. Android behavior is byte-for-byte unchanged. `docs/tools.md` no longer describes the fold as Android-only.

## Tests
`bun test test/features/observe` (2682 pass) and `bun test test/server test/utils test/lint` (7725 pass): new "iOS keyboard output projection (#7214)" suite — UITextField + UIKeyboard with 32 keys + Done button → exactly one `<ime>` row with the expected label/bounds/affordances, no per-key rows, text field and button still present, `project: "full"` still lists every key, keyboard summary set. typecheck, lint, format:check clean.

Closes #7214


kaeawc:
Verified on-device on iOS 26.2 and iOS 18.6 (plus an Android API 36 non-regression check). Evidence is on #7214.

Issue #7144, opened 2026-09-14
Track Android/iOS parity for stable device acquisition, doctor recovery, provisioning budgets, and CLI-to-MCP readiness


kaeawc:
## Required outcome
Track the complete, cross-platform workflow: provision an exact configured virtual device; reacquire it later by Android AVD name or iOS simulator UUID through the CLI, whether stopped or booted; return only when the requested readiness is achieved; operate that same device through MCP. Doctor must have an explicit path to recover the shared daemon/socket infrastructure.

This is an acceptance and regression-coverage issue, not another report of the linked defects. A merged platform-specific fix is not proof that the corresponding end-to-end contract works on the other platform.

Audit baseline: main [887079dd4f1119d7de4934d69538463b8cd2fb80](https://github.com/kaeawc/auto-mobile/commit/887079dd4f1119d7de4934d69538463b8cd2fb80); issues created and PRs merged since 2026-09-14 11:08 UTC, refreshed before filing. Unit evidence below uses fakes; live platform acceptance remains required.

## Parity matrix
| Contract | Android | iOS | Remaining work |
|---|---|---|---|
| Exact durable device selection | AVD name, separate from current ADB serial/ports; ambiguous live names currently pick a discovery-order winner | UUID, never display-name fallback; exact-UDID fix landed | [Ambiguous AVD issue](https://github.com/kaeawc/auto-mobile/issues/7142); retain [exact-UDID fix](https://github.com/kaeawc/auto-mobile/pull/7098) coverage |
| Correct runtime/configuration | Exact minor API identity fixed; generic bounds and minor-runtime ordering still open | Component-exact matching, runtime bounds, and device-family fallback fixed | [Android bound mismatch](https://github.com/kaeawc/auto-mobile/issues/7133), [Android minor-runtime truncation](https://github.com/kaeawc/auto-mobile/issues/7134) |
| Bounded provisioning and cancellation | Pending config write can corrupt a replacement AVD; final operation persistence remains unbounded | Same final operation persistence path; no claim of the Android file-write defect on iOS | [AVD write fence](https://github.com/kaeawc/auto-mobile/issues/7126), [provision persistence deadline](https://github.com/kaeawc/auto-mobile/issues/7135) |
[...]

Issue #6724, opened 2026-09-09
ios: expose typed bound-session loss across shared daemon restart


kaeawc:
## Summary

On iOS, a client holding a bound device session can receive a generic `session-not-found` response after the shared daemon restarts or the bound session disappears. The response does not let the caller distinguish terminal ownership loss from a retryable transport or loopback-session failure.

## Expected behavior

A bound client should receive a structured, terminal ownership-loss result that identifies the lifecycle reason when available. The result must make clear that the former session cannot be resumed; callers must deliberately acquire a fresh device session before continuing.

## Reproduction scenario

1. Start multiple concurrent iOS simulator clients, each with a distinct active bound device session.
2. Restart the shared daemon while the sessions are active.
3. Make a session-bound operation from each client.
4. Observe whether each client receives a typed ownership-loss result rather than an ambiguous `session-not-found` response or silent retargeting.

## Acceptance criteria

- iOS coverage verifies the behavior of active bound sessions across a shared-daemon restart.
- A lost bound session produces a structured terminal ownership-loss result with an actionable reason when available.
- No request silently resumes or retargets the former session.
- A caller can reliably distinguish bound-session loss from recoverable transport and loopback-session failures.
- Documentation or integration coverage states the required fresh-acquisition recovery path.

Related: #6336, #6383

PR #7255, opened 2026-09-18
fix(daemon): rehydration follow-ups — graceful handoff, autolock identity, ownership ordering, retention/expiry, bounded read

kaeawc:
Follow-up for the six threads deferred from [#7245](https://github.com/kaeawc/auto-mobile/pull/7245) (session rehydration) under the two-round rule. Reproduced live today: a `daemon-restart` row from 2026-09-17 was re-rehydrated as `awaiting-owner` on every daemon restart and blocked the #7144 acceptance-session restart with `unrelated_sessions` until the 600 s reap.

- **Graceful handoff (P1):** `DaemonMcpProxy.handleSessionReleasedNotification` no longer permanently fences the bound UUID when the release reason is `daemon-restart`/`daemon-shutdown`; it marks a recoverable handoff, and `withRecoverableReconnect` resets the transport and reclaims the awaiting-owner session on the next call/heartbeat. Non-recoverable reasons still fence.
- **Autolock identity (P1):** recovery carries the persisted source/autolock/MCP metadata through `createSession`; `DevicePool` restores `device.autolockSessionId`, so `attachAutolockSessionToMcpSession` no longer bails.
- **Ownership ordering (P1):** the persisted liveness-owner token is restored inside `createSession`, before the session is published, so a reconnect's `claimLivenessOwnership()` can't be overwritten by the stale token.
- **Retention / expiry (P2 ×2):** `listRecoverableSessions` applies the seven-day `released_at_ms` cutoff and never revives a row past its `expires_at_ms` — those are terminalized as `expired` first (injected Timer).
- **Bounded list read (P2):** the initial persistence read runs inside the 15 s startup deadline; on timeout the sweep reports `timedOut: true` and the late promise gets a `.catch`.

One FakeTimer test per finding (proxy handoff vs. fence, autolock restore, ownership race, retention/expiry, hung read).

Verified: `bun test test/daemon test/db test/server` (7372 pass), `bash scripts/prepush-node.sh --changed` after rebase onto `origin/main`.


kaeawc:
Round-3 note (contract change, deliberate): the result-mint exclusion is gone entirely — an ordinary `getAndroid`/`getApple`-minted binding now treats a `daemon-restart` **or** `daemon-shutdown` release as a recoverable handoff and reclaims the same session on its next call. That supersedes the pre-rehydration contract from #6724/#6336 ("the former session cannot be resumed; acquire a fresh one"); those tests now use a non-recoverable reason (`device-killed`) for the typed terminal-loss assertions and have sibling tests proving reconnect-and-reclaim of the same UUID, including two independent iOS clients across one shared-daemon restart. On-demand recovery (`createUnseenSession`) now shares the startup admission predicate (expiry + retention), and rows already past retention are pruned rather than re-aged. Rebased onto `19da96163`.

PR #7284, opened 2026-09-19
feat(devices): populate model and architecture in device descriptions

kaeawc:
## PA4 — model/architecture enrichment

Device descriptions emitted `model` and `architecture` as null on both platforms. This was a **wiring gap, not a design gap**: the builder already merges them (`deviceBootService`: `device.X ?? image.X`) and `DeviceInfo` already declares the fields — the sources just weren't fed in. Wired from the existing (already-cached) sources:

- **Android model** ← `getprop ro.product.model` (cached per serial, e.g. `sdk_gphone64_arm64`)
- **Android architecture** ← the system-image ABI (`parseAndroidSystemImageRuntime`, e.g. `arm64-v8a`), falling back to the AVD config / `getprop ro.product.cpu.abi`
- **iOS model** ← the simulator device-type profile `modelIdentifier` (same profile already used for display dims)
- **iOS architecture** ← host arch (simulators execute on the host)

Additive — no shape change; the existing merge precedence is preserved. Verified: deviceDescription + client + platform-prep suites 170 pass; typecheck/lint/format green.


kaeawc:
**Device-verified with one gap** — nightly `manual-test-daily`, `main` d5087497b (0.0.76).

**Android, booted acquisition — PASS.** `getAndroid`, `listDevices` and the `automobile:devices/booted` resource all agree with adb:

| field | tool | `adb shell getprop` |
|---|---|---|
| `model` | `sdk_gphone64_arm64` | `ro.product.model` = `sdk_gphone64_arm64` |
| `architecture` | `arm64-v8a` | `ro.product.cpu.abi` = `arm64-v8a` |

**iOS — PASS on both configs, including cold boot.** `getApple` carries `model` / `architecture` on the cold-boot response itself: iOS 26.2 -> `iPhone18,3` / `arm64`; iOS 18.0 -> `iPhone15,4` / `arm64`.

**Gap: Android cold-boot omits `model`.** On the cold-boot `getAndroid` response — the one that simultaneously reports `isReady: true` and `runtime.lifecycle.state: "booted"` — `architecture` is populated but `model` comes back `null`, while `adb shell getprop ro.product.model` answers correctly at that same moment. Reproduced 2/2; re-acquiring the same (now booted) device immediately returns `sdk_gphone64_arm64`. Filed as #7321 with the `modelNameCache` root cause.

Everything else from this PR checks out; the cold-boot path is the only surface where the two fields disagree.

PR #7288, opened 2026-09-19
fix(daemon): listDevices reads are side-effect-free (no ownership reclaim)

kaeawc:
## PA3b — a read must not reclaim ownership

`automobile:devices/booted` (read via a non-owning DaemonClient) honestly reported a rehydrated session as `awaiting-owner`, but `--cli listDevices --session-uuid X` **promoted it to `owned`** as a side effect of the read. Two promotion routes are now closed:

- **Proxy bind/heartbeat**: `daemonMcpProxy.rememberSessionUuid` — and the sibling failure-path binders `refreshReplayLeaseAfterAdmittedFailure` / `mayBindSessionFromErrorResult` — now skip `listDevices` (mirroring the existing `setToolEnabled` carve-out), so a read never starts a bound-session heartbeat whose keeper tick promoted `awaiting-owner`→`owned`.
- **Not-yet-resident join race**: `SessionManager` threads a `read-only` access mode through `getOrCreateSession`/`startOrJoinUnseenAssignment`, so a session joined by a read-only admission before rehydration isn't promoted either.

`listDevicePayloads` now projects the session's **real** ownership (via `SessionManager.getSession()`) instead of defaulting to `owned`, so `listDevices` and `devices/booted` report identical ownership at the same instant.

Genuine acquisition (getAndroid/startDevice/tapOn/setActiveDevice) still binds, heartbeats, and promotes — a real-`DaemonMcpProxy` regression test pins that contrast and fails without the fix.

Verified: `bun test test/daemon/ test/server/ test/cli/` 6438 pass; typecheck/lint/format green. (Advisor-reviewed; this is the corrected fix after the first pass missed the heartbeat-keeper promotion.)


kaeawc:
**Device-verified** on the nightly `manual-test-daily` run — `main` d5087497b (0.0.76), `am-api36-ga-arm64` (API 36).

Held a device session in one process, then issued `listDevices` from a **separate** process three times while the session was live, then went back to the holder:

```
process A: getAndroid { avdName: "am-api36-ga-arm64" }   -> ready (booted), session owned
process B: listDevices  -> "Found 1 booted device"
process B: listDevices  -> "Found 1 booted device"
process B: listDevices  -> "Found 1 booted device"
process A: observe      -> served normally (7214-byte payload, no isError)
```

The holder's session survived all three cross-process reads — ownership was not reclaimed. Before this fix that read path was the one that could take the device out from under a live owner.

Separately, `listDevices` reported the owning session correctly throughout: `runtime.session = { sessionUuid: "618138d8-…", ownership: "owned" }`.

Issue #7279, opened 2026-09-19
getAndroid acquisition result drops the AVD image link (image.path/target/basedOn all null) that listDevices reports for the same device


kaeawc:
Found by the nightly manual-test routine (2026-09-19), on a daemon built from `bf28e4fe0`.

### Observed
On the same booted AVD, a few seconds apart, the tools report different `image` fields:

| surface | `image` for `am-api36-ga-arm64` (emulator-5554) |
|---|---|
| `listDevices {platform: android}` | `{path: "/Users/…/.android/avd/am-api36-ga-arm64.avd", target: "Google APIs (Google Inc.)", basedOn: "Android 16.0 (\"Baklava\") Tag/ABI: google_apis/arm64-v8a"}` |
| `automobile:devices/booted` | populated, same as above |
| `getAndroid {deviceId: emulator-5554}` (acquisition `already-booted`) | `{path: null, target: null, basedOn: null}` |

This also reproduces on a second AVD (`acceptance-pixel-6`, emulator-5556). The acquisition result already knows the backing image, because `runtimeId: "system-images;android-36;google_apis;arm64-v8a"` and `deviceType: "pixel_6"` are populated in the same response. Only the image link is dropped.

Related gap in the same response: `runtime.deviceSessionUuid` is `null` in the `getAndroid` result, while `listDevices` reports it (`2f66055c-…`) for the same pool entry.

### Expected
#7275 (E2) says: "`getAndroid`/`getApple` acquisition results and `provisionDevice.device` now carry the same `display`/`capabilityInventory`/image facts as `listDevices` and `devices/booted`". Under E5, the booted↔image link is "genuinely null only without a backing image".

### Root cause
`buildBootedResponse` (`src/server/deviceTools.ts:9279-9300`) calls `describeDevice({kind: "booted", device, pooled, discovery: sourceImage, session, serviceStatus})`. It never passes `configured:`, and it never passes `deviceSessionUuid:`.

`describeBooted` derives the link from `imageLinkWithConfiguredFallback(admittedImage, configured)` (`src/server/deviceDescription.ts:305`, `:366-375`). `sourceImage` is a `DeviceInfo` with no `image` sub-object, and `configured` is `undefined`, so every field falls through to `null`.

`listDevicePayloads` (`deviceTools.ts:1045-1060`) passes both `configured: configuredImageForBootedDevice(device, configuredImages)` and `deviceSessionUuid`, which is why `listDevices` is correct.

[...]

PR #7297, opened 2026-09-19
fix(devices): warm Android provenance during boot so cold-cache getAndroid carries the image link

kaeawc:
## Completes #7279 (cold-cache case)

#7279 made getAndroid/getApple carry the AVD image link, reading it on the hot path from `AndroidAvdProvenanceCache.getCachedByName()` (cache-only, no scan). Gap: a **cold cache** — getAndroid / Android startDevice as the first inventory call after daemon start — left `image.{path,target,basedOn}` null (CodeRabbit #7283 r4052961672).

## Fix
`bootAndPrepareDevice` now kicks off `androidProvenanceByAvdName` (the existing generation-guarded / in-flight-shared / bounded `getByName` wrapper — **no second provenance path**) **concurrently with boot**, and awaits it just before `buildBootedResponse` projects. So provenance is warm by projection time without a new serial scan and **without regressing the prompt-settling contract** #7279 restored. iOS unchanged.

## Test
New `deviceTools.startDevice.test.ts` case: a **cold** `AndroidAvdProvenanceCache` (no pre-warm) acquisition still returns `image` populated. 234 tests pass across the device suites incl. the platformPreparation prompt-settling test; typecheck/lint clean.


kaeawc:
Superseding this boot-path warm approach. The bounded 50ms best-effort await can't complete a cold avdmanager scan (~1.5-3s), so it doesn't reliably populate the cold-cache image link and it adds warm work to the acquire path. Replacing with a daemon-startup fire-and-forget warm (Option C) so the provenance cache is warm before acquisitions, leaving the acquire path a pure cache-only read.

Issue #7361, opened 2026-09-20
systemTray(clearAll): #7332 widening filters on ownership==="dumpsys" but list reports these rows as "header", so nothing is ever cleared


kaeawc:
Follow-up to #7320 / #7332, found by the nightly `manual-test-daily` routine on `main` 25948653c, Android emulator `am-api36-ga-arm64` (API 36).

### What #7332 fixed, and what it did not

The **honest-reporting half works** — this is a real improvement and the headline harm (a false `success: true`) is gone:

```
systemTray {action:"clearAll", notification:{appId:"com.android.shell"}}
-> success: false
   "Cleared 0 of 3 notification(s) for com.android.shell; 3 could not be matched on screen."
```

The **functional half does not**. Nothing is dismissed: `dumpsys notification` still reports all 3 afterwards, including when the shade is explicitly expanded first and rows are confirmed visible in a `uiautomator dump` at that moment. Reproduced 3/3.

### Root cause: the widening filter never matches

#7332 widens the swipe loop's match texts with dumpsys-correlated row texts, but only for rows whose `ownership` is exactly `"dumpsys"`:

`src/server/interactionTools.ts` (clearAll)
```ts
clearMatchTexts = [
  ...new Set([
    ...appMatchTexts,
    ...listed.notifications
      .filter((listedNotification) => listedNotification.ownership === "dumpsys")
      .flatMap((listedNotification) => listedNotification.texts),
  ]),
];
```

On this device `list` reports every one of those rows as `ownership: "header"`, not `"dumpsys"`:

```
systemTray {action:"list", notification:{appId:"com.android.shell"}}
-> "Listed 3 notifications for com.android.shell"
   ownership='header'  texts=["MT7332-3","Body"]  appId=com.android.shell
   ownership='header'  texts=["MT7332-2","Body"]  appId=com.android.shell
   ownership='header'  texts=["MT7332-1","Body"]  appId=com.android.shell
```

So the filter yields an empty array, `clearMatchTexts` collapses back to `appMatchTexts`, and the swipe loop behaves exactly as it did before #7332 — it breaks on the first iteration with no match. The texts needed to isolate these rows (`"MT7332-1"`, …) are right there in the `list` payload; they are simply excluded by the ownership predicate.

### Repro

```
for i in 1 2 3; do
  adb -s emulator-5554 shell cmd notification post -S bigtext -t "MT7332-$i" "mt7332_$i" "Body $i"
done
# then, in one MCP session:
[...]

PR #7392, opened 2026-09-21
fix(systemTray): clearAll success from post-clear remaining count, not swipe count (#7383)

### LINKED ISSUE #7383: systemTray(clearAll): reports "Cleared 1 of 3 ... could not be matched" with success:false after dismissing all 3 — dismissed counts swipes, expectedCount counts notifications
Follow-up to #7361 / #7365, found by the nightly `manual-test-daily` routine on `main` 658a3f0c0, emulator `am-api36-ga-arm64` (API 36).

### The functional half is fixed — the accounting inverted

#7365 works: the notifications now actually get dismissed, which was the whole point of #7361. But a **complete** success is now reported as a **partial failure**:

```
systemTray {action:"list",     notification:{appId:"com.android.shell"}}  -> "Listed 3 notifications"
systemTray {action:"clearAll", notification:{appId:"com.android.shell"}}
  -> success: false
     "Cleared 1 of 3 notification(s) for com.android.shell; 2 could not be matched on screen."

adb shell dumpsys notification --noredact | grep -c MT7365   ->  0
```

The tray is empty. All three are gone. The caller is told the operation failed.

Reproduced **3/3** with a fresh set of three notifications each time (before=3, after=0, same message every run).

### Cause

`dismissed` counts **swipes issued**, while `expectedCount` counts **dumpsys rows**. Android collapses three same-app notifications into one group, and a single swipe on the group dismisses all three. The loop then re-matches, finds nothing left (correctly — they are gone), breaks with `dismissed === 1`, and `formatClearAllResult` compares the two incommensurable numbers:

`src/server/interactionTools.ts:2107-2113`
```ts
if (dismissed >= expectedCount) {
  return { message: `Cleared ${dismissed} notification(s) for ${appId}`, success: true };
}
return {
  message:
    `Cleared ${dismissed} of ${expectedCount} notification(s) for ${appId}; ` +
    `${expectedCount - dismissed} could not be matched on screen.`,
  success: false,
};
```

`expectedCount` is a sound input; `dismissed` is not a comparable quantity. This is the same swipes-vs-notifications conflation as #6482, now load-bearing for the `success` flag rather than just a count.

### Impact

`success: false` on a fully-cleared tray is worse than an inaccurate count: it is the field callers gate on. A caller that retries on failure will loop against an empty tray, and one that surfaces the error will report a spurious failure for an operation that did exactly what was asked. Grouped notifications are the common case for any app posting more than one, so this fires routinely rather than at an edge.

Note the interaction with #7361: the honest-reporting behaviour added there was correct and valuable when nothing was being dismissed. Now that dismissal works, the same comparison produces a false negative. Both halves need the same ground truth.

### Suggested fix

Re-read the dumpsys-backed count after the swipe loop and report against **that**, not against swipes issued:

- post-loop remaining `=== 0` -> `success: true`, "Cleared N notification(s)" where N is `expectedCount - remaining`;
- remaining `> 0` and `< expectedCount` -> genuine partial, report `remaining` as what could not be matched;
- remaining `=== expectedCount` -> nothing was dismissed, which is the #7361 case and correctly `success: false`.

That also fixes #6482 as a side effect, since the count stops being derived from swipes at all.

### Verification

Post 3 notifications from one app (they will group), `clearAll` by appId: `success` must be `true`, the reported count must be 3, and `dumpsys notification` must show 0. A test with a fake where one swipe removes three correlated rows would pin the grouped case.

Related: #7361, #7365, #6482, #6875, #6883.

kaeawc:
Fixes #7383 (gap in merged #7332/#7361).

`clearAll` reported `Cleared 1 of 3 ... could not be matched` with `success:false` **after** dismissing all 3: `dismissed` counted **swipes** while `expectedCount` counted **notifications**, so a single swipe clearing a grouped notification (multiple dumpsys rows) under-counted.

## Fix
`formatClearAllResult` now takes `remainingCount`: `cleared = expectedCount - remaining`, `success = remaining === 0`. The handler re-runs `listSystemTrayNotifications` for the appId after the swipe loop (only when `expectedCount > 0`) to get the real remaining count. `dismissedCount` stays the raw swipe count (diagnostic field). The #7361 widening, collapsed-group handling, and honest-reporting contract are preserved.

## Tests
Grouped swipe clearing 3 notifications → `success:true`, "Cleared 3"; a genuine partial (1 of 3 remains) → `success:false`, "Cleared 2 of 3 … 1 could not be matched". 61 systemTray-list tests + 732 across all systemTray suites pass; typecheck/lint/format clean.


kaeawc:
Superseded by #7389, which already merged to main and closes #7383 with a more robust fix (stable dumpsys notification identities before/after clearAll, counting grouped notifications independently of swipe commands, excluding API-36 group summaries, and reporting concurrent arrivals honestly). This PR's re-list-and-count approach is redundant; closing.

PR #7408, opened 2026-09-21
fix(daemon): bounded reconciliation for unreadable AVD name instead of immediate identity quarantine (#7403)

### LINKED ISSUE #7403: Recover recorded AVD identity after incomplete runtime discovery
## Current behavior

An Android emulator can remain recorded in the shared daemon under a known AVD name while a later discovery pass temporarily cannot read that AVD name from the runtime. A subsequent device-scoped operation is then rejected with the equivalent of:

> this daemon has the device recorded as AVD `<name>`, but discovery could not read the AVD name from the runtime, so its identity is unresolved

This can occur after the device was successfully acquired, the application was installed, and later setup work begins. The caller is told to retry or reacquire, but the existing session does not reliably recover its identity before failing.

## Expected behavior

A temporary inability to read the runtime AVD name should preserve the last confirmed identity as non-destructive evidence while bounded reconciliation retries discovery. Operations must remain fenced from a genuinely different emulator, but a healthy device should not become permanently unusable after one incomplete observation.

## Details

This is related to #6406, which covers validating identity at cold-boot return time. This report concerns continuity after a previously confirmed daemon mapping already exists and a later observation becomes incomplete.

## Reproduction

1. Acquire a named Android AVD and confirm its serial/name mapping.
2. Make both runtime name probes temporarily return empty for that serial during a later discovery cycle.
3. Invoke a device-scoped resource or tool through the existing session.
4. Restore the runtime name probe and repeat.

Today the intermediate operation is rejected as unresolved; verify whether the session can recover without release/reacquisition.

## Acceptance criteria

- Incomplete runtime-name discovery is distinguished from confirmed identity mismatch.
- The last confirmed serial/AVD mapping is retained as evidence but is not used destructively.
- Bounded reconciliation can restore the existing session when the same runtime identity becomes observable again.
- Confirmed mismatch still fences operations and cannot be bypassed by stale identity.
- Tests cover temporary omission, recovery, and true mismatch.

kaeawc:
Closes #7403

## Problem
When a discovery sweep reads the `Unknown (<serial>)` placeholder (runtime AVD-name probe returned empty) for a pooled emulator that already has a confirmed AVD name, the daemon immediately quarantines it (`identityUnresolved = true`), and FUNNEL 2 (`assertDeviceActionable`) then rejects every later device-scoped operation — bricking a healthy session until some later sweep happens to resolve the name. It conflates "incomplete discovery" with "confirmed mismatch."

## Fix (in `devicePool.ts`)
- New `PooledDevice` fields `identityReconcileAttempts` / `identityReconcileStartedAt` (kept separate from the terminal `identityUnresolved`).
- `shouldQuarantineUnresolvedEmulatorName` now only quarantines immediately when there is **no** confirmed `avdName` or the entry is already terminal. For a confirmed entry seeing an unreadable name, `shouldReconcileUnresolvedEmulatorName`/`reconcileUnresolvedEmulatorName` run a **bounded 3-attempt** reconciliation (via `this.retryExecutor`, `DEFAULT_RETRY_OPTIONS.maxAttempts`, cache-bypassing rediscovery on attempts 2–3), preserving the last confirmed identity as non-destructive evidence (`identityObservedAt` is **not** advanced — "no new evidence").
- During reconciliation the device stays actionable (FUNNEL 2 only rejects on the terminal flag). A resolved-but-**different** name still routes to the unchanged immediate-fence `quarantineDisagreeingPooledIdentity`. Exhausting the bound still calls `enterPooledIdentityQuarantine` (terminal fallback), so a permanently-gone device is still fenced. A concurrent-epoch guard prevents a stale in-flight retry from clobbering a newer resolution/replacement.

## Tests
`devicePool.test.ts`: recovery after temporary omission (stays actionable, session preserved, never terminally quarantined), immediate fence on a different resolved name during retries, terminal quarantine after bound exhaustion; updated the old immediate-quarantine case. 266 pass across devicePool/quarantine/fence suites; funnel-allowlist bumped 8→9.

## Validation
`bash scripts/prepush-node.sh --changed` — all gates PASS.


kaeawc:
**Design note (deliberate trade), from a fresh-eyes review:** `assertDeviceActionable` is intentionally left permissive during the bounded reconciliation window. So for up to the 2 cache-bypassing rediscoveries (attempts 2–3), a pooled entry with a confirmed identity stays actionable while its runtime name is temporarily unreadable. The edge case this accepts: a *same-serial replacement* emulator whose console hasn't answered yet produces the same `Unknown (<serial>)` placeholder, so an in-flight op could target the replacement until attempt 2/3 resolves a *different* name and `quarantineDisagreeingPooledIdentity` fences + cancels executions. This is the exact continuity-vs-fencing trade #7403 asks for — a healthy device must not be bricked by one incomplete observation — and in practice the old emulator's death reaches the missing-device path first. A genuinely different *resolved* name still fences immediately; only the unreadable-placeholder case gets the grace window. Termination is guaranteed (3 failures → quarantine).

PR #7456, opened 2026-09-22
fix(android): route sendKeys replace of WYSIWYG-trigger text through per-key delivery

kaeawc:
## Problem
`sendKeys {action:"type", operation:"replace"}` resolved to the batch accessibility set-text path, which sets the field value wholesale and bypasses the app's per-keystroke input pipeline. A rich-text/markdown editor that formats or autocompletes as you type (code fences from ```, bold from `*`, italic from `_`, strikethrough from `~`) never sees those keystrokes, so the WYSIWYG/autocomplete never triggers. (`operation:"insert"` already used the per-key path.)

## Fix
Route replace text that contains a WYSIWYG trigger char (backtick, asterisk, underscore, tilde) through the existing per-key `eventAll` path instead of batch `a11y`. New `src/features/action/wysiwygTriggerChars.ts` exports `WYSIWYG_TRIGGER_CHARS` + `containsWysiwygTriggerChar`. Ordinary text and explicit `mode` callers are unchanged (batch stays the fast default); iOS is unaffected. Shifted trigger chars need API 31+ for `input keycombination`; older APIs degrade via the existing `eventAll` per-character fallback.

## Verification
- 27 unit tests: replace+trigger-char → `eventAll` per char; ordinary replace → `a11y`; insert unchanged; explicit `mode:"a11y"` wins; predicate unit tests.
- On a physical API-36 device, per-key (`eventAll`) delivery was confirmed to trigger the editor's autoformat (the opening fence is consumed into a code block) — this change routes the replace path through that same proven per-key path.
- prepush-node --changed green.


kaeawc:
**pr-queue-steward:** flagged `needs-human`. This PR has two blockers and has not moved in ~20h (created/last-pushed 2026-09-22 16:06Z), with no owning-routine refire:

1. **Merge conflict** — `mergeable_state: dirty` against `main`. A burn-down routine would resolve this by merging base into the head; it hasn't refired to do so.
2. **Unresolved P1 review thread** — [codex's "Preserve password replacements on pre-API-31 Android"](https://github.com/kaeawc/auto-mobile/pull/7456#discussion_r4076908197): on API ≤30, auto-routing shifted WYSIWYG triggers (`*` `_` `~`) to `eventAll` can destructively half-replace a password field (the trigger falls through to `textClient.insert`, which rejects password nodes), where the old atomic `requestSetText` path did not.

Per steward policy I don't push code — the owning routine repairs its own PR when it next fires. Flagging so a human can re-sequence or kick it if the routine isn't scheduled to return.

---

kaeawc:
Superseded by https://github.com/kaeawc/auto-mobile/pull/7500. This PR's premise (per-key `eventAll` triggers the composer autoformat) was disproven on-device this session — only IME `commitText`/`mode:"ime"` triggers it. The corrected version auto-routes trigger-char text to `mode:"ime"` with a fallback. Closing in favor of that.

